PR #55, opened 2020-10-07
Use urlencode for search queries (fixes #10)

### LINKED ISSUE #10: URL search is broken
Partially a URI encode / decode issue.

Nutomic:
This fixes the search for URLs like `http://lemmy-beta:8551/c/main`, and correctly displays the community in the search results. Both when searching from the navbar and from the search page.

I also tested the HTTP API using `curl "http://localhost:8540/api/v1/search?q=test%3F&type_=All&sort=TopAll"`, which is logged on the backend as `q: "test?"`, so it looks like actix is already decoding parameters for us, and we dont have to change anything there.

I think this also needs to be fixed [here in lemmy-js-client](https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L137).

### DIFF
--- a/src/shared/components/navbar.tsx
+++ b/src/shared/components/navbar.tsx
@@ -137,8 +137,9 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
     if (searchParam === '') {
       this.context.router.history.push(`/search/`);
     } else {
+      const searchParamEncoded = encodeURIComponent(searchParam);
       this.context.router.history.push(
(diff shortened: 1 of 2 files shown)


dessalines:
I think the lemmy-js-client should be okay, its encoding the params https://github.com/LemmyNet/lemmy-js-client/blob/main/src/http.ts#L380

I'll test out the search page locally tho just to make sure.

dessalines:
I did some logging to test as well, looks good.

PR #62, opened 2020-10-23
COPY requires / suffix

revmischa:
I got this error when building:
`When using COPY with more than one source file, the destination must be a directory and end with a /`


dessalines:
Mmmk thanks, must be a new docker version thing.

PR #65, opened 2020-10-25
Use Webpack 5 filesystem cache

0rvar:
Project rebuilds in 3.5 seconds instead of 10 seconds.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -12,6 +12,8 @@ const banner = `
   `;
 
 module.exports = function (env, _) {
+  const platform = env.platform || 'server';
+
   const base = {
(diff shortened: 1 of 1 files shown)


dessalines:
Thanks! @0rvar I'm a total webpack noob, is there any way to combine the builds into one, instead of running the whole thing twice like I do now?

0rvar:
> Thanks! @0rvar I'm a total webpack noob, is there any way to combine the builds into one, instead of running the whole thing twice (once for server, once for client) like I do now?

I'm looking into it, it might be possible. I'll post as a separate PR

0rvar:
@dessalines implemented in #67

PR #68, opened 2020-10-26
Don't flash default theme on page load

### LINKED ISSUE #45: Page reload flashes default theme
This can be very noticable when a theme different from the default `lightly` is configured. Usually the page reload is almost instant, but sometimes it takes a second or more for the correct theme to be applied, so that buttons flash up in green. Sometimes even the background stays white for a noticable time before darkly is applied.

This problem only started to happen after the isomorphic ui rewrite, not before.

Happens with Firefox and Firefox Android.

0rvar:
Fixes #45 

Also initialize more things in the entry points instead of inside Navbar

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -13,6 +13,7 @@ import { lemmyHttp, setAuth } from '../shared/utils';
 import { GetSiteForm, GetSiteResponse } from 'lemmy-js-client';
 import process from 'process';
 import { Helmet } from 'inferno-helmet';
+import { initializeSite } from '../shared/initialize';
 
 const server = express();
 const port = 1234;
@@ -59,6 +60,19 @@ server.get('/*', async (req, res) => {
     lang,
   };
 
+  initializeSite(site);
+  const user = site.my_user;
+  const userTheme =
+    user &&
+    user.theme &&
+    `
+      <link
+        rel="stylesheet"
+        type="text/css"
+        href="/static/assets/css/themes/${user.theme}.min.css"
+      />
+    `;
+
   const wrapper = (
     <StaticRouter location={req.url} context={isoData}>
       <App site={isoData.site} />
@@ -95,8 +109,15 @@ server.get('/*', async (req, res) => {
 
            <!-- Styles -->
            <link rel="stylesheet" type="text/css" href="/static/styles/styles.css" />
+           ${userTheme || ''}
+           ${
+             userTheme
+               ? ''
+               : `
            <link rel="stylesheet" type="text/css" href="/static/assets/css/themes/litely.min.css" id="default-light" media="(prefers-color-scheme: light)" />
            <link rel="stylesheet" type="text/css" href="/static/assets/css/themes/darkly.min.css" id="default-dark" media="(prefers-color-scheme: no-preference), (prefers-color-scheme: dark)" />
+           `
+           }
            </head>
 
            <body ${helmet.bodyAttributes.toString()}>
@@ -105,7 +126,7 @@ server.get('/*', async (req, res) => {
                  <b>Javascript is disabled. Actions will not work.</b>
                </div>
              </noscript>
-            
+
              <div id='root'>${root}</div>
              <script defer src='/static/js/client.js'></script>
            </body>
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -89,7 +89,7 @@ server.get('/*', async (req, res) => {
            <!DOCTYPE html>
            <html ${helmet.htmlAttributes.toString()} lang="en">
            <head>
-           <script>window.isoData = ${serialize(isoData)}</script>      
+           <script>window.isoData = ${serialize(isoData)}</script>
 
            ${helmet.title.toString()}
            ${helmet.meta.toString()}
@@ -109,14 +109,12 @@ server.get('/*', async (req, res) => {
 
            <!-- Styles -->
            <link rel="stylesheet" type="text/css" href="/static/styles/styles.css" />
-           ${userTheme || ''}
            ${
-             userTheme
-               ? ''
-               : `
+             userTheme ??
+             `
            <link rel="stylesheet" type="text/css" href="/static/assets/css/themes/litely.min.css" id="default-light" media="(prefers-color-scheme: light)" />
            <link rel="stylesheet" type="text/css" href="/static/assets/css/themes/darkly.min.css" id="default-dark" media="(prefers-color-scheme: no-preference), (prefers-color-scheme: dark)" />
-           `
+           `.trim()
            }
            </head>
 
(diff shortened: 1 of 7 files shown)


src/shared/utils.ts
```diff
@@ -443,32 +443,34 @@ export function getMomentLanguage(): string {
...
 export function setTheme(theme: string, forceReload: boolean = false) {
-  if (isBrowser() && (theme !== 'browser' || forceReload)) {
-    // This is only run on a force reload
-    if (theme == 'browser') {
-      theme = 'darkly';
-    }
+  if (!isBrowser()) {
+    return;
+  }
+  if (theme === 'browser' && !forceReload) {
```
dessalines, line 449: You can combine these.

src/shared/components/app.tsx
```diff
@@ -8,6 +8,7 @@ import { Navbar } from '../../shared/components/navbar';
 import { Footer } from '../../shared/components/footer';
 import { Symbols } from '../../shared/components/symbols';
 import { GetSiteResponse } from 'lemmy-js-client';
+import { UserService } from '../../shared/services';
```
dessalines: This is probably unused here.

src/shared/initialize.ts
```diff
@@ -0,0 +1,9 @@
+import { GetSiteResponse } from 'lemmy-js-client';
+import { UserService } from './services';
+import { i18n } from './i18next';
+import { getLanguage } from './utils';
+
+export const initializeSite = (site: GetSiteResponse) => {
+  UserService.Instance.user = site.my_user;
+  i18n.changeLanguage(getLanguage());
```
dessalines, line 8: This should be good in an isomorphic setting, but I'll test this too.

src/server/index.tsx
```diff
@@ -95,8 +109,15 @@ server.get('/*', async (req, res) => {
 
            <!-- Styles -->
            <link rel="stylesheet" type="text/css" href="/static/styles/styles.css" />
+           ${userTheme || ''}
+           ${
+             userTheme
+               ? ''
```
dessalines: Maybe forgot to delete something here.

dessalines: Or how about `${userTheme || <link ...`

src/shared/initialize.ts
```diff
@@ -0,0 +1,9 @@
+import { GetSiteResponse } from 'lemmy-js-client';
+import { UserService } from './services';
+import { i18n } from './i18next';
+import { getLanguage } from './utils';
+
+export const initializeSite = (site: GetSiteResponse) => {
```
dessalines, line 6: I'm not a fan of closures where they aren't necessary. Instead of this new file, just move this function to `utils.ts`, and 

`export function initializeSite(site: GetSiteResponse)...`

dessalines (changes requested):

src/server/index.tsx
0rvar: This diff is updated now, moved this stuff into it's own component

src/shared/components/app.tsx
0rvar: Fixed in the latest diff

src/shared/initialize.ts
0rvar: I'll change to a named function. Babel does give the closure a name in this case though, just FYI.

Regarding utils.ts, I think it's very busy. I'd much rather see a `utils` directory with smaller, more categorized files. I can make a detailed suggestion of what I have in mind in an issue for discussion if you want

src/shared/utils.ts
0rvar: I think it reads easier to have them separate. The first one says "This code is not run in SSR", while the second deals with the logic of default themes and force reloading

dessalines: I'm good with that.

src/shared/initialize.ts
dessalines: Sure, I agree utils is a bit busy. A lot of things in there are only used by a single component, in which case they should probably just be private functions there. 

But overall its nice to only have to worry about a single import for logic functions that aren't components.

src/shared/initialize.ts
dessalines: I still prefer this get moved to utils.ts until we think out a bigger re-org.

src/shared/components/app.tsx
```diff
@@ -18,12 +19,12 @@ export class App extends Component<AppProps, any> {
   constructor(props: any, context: any) {
     super(props, context);
   }
-
   render() {
     return (
       <>
         <Provider i18next={i18n}>
           <div>
+            <Theme user={this.props.site.my_user} />
```
dessalines, line 27: Nice, this seems clearer.

src/shared/components/theme.tsx
```diff
@@ -0,0 +1,36 @@
+import { User } from 'lemmy-js-client';
+import { Helmet } from 'inferno-helmet';
+
+export const Theme = (props: { user: User | undefined }) => {
```
dessalines, line 4: No closure, use the component style like the other components. And define the props object like the others, with user as optional.

src/shared/components/theme.tsx
```diff
@@ -0,0 +1,36 @@
...
+            type="text/css"
+            href="/static/assets/css/themes/darkly.min.css"
+            id="default-dark"
+            media="(prefers-color-scheme: no-preference), (prefers-color-scheme: dark)"
+          />
+        </>
+      )}
+    </Helmet>
+  );
+};
```
dessalines, line 36: Looks good, nice.

dessalines (changes requested):

src/shared/initialize.ts
0rvar: Switched to function. The main problem right now is that if I move initializeSite into utils, there are dependency cycles which breaks getLanguage(), due to initializeSite needing to import `i18next`. That's another drawback of keeping all the utils in the same file.

src/shared/components/theme.tsx
0rvar: I changed to component style.

Optional and allowing undefined are distinctly different. Optional means you don't have to pass the prop at all. Non-optional but allowing undefined means we have to pass the prop, but the value can be undefined. In this case, this latter version fits better, because we need the user object, but we know it can sometimes be undefined.

src/shared/initialize.ts
0rvar: See comment in other chain, can't move to utils for now

src/shared/components/theme.tsx
dessalines: Gotcha.

dessalines:
This didn't work right away, I had to fix an issue where a logged-in user has the default `browser` theme, I got it tho. Very nice job.

Issue #79, opened 2020-11-17
Downvote icon still visible in mobile UI when downvoting disabled


m52go:
I have downvoting disabled, and the downvote arrow does not show in desktop mode. But it does show in mobile mode. Clicking it shows the downvote animation as well as error message that the downvote failed because the function is disabled.

The downvote shows in UI until the site is reloaded.

![Screenshot from 2020-11-17 13-56-53](https://user-images.githubusercontent.com/735155/99434592-e09f2200-2906-11eb-93e2-e13f24171efa.png)

dessalines:
Thanks for this. In the future

PR #80, opened 2020-11-17
Hide downvote on mobile when downvotes disabled

m52go:
To address #79.


dessalines:
Thanks for this. In the future pull down the project and run yarn so that a commit has the chance to run prettier, I had to format this.

m52go:
Noted, will do. Excellent work on this project. I'm really enjoying working with it :)

PR #156, opened 2021-01-30
Add aria attributes where possible

### LINKED ISSUE #33: Screen reader support / aria tags

MitchLillie:
Alright, so full disclosure here: I actually had the project up and running a few months ago, but recently I got one of the new M1 Macs, so I was not able to actually build this locally due to some architecture issues with local development and the fact that Docker isn't yet supported on M1. What I did do was pull up lemmy.ml, run a Google Lighthouse report on the various pages, and do as much as possible to fill in missing aria attributes. 

I might try to get the code running on AWS later this evening or tomorrow. But feel free to pull and verify locally if you have the build working.

Closes #33 kinda


dessalines:
Thanks! I'll look at this soon.

dessalines:
Could you unstage / remove lemmy-translations from this PR, it shouldn't be in here.

MitchLillie:
I can remove the lemmy-translations, but I did add a few translations there

Nutomic:
You will have to add translations directly in the https://github.com/LemmyNet/lemmy-translations repo.

src/shared/components/banner-icon-header.tsx
```diff
@@ -14,12 +14,13 @@ export class BannerIconHeader extends Component<BannerIconHeaderProps, any> {
...
     return (
       <div class="position-relative mb-2">
-        {this.props.banner && <PictrsImage src={this.props.banner} />}
+        {this.props.banner && <PictrsImage src={this.props.banner} alt="" />}
         {this.props.icon && (
           <PictrsImage
             src={this.props.icon}
             iconOverlay
             pushup={!!this.props.banner}
+            alt=""
```
dessalines, line 23: What's the preferred way to do this? Maybe just have the alt the same as the src? Or maybe use the post title?

src/shared/components/comment-node.tsx
```diff
@@ -667,9 +689,12 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
         {this.state.showBanDialog && (
           <form onSubmit={linkEvent(this, this.handleModBanBothSubmit)}>
             <div class="form-group row">
-              <label class="col-form-label">{i18n.t('reason')}</label>
+              <label class="col-form-label" htmlFor="mod-ban-reason">
+                {i18n.t('reason')}
+              </label>
               <input
                 type="text"
+                id="mod-ban-reason"
```
dessalines, line 697: These will probably need to use the random string generated above, because there can be many comment nodes and many ban forms open up in one page.

src/shared/components/login.tsx
```diff
@@ -326,6 +326,7 @@ export class Login extends Component<any, State> {
               class="rounded-top img-fluid"
               src={this.captchaPngSrc()}
               style="border-bottom-right-radius: 0; border-bottom-left-radius: 0;"
+              alt={i18n.t('captcha')}
             />
             {this.state.captcha.ok.wav && (
               <button
```
dessalines, line 332: I don't know much about aria, but why doesn't this button need an aria-label? Is it because it has a title?

src/shared/components/pictrs-image.tsx
```diff
@@ -71,4 +73,11 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
 
     return out;
   }
+
+  alt(): string {
+    if (this.props.icon) {
+      return '';
+    }
+    return this.props.alt || '';
+  }
```
dessalines, line 82: I'm not sure I understand this logic. If its an icon, it doesn't need an alt? I thought all pictures needed an alt.

src/shared/components/post-listing.tsx
```diff
@@ -169,6 +169,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
       <PictrsImage
         src={src}
         thumbnail
+        alt=""
```
dessalines, line 172: If its an optional prop, you don't need to include it here.

dessalines (changes requested):
One other thing: grepping the code here for `onClick` rather than for `<button` would probably give a better indication of where aria-label's are required. I unfortunately have a lot of `span`s acting as buttons, when I should change these to proper inline buttons. 

One other thing is lighthouse is telling me [links don't have discernable names](https://web.dev/link-name/?utm_source=lighthouse&utm_medium=devtools)? Not sure how to get around that, its an svg link:

image


Is there a guide somewhere so we can see what else besides aria labels we need to add?

src/shared/components/banner-icon-header.tsx
MitchLillie: The post title could work, but would be redundant: screen readers would just hear the same text repeated. The src definitely wouldn't as it takes a *long* time to read out a URL, and it usually doesn't describe the image in question.

The absolute best way to be accessible would be to have an option for users to submit alt text describing their images... but that would be a bit beyond the scope of this ticket.

Empty alt text is used for "decorative images", for example I used it in a few other places where an icon appears next to text. Since the icon is already described by the text it's considered purely decorative. In this case I used the empty alt text because the banner and icon didn't seem to add much to the page for vision impaired users, so this is a clear way of telling them "just skip this".

Further reading: https://www.w3.org/WAI/tutorials/images/decorative/

src/shared/components/comment-node.tsx
MitchLillie: Yep good call, I will add the id at the end of the string here

src/shared/components/login.tsx
MitchLillie: Yeah exactly. Title isn't the absolute best way (that would be just having text inside the button) but it definitely works

src/shared/components/pictrs-image.tsx
MitchLillie: Icons are decorative; see the novel I wrote in the other comment

src/shared/components/post-listing.tsx
MitchLillie: Will remove

MitchLillie:
@dessalines There are tons of guides on what can be done, but having done a few of these a11y on big for-profit projects, it's really an ongoing effort. Lighthouse provides a nice benchmark that could even be integrated into CI/CD pipelines to track it over time. Another more thorough resource is Axe by Deque, which is an extension only for Chrome. But if you look at the reports of both tools, there is a large "things to manually check" section. This PR is not going to be a "one-and-done" for a11y, it's more akin to performance improvements: there are always ways to improve, some of which make sense, some of which don't make sense, and some make sense but would be too much effort to be worth it.

MitchLillie:
> You will have to add translations directly in the https://github.com/LemmyNet/lemmy-translations repo.

OK, I opened an MR here, hope that's ok: https://github.com/LemmyNet/lemmy-translations/pull/6

MitchLillie:
OK, should be ready for another look.

I addressed the comments, and also did an audit of all of the `onClick` handlers, ensuring they had labels and `role="button"` if they were divs/spans

dessalines:
I'll get to this soon.

dessalines:
This is failing currently because of the merge conflict.

MitchLillie:
Should be all set, but let me know if there's anything else. I'm giving the Apple Silicon Docker Preview a shot today as well, hopefully can get local development working.

dessalines:
Sweet! Ya def let me know if there are any problems with it, esp since we do have docker images now for arm64 for lemmy and lemmy-ui.

src/shared/components/banner-icon-header.tsx
dessalines: Sweet. I'll re-open an old issue on lemmy for us to add an optional image_description as that is entirely possible for post images at least: https://github.com/LemmyNet/lemmy/issues/1086

dessalines:
Looks great! Thx a ton for this. As you said its an ongoing effort. Another TODO I have is that I need to change all these faked `<span>` buttons into real buttons, and make sure aria-labels are added for them.

PR #167, opened 2021-02-07
Make WS host configurable

revmischa:
I want to use a different domain for the WS backend from the frontend. The reason is that I want to put a CDN in front of the frontend, and those don't work well with WebSockets.

This adds an optional `LEMMY_WS_HOST` and sets up a simple way of passing configuration from the server to the browser page. I tried to clean up some of the logic regarding browser/not browser because it was a nest of ternaries.  

ref: https://github.com/jetbridge/lemmy-cdk/blob/master/lib/cdn.ts


dessalines:
How exactly does a CDN work with an isomorphic app, which needs to do a local fetch to get data?

Why does wsHost here need to be different from the regular host? The entire back end should be at the same host. 

What's going on with the window function here?

revmischa:
> How exactly does a CDN work with an isomorphic app, which needs to do a local fetch to get data?
By local fetch are you referring to the frontend speaking to the backend directly on the server side? That is just inter-container communication, the CDN isn't involved there.

> What's the purpose of a CDN here?
It will cache static files like the JS bundle and images from pictrs at edge locations, handles SSL termination, handle the reverse proxying, some other nice things. Mainly it's needed here to perform the same role nginx does, which is [rewriting](https://github.com/jetbridge/lemmy-cdk/blob/master/lib/edge-lambda/cdn-rewrite-paths.js#L4) the `/iframely/(.*)` URLs.

> Why does wsHost here need to be different from the regular host? The entire back end should be at the same host.
Because I want a CDN in front of all the requests the browser is making, except for the websocket requests, as I need to rewrite iframely URLs.
Anything requested goes to the CDN, which reverse proxies to either the iframely or lemmy load balancers. The Lemmy LB routes /api/* or /pictrs/* to the lemmy container otherwise sends everything else to the lemmy-ui containers.

> What's going on with the window function here?
I'm using it to pass configuration from server to browser. This is already being done for `window.isoData` the same way so I'm continuing that pattern. There is a module declaration to keep it typesafe.

dessalines:
So basically you're using a CDN to replace what nginx does? It already handles file and picture caching, as well as routing to iframely, and you'll need to run it no matter what for federation purposes. I spose I just don't understand the point of a CDN for smaller federated services. 

> Because I want a CDN in front of all the requests the browser is making, except for the websocket requests, as I need to rewrite iframely URLs.

I think I understand, but I'll make some comments on how this is implemented here.

src/shared/env.ts
```diff
@@ -2,30 +2,39 @@ import { isBrowser } from './utils';
...
+  host = externalHost;
+  wsHost = lemmyConfig.wsHost || host;
+  secure = window.location.protocol == 'https:' ? 's' : '';
+} else {
+  // server-side
+  externalHost = process.env.LEMMY_EXTERNAL_HOST || testHost;
+  host = internalHost;
+  wsHost = process.env.LEMMY_WS_HOST || host;
+  secure = process.env.LEMMY_HTTPS == 'true' ? 's' : '';
+}
```
dessalines, line 34: My main problem with this is that the actual host, both the http and websocket, are not the issue here. They both should not be pointing at anything but the real host. 

Your issue is that you want some of the static resources, and bundles, to use your own custom CDN. That should be done in the `server/index.tsx`. You've tried to get around this issue by making the CDN your "real host", and then adding a config to workaround that.

dessalines (changes requested):

src/shared/env.ts
revmischa: CloudFront (the CDN) is the "real host" because that's what the DNS that someone types into the browser should resolve to.
CloudFront does a fair bit of what one would do with nginx, except you don't have to run anything. It handles caching, reverse proxying, SSL termination, etc. It's a good practice to put it in front of everything, although it is no good for WS. 
I get your point that it should be used for just the static files, but it needs to be what the browser talks to. 

IMO the fewer containers and services I have to run and the more I can offload onto AWS the better. That's generally how this cloud stuff works; it's not just a datacenter it's a set of managed services you can configure and get some great benefits vs. managing your own services. 

I didn't get to federating yet so I'm unclear on why nginx is needed to federate.

dessalines: Nginx is 100% needed for federating. And cloudfront is not running the lemmy server.

revmischa: I'm okay with having my infrastructure on AWS. Maybe you are thinking of CloudFlare?

dessalines: Ah sry you're right lol. This is new territory because I have no idea how you'd write the nginx rerouting rules we have for AWS. But I think this should mostly be done through that. I

Lemmy's backend includes both an http and websocket API, and I don't think extricating them is the problem you're trying to solve. You're trying to figure out how to get static resources, served up by your CDN. Those are included in `server/index.tsx`.

revmischa: What I'm trying to do in this PR is simply make the websocket host optionally configurable. People may have good reasons to want to do this, such as separating out the services hosting the frontend and backend, or load balancing WS (network load balancers) differently from HTTP (application load balancers) for example. 

In my case I would rather do my rewrite rules with cloud-native infrastructure, because it's one less server for me to run, has caching at 225 POPs, handles SSL termination, supports HTTP/2, compression, IPv6, DNS alias integration, and a lot of other wonderful features.
I'm looking at https://raw.githubusercontent.com/LemmyNet/lemmy/main/ansible/templates/nginx.conf right now, is this the config that is needed to run lemmy and federation properly? I think these rules are quite translatable to Cloudfront. What I've been working on is trying to translate the configuration needed to run lemmy into a reusable IaC definition for AWS so anyone can deploy a highly available lemmy instance with the minimum amount of effort.

Nutomic: Nginx is definitely not required for federating, but you will have to translate the nginx config for your provider. Mainly the parts which go to the backend (everything with `{{ lemmy_port }}` in that template).

revmischa: Ok I've done that here https://github.com/jetbridge/lemmy-cdk/blob/master/lib/lemmy/loadbalancer.ts#L103

dessalines: I'll test this locally, but I have reservations about merging this. Its essentially supporting an amazon-specific limitation since they don't support websockets, and you have to offload that work to its own container, whilst amazon does everything else? Websockets is the main way lemmy-ui communicates, so it seems odd to have AWS handle everything BUT the main way lemmy communicates.

revmischa: There's nothing AWS-specific here, just making a hostname configurable. Other people may have other setups where they want different hostnames or ports for frontend/backend. The small refactoring and ability to pass through configuration from environment -> server.js -> browser will probably come in handy for future frontend configuration options as well.

src/shared/utils.ts
```diff
@@ -55,6 +55,16 @@ import {
...
+export interface ILemmyConfig {
+  wsHost?: string;
+}
+
+declare global {
+  interface Window {
+    lemmyConfig?: ILemmyConfig;
+  }
+}
+
```
dessalines, line 67: Move this to `interfaces.ts`, the other global is already defined there.

dessalines (changes requested):
Okay, I'll merge this after the other requested fix.

dessalines (approved):

PR #169, opened 2021-02-08
Comments and posts no longer live-sorted. Fixes #51

### LINKED ISSUE #51: [Bug?] Voting on comments causes them to jump around due to ranking.

dessalines:
- Only changing the view or sort, or reloading the page will resort.
- New comments by default added to the top, and left unsorted.

### DIFF
--- a/src/shared/components/comment-node.tsx
+++ b/src/shared/components/comment-node.tsx
@@ -15,15 +15,10 @@ import {
   AddAdmin,
   TransferCommunity,
   TransferSite,
-  SortType,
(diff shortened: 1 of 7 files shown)


dessalines:
This was very usable, when not resorting comments and unshifting them to the top, firefox handles the scroll brilliantly. If you are looking at an area that doesn't have new content, it won't change your view at all, you'll simply see your scrollbar grow. When looking at an area of the tree where new data is pushed, it will alter that view only.

PR #213, opened 2021-03-01
Feature/i18n type constraints

shilangyu:
This makes the `i18n` keys strongly typed. Results in static checks of keys, and nice autocomplete of the available keys:

image

Running `yarn lint` found a key [`"disclaimer"`](https://github.com/LemmyNet/lemmy-ui/blob/1916483ce9eaabdc869fac424d13b5b4f1179e63/src/shared/components/private-message-form.tsx#L122) being used that is not present in the translations (this might cause the CI to fail). All other strings are correct.

Technically speaking this just generates a union type of all i18n keys and constraining i18n.t() to those keys.

### DIFF
--- a/generate_translations.js
+++ b/generate_translations.js
@@ -1,27 +1,64 @@
-fs = require('fs');
+const fs = require("fs");
 
-let translationDir = 'lemmy-translations/translations/';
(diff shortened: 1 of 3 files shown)


dessalines:
Works, great, thx!

PR #217, opened 2021-03-08
Add translators' credits

skrlet13:
I made a .json file with translators and their links (if they had). They are grouped by language.

### DIFF
--- a/src/assets/jtrans.json
+++ b/src/assets/jtrans.json
@@ -1,0 +1,167 @@
+{
+    "ar": [
+        {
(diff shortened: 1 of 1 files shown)


dessalines:
Sweet thx! Could you redo this PR to https://github.com/LemmyNet/lemmy-translations , so that other repos can use these credits? It can be in the top level, and you can call it `translators.json`

dessalines:
Oh also, you can remove the empty link fields, it can be an optional field.

PR #221, opened 2021-03-15
user_ -> person table migration.

dessalines:


src/server/index.tsx
```diff
@@ -48,7 +48,18 @@ server.get("/*", async (req, res) => {
...
+  let try_site: any = await initialFetchReq.client.getSite(getSiteForm);
+  if (try_site.error == "not_logged_in") {
+    console.error(
+      "Incorrect JWT token, skipping auth so frontend can remove jwt cookie"
+    );
+    delete getSiteForm.auth;
+    delete initialFetchReq.auth;
+    try_site = await initialFetchReq.client.getSite(getSiteForm);
+  }
+  let site: GetSiteResponse = try_site;
```
Nutomic, line 62: Ah so this is to handle the invalidated jwts from the backend change :+1:

dessalines: Yes... I ran into this while using incorrect JWT tokens on the federation test instances. Its necessary because the server can't reset your cookies, only the browser can, and this at least lets the server-side do an initial user-less fetch without errors, so that the front-end can run into one, and reset the cookies.

Nutomic: Might be worth mentioning this in the API docs.

PR #229, opened 2021-03-26
Set listen host and port with environment variables

stemid:
At the moment the process listens on localhost, hard coded. This makes it hard for me to deploy in Kubernetes.

So I would very much appreciate if the server could take listen host and port from environment.


dessalines:
Seems fine to me, thanks!

Issue #265, opened 2021-04-16
Secondary languages not showing


dessalines:
If the language list has several languages, but the top one isn't supported, it will default to english.

Nutomic:
Even the primary language is ignored in some cases. For example when I set my browser language to Korean, it sends the header `Accept-Language: ko,es-ES;q=0.5`, but Lemmy is shown in English. Tested while being logged out, so it cant be related to user settings.

Nutomic:
Ah its because Korean localization isnt enabled, even though its 100% finished. Wouldnt it be possible to use all translations automatically if they exist, instead of having to enable them manually?

dessalines:
Unfortunately not, they do need manual imports. I'll do that now for ko.

Nutomic:
Guess I am too spoiled from Android development, where the SDK takes care of that and you barely ever have to think about localization.

PR #267, opened 2021-04-21
Add check for unused languages in update_translations.sh

Nutomic:
This fixes the problem I mentioned in https://github.com/LemmyNet/lemmy-ui/issues/265#issuecomment-823419697

### DIFF
--- a/update_translations.sh
+++ b/update_translations.sh
@@ -1,9 +1,18 @@
 #!/bin/bash
-pushd ../lemmy-translations
+pushd ../lemmy-translations || exit
 git fetch weblate
 git merge weblate/main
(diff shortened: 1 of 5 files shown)


update_translations.sh
```diff
@@ -1,9 +1,18 @@
 #!/bin/bash
-pushd ../lemmy-translations
+pushd ../lemmy-translations || exit
```
Nutomic, line 2: Due to this lint: https://github.com/koalaman/shellcheck/wiki/SC2164

dessalines:
Whoa nice, much better. I'll get this conflict.

PR #270, opened 2021-04-23
Change from using Link to NavLink. resolve #269

### LINKED ISSUE #269: Using NavLink instead of Link in navbar

It's related to navbar active page, in `src/shared/components/navbar.tsx`

https://github.com/LemmyNet/lemmy-ui/blob/1bd2473cdc65357c18a45c8f1afb532c898b66f6/src/shared/components/navbar.tsx#L175-L195

(Write your answer here.)

we should use `NavLink` instead of `Link` from inferno-router

the `NavLink` is a `Link` wrapper that knows if it's active or not

hermantolim:
there is some issue with `title` attribute. So it will not pass the lint, but everything is working fine. Since `NavLink` only accepted `NavLinkProps`, this is issue with the `inferno-router` itself. .

### DIFF
--- a/src/shared/components/footer.tsx
+++ b/src/shared/components/footer.tsx
@@ -1,5 +1,5 @@
 import { Component } from "inferno";
-import { Link } from "inferno-router";
+import { NavLink } from "inferno-router";
 import { i18n } from "../i18next";
 import { repoUrl, joinLemmyUrl, docsUrl } from "../utils";
 import { GetSiteResponse } from "lemmy-js-client";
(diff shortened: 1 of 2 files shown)


dessalines:
It does need to pass lint. If inferno is missing that field, you'll need to open up an issue there and reference this one.

hermantolim:
It will be fixed in the next inferno release [#1563](https://github.com/infernojs/inferno/issues/1563)

dessalines:
Thanks!

PR #278, opened 2021-04-25
Add UI version to UI via docker. Fixes #263

### LINKED ISSUE #263: Add UI version info to UI
Right now, we're only showing the backend version in the UI, we should show the front end too.

https://github.com/LemmyNet/lemmy/pull/1567

dessalines:


Nutomic:
You need to run `git fetch --tags` at the start of CI, otherwise this is not going to work there.

https://github.com/LemmyNet/lemmy/pull/1590

Nutomic (approved):

Issue #164, opened 2021-02-03
Clicking the "x comments" link on a post should scroll down to the comments


Nutomic:
This would be especially useful when you want to read new comments on a long text post. At the moment you have to scroll down a lot just to reach the first comment.

PR #312, opened 2021-06-26
Scroll to comments on post's x comments button

hammsvietro:
This adds the enhancement mentioned in #164

### DIFF
--- a/src/shared/components/post-listing.tsx
+++ b/src/shared/components/post-listing.tsx
@@ -1,4 +1,4 @@
-import { Component, linkEvent } from "inferno";
+import { Component, linkEvent, RefObject } from "inferno";
 import { Link } from "inferno-router";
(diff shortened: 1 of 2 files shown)


dessalines:
Finally looking at this, apologies. 

Its close, but the main issue is that you want it to skip to the comment section from the front page also. Since its usually from a different route, the only way I can think to do that, is to use a URL param, which is how we do it for comment scrolling, which has an example in `post.tsx`.

hammsvietro:
Hi, I've been kinda busy, I changed as you suggested, I'm not used with the live view/inferno structure, but I think is working now, inside and outside the post component :D

dessalines (approved):
Just tested and it works great, nice job!

hammsvietro:
thanks! always great to help this awesome project :)

PR #315, opened 2021-07-07
Fix the thumbnail size

Mehran-Baghi:
This commit removes the height restriction and makes
thumbnails fill up their designated place.

lemmy


Mehran-Baghi:
I've just realized that this issue only occurs in Firefox.
So I've reported #316 and I'm closing this pr.

PR #362, opened 2021-08-10
Feature/user community block

dessalines:


dessalines:
Here's what this looks like currently:

image

Nutomic:
Besides the dropdown looking weird (I would expect a simple text input), avatars also mess up the layout. Maybe leave an empty margin to the left if a user/community has no icon, so that the names line up (makes it easier to read). I would also make the `x` buttons bigger, and have them line up in the same column.

dessalines:
> Besides the dropdown looking weird (I would expect a simple text input),

The dropdown thing I can't mess with the styling too much, because its a component called choices that has its own complications.

I tried right aligning the user icons and X button into a table, and it didn't look nearly as clean (and tables aren't as mobile friendly ) . I could add some extra spacing between the user link and X tho.

Nutomic:
You could also try putting the X button on the left.

dessalines:
That just looks confusing.

PR #401, opened 2021-09-05
Add robots.txt file

mahanstreamer:
Reasoning for patch:

Some search engines and crawler bots rely on the robots.txt file to know what parts of the site to crawl or not crawl. Without one, it will proceed to crawl the entire site by deafult. However, it doesn't make sense for bots to crawl parts that require a login or are user directed , like "create_community" or "login". Some poorly programmed bots will constantly hit an endpoint that doesn't make sense, causing general slowness for everyone.

Suggestions for other protected endpoints welcome.


dessalines:
Looks good! I fixed lint and added a few more endpoints. Thanks!

Issue #396, opened 2021-09-03
Add community and user blocking from those profile pages.


dessalines:
IE, have a block user button on their profile page, a block community button on the community pages.

mahanstreamer:
Will do this if no one else will now that the youtube issue has been fixed already

dessalines:
Sure! Let me know if you need any assistance.

dessalines:
BTW our dev chat is: https://matrix.to/#/#lemmydev:matrix.org

Nutomic:
This is half implemented, there is a block button on user profiles. But community profiles dont have any block button yet, so that should be added.

Nutomic:
Implemented now.

PR #408, opened 2021-09-07
Added user page block functionality

mahanstreamer:
This adds support for blocking on user page functionality. I will submit community page blocking functionality in a separate PR, where then feature request #396 would be fully complete.

Comments on code style, code methods, etc.. is appreciated as I am still new to the project.

### DIFF
--- a/src/shared/components/person/profile.tsx
+++ b/src/shared/components/person/profile.tsx
@@ -3,6 +3,7 @@ import { Link } from "inferno-router";
 import {
   AddAdminResponse,
   BanPersonResponse,
+  BlockPerson,
   BlockPersonResponse,
   CommentResponse,
   GetPersonDetails,
@@ -10,6 +11,7 @@ import {
   GetSiteResponse,
   PostResponse,
   SortType,
+  PersonBlockView,
   UserOperation,
 } from "lemmy-js-client";
 import moment from "moment";
@@ -57,6 +59,7 @@ interface ProfileState {
   sort: SortType;
   page: number;
   loading: boolean;
+  personBlocks: PersonBlockView[];
   siteRes: GetSiteResponse;
 }
 
@@ -84,6 +87,7 @@ export class Profile extends Component<any, ProfileState> {
     view: Profile.getViewFromProps(this.props.match.view),
     sort: Profile.getSortTypeFromProps(this.props.match.sort),
     page: Profile.getPageFromProps(this.props.match.page),
+    personBlocks: UserService.Instance.myUserInfo.person_blocks,
     siteRes: this.isoData.site_res,
   };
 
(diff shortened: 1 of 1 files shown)


src/shared/components/person/profile.tsx
```diff
@@ -57,6 +59,7 @@ interface ProfileState {
   sort: SortType;
   page: number;
   loading: boolean;
+  personBlocks: PersonBlockView[];
```
dessalines, line 62: The entire array of blocks isn't really needed here. All that's needed is either a boolean for `personBlocked`, or a single `personBlockView`

src/shared/components/person/profile.tsx
```diff
@@ -125,6 +129,15 @@ export class Profile extends Component<any, ProfileState> {
...
     );
   }
+  isBlocked() {
+    // this.state.personRes?.person_view.person.name
+    for (var k of this.state.personBlocks) {
+      if (k.target.id == this.state.personRes.person_view.person.id) {
+        return true;
+      }
+    }
+    return false
```
dessalines, line 139: Use `.map` and `includes`, no need to do a loop here.

src/shared/components/person/profile.tsx
```diff
@@ -125,6 +129,15 @@ export class Profile extends Component<any, ProfileState> {
       this.state.personRes.person_view.person.id
     );
   }
+  isBlocked() {
```
dessalines, line 132: Use a `get`, then you don't have to call `isBlocked()`, you can just use `isBlocked`

src/shared/components/person/profile.tsx
```diff
@@ -383,11 +414,31 @@ export class Profile extends Component<any, ProfileState> {
                     {i18n.t("send_secure_message")}
                   </a>
                   <Link
-                    className={"d-flex align-self-start btn btn-secondary"}
+                    className={"d-flex align-self-start btn btn-secondary mr-2"}
                     to={`/create_private_message/recipient/${pv.person.id}`}
                   >
                     {i18n.t("send_message")}
                   </Link>
+                  {!this.isBlocked() && (
```
dessalines, line 422: Use a ternary like the other examples in this codebase. `this.isBlocked ? () : ()`

src/shared/components/person/profile.tsx
```diff
@@ -383,11 +414,31 @@ export class Profile extends Component<any, ProfileState> {
...
+                  <a
+                    id="blockbtn"
```
dessalines, line 424: id isn't required here. Use the other examples on this page.

dessalines (changes requested):

mahanstreamer:
Will push updated version shortly

src/shared/components/person/profile.tsx
```diff
@@ -127,6 +130,11 @@ export class Profile extends Component<any, ProfileState> {
       this.state.personRes.person_view.person.id
     );
   }
+  get isBlocked() {
+    this.state.personBlocks = UserService.Instance.myUserInfo.person_blocks.
+    map(a => a.target.id).includes(this.state.personRes.person_view.person.id) 
+    return this.state.personBlocks
+  }
```
dessalines, line 137: You don't want to set that state here, do it below on line 631:

```ts
let blocks = updatePersonBlock(data);
this.state.personBlocks = blocks.map(a => a.target.id).includes(this.state.personRes.person_view.person.id);
this.setState(this.state);
```
Then you don't need an `isBlocked()` function, just `this.state.personBlocks` should work.

Also you should probably rename that to `personBlocked` to be more clear.

dessalines (changes requested):

mahanstreamer:
Will push updated version shortly

PR #414, opened 2021-09-17
fix: bug in profile that broke pagiantion

LunaticHacker:
pagination was broken because handlePageChange wasn't bound to Profile


dessalines:
Thanks for pointing this out, I've added the bind in another spot!

PR #434, opened 2021-09-28
Reporting

dessalines:


Nutomic:
Great! Lets make a new release then, and I will start working on federation for reports.

dessalines:
Cool, I'll start doing a `v0.13.0-rc.1` release, so we can test it out on the federation servers.

PR #447, opened 2021-10-03
Hide mod actions appropriately

### LINKED ISSUE #441: Post actions that need reasons (remove, ban, report, etc) weirdly stack their text fields
### Issue Summary

Post actions that require more data, such as reasons, stack their input fields in a weird way; not aligned and not in the order they were opened.

Furthermore, I think that it would be much better to not even allow multiple of these fields at the same time, as it can create confusion. Just hide whatever field is there already when another action is triggered.

### Steps to Reproduce

1. Open a post on a community you moderate
2. Click the "remove", "ban" and "report".

### Technical details
N/A

### Screenshots

imagem

LunaticHacker:


LunaticHacker:
the tests kept running forever. closing and reopening to see if it fixes that

dessalines:
Thx!

PR #482, opened 2021-11-12
Remove touch events

dayinjing:
Remove touch events, it breaks clicking links on mobile. Tested this on a iphone 7


dessalines:
You missed one, but other than that looks good. Thx!

PR #510, opened 2021-11-25
Fix docker latest

dessalines:


.drone.yml
```diff
@@ -63,6 +63,23 @@ steps:
...
       - refs/tags/*
 
+  - name: publish latest release docker manifest
+    image: plugins/manifest
+    settings: 
+      username:
+        from_secret: docker_username
+      password:
+        from_secret: docker_password
+      target: "dessalines/lemmy-ui:latest"
```
Nutomic, line 73: Instead of target, we should be able to use the following params, to avoid copying the whole block (and possibly even building twice).

```
    repo: dessalines/lemmy-ui
    tags:
      - latest
      - "${DRONE_TAG}"
```

https://docs.drone.io/plugins/popular/docker/

dessalines: I tried that, it didn't work.

Nutomic: We arent using auto_tag though as far as i can see. Anyway, at least it seems to cache the build result.

PR #523, opened 2021-12-09
Private instances

dessalines:


Nutomic:
You should use another icon for registration applications, at the moment its the same one as editing a post/comment, which is confusing.

It would also be nice if we could add some extra description for the different site configuration options.

dessalines:
> You should use another icon for registration applications, at the moment its the same one as editing a post/comment, which is confusing

There aren't too many icons to choose from, but I'll see if I can find something else.

> Also it shouldnt be possible to change the deny reason later.

This one I think it should, because applications can be approved, then denied, and back and forth by different admins. So a discussion could develop that ends up with a different deny reason than initially.

> And it would be nicer if the "deny reason" input field was only shown after clicking the "deny" button.

I'll try to refactor this so it looks prettier.

> It would also be nice if we could add some extra description for the different site configuration options.

The descriptions we can add later once ppl have some more clarifying questions about what they do... they all seem self-expalanatory to me at the moment.

dessalines:
> Still not great. Maybe it would be better to have a single link for "administration", which has tabs or something similar for reports, registration applications, instance config etc. I dont think its necessary to access this stuff from every page.

One issue here is that applications are only for admins to see, while reports are for both admins and mods. So I'd rather have them be at different routes. If we have more "only admin" notifs, or "both admin/mod" notifs, then we can think of better ways to combine them later.

> Its a really bad format for discussion...

I agree, that isn't the place to do discussion. However after discussion is done, it should be possible to reverse the decision on applications. There's no reason to take away the ability to deny, then approve an application. We wouldn't help anyone by making application decisions irreversible.

> Now it seems you have to press deny, then enter the deny reason, and press deny again, not intuitive. 

I liked it the first way initially, but you asked me to change it to this. I can either always show the deny_reason textarea, or have you reveal that text box with another button. I can't do both.

> Also the approve/deny buttons should probably be hidden at least on accepted applications.

I already did this for the deny button, but can probably do this for the approve button as well, for ones that are already approved.

Nutomic:
> I liked it the first way initially, but you asked me to change it to this. I can either always show the deny_reason textarea, or have you reveal that text box with another button. I can't do both.

It should be a separate button for the second action. Possibly show the text input + button in a popup window.

dessalines:
> True, but its not necessary to change the deny reason in that case (or in any other case that i can think of).

I changed it so you can't deny twice.

> It should be a separate button for the second action. Possibly show the text input + button in a popup window.

I can't do a popup modal, because that'd be the first we'd have. This is actually the first case of hiding a box, all the others show the `Reason` textarea. So if you want me to use a different button than deny to show the deny reason textarea, you need to be more specific about where that would go, and what it would say.

Nutomic:
How about the same kind of input thats used to give a ban reason?

dessalines:
The ban reason is always shown: 

IMG_20211217_121352

Which is how I first did this.

Nutomic:
I mean it has one button which opens the text input + ban button. In comparison, your first version iirc you had to press reject which opened the text input, and then had to press the same reject button again. So its a matter of having different buttons for different actions.

dessalines:
We can always tweak this later. It is working as is.

PR #535, opened 2021-12-24
Add some fallback properties for display in older browsers

marek-lach:


src/assets/css/main.css
```diff
@@ -112,6 +112,7 @@
 }
 
 .icon {
+  display: inline-grid;
   display: inline-flex;
```
dessalines, line 116: This doesn't seem like valid css, could you do `display: inline-flex inline-grid`.

And what browser doesn't support flex?

marek-lach: > This doesn't seem like valid css,

Apparently [it is](https://caniuse.com/?search=inline-grid#:~:text=This%20feature%20is%20deprecated%2Fobsolete%20and%20should%20not%20be%20used.)...?

> could you do display: inline-flex inline-grid.

Well, when doing fallback code, apparently the fallback should be put first, than the current code under it...

> And what browser doesn't support flex?

This is a very minor use-case yes, but the one SailfishOS uses as it's default.

PR #554, opened 2022-01-26
Move lemmy.ml signup message to top of registration page

Nutomic:
Move this to the top so that people see it before filling in all the information. 

I also noticed that this message is supposed to be highlighted (with classes `alert alert-light`), but in the default theme its just gray text on normal white background. It should be changed to be easily visible, with yellow background for example. Dont know how to do that though.


dessalines:
Just change it from `alert-light` to `alert-warning`

Nutomic:
Updated, much better now.

Screenshot_20220127_173410

PR #575, opened 2022-02-14
Add federated post and comment links. Fixes #569

### LINKED ISSUE #569: Show Activitypub ID for posts and comments
For fetching a post/comment from another instance, it is necessary to search for its activitypub id. This works by copying the url from the browser if its a local item, but the same will not work if its a remote post or comment. In that case, the only way to get the apub id is via browser console or api. 

For example, searching `https://feddit.de/post/83962` will fail, and it is necessary to search `https://fapsi.be/post/36912` instead (see https://github.com/LemmyNet/lemmy/issues/2090).

So there should be a link to expose the activitypub id of each post and comment, maybe using the [fediverse icon](https://en.wikipedia.org/wiki/File:Fediverse_logo_proposal.svg).

dessalines:


src/shared/components/comment/comment-node.tsx
```diff
@@ -849,14 +850,30 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
-      </Link>
+      <>
+        <Link
+          className={classnames}
+          to={`/post/${cv.post.id}/comment/${cv.comment.id}`}
+          title={title}
+        >
+          <Icon icon="link" classes="icon-inline" />
+        </Link>
+        {/* TODO comment ap_ids are currently broken anyway, so use post.ap_id, and wait until comment tree / endpoint refactor */}
```
Nutomic, line 870: This will be pretty confusing. And ap_ids arent broken at all, there is just no frontend route for them. So it should just show the comment ap_id, as thats what you need to search on another platform.

src/shared/components/comment/comment-node.tsx
```diff
@@ -849,14 +850,30 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
+        {!cv.comment.local && (
+          <a className={classnames} title={title} href={cv.post.ap_id}>
+            <Icon icon="fedilink" classes="icon-inline" />
+          </a>
```
Nutomic, line 874: It should probably be shown for local comments/posts as well, otherwise it will also be confusing that some dont have it.

Nutomic: In fact, i dont think it makes sense to show separate icons for link and activitypub link. If you want to share a specific post/comment with someone, it should be using the link where it was originally posted, not on the instance where you happen to browse. So it should show the ap_id in all cases (not sure which icon though). 

That means we definitely have to implement a view under `/post/123/comment/456`, but that should be really easy if its simply a redirect to `/comment/456`.

src/shared/components/comment/comment-node.tsx
dessalines: The comment ap_ids are broken. This goes nowhere currently: https://lemmy.ml/comment/93927

That's waiting on a proper comment endpoint which we should probably hold off on until I start tree-views.

src/shared/components/comment/comment-node.tsx
dessalines: > It should probably be shown for local comments/posts as well, otherwise it will also be confusing that some dont have it.

In that case it'd just be the same link twice, which would be pointless. The clutter will start to pile up on mobile especially. 

> If you want to share a specific post/comment with someone, it should be using the link where it was originally posted, not on the instance where you happen to browse. So it should show the ap_id in all cases (not sure which icon though).

Not sure abt this one. I could see some cases where you'd want to link a specific instance's cached version of that post, esp if one instance goes down, gets blocked, or you're linking to someone you know doesn't have an account on the original. But I do agree in most cases the federated one makes sense. Maybe for federated / non-local content, just don't show the original link?

> That means we definitely have to implement a view under /post/123/comment/456, but that should be really easy if its simply an alias for /comment/456

The actual comment `ap_id` is just `instance/comment/number`, which doesn't include the post. The `/post/number/comment/number` is a front end scrolling hack that I'm starting to regret adding, because its really not the right way to refer to comments. I'd rather this just wait on the proper comment endpoint than continue working building on the hack.

Issue #578, opened 2022-02-14
Improve style of community list


fadelkon:
### Is your proposal related to a problem?

I get confused by the way that long community names overflow their line. Sometimes the chunks of the same title, look like two different communities.

![lemmy-community-list-packed](https://user-images.githubusercontent.com/20354477/153943629-476f37bd-c597-4490-913b-9bfb37ccca0f.png)

Also, there's not a pixel of vertical space between community avatars, and this is uncomfortable to my eye.

And finally, when a community does not have an avatar, the text gets aligned to the left, and it doesn't get any vertical space either.

![lemmy-community-list-no-avatar](https://user-images.githubusercontent.com/20354477/153944554-d190974b-b42c-48e2-9c35-5806f89c7a71.png)

### Describe the solution you'd like

I know enough css to fix it, by I don't know anything about Inferno or this project's structure. If somebody hints me where to write, I can  try to write a simple PR to fix that.

![lemmy-community-list-spaced](https://user-images.githubusercontent.com/20354477/153943903-e8606e60-c68f-45f9-8d68-809baf078591.png)

I didn't do the browser test to fix the missing avatar issue, but I think it would be as easy as showing  a void circle, with some theme default color.

Otherwise, one could spare the space that the avatar takes and don't use it, but at least community titles would be all left aligned to the same vertical imaginary guide.

### Drawbacks

My proposed solution would take more vertical space. Though I firmly think that it should. Or else avatars should be smaller.

dessalines:
Sure, here's the relevant HTML: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/home/home.tsx#L400

I'm not a fan of generated avatars for the most part, but I spose I'd be fine with them here, as long as its not obtrusive. The `CommunityLink` component is where you'd add that.

dessalines:
Dupe of #546

PR #579, opened 2022-02-14
Improve community lists style 

fadelkon:
fixes https://github.com/LemmyNet/lemmy-ui/issues/578

I'm editing in the browser, so this is not tested at all yet


fadelkon:
How can I test it?

In a context with the other 4 services up, I would do `docker build .` and then... `docker run SOMENAME` (?)

dessalines:
https://join-lemmy.org/docs/en/contributing/local_development.html#run-a-local-development-instance

fadelkon:
I achieved what I wanted, but it was difficult for me to contain a chain of unwanted changes, but I'm happy i did it.

I tried directly the changes on https://info.prou.be and they are online already for you to check. I loaded the dist/ folder of my local setup into the lemmy-ui internal filesystem.

However, there is one major issue that I need help with because I can't find how I broke it: the  navbar's user dropdown list, and the "show more" button of the post listings bottom toolbox, both made with tippy.js , stopped working, they don't unfold when clicked.

Also translations are not shown now, I suspect that both issues could be related with my lousy testing strategy.

There's also one minor style issue: the post listing's bottom toolbox text are now in a slight vertical unalignment. They are supposed to align on baseline, but it's not pixel perfect now.

I attach edited before/after screenshots in the next comment

fadelkon:
In pink, what is working. In yellow, what it's not.

The contribution:

1. Communities without a logo, now show a semitransparent grey logo as a placeholder. Actually it's a simple div with background color.
2. Communities with names so long that become multiline, align naturally with their logo, centering vertically the logo with all the name lines.
3. There's some vertical space now for community logos to breath
4. And non visual change: post listing toolbox is now a flex container, and I split some elements with their own `<li>` so that they can align mutually. This also gives us more control on line overflow and how much space each element can take.

## Before
before

## After
after

fadelkon:
Change number 4. was necessary because with the changes I did to "CommunityLink" to be displayed better on the vertical lists, was doing line breaks in this horizontal list.

I made up the term "toolbox". Maybe it's more of an info-line? anyway. It's the underlined div in the central page.

Nutomic:
In general the changes look great, but the grey community icon placeholder looks quite ugly. Maybe you could use a placeholder icon similar to [this one](https://www.roganscale.com/wp-content/uploads/placeholder.jpg), though i'm not sure what exactly the icon should be. Or you could use the site icon as placeholder, but that wont work if its missing as well, and might be weird for remote communities.

src/assets/css/main.css
```diff
@@ -213,6 +213,26 @@ hr {
...
+
+.list-flex-v .avatar-name-container {
+    margin-top: 0.5rem;
+}
+
+.list-flex-h .avatar-name-container {
+    margin-left: 0.5rem;
+    margin-right: 0.5rem;
+}
+
```
dessalines, line 231: These custom flex classes shouldn't be necessary: use bootstraps flex utilities: https://getbootstrap.com/docs/4.0/utilities/flex/

src/shared/components/common/pictrs-image.tsx
```diff
@@ -13,38 +13,47 @@ interface PictrsImageProps {
   nsfw?: boolean;
   iconOverlay?: boolean;
   pushup?: boolean;
+  placeholder?: boolean;
```
dessalines, line 16: This probably shouldn't be here, but in community-link, person-link, etc. It can then use the placeholder source picture, only if there is no icon given.

src/shared/components/post/post-listing.tsx
```diff
@@ -321,13 +321,17 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
-              <CommunityLink community={post_view.community} />
-            </span>
-          )}
         </li>
+        {this.props.showCommunity && (
+          <li className="list-inline-item">
+            <span class="mx-1"> {i18n.t("to")} </span>
+          </li>
+        )}
+        {this.props.showCommunity && (
```
dessalines, line 330: Doesn't need to do this same check twice.

dessalines (changes requested):

src/shared/components/post/post-listing.tsx
fadelkon: I agree. However, the template parser complained if I did something like
```
{this.props.showCommunity && (
  <li>...</li>
  <li>...</li>
)}
```
It said it only accepted one tag as result. Would something like...
```
{this.props.showCommunity && (
  <>
  <li>...</li>
  <li>...</li>
  </>
)}
```
... produce the same output as my repeated check strategy?

I'm new to these frameworks, sorry :/

dessalines: Yep that works.

fadelkon:
> In general the changes look great, but the grey community icon placeholder looks quite ugly. Maybe you could use a placeholder icon similar to [this one](https://www.roganscale.com/wp-content/uploads/placeholder.jpg), though i'm not sure what exactly the icon should be. Or you could use the site icon as placeholder, but that wont work if its missing as well, and might be weird for remote communities.

Usually placeholder images say a lot of the software or platforms, and thus I didn't want to propose anything in particular. Also I'm not a designer. Aseptic pictures also mean things; in facebook it looks like an id card, and that's what facebook is about. I could use something like this [3 people is a community](https://forkaweso.me/Fork-Awesome/icon/users/), but it's boring.

At most, what about lemmy logo itself?

Something I can try is using the community icon with lowered contrast and higher brightness. Some kind of watermark effect. But may not work with all icons, for instance, what about fair icons with lot of transparent?

Nutomic:
I'm also not sure whats the best option for a placeholder. I suppose it can be done later, you could open an issue to present some options and discuss them.

dessalines:
@fadelkon any updates?

fadelkon:
I'm haven't worked on it again yet. When I can slot some hours on it, I 
will let you know the results. I'm aware of this, just it's been not the 
time yet. Is it ok for you?
I see there are some conflicts now, but I'll rebase the changes taking 
care not to destroy anything.

El 15/4/22 a les 23:40, Dessalines ha escrit:

PR #580, opened 2022-02-15
Fix language names

Nutomic:
As described in https://github.com/LemmyNet/joinlemmy-site/pull/76#issuecomment-1040200021

There is still a problem in that all language names are rendered left to right, even those which should be rtl (like Arabic).

### DIFF
--- a/src/shared/components/person/settings.tsx
+++ b/src/shared/components/person/settings.tsx
@@ -18,7 +18,7 @@ import {
   UserOperation,
 } from "lemmy-js-client";
 import { Subscription } from "rxjs";
-import { i18n } from "../../i18next";
+import { i18n, languages } from "../../i18next";
 import { UserService, WebSocketService } from "../../services";
 import {
   authField,
@@ -30,10 +30,8 @@ import {
   elementUrl,
   fetchCommunities,
   fetchUsers,
-  getLanguage,
-  getNativeLanguageName,
+  getLanguages,
   isBrowser,
-  languages,
   personSelectName,
   personToChoice,
   setIsoData,
@@ -527,7 +525,7 @@ export class Settings extends Component<any, SettingsState> {
                   .sort((a, b) => a.code.localeCompare(b.code))
                   .map(lang => (
                     <option value={lang.code}>
-                      {getNativeLanguageName(lang.code)}
+                      {getLanguages(lang.code)[0]}
                     </option>
                   ))}
               </select>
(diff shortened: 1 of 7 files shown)


src/shared/utils.ts
```diff
@@ -450,119 +346,21 @@ export function getLanguage(override?: string): string {
...
   let langs = languages ? languages.map(l => l.code) : ["en"];
 
   // NOTE, mobile browsers seem to be missing this list, so append en
   let allowedLangs = navigator.languages
     .concat("en")
     .filter(v => langs.includes(v));
-  return allowedLangs[0];
-}
-
-export function getMomentLanguage(): string {
```
Nutomic, line 470: This function is not necessary at all, as momentjs does it much smarter on its own.

https://momentjs.com/docs/#/i18n/changing-locale/

src/shared/utils.ts
```diff
@@ -450,119 +346,21 @@ export function getLanguage(override?: string): string {
       : "browser");
 
   if (lang == "browser" && isBrowser()) {
-    return getBrowserLanguage();
+    return getBrowserLanguages();
   } else {
-    return lang;
+    return [lang];
   }
 }
```
Nutomic, line 353: Makes more sense to return a list of preferred languages here, momentjs uses this. Unfortunately i18next doesnt, it can only set a single language. So if a word isnt translated into the browser language, it will always fallback to English, even if the user specified another language as second priority.

src/shared/i18next.ts
```diff
@@ -18,88 +15,57 @@ import { fi } from "./translations/fi";
...
+  { resource: nl, code: "nl", name: "Nederlands" },
+  { resource: pl, code: "pl", name: "Polski" },
+  { resource: pt, code: "pt", name: "Português" },
+  { resource: pt_BR, code: "pt_BR", name: "Português (Brasil)" },
+  { resource: ru, code: "ru", name: "Русский" },
+  { resource: sv, code: "sv", name: "Svenska" },
+  { resource: vi, code: "vi", name: "Tiếng Việt" },
+  { resource: zh, code: "zh", name: "中文" },
+  { resource: zh_Hant, code: "zh_Hant", name: "文言" },
+];
```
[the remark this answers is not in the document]
dessalines: This is much better.

src/shared/components/person/settings.tsx
```diff
@@ -527,7 +525,7 @@ export class Settings extends Component<any, SettingsState> {
                   .sort((a, b) => a.code.localeCompare(b.code))
                   .map(lang => (
                     <option value={lang.code}>
-                      {getNativeLanguageName(lang.code)}
+                      {getLanguages(lang.code)[0]}
```
dessalines, line 528: Use the language name here, and I'm not sure why this needed to call getLanguages again.

dessalines (changes requested):

src/shared/components/person/settings.tsx
Nutomic: True, i changed this before merging the language lists.

dessalines (approved):

PR #584, opened 2022-02-18
Custom themes

Nutomic:
Allow loading additional theme from a folder on the server (default `./extra_themes`, configurable via env var `LEMMY_UI_EXTRA_THEMES_FOLDER`.

### DIFF
--- a/src/shared/components/home/site-form.tsx
+++ b/src/shared/components/home/site-form.tsx
@@ -6,7 +6,7 @@ import { WebSocketService } from "../../services";
 import {
   authField,
   capitalizeFirstLetter,
-  themes,
+  fetchThemeList,
   wsClient,
 } from "../../utils";
 import { Spinner } from "../common/icon";
@@ -21,6 +21,7 @@ interface SiteFormProps {
 interface SiteFormState {
   siteForm: EditSite;
   loading: boolean;
+  themeList: string[];
 }
 
 export class SiteForm extends Component<SiteFormProps, SiteFormState> {
@@ -40,6 +41,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
       auth: authField(),
     },
     loading: false,
+    themeList: [],
   };
 
   constructor(props: any, context: any) {
@@ -76,6 +78,11 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
         auth: authField(),
       };
     }
+  }
+
+  async componentDidMount() {
+    this.state.themeList = await fetchThemeList();
+    this.setState(this.state);
   }
 
   // Necessary to stop the loading
@@ -336,7 +343,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
                 class="custom-select w-auto"
               >
                 <option value="browser">{i18n.t("browser_default")}</option>
-                {themes.map(theme => (
+                {this.state.themeList.map(theme => (
                   <option value={theme}>{theme}</option>
                 ))}
               </select>


src/server/index.tsx
```diff
@@ -46,6 +49,54 @@ server.get("/robots.txt", async (_req, res) => {
...
+    } else {
+      res.send(builtinThemes);
+    }
+  });
+});
+
+server.get("/css/themes/:name", async (req, res) => {
+  res.contentType("text/css");
+  const theme = req.params.name;
+  if (!theme.endsWith(".min.css")) {
```
Nutomic, line 87: It might be confusing that themes wont work unless the name ends with .min.css. Maybe better to remove the .min part from builtin themes.

src/server/index.tsx
```diff
@@ -46,6 +49,54 @@ server.get("/robots.txt", async (_req, res) => {
...
+    "cyborg",
+    "darkly",
+    "journal",
+    "sketchy",
+    "vaporwave",
+    "vaporwave-dark",
+    "i386",
+    "litely",
+    "nord",
+  ];
```
dessalines, line 68: You should have access to the shared exported themes already: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils.ts#L78

src/server/index.tsx
```diff
@@ -46,6 +49,54 @@ server.get("/robots.txt", async (_req, res) => {
...
+        .map(d => d.replace(".min.css", ""));
+      data = builtinThemes.concat(data);
+      // use set to remove duplicate values
+      data = Array.from(new Set(data));
+      res.send(data);
+    } else {
+      res.send(builtinThemes);
+    }
+  });
+});
```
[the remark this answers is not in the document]
dessalines: Yep it the theme list can be built on startup. Its better performance wise and we can assume that they need to restart the server if they want to add a theme.

Have the function that builds the theme list be at top level, not within a get.

src/server/index.tsx
dessalines: Hrm.... I spose it should be able to read any css file, these are custom themes after all, and nothings going in this repo anyway. So not too necessary to do this check.

src/server/index.tsx
```diff
@@ -46,6 +49,54 @@ server.get("/robots.txt", async (_req, res) => {
...
+  const theme = req.params.name;
+  if (!theme.endsWith(".min.css")) {
+    res.send("Theme must be a css file");
+  }
+
+  const customTheme = path.resolve(`./${extraThemesFolder}/${theme}`);
+  if (fs.existsSync(customTheme)) {
+    res.sendFile(customTheme);
+  } else {
+    const internalTheme = path.resolve(`./dist/assets/css/themes/${theme}`);
```
dessalines, line 95: This will pry need to be `static/assets/css/themes`, like below.

src/shared/utils.ts
```diff
@@ -166,22 +166,23 @@ export const languages = [
...
-  "darkly",
-  "journal",
-  "sketchy",
-  "vaporwave",
-  "vaporwave-dark",
-  "i386",
-  "litely",
-  "nord",
-];
+export function fetchThemes(): Promise<string[]> {
```
dessalines, line 169: `export async function...`

and then when calling it do `await fetchThemes()...`

dessalines (commented):
Let me know if you have other questions, also change this to a draft PR till its ready.

src/server/index.tsx
Nutomic: .min.css is hardcoded here: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/app/theme.tsx#L20

Nutomic:
Redid the changes from scratch using `import { themeList } from "../../../server/index";`. As you can see, it fails on yarn build.

```
ERROR in ./src/shared/utils.ts 61:0-35

Module not found: Error: Can't resolve 'server' in '/drone/src/src/shared'

resolve 'server' in '/drone/src/src/shared'

  Parsed request is a module

  using description file: /drone/src/package.json (relative path: ./src/shared)

    resolve as module

      /drone/src/src/shared/node_modules doesn't exist or is not a directory

      /drone/src/src/node_modules doesn't exist or is not a directory

      looking for modules in /drone/src/node_modules

        single file module

          using description file: /drone/src/package.json (relative path: ./node_modules/server)

            no extension

              /drone/src/node_modules/server doesn't exist

            .js

              /drone/src/node_modules/server.js doesn't exist

            .jsx

              /drone/src/node_modules/server.jsx doesn't exist

            .ts

              /drone/src/node_modules/server.ts doesn't exist

            .tsx

              /drone/src/node_modules/server.tsx doesn't exist

        /drone/src/node_modules/server doesn't exist

      /drone/node_modules doesn't exist or is not a directory

      /node_modules doesn't exist or is not a directory

 @ ./src/server/index.tsx 18:0-66 124:12-27 157:12-26
```

src/server/index.tsx
Nutomic: `server.use("/static", express.static(path.resolve("./dist")));`

So the http route `/static` is served from local path `./dist`.

dessalines:
Check out the recent PR I did to this, I fixed these issues now.

dessalines:
Change from draft to ready, I tested and it seems good.

dessalines:
Did you test this when setting a default theme that's not a custom one?

I'm getting:

`Error: ENOENT: no such file or directory, stat '/home/.../git/lemmy-ui/dist/assets/css/themes/undefined.min.css'`

PR #586, opened 2022-02-25
Add option to set site default theme

Nutomic:
It works, but the input looks a bit ugly. Not sure how to fix that.

Screenshot_20220225_143642

### DIFF
--- a/src/shared/components/app/app.tsx
+++ b/src/shared/components/app/app.tsx
@@ -26,7 +26,7 @@ export class App extends Component<AppProps, any> {
       <>
         <Provider i18next={i18n}>
           <div>
-            <Theme myUserInfo={siteRes.my_user} />
+            <Theme myUserInfo={siteRes.my_user} site={siteRes.site_view.site} />
             {siteRes &&
               siteRes.site_view &&
               this.props.siteRes.site_view.site.icon && (
--- a/src/shared/components/app/app.tsx
+++ b/src/shared/components/app/app.tsx
@@ -26,7 +26,10 @@ export class App extends Component<AppProps, any> {
       <>
         <Provider i18next={i18n}>
           <div>
-            <Theme myUserInfo={siteRes.my_user} site={siteRes.site_view.site} />
+            <Theme
+              myUserInfo={siteRes.my_user}
+              defaultTheme={siteRes.site_view.site.default_theme}
+            />
             {siteRes &&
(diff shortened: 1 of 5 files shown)


src/shared/components/app/app.tsx
```diff
@@ -26,7 +26,7 @@ export class App extends Component<AppProps, any> {
       <>
         <Provider i18next={i18n}>
           <div>
-            <Theme myUserInfo={siteRes.my_user} />
+            <Theme myUserInfo={siteRes.my_user} site={siteRes.site_view.site} />
```
dessalines, line 29: I'd prefer you just pass in the default theme from site, and have the param be something like `defaultSiteTheme` .

My convention, and I'm doing this for android too, is to always use the lowest level possible, then once you need extra data from a few levels up, then select the bigger dataset. But only do that if you really need that other data.

src/shared/components/app/theme.tsx
```diff
@@ -20,22 +21,11 @@ export class Theme extends Component<Props> {
...
-              href="/static/assets/css/themes/darkly.min.css"
-              id="default-dark"
-              media="(prefers-color-scheme: no-preference), (prefers-color-scheme: dark)"
-            />,
-          ]
+          <link
+            rel="stylesheet"
+            type="text/css"
+            href={`/static/assets/css/themes/${this.props.site.default_theme}.min.css`}
+          />
```
dessalines, line 28: iirc the default theme should be "browser", which isn't actually a theme, but uses either litely or darkly. So this should be checking for that, not removing the prefers color scheme stuff.

src/shared/components/home/site-form.tsx
```diff
@@ -314,14 +321,35 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
               </div>
             </div>
           </div>
+          <div class="form-group row">
+            <div class="col-12">
+              <label
+                class="form-check-label"
+                htmlFor="create-site-default-theme"
+              >
+                {i18n.t("theme")}
```
dessalines, line 330: Just add a `mr-2` or something if they look too close together.

dessalines (changes requested):

dessalines (approved):

dessalines:
Looks good, I'll handle the lemmy-js-client issues.

PR #596, opened 2022-03-14
Fix error during new site setup

Nutomic:
New site setup failed because it tried to read the default theme from site.

### DIFF
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -225,7 +225,6 @@ export class Modlog extends Component<any, ModlogState> {
     combined.push(...transferred_to_community);
     combined.push(...added);
     combined.push(...banned);
-    console.log(combined);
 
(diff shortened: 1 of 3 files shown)


src/shared/components/modlog.tsx
```diff
@@ -224,6 +224,7 @@ export class Modlog extends Component<any, ModlogState> {
     combined.push(...transferred_to_community);
     combined.push(...added);
     combined.push(...banned);
+    console.log(combined);
```
dessalines, line 227: Could you remove this logging line?

dessalines (changes requested):

src/shared/components/modlog.tsx
Nutomic: Sorry, fixed.

dessalines (approved):

PR #597, opened 2022-03-14
Differentiate between mods and admins in mod log

Nutomic:
Also differentiate between admin and mod actions for users who
are not logged in.


dessalines (commented):
I'm not sure we should merge this, due to harrassment reasons, I'll comment in the issue.

Nutomic:
updated

dessalines (approved):

PR #608, opened 2022-04-05
Set content security policy http header for all responses

Nutomic:
The current implementation is broken due to wrong encoding, and doesnt apply to all endpoints.
`<meta http-equiv="Content-Security-Policy" content="default-src data: &#039;self&#039;; connect-src * ws: wss:; frame-src *; img-src * data:; script-src &#039;self&#039;; style-src &#039;self&#039; &#039;unsafe-inline&#039;; manifest-src &#039;self&#039">`


dessalines (approved):

dessalines:
I should not have merged this, because it broke lemmy-ui.

Nutomic:
Damn. I only checked locally that the header was set correctly. Whats the problem?

dessalines:
Could not load script-src.

PR #612, opened 2022-04-06
Dont allow community urls like /c/{id} (fixes #611)

### LINKED ISSUE #611: Communities that are numbers only cannot be accessed via url
### Issue Summary

When you enter a community url, and the commnity id only contains numbers, you get an error.

### Steps to Reproduce

1. click the following link: [https://lemmy.ml/c/195](https://lemmy.ml/c/195)
2. you should get the error: `404: couldnt_find_community`
3. however, if you search for "195" and click on the community link, it works
4. once you're on the community page, if you press "refresh" in the browser, you'll get the same error again

Nutomic:
If the community identifier in url can be interpreted as a valid integer, lemmy-ui considers it as community id, which breaks things if a community has a name with only numbers.

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -166,10 +166,8 @@ export class Community extends Component<any, State> {
     let pathSplit = req.path.split("/");
     let promises: Promise<any>[] = [];
 
-    let idOrName = pathSplit[2];
-    let name_ = idOrName;
-
-    let communityForm: GetCommunity = { name: name_ };
+    let communityName = pathSplit[2];
+    let communityForm: GetCommunity = { name: communityName };
     setOptionalAuth(communityForm, req.auth);
     promises.push(req.client.getCommunity(communityForm));
 
@@ -197,7 +195,7 @@ export class Community extends Component<any, State> {
         saved_only: false,
       };
       setOptionalAuth(getPostsForm, req.auth);
-      this.setName(getPostsForm, name_);
+      this.setName(getPostsForm, communityName);
       promises.push(req.client.getPosts(getPostsForm));
     } else {
       let getCommentsForm: GetComments = {
@@ -207,7 +205,7 @@ export class Community extends Component<any, State> {
         type_: ListingType.Community,
         saved_only: false,
       };
-      this.setName(getCommentsForm, name_);
+      this.setName(getCommentsForm, communityName);
       setOptionalAuth(getCommentsForm, req.auth);
       promises.push(req.client.getComments(getCommentsForm));
     }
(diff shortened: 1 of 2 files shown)


dessalines:
This is also an issue for usernames, which is in `profile.tsx`. There are a few options:

- Now that fully qualified names are working ( IE /c/announcements@lemmy.ml ), then we can remove the number checks, and also remove numeric ID fetching from the back end completely, and from lemmy-js-client.
- Leave the numeric fetching in place, but add `/username/{number}` and `/community/{number}` in addition to the /c/ and /u/ front end routes. Then when doing that check, use the split[3] == "username" or whatever it is to determine whether it should fetch the number or string version. 

I kinda think the 2nd option is best.

Nutomic:
Removed it from profile.tsx as well. I dont think it makes sense to keep around legacy code because "it might still be used somewhere". That way we will just keep adding more logic that needs to be maintained, with no benefit for anyone.

src/shared/components/community/community.tsx
```diff
@@ -166,17 +166,10 @@ export class Community extends Component<any, State> {
...
-    // It can be /c/main, or /c/1
     let idOrName = pathSplit[2];
-    let id: number;
-    let name_: string;
-    if (isNaN(Number(idOrName))) {
-      name_ = idOrName;
-    } else {
-      id = Number(idOrName);
-    }
+    let name_ = idOrName;
```
dessalines, line 170: You can just do 

`let name_ = pathSplit[2];`

src/shared/components/person/profile.tsx
```diff
@@ -173,21 +163,14 @@ export class Profile extends Component<any, ProfileState> {
       saved_only: view === PersonDetailsView.Saved,
       page,
       limit: fetchLimit,
+      username: pathSplit[2],
     };
     setOptionalAuth(form, req.auth);
-    this.setIdOrName(form, person_id, username);
+    form.username;
```
dessalines, line 169: This line might have got in by accident.

dessalines (changes requested):

src/shared/components/person/profile.tsx
```diff
@@ -173,21 +164,14 @@ export class Profile extends Component<any, ProfileState> {
       saved_only: view === PersonDetailsView.Saved,
       page,
       limit: fetchLimit,
+      username: username,
     };
     setOptionalAuth(form, req.auth);
-    this.setIdOrName(form, person_id, username);
+    form.username = username;
```
dessalines, line 170: This line is redundant now since you have the username in the form above.

dessalines (approved):

PR #615, opened 2022-04-09
add new themes

2xx04:


Nutomic:
Could you include screenshots for each theme? I see some of them in your [lemmy-ui-themes](https://github.com/2xx04/lemmy-ui-themes) repo, but not all of them. Btw, in the future it might also be nice to include screenshots directly in lemmy-ui to make it easier for users to pick a theme.

Its also worth noting that after this PR, there will be 20 themes included with lemmy-ui, which is a lot. It might be better to limit builtin themes to a more managable number, like 10 or so. But its okay if we merge this first and then consider removing some unpopular themes later.

2xx04:
the lora themes are basically the same (with variants, for an example in compact I hid all avatars and banners), all screenshots are there/

dessalines (changes requested):
These also need to be added to `src/server/index.tsx`, line 90. And make sure these work as users change their user settings to the new themes.

2xx04:
When is the next release? none of the themes are tested through a real lemmy install except for winternord. Also darkspace still has missing search background because I haven't added the class to search.tsx yet.

Until I've done so, we shouldn't merge this pull request.

2xx04:
closing, for now.

Nutomic:
FYI you can also mark a pr as draft so it cant be merged (top right, near reviewers).

2xx04:
Oh, I didn't know that. thanks.

dessalines:
I've converted it to a draft for you. I'd prefer you bundle all these changes as one testable PR, would make it a lot easier for me.

2xx04:
How would I do that?

dessalines:
You need to add css classes / changes for these themes to work correctly, right? Add those changes to the necessary files, and commit them to this branch.

PR #620, opened 2022-04-11
Adds OC

ensag-dev:
For the translations are back


Nutomic:
Thanks! In general we want to include all languages which are at least 70% translated. Would you be so helpful to go through the [weblate translations](https://weblate.yerbamate.ml/projects/lemmy/lemmy/) and also add others which are over 70%, if any?

dessalines (approved):

PR #621, opened 2022-04-11
Set content security policy http header for all responses

Nutomic:
Tested with [this browser plugin](https://github.com/april/laboratory) that it works without any errors. 

[This site](https://cspscanner.com/?q=default-src+%27none%27%3B+connect-src+%27self%27%3B+img-src+*+data%3A%3B+script-src+%27self%27+%27unsafe-inline%27%3B+style-src+%27self%27+%27unsafe-inline%27%3B+form-action+%27self%27%3B+base-uri+%27self%27) analyzes the policy. Main problems are `frame-ancestors` (do we really want to block iframes?) and `upgrade-insecure-requests` (not sure if this would break http images).

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -11,7 +11,7 @@ import process from "process";
 import serialize from "serialize-javascript";
 import { App } from "../shared/components/app/app";
 import { SYMBOLS } from "../shared/components/common/symbols";
-import { httpBaseInternal } from "../shared/env";
+import { httpBaseInternal, wsUriBase } from "../shared/env";
 import {
   ILemmyConfig,
   InitialFetchRequest,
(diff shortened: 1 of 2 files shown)


dessalines:
Broke it again. Check the browser console.

Nutomic:
I checked it. What error do you see?

dessalines:
Content Security Policy: The page’s settings blocked the loading of a resource at eval (“script-src”). [client.js:675](http://localhost:1234/static/js/client.js)

Nutomic:
I cant reproduce that, but added unsafe-eval. Would be good if we could get rid of all the unsafe btw.

dessalines:
Now I'm getting:

```
Content Security Policy: The page’s settings blocked the loading of a resource at ws://localhost:8536/api/v3/ws (“connect-src”).
```

To reproduce this, you have to be running `yarn start` in the lemmy-ui folder, and going to http://localhost:1234

Nutomic:
Found a way to fix it.

Nutomic:
This should be good to merge.

dessalines (approved):

dessalines:
Okay, but we gotta remember to check the federation test instances before we do our next deploy to make sure it didn't break websocket.

dessalines:
Nevermind, I'm reverting this again as it broke development over a local network.

dessalines:
Actually there's another way to do it, I'll include that with another PR.

PR #640, opened 2022-04-28
Adding site sidebar for remote communities. Fixes #626

### LINKED ISSUE #626: Display remote instance sidebar below remote community sidebar
Since other instances can have different rules, it makes sense, for remote communities, to display the other instance sidebar below the community one. This is dependent on GetCommunity returning the other site, which should be added soon.

Discussion: https://github.com/LemmyNet/lemmy/issues/2208

dessalines:
This has limitations: 

Since only the site, and not the site_view, or admins, come back from GetCommunity, it really only shows the sidebar, and no activity counts.

### DIFF
--- a/src/shared/components/home/site-sidebar.tsx
+++ b/src/shared/components/home/site-sidebar.tsx
@@ -1,0 +1,267 @@
+import { Component, linkEvent } from "inferno";
+import { Link } from "inferno-router";
+import { PersonViewSafe, Site, SiteAggregates } from "lemmy-js-client";
+import { i18n } from "../../i18next";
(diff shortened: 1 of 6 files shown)


src/shared/components/home/site-sidebar.tsx
```diff
@@ -0,0 +1,267 @@
+import { Component, linkEvent } from "inferno";
```
dessalines, line 1: Moved the SiteSidebar into its own component, since it's now used elsewhere.

Nutomic (approved):

PR #644, opened 2022-05-13
Fix local development.

dessalines:


Nutomic (approved):
Looks better :+1:

Nutomic (commented):
This means that we wont notice csp issues during development.

PR #647, opened 2022-05-15
Adding Option Types.

### LINKED ISSUE #646: Add Option and Result Rust types to lemmy-ui
A few *options* 😆  : 

- https://github.com/karen-irc/option-t
- https://github.com/swan-io/boxed
- https://github.com/sniptt-official/monads

Monads is my fav currently.

dessalines:
Fixes #646


src/shared/utils.ts
```diff
@@ -1271,3 +1384,56 @@ export function pushNotNull(array: any[], new_item?: any) {
     array.push(...new_item);
   }
 }
+
+/**
+ * Converts an undefined or null value to an Option
+ *
+ * Note: You cannot use a prototype, because it won't be able to handle null values
+ */
+export function toOption<T>(val: T): Option<T> {
```
dessalines, line 1393: This is the main change, converting nulls and undefineds to rust-style options.

Because typescript doesn't have first class support for these monads, I have to do:

Receive API with nulls <-> Lemmy UI with Monads <--> Send API with nulls.

Even using the Options internally helped me check and fix several bugs during this process, and clean up a LOT of code.

Nutomic:
Does this have any lints or checks which prevent usage of null etc?

src/shared/utils.ts
Nutomic: You could consider adding this library to lemmy-js-client as well.

dessalines:
Yep, the Option types effectively do compile time checks for nulls, and will fail lints if not used. 

The next step might be to use something [like this](https://github.com/JohnWeisz/TypedJSON) to add the Monads to lemmy-js-client, and then the `toOption` function could be removed in this repo.

src/shared/utils.ts
dessalines: It'd require something like this, and might be a lot of work / boilerplate to add: https://github.com/JohnWeisz/TypedJSON

dessalines: That library didn't work, but I found a way to do this using this library: https://github.com/typestack/class-transformer

Adds a good amount of boilerplate, but it fully works deserializing and serializing.

dessalines:
Closed in favor of #689

PR #653, opened 2022-05-20
Add legal information (fixes #652)

### LINKED ISSUE #652: Add legal information field.
https://github.com/LemmyNet/lemmy/pull/2273

Nutomic:
This should work as far as i can tell, but there are two problems:
- legal information field on /admin is not populated when browsing from another page (its populated fine after reload)
- /legal page fails to load with error "TypeError: _utils__WEBPACK_IMPORTED_MODULE_3__.md is not a function"
- initial setup seems to be broken by this change (says `404: TypeError: Cannot read properties of null (reading 'site')` but i cant figure out why)

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "husky": "^7.0.4",
     "import-sort-style-module": "^6.0.0",
-    "lemmy-js-client": "0.16.4-rc.1",
+    "lemmy-js-client": "git+https://github.com/LemmyNet/lemmy-js-client.git#legal-info",
     "lint-staged": "^12.4.1",
     "mini-css-extract-plugin": "^2.6.0",
     "node-fetch": "^2.6.1",
--- a/src/shared/components/app/footer.tsx
+++ b/src/shared/components/app/footer.tsx
@@ -32,6 +32,13 @@ export class Footer extends Component<FooterProps, any> {
                 {i18n.t("modlog")}
               </NavLink>
             </li>
+            {this.props.site.site_view.site.legal_information && (
+              <li className="nav-item">
+                <NavLink className="nav-link" to="/legal">
+                  {i18n.t("legal_information")}
+                </NavLink>
+              </li>
+            )}
             {this.props.site.federated_instances && (
               <li class="nav-item">
                 <NavLink className="nav-link" to="/instances">
--- a/src/shared/components/home/admin-settings.tsx
+++ b/src/shared/components/home/admin-settings.tsx
@@ -2,6 +2,7 @@ import autosize from "autosize";
 import { Component, linkEvent } from "inferno";
 import {
   BannedPersonsResponse,
+  EditSite,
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -74,7 +74,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "husky": "^7.0.4",
     "import-sort-style-module": "^6.0.0",
-    "lemmy-js-client": "0.16.4-rc.1",
+    "lemmy-js-client": "git+https://github.com/LemmyNet/lemmy-js-client.git#legal-info",
```
Nutomic, line 77: This needs to be changed before merging.

src/shared/components/home/admin-settings.tsx
```diff
@@ -36,6 +37,7 @@ interface AdminSettingsState {
   siteRes: GetSiteResponse;
   siteConfigRes: GetSiteConfigResponse;
   siteConfigHjson: string;
+  legalInfo: string;
```
dessalines, line 40: This field needs to be on SiteForm, with the rest of the site fields.

dessalines (changes requested):

src/shared/components/home/admin-settings.tsx
Nutomic: Done, doesnt seem to fix anything.

src/shared/components/app/footer.tsx
```diff
@@ -32,6 +32,13 @@ export class Footer extends Component<FooterProps, any> {
                 {i18n.t("modlog")}
               </NavLink>
             </li>
+            {this.props.site.site_view.site.legal_information && (
+              <li className="nav-item">
+                <NavLink className="nav-link" to="/legal">
+                  {i18n.t("legal_information")}
```
dessalines, line 38: Can we just have this be `Legal` or `Privacy`? I don't want the navbars to get crowded with long text.

src/shared/components/home/admin-settings.tsx
dessalines: No I mean remove the field from here, and move it to SiteForm, along with all the other EditSite columns.

src/shared/components/home/site-form.tsx
```diff
@@ -76,6 +77,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
         application_question: site.application_question,
         private_instance: site.private_instance,
         default_theme: site.default_theme,
+        legal_information: site.legal_information,
```
dessalines, line 80: You'll notice this is the component that handles EditSite, not admin-settings.tsx

src/shared/components/app/footer.tsx
Nutomic: This is just the translation key, made it longer for clarity. English text is "Legal".

dessalines:
I'll make a new pr based on this soon.

dessalines:
Closed in favor of #666

PR #656, opened 2022-05-22
Add LEMMY_UI_CUSTOM_SCRIPT env var. Fixes #655

### LINKED ISSUE #655: Allow custom javascript to be injected server-side.
I would like to add Matomo to my Lemmy instance.
I use Matomo for my other web resources.
It's simple, but Matomo works with a javascript to embed in the main page or in any single page.
Sincerely, I don't know how and where to add the Matomo script.

dessalines:


src/server/index.tsx
```diff
@@ -201,6 +203,9 @@ server.get("/*", async (req, res) => {
            <!-- A remote debugging utility for mobile -->
            ${erudaStr}
 
+           <!-- Custom injected script -->
+           ${customScript}
```
Nutomic, line 207: Should be called like `customHtmlHeader` if its like this. For `customScript`, i would expect something like `<script src="${customScript}"></script>`.

dessalines: Agree... how about `LEMMY_UI_CUSTOM_HTML`? 

I'll make an issue to document the ENV vars.

Nutomic: I would use `LEMMY_UI_CUSTOM_HTML_HEADER` to make it a bit clearer.

Nutomic (approved):

PR #657, opened 2022-05-22
Turn off html in markdown. Fixes #650

### LINKED ISSUE #650: The audio of an embedded video can be made to auto-play outside of the post
### Issue Summary

A summary of the bug.

If you are on a page which contains the link to a post, and that post contains a video embedded using iframe, the audio of the video will be heard (if you access via a URL externally, then the auto play will occur as soon as an action is taken on that page, for example, changing from "All" to "Local", or from "New" to "Hot".) This autoplay occurs even if the browser is set to block autoplay. 

### Steps to Reproduce

Create a post and embed a video beginning with a block such as the following:

```
<div style='position:relative; padding-bottom:calc(50%)'><iframe src='https://radiz.nl/dancingbug.mp4' frameborder='0' scrolling='no' width='100%' height='100%' style='position:absolute;top:0;left:0;' allowfullscreen></iframe></div>
```

Then, go to the sub-lemmy that contains that post. If the audio does not autoplay, change the "Sort type". 

You can access an example by going to:

https://mander.xyz/c/test

Then change the "Sort type" to something else.

EDIT: I have noticed the following. If the post has a first sentence, then a space, and after this the embedding block, then the autoplay will not happen. For example:

```
a

<div style='position:relative; padding-bottom:calc(50%)'><iframe src='https://radiz.nl/dancingbug.mp4' frameborder='0' scrolling='no' width='100%' height='100%' style='position:absolute;top:0;left:0;' allowfullscreen></iframe></div>
```

Won't autoplay. Maybe this provides another clue about what is going on. 

### Technical details

I have tested this in Firefox both for arch linux and an android phone.

dessalines:


Nutomic (approved):
This can be considered a breaking change, but also fixes potential security issues. Do you plan to release it in 0.16.4 or 0.17.0?

dessalines:
There are no API changes or anything like that, its only a change in how markdown is rendered. I'm up for either, but IMO sooner than later, because some people are injecting a lot of weird things into lemmy like tooltips that block other buttons and such.

Nutomic:
Okay sounds fine.

PR #663, opened 2022-05-24
Add syntax highlighting for code blocks

Thann:
Each theme will have to import an appropriate highlight.js theme to work. (like I did for darkly & litely)

Ill do that for more themes if this direction is approved.


src/server/index.tsx
```diff
@@ -58,19 +58,24 @@ server.get("/robots.txt", async (_req, res) => {
   res.send(robotstxt);
 });
 
-server.get("/css/themes/:name", async (req, res) => {
+server.get("/css/themes/:set?/:name", async (req, res) => {
```
Thann, line 61: This is probably the most controversial change but it allows us to import nicely from themes, i.e. `@import "hljs/default.css"`

src/server/index.tsx
```diff
@@ -58,19 +58,24 @@ server.get("/robots.txt", async (_req, res) => {
...
   const theme = req.params.name;
   if (!theme.endsWith(".css")) {
     res.send("Theme must be a css file");
   }
 
-  const customTheme = path.resolve(`./${extraThemesFolder}/${theme}`);
-  if (fs.existsSync(customTheme)) {
-    res.sendFile(customTheme);
+  if (req.params.set === 'hljs') {
+    const hljsTheme = path.resolve(`./node_modules/highlight.js/styles/${theme}`);
```
Thann, line 69: IDK how you feel about pulling files from the node_modules folder, but we could webpack CopyPlugin the files to `src/assets/css/themes/hljs` or something if you prefer.

src/assets/css/themes/darkly.css
```diff
@@ -1 +1,2 @@
+@import "hljs/dark.css";
```
dessalines, line 1: How about instead of editing every css file, load the appropriate light / dark one based on prefers-color-scheme? 

You could do that in this file somewhere: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/app/theme.tsx#L39

src/server/index.tsx
dessalines: I'd prefer it be hljs-dark.css , and hljs-light.css, and have them loaded in `theme.tsx` if possible. I'm not sure if symlinking is possible there, but its worth a try. If not we could always have a job to copy the theme from node_modules into the assets/css/themes folder.

src/server/index.tsx
dessalines: Yep I like that better. hljs-light.css and hljs-dark.css . Also a symlink might work but I'm not sure.

src/assets/css/themes/darkly.css
Thann: There are a bunch of hljs themes, so it seems like a shame to only include two =\

I think the ultimate solution would be to have a user-setting like for the site-themes.

I could implement it using local storage while we debate if it warrants inclusion in the API proper

Thann:
Sorry for dragging my feet with this, I've been busy with work.

Here's a work-in-progress commit, so we can discuss some architectural decisions.

src/shared/components/app/theme.tsx
```diff
@@ -14,45 +15,110 @@ export class Theme extends Component<Props> {
...
+            />
+            <link
+              rel="stylesheet"
+              type="text/css"
+              href="/css/themes/darkly.css"
+              id="default-dark"
+              media="(prefers-color-scheme: no-preference), (prefers-color-scheme: dark)"
+            />
+          </Helmet>
+          <CodeTheme
```
Thann, line 70: There are three cases for themes: `user-selected`, `default`, and `browser`.
There are the same choices, for code themes, so I split the code into its own component.

The question is: should that component be embedded in Theme or be a peer?
I'm leaning to it being a peer, but Ill defer to yo.

Thann:
Currently this will assume `browser` for code-theme, but supports the future addition of a user preference.

src/shared/components/app/app.tsx
```diff
@@ -29,6 +29,7 @@ export class App extends Component<AppProps, any> {
             <Theme
               myUserInfo={siteRes.my_user}
               defaultTheme={siteRes?.site_view?.site?.default_theme}
+              defaultCodeTheme={siteRes?.site_view?.site?.default_code_theme}
```
dessalines, line 32: Lint is failing because this field doesn't exist in the back end.

Rather than have this be customizeable in the back end, why not just do one of the following:

- Have a default light and dark theme, and maybe default to dark in other cases.
- Possibly create a javascript map from theme to hljs theme ( in src/server/index.tsx ) , although this seems overkill to me.

src/shared/components/app/theme.tsx
dessalines: Splitting it into its own component is way overkill, **exactly because** it uses the same three options as theme. 

Just add an extra one of these to each of the 3 cases which already have blocks above.

```
<link
              rel="stylesheet"
              type="text/css"
              href={`location_of_hljs_theme`}
            />
```

Thann: Its the same options, but I wanted them to be independent, so there would be 9 possible combos.

src/shared/components/app/app.tsx
Thann: If were to have a 1->1 relation between themes and code-themes, I thought my previous solution was better because 1. mainline theme creators could pick the most appropriate theme, and 2. custom themes can include an appropriate theme. 

Another possibility would be to measure the human-perceived luminosity of the themes background color and pick between light and dark code-themes. I've done this before and it worked well, but I would prefer to have user-selectable themes =/

dessalines (commented):
Since `code_theme` needs to be added to the back end, then lemmy-js-client, this PR will have to wait on those to be added first.

Thann:
backend setting for code-theme: https://github.com/LemmyNet/lemmy/issues/2307

src/shared/components/app/app.tsx
Nutomic: I explained here why i dont think it makes sense to add a setting for this:

https://github.com/LemmyNet/lemmy/issues/2307#issuecomment-1151615448

dessalines: I think my first proposal makes sense: pick a good light and dark theme, and use `prefers-color-scheme` to decide the right one to load.

Thann: If people set a UI-theme that conflicts with their browser theme the only solution is to change their browser theme.  Any sites that set a default theme will have decoherent themes for a lot of users  =\

Additionally, there is no customizability and themers have their hands tied. I think importing hljs themes from UI-themes is more extensible, and doesn't pose a similar usability issue.

Nutomic: I think the initial implementation with `@import` made most sense, because it allows each theme to explicitly specify a matching code theme. Plus if there is none that matches, parts of the code theme could be overwritten, or an entirely new theme included. `prefers-color-scheme` is the same thing thats still broken in Firefox, which makes Lemmy always use the light theme, right? I would hate to rely on something thats still completely broken for many users.

dessalines: Yep some browsers don't fully support it yet, I thought firefox did tho, at least FF mobile does. I was thinking the default for code would be dark. The @import s on each of the themes could work for a short term fix too tho.

Ideally, lemmy-ui would not store any minified themes, and only work with bootstrap `variables.scss` files, then generate the themes itself. That way we could get rid of them from the codebase. But its not necessary for this issue.

richinseattle:
https://github.com/LemmyNet/lemmy/issues/2307 was closed in January. Can we get hljs syntax highlighting for lightly/darkly merged? Its important for a server I want to run that is focused around development communities.

dessalines:
@richinseattle you could open up a PR that uses the simple `@import` solution, it would likely only be a few lines.

Issue #669, opened 2022-05-29
login broken in webkit


dayinjing:
### Issue Summary

Login does not work in the 16.4 release of lemmy-ui for webkit browsers. ios and osx

### Steps to Reproduce

1. on osx or iphone go to lemmy.ml
2. try to login
3. you can't

### Technical details

Console says something about 'undefined is not an object (evaluating 'this.we.send')'

dayinjing:
Content security policy change here https://github.com/LemmyNet/lemmy-ui/compare/0.16.3...0.16.4#diff-a9ad80ae074de023f610c915d6e37938213c456d1f54fa0e58bbab181045bb59R30

Seems to have broken login on webkit devices

dayinjing:
Looks like this value is comming out wrong in the ui `${wsUriBase}` I guess

nicfab:
It could be the same of my issue https://github.com/LemmyNet/lemmy/issues/2286

dayinjing:
> It could be the same of my issue [LemmyNet/lemmy#2286](https://github.com/LemmyNet/lemmy/issues/2286)

Yeah looks like the same issue; although you should have opened it here because it is a ui issue. I'm working on a fix.

dayinjing:
Ok, this is what needs to be done to fix it,

```
if (!process.env["LEMMY_UI_DEBUG"]) {
  console.log("wsUriBase:",wsUriBase);
  server.use(function (_req, res, next) {
    res.setHeader(
      "Content-Security-Policy",
      `default-src 'none'; connect-src 'self' ${wsUriBase} wss://DOMAINNAME/api/v3/ws; img-src * data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; base-uri 'self'`
    );
    next();
  });
}

```

**DOMAINNAME** needs to be your domain name and then it will work on web-kit browsers. I just hard coded it and tested it on my instance and it works fine. 

My kids woke up So I need to go make breakfast. Hopefully someone else can pick this up Or i'll open a pr later to make it more general.

If anyone knows where the domain name is accessable on the front end and can put it here that would speed it up.

dayinjing:
https://github.com/LemmyNet/lemmy-ui/pull/671

PR to fix here

dessalines:
[...]

PR #671, opened 2022-05-30
Fix web socket being rejected in webkit browsers

dayinjing:
fixes https://github.com/LemmyNet/lemmy-ui/issues/669

### DIFF
--- a/src/shared/env.ts
+++ b/src/shared/env.ts
@@ -35,8 +35,9 @@ if (isBrowser()) {
 
 export const httpBaseInternal = `http://${host}`; // Don't use secure here
 export const httpBase = `http${secure}://${host}`;
-export const wsUriBase = `ws${secure}://${wsHost}`;
+export const wsUriBase = `ws${secure}://${externalHost}`;
 export const wsUri = `${wsUriBase}/api/v3/ws`;
+export const wsWebKitUri = `wss${secure}://${externalHost}/api/v3/ws`;
 export const pictrsUri = `${httpBase}/pictrs/image`;
 export const isHttps = secure.endsWith("s");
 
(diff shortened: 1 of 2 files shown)


src/shared/env.ts
```diff
@@ -35,7 +35,7 @@ if (isBrowser()) {
 
 export const httpBaseInternal = `http://${host}`; // Don't use secure here
 export const httpBase = `http${secure}://${host}`;
-export const wsUriBase = `ws${secure}://${wsHost}`;
+export const wsUriBase = `ws${secure}://${externalHost}`;
```
dessalines, line 38: Failing CI, and this would also break dev builds. 

Why are apple devices not running the code in lines 25-26 here, but android is?

dayinjing: I'm looking. My initial guess is webkit is just more strict about enforcing the rules on websockets. I'm still pretty new to inferno.  I have a test instance and a mac I can test with if you guys want to send me a image though.

dayinjing: I think i see, should be wsUri not wsUriBase on index.tsx. maybe webkit is the only one enforcing it.

dessalines: Gotcha, if u have a mac that'd be helfpul if you could see if that works.

For now i have the 0.16.5 release with the those headers commented out, but let me know if wsUri works.

bigOconstant: I went and tested it this morning and **wsUri** evaluates to `wss://lemmy:8536/api/v3/ws` because `process.env.LEMMY_WS_HOST` is undefined. So that won't work.

It works if we change `wsUrliBase` to `wsUri` AND update lemmy_ansible to include `- LEMMY_WS_HOST={{ domain }}` under lemmy-ui in **docker-compose.yml**

LEMMY_WS_HOST is not defined anywhere.

dessalines: Could you try this line? Replace L32 with:

`wsHost = process.env.LEMMY_WS_HOST || externalHost;`

dayinjing: > `wsHost = process.env.LEMMY_WS_HOST || externalHost;`

That appears to work yes.

dessalines:
Closed in favor of #678

PR #678, opened 2022-06-01
Fixing CSP for iOS devices. Fixes #669

### LINKED ISSUE #669: login broken in webkit

dessalines:


src/server/index.tsx
```diff
@@ -27,16 +27,15 @@ const [hostname, port] = process.env["LEMMY_UI_HOST"]
...
-// }
+if (!process.env["LEMMY_UI_DEBUG"]) {
+  server.use(function (_req, res, next) {
+    res.setHeader(
+      "Content-Security-Policy",
+      `default-src 'none'; connect-src 'self' ${wsUriBase}; img-src * data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; base-uri 'self'`
+    );
+    next();
+  });
+}
```
Nutomic, line 38: So this means CSP only runs in prod, not in debug right? Like i said before, i dont like this because we might break things during dev, and not notice until its deployed. Why not simply set a different `connect-src` depending on the debug env var?

dessalines: Correct. Eruda can't work with CSP, its remote javascript, so I need a flag to be able to disable it during dev.

dessalines: What's sad too is I at least found a good repo that replicates rust Results and Options, but it doesn't deserialize correctly :(

makotech222:
Just want to note, this change breaks websocket connections on local dev testing on Windows machine. I had to revert in order to get it in working condition again. Should we be setting environment variable for 'LEMMY_UI_DEBUG' now?

dessalines:
Yep, please set that for local dev. `export LEMMY_UI_DEBUG=true`

PR #682, opened 2022-06-13
Add Modlog Filters

makotech222:
See https://github.com/LemmyNet/lemmy/pull/2313

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -375,6 +375,10 @@ br.big {
   border-radius: 4px;
 }
 
+.modlog-choices-font-size {
+  font-size: .9375rem !important;
+}
+
 .preview-lines {
   overflow: hidden;
   text-overflow: ellipsis;
--- a/src/shared/components/home/site-form.tsx
+++ b/src/shared/components/home/site-form.tsx
@@ -54,6 +54,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
       description: None,
       community_creation_admin_only: None,
       auth: undefined,
+      hide_modlog_mod_names: Some(true),
     }),
     loading: false,
     themeList: None,
@@ -98,6 +99,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
           default_theme: Some(site.default_theme),
           default_post_listing_type: Some(site.default_post_listing_type),
           legal_information: site.legal_information,
+          hide_modlog_mod_names: site.hide_modlog_mod_names,
           auth: undefined,
         });
       },
@@ -439,6 +441,25 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
           </div>
           <div class="form-group row">
             <div class="col-12">
+              <div class="form-check">
+                <input
+                  class="form-check-input"
+                  id="create-site-hide-modlog-mod-names"
+                  type="checkbox"
+                  checked={toUndefined(this.state.siteForm.hide_modlog_mod_names)}
+                  onChange={linkEvent(this, this.handleSiteHideModlogModNames)}
+                />
+                <label
+                  class="form-check-label"
+                  htmlFor="create-site-hide-modlog-mod-names"
+                >
+                  {i18n.t("hide_modlog_mod_names")}
+                </label>
+              </div>
+            </div>
+          </div>
+          <div class="form-group row">
+            <div class="col-12">
               <button
                 type="submit"
                 class="btn btn-secondary mr-2"
@@ -495,6 +516,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
         default_theme: sForm.default_theme,
         default_post_listing_type: sForm.default_post_listing_type,
         auth: auth().unwrap(),
+        hide_modlog_mod_names: sForm.hide_modlog_mod_names
       });
       WebSocketService.Instance.send(wsClient.createSite(form));
     }
@@ -558,6 +580,11 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
 
   handleSitePrivateInstance(i: SiteForm, event: any) {
     i.state.siteForm.private_instance = Some(event.target.checked);
+    i.setState(i.state);
+  }
+
+  handleSiteHideModlogModNames(i: SiteForm, event: any) {
+    i.state.siteForm.hide_modlog_mod_names = Some(event.target.checked);
     i.setState(i.state);
   }
 
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -581,14 +581,19 @@ export class Modlog extends Component<any, ModlogState> {
         ) : (
           <div>
             <h5>
-              {this.state.communityName.isSome() && (
-                <Link
-                  className="text-body"
-                  to={`/c/${this.state.communityName}`}
-                >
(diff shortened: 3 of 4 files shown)


src/shared/components/common/dropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+      </>
+    );
+  }
+
+  handleToggleDropdown(i: Dropdown, event: any) {
+    event.preventDefault();
+    i.state.showDropdown = !i.state.showDropdown;
+    i.setState(i.state);
+  }
+}
```
dessalines, line 52: None of this is necessary, check out the search.tsx for how to do user searching and filtering. Don't bother with array filtering, an `All` and then the specific types are enough.

src/shared/components/modlog.tsx
```diff
@@ -609,7 +612,7 @@ export class Modlog extends Component<any, ModlogState> {
                 <option value={ModlogActionType.ModAdd}>Adding Mod to Site</option>
                 <option value={ModlogActionType.ModBan}>Banning From Site</option>
               </select>
-              {!this.state.site_view.site.hide_modlog_mod_names && <select
+              {this.state.siteRes.site_view.isSome() && !this.state.siteRes.site_view.unwrap().site.hide_modlog_mod_names && <select
```
dessalines, line 615: Use a match, there should be lots of examples, especially in search.tsx 

In general if you see an unwrap, it means there's a better way to do it, because an unwrap could crash things.

src/shared/components/modlog.tsx
```diff
@@ -83,10 +84,11 @@ interface ModlogState {
   res: Option<GetModlogResponse>;
   communityId: Option<number>;
   communityMods: Option<CommunityModeratorView[]>;
+  communityName: Option<string>,
   page: number;
   siteRes: GetSiteResponse;
   loading: boolean;
-  filter_action?: ModlogActionType,
+  filter_action: ModlogActionType,
   filter_user?: number,
```
dessalines, line 92: I highly recommend using `filter_user_id: Option<number>`, and changing all the other typescript optionals to rust-style Option's. The only case where I've found it doesn't work perfectly, is when we have to use derivedState.

src/shared/components/modlog.tsx
```diff
@@ -57,6 +57,7 @@ import { PersonListing } from "./person/person-listing";
 type ModlogType = {
   id: number;
   type_: ModlogActionType;
+  moderator: PersonSafe;
```
dessalines, line 60: This moderator is optional now.

src/shared/components/modlog.tsx
```diff
@@ -549,7 +552,7 @@ export class Modlog extends Component<any, ModlogState> {
 
   modOrAdminText(person: PersonSafe): string {
     if (
-      this.isoData.site_res.admins.map(a => a.person.id).includes(person.id)
+      person != null && this.isoData.site_res.admins.map(a => a.person.id).includes(person.id)
```
dessalines, line 555: Now that we're using rust-style options, this shouldn't be necessary.

src/assets/css/main.css
```diff
@@ -375,6 +375,10 @@ br.big {
   border-radius: 4px;
 }
 
+.default-font-size {
+  font-size: .9375rem !important;
+}
```
dessalines, line 380: What's this doing here?

src/shared/components/home/site-form.tsx
```diff
@@ -437,6 +439,25 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
               </div>
             </div>
           </div>
+          <div class="form-group row">
```
dessalines, line 442: Looks good, after the backend is ready to go I can test this.

src/shared/components/common/dropdown.tsx
```diff
@@ -0,0 +1,52 @@
+import { Component, linkEvent } from "inferno";
```
dessalines, line 1: This isn't used now.

src/shared/components/modlog.tsx
```diff
@@ -628,6 +684,94 @@ export class Modlog extends Component<any, ModlogState> {
...
 
+  setupUserFilter() {
+    if (isBrowser()) {
+      let selectId: any = document.getElementById("filter-user");
+      if (selectId) {
+        let config = choicesConfig;
+        config.classNames.containerOuter = "choices mb-2 custom-select col-4 px-0";
+        config.classNames.containerInner = config.classNames.containerInner + " py-0 default-font-size";
+        config.classNames.listSingle = config.classNames.listSingle + " py-0";
+        config.classNames.inputCloned = config.classNames.inputCloned + " w-100";
```
dessalines, line 695: These should be set in the choicesConfig in utils.ts , so they apply everywhere.

src/assets/css/main.css
makotech222: Not sure where else to put it? Its specifically for the text in the 'filter by mod/user' dropdown, which has a fontsize of '14px' from '.choices__inner' style.

src/shared/components/modlog.tsx
makotech222: These are specifically for the modlog dropdowns for filter by user/mod so that the style matches the filter by action dropdown (which is a different component).

dessalines: Make a single new choices config in utils.ts and use that for both of these. You have this single config copy-pasted twice.

src/assets/css/main.css
```diff
@@ -375,6 +375,10 @@ br.big {
   border-radius: 4px;
 }
 
+.default-font-size {
```
dessalines, line 378: Rename this to .modlog-choices-font-size or something, that's what its for.

src/shared/components/modlog.tsx
```diff
@@ -628,6 +684,94 @@ export class Modlog extends Component<any, ModlogState> {
...
+          },
+          false
+        );
+        this.userChoices.passedElement.element.addEventListener(
+          "search",
+          debounce(async (e: any) => {
+            try {
+              console.log(e.detail.value);
+              let users = (await fetchUsers(e.detail.value)).users;
+              console.log(users);
```
dessalines, line 712: Remove these debugging console logs once things are working.

stevencherry-uu:
Okay this has been updated.

src/shared/components/modlog.tsx
```diff
@@ -578,9 +580,57 @@ export class Modlog extends Component<any, ModlogState> {
           </h5>
         ) : (
           <div>
-            <div class="table-responsive">
-              <table id="modlog_table" class="table table-sm table-hover">
-                <thead class="pointer">
+            <h5>
+              {this.state.communityName.isSome() && (
```
dessalines, line 584: Use a `this.state.communityName.match({some: communityName => ...` here.

dessalines:
Also failing because lemmy-translations has some merge conflicts, I'll see if I can fix up.

src/shared/components/modlog.tsx
stevencherry-uu: Okay, updated

makotech222: Okay, updated

dessalines (approved):
Looks good ( outside of the merge conflict, but don't worry I can get that ).

Once the back-end is merged, I'll merge this.

PR #706, opened 2022-06-30
Clicking "subscribe pending" button performs unsubscribe (fixes #705)

### LINKED ISSUE #705: Allow unfollowing a community while status is "pending"
In case a follow failed to go through, it would be useful to be able to cancel it, and try again. This could be implemented simply by turning the "subscribe pending" text into a button, which does unfollow when clicked.

Screenshot_20220630_120029

Nutomic:


src/shared/components/community/sidebar.tsx
```diff
@@ -130,9 +130,13 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
             </a>
           )}
           {subscribed == SubscribedType.Pending && (
-            <div class="badge badge-warning mr-2">
+            <a
+              class="badge badge-warning mr-2"
```
dessalines, line 134: Could you turn this into a `btn btn-warning`

Nutomic: I changed the class, but dont understand what you mean by removing <a> tag. The previous item also uses <a>.

src/shared/components/community/sidebar.tsx
```diff
@@ -130,9 +130,13 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
...
           {subscribed == SubscribedType.Pending && (
-            <div class="badge badge-warning mr-2">
+            <a
+              class="btn btn-warning mr-2"
+              href="#"
+              onClick={linkEvent(this, this.handleUnsubscribe)}
+            >
               {i18n.t("subscribe_pending")}
-            </div>
+            </a>
```
dessalines, line 139: This works for now, at some point tho I need to change all these a tags to `<button`s, a tags should only be used for links.

dessalines (approved):

PR #724, opened 2022-07-24
Fix Notification browser fetch

ernestwisniewski:
Checking if the icon exists, otherwise no parameter is passed.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -686,21 +686,18 @@ export function notifyPrivateMessage(pmv: PrivateMessageView, router: any) {
 function notify(info: NotifyInfo, router: any) {
   messageToastify(info, router);
 
-  // TODO absolute nightmare bug, but notifs are currently broken.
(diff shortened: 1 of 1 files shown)


src/shared/utils.ts
```diff
@@ -686,21 +686,18 @@ export function notifyPrivateMessage(pmv: PrivateMessageView, router: any) {
...
-  //   notification.onclick = (ev: Event): any => {
-  //     ev.preventDefault();
-  //     router.history.push(info.link);
-  //   };
-  // }
+  if (Notification.permission !== "granted") Notification.requestPermission();
+  else {
+    var notification = new Notification(info.name, {
+      ...{body: info.body},
+      ...(info.icon && {icon: info.icon})
```
dessalines, line 693: icon is an Option type, so .isSome()

dessalines (changes requested):

dessalines:
This is merged now.

Issue #667, opened 2022-05-26
translate title of zh_Hant translate is incorrect and not auto selected


ianiiaannn:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

1. zh_Hant is NOT "文言" but "中文（繁體）". 文言(Classical Chinese) is old language Chinese use. 中文（繁體）(Traditional Chinese) is Chinese common used in Taiwan and some other country. zh_Hans 中文（簡體）(Simplified Chinese) is characters modified by China, and used in China now. Lemmy ui uses it as 中文.
2. When Accept-Language (or browser language? I'm not sure.) is "zh-Hant", zh_Hant should be selected. http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

Nutomic:
Thanks for the report! Im a bit confused by the description though, and dont understand which language names are correct. You can have a look at the [full language list](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/i18next.ts#L66), and even make a pull request to fix it.

About the second issue, which language is actually selected in your case?

ianiiaannn:
After analyze the code, I found lemmy reads navigator.languages. But in my browser ,navigator.languages return zh-TW(Taiwan, where used to use Traditional Chinese) instead of zh-Hant.
Maybe we should try alias?

```>navigator.languages
(4) ['zh-TW', 'en-US', 'zh', 'en']
0: "zh-TW"
1: "en-US"
2: "zh"
3: "en"
```
Browser return - but Lemmy's language code use _, so the filter can't find right language and select less specific "zh".

dessalines:
The browser not selecting the correct language, might be an issue with our `getLanguage` function. I'll test this soon.

We don't have zh_TW yet, so that might be a separate issue.

ianiiaannn:
zh-TW and zh-HK (region) uses zh_Hant (language).
https://datatracker.ietf.org/doc/html/rfc5646
Maybei 18next.ts can be cahnged like this?
```{ resource: zh_Hant, code: "zh_Hant", name: "中文 (繁體)", tags:["zh-TW", "zh-HK"] },```

My environment is Chrome Version 102.0.5005.63 (Official Build) (64-bit).

dessalines:
Is this still an issue? Can re-open if it is.

PR #725, opened 2022-07-26
Use zh-TW for language code, instead of zh_Hant.

sonata-chen:
This pull request fixed issue [#667](https://github.com/LemmyNet/lemmy-ui/issues/667).

### DIFF
--- a/src/shared/i18next.ts
+++ b/src/shared/i18next.ts
@@ -63,7 +63,7 @@ export const languages = [
   { resource: sv, code: "sv", name: "Svenska" },
   { resource: vi, code: "vi", name: "Tiếng Việt" },
   { resource: zh, code: "zh", name: "中文 (简体)" },
-  { resource: zh_Hant, code: "zh_Hant", name: "中文 (繁體)" },
+  { resource: zh_Hant, code: "zh-TW", name: "中文 (繁體)" },
 ];
 
 const resources: Resource = {};


sonata-chen:
This pull request fix issue [#667](https://github.com/LemmyNet/lemmy-ui/issues/667).

src/shared/i18next.ts
```diff
@@ -63,7 +63,7 @@ export const languages = [
   { resource: sv, code: "sv", name: "Svenska" },
   { resource: vi, code: "vi", name: "Tiếng Việt" },
   { resource: zh, code: "zh", name: "中文 (简体)" },
-  { resource: zh_Hant, code: "zh_Hant", name: "中文 (繁體)" },
+  { resource: zh_Hant, code: "zh-TW", name: "中文 (繁體)" },
```
dessalines, line 66: Must use an underscore.

dessalines (changes requested):

src/shared/i18next.ts
sonata-chen: Could you elaborate more why we must use an underscore here?
I have tested on my machine. It seems that it's not OK to use underscores for both zh_TW and  pt_BR.
Lemmy (without a login) will not pick the correct language based on browser settings.

sonata-chen: My browser setting
`navigator.languages:   [ "zh-TW", "en-US", "en" ]`

dessalines: Ah... I was under the impression it needed one, coulda swore I tested that with pt-BR. In that case I'll trust your judgement here, thx!

dessalines (approved):

PR #737, opened 2022-08-03
Change CSP rule for connect-src (websocket) to wildcard (fixes #730)

### LINKED ISSUE #730: Lemmy almost completely broken on iOS and MacOS devices
Lemmy on iOS and MacOS devices is broken a lot of way. I think it's couse some JS error, but I'am not a developer.
E.g.: 
- Only could open a post after you click two times (!) on the link.
- Login is not work.
- Registration not work.
- Up/downvote not work.
- Top menu is not work.
- Logo/website name is unclickable, so you can't go back to the frontpage 
- etc.

I think nothing work on iOS and MacOS what is javascript-like on the website.

I checked from Safari browser and it show the follow errors:
- TypeError: undefined is not an object (evaulating 'this.ws.send') // error location is /static/js/client.js file.
- TypeError: null is non an object (evaulating '.$Li') // error location is /static/js/client.js file too.

However, this error is not affected every Lemmy site: so as I seen e.g. Lemmy.ml and Feddit.de is work fine on Apple devices, but mander.xyz, beehaw.org, lemmy.rollenspiel.monster, etc. is no.

I install Lemmy to my VPS too with Docker, and I have this JS problem on Apple devices too. Have you any idea what's the problem or how can I solve it. What couse is Lemmy.ml works fine while some others is no? Thanks for your help, I really like Lemmy and would like to solve this problem on my VPS too.

P.s.: I read in other github issue is somebody in Dev is not have Apple devices, in this cause I recommend this website: https://live.browserstack.com/dashboard
At here after a fast and free registration you can check any website from any devices, Apple devices too. I checked this problem from here too, I do not have MacOS too. 

I made some screenshots from the error, maybe help:

Nutomic:
Also change the env var name to LEMMY_UI_DISABLE_CSP as debug with csp generally works fine now, and this name is much clearer.

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -13,7 +13,7 @@ import process from "process";
 import serialize from "serialize-javascript";
 import { App } from "../shared/components/app/app";
 import { SYMBOLS } from "../shared/components/common/symbols";
-import { httpBaseInternal, wsUriBase } from "../shared/env";
+import { httpBaseInternal } from "../shared/env";
 import {
   ILemmyConfig,
   InitialFetchRequest,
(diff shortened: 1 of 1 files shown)


dessalines (approved):

Nutomic:
Might be worth another patch release.

PR #743, opened 2022-08-05
Disable CSP when in debug mode.

dessalines:


Nutomic:
Why?

dessalines:
Eruda needs to run a remote script to work, otherwise just setting LEMMY_UI_DEBUG won't work on its own.

Nutomic:
Ah but LEMMY_UI_DEBUG wont be set automatically when you just run yarn start, right? In that case its fine, but the name is confusing. I would change it to LEMMY_UI_ERUDE_DEBUG.

dessalines:
Correct, you still need to set that one env var now. It should pry stay as just `LEMMY_UI_DEBUG` because I don't know what else I might add besides eruda in the future to help debug.

PR #744, opened 2022-08-05
Adding mod / admin distinguish.

dessalines:
Looks like this:

Screenshot_2022-08-05-12-22-09-836_mark via gp

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "eslint-plugin-prettier": "^4.2.1",
     "husky": "^8.0.1",
     "import-sort-style-module": "^6.0.0",
-    "lemmy-js-client": "0.17.0-rc.39",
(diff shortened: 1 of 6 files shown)


Nutomic:
So other users will see this post as "distinguised" with the same shield icon? Imo the icon doesnt make any sense. I would go with something like a yellow/golden background for the entire comment (and another icon for the button, or just use text).

dessalines:
> So other users will see this post as "distinguised" with the same shield icon? 

They'll see both the shield, and the comment has the `mark` class, highlighting it a bit. I tried various card highlights, and all of them looked really ugly. 

I could possibly add text or a different icon, but I dont' know what would work. The strings I added are "speak as mod" and "speak as user".

PR #762, opened 2022-08-26
Default translation

Bublikhead:
* Ukrainian is default language 
* English and Russian still can be selected
* Other languages are removed


Bublikhead:
Wrong detestation

PR #784, opened 2022-09-19
Cantarell for darkly/darkly-red. Fixes #779

### LINKED ISSUE #779: Use Cantarell instead of Liberation Sans for GNOME users on darkly/darkly-red themes.

When I use Lemmy on my computers, the typeface displayed is Liberation Sans on the `darkly` and `darkly-red` themes. It is not a bad option per sé, but in Windows the font shown is the one used in the whole operating system (Segoe UI) and it would be nice if the same thing happened with GNOME to ensure that the interface is as consistent as possible and as close to the native interface as possible. For Plasma users it is enough to use Noto Sans too.

The CSS file would look like this:

```CSS
body {
	font-family: Lato,-apple-system,BlinkMacSystemFont,"Segoe UI",Roboto,"Helvetica Neue", Cantarell, Noto Sans, Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol";
	font-size: .9375rem;
	font-weight: 400;
	line-height: 1.5;
	color: #dee2e6;
	text-align: left;
}
```

**Screenshot showing default font**
darklywithdefaultfont

**Screenshot showing Cantarell font (my proposed change)**
darklywithcantarellfont

I am aware that Cantarell is hated by many users and perhaps my wish conflicts with the wishes of other users, so perhaps the best thing would be to let the user choose the one they prefer, using a method similar to what DuckDuckGo does, which is to give a few options to choose from and give the option to indicate a specific font (whatever the user has installed locally).

ddgimplementation

ddgimplementationofcustomfonts

### Additional context

The intention of this change is that those of us who use GNOME will feel more comfortable navigating Lemmy, creating a more consistent experience.

IngrownMink4:
More info about this change here: https://github.com/LemmyNet/lemmy-ui/issues/779


dessalines:
Looks good, thanks.

dessalines (approved):

PR #811, opened 2022-09-30
Fix browser popup warning when deleting image

sam365724:
Small fix for avoiding an additional popup when clicking on image delete.
Also handles error case when deleting image fails.
Adds the user supplied image filename for clear identification (this will be relevant for multiple image upload coming).

image


sam365724:
@dessalines How do i create entries for newly created language strings?

sam365724:
> @dessalines How do i create entries for newly created language strings? Can you please add: ` "failed_to_delete_picture": "Failed to delete picture.",`

https://github.com/LemmyNet/lemmy-translations/pull/29

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -386,8 +386,9 @@ export class MarkdownTextArea extends Component<
           let textarea: any = document.getElementById(i.id);
           autosize.update(textarea);
           pictrsDeleteToast(
-            i18n.t("click_to_delete_picture"),
-            i18n.t("picture_deleted"),
+            i18n.t("click_to_delete_picture").concat(' (', file.name,')'),
```
dessalines, line 389: Use string literals, not concat. I'd prefer these to be colons as well, so:

```
`${i18n.t("click_to_delete_picture)}: ${file.name}`
```

Same for the other two.

src/shared/utils.ts
```diff
@@ -552,9 +553,16 @@ export function pictrsDeleteToast(
       duration: 10000,
       onClick: () => {
         if (toast) {
-          window.location.replace(deleteUrl);
-          alert(deletePictureText);
-          toast.hideToast();
+          fetch(deleteUrl, {})
```
dessalines, line 556: Why are you doing a fetch here, doesn't it delete the picture elsewhere? This should just be the toast.

src/shared/components/common/markdown-textarea.tsx
sam365724: Alright, we need to update translations then. I'll do a pull request.

sam365724: I did this change in the asyn image upload branch now though.

src/shared/utils.ts
sam365724: No, it doesn't here we had a simple window.location() which the browser does not like.

dessalines: In that case use a callback function for onClick, because this should really just have toast-related code.

sam365724: I don't really understand the difference, can you make the change or a proposal?

dessalines:
```ts
export function pictrsDeleteToast(...
 onClick: () => void
) {
...
  onClick: () => {
    toast.hideToast();
    onClick();
  }
}
...
pictrsDeleteToast(.... () => { do fetching stuff here }
```

sam365724: I think you misunderstand the logic here. The picture is deleted when the user clicks on the toast message itself. So it must be inside onClick?! Please just fix it yourself, I'm losing interest like this. It's not worth it for such small but annoying bug. :(

PR #817, opened 2022-10-04
Support new video embed api format (fixes #709)

### LINKED ISSUE #709: Support for video embeds
[This change](https://github.com/LemmyNet/lemmy/pull/2261) removed API field post.embed_html, and added embed_url instead. This should work for Youtube, Peertube and and other sites that set the opengraph attribute in their html.

So lemmy-ui should show a video embed whenever that field is set. At the moment the implementation is a bit strange, because it only shows the embed if post.embed_title is set, which isnt always the case. I would also suggest to change the "plus" button behind which the embed is hidden, because its not clear at all what it does. Maybe just change it to a text label "watch video".

Nutomic:


src/shared/components/post/metadata-card.tsx
```diff
@@ -76,9 +76,8 @@ export class MetadataCard extends Component<
                             <button
                               className="mt-2 btn btn-secondary text-monospace"
                               onClick={linkEvent(this, this.handleIframeExpand)}
-                              data-tippy-content={i18n.t("expand_here")}
                             >
-                              {this.state.expanded ? "-" : "+"}
+                              {i18n.t("expand_here")}
```
Nutomic, line 80: Only showing `+` on the button was confusing, so i changed it to show the text directly (not in tooltip). Also `-` was never shown because the button gets hidden when embed is displayed.

dessalines (approved):

PR #825, opened 2022-10-20
Search button and input style fixes

### LINKED ISSUE #500: Nav bar search input field very broken
### Issue Summary

Something in the new release broke the nav bar search input field. It has three issues:

1. When opened it pushes the layout in a weird way. 
2. The magnifying class icon is slightly darker than the other nav bar icons and the padding is smaller on the left than it is on the right.
3. The input field is actually selectable without it being expanded, which is weird.

### Steps to Reproduce

1. Log onto a Lemmy account (you need to be logged for the icons to appear and the layout shift to happen).
2. Click the search icon on the navigation bar.

### GIF

Peek 2021-11-20 23-46

rgroothuijsen:
This is a set of style fixes for the search button and its associated input fields, based on the issues brought up in #500. 

- The search input jumps out of place when expanded: I could only reproduce this one by making my browser screen as small as possible without making the search button disappear, which can technically still occur with the proposed solution. Nonetheless, I thought it could be mitigated a small bit by making the search bar width dependent on view width.
- Search button is misaligned compared to the other navbar elements: This was fixed by reducing the margin on the right of the button.
- Search button has the wrong color: This was fixed by wrapping the existing HTML in the same list tags and CSS classes as all the other navigation bar items.
- Input field still selectable and editable when hidden: This was fixed by explicitly setting the `disabled` attribute on the text input when it is hidden. This prevents it from being either selectable or editable.

Fixes #500.

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -306,7 +306,7 @@ pre {
 }
 
 .show-input {
-  width: 13em !important;
+  width: 13vw !important;
 }
 .hide-input {
   background: transparent !important;
--- a/src/shared/components/app/navbar.tsx
+++ b/src/shared/components/app/navbar.tsx
@@ -314,35 +314,39 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
             {!this.context.router.history.location.pathname.match(
               /^\/search/
             ) && (
-              <form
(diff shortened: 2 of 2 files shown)


dessalines (approved):
Looks great, thanks!

Issue #819, opened 2022-10-06
Can't log out from account within a private window session.


zksmk:
### Expected behavior: 

Log in into account in a private window. Then log out. You're logged out.

### What happens: 

Log in into account in a private window. Then log out. You're still logged in, plus the UI is a bit glitchy about it.

### Why is this a problem: 

In general, this never happens with any other website I log into in a browser private window. It's not expected behavior. It's kinda bad from a privacy/security perspective. If you do this on somebody else's PC.

For me, personally, it's not a huge problem. However, I frequently browse lemmy/internet from a private window, and then get compelled to log in to comment/vote, where I have some communities blocked, for reasons. And then sometimes I want to log out, to see blocked communities again, for example. And I don't want to close/reopen the private window because I have other websites/tabs open. Opening a new private window without closing the first one doesn't fix the reported issue.

### Technical details:

Browser: Firefox 91.9esr
OS: Windows 8.1

dessalines:
I can't seem to verify, and I just tested this both on FF and chrome private tabs. 

Other people have seen this bug tho too, but I need to see a console, or have some way to reproduce it, to be able to fix it.

zksmk:
I can't seem to reproduce it consistently. It happens about a third or fourth of the time. I think it happens when I first login from a thread instead of the frontpage, and then try to log out, but I really don't know.

This is what the console showed when I did just that:

`Exception { name: "NS_ERROR_NOT_AVAILABLE", message: "Component returned failure code: 0x80040111 (NS_ERROR_NOT_AVAILABLE) [nsIDocShell.domWindow]", result: 2147746065, filename: "resource://devtools/server/actors/targets/browsing-context.js", lineNumber: 413, columnNumber: 0, data: null, stack: "get window@resource://devtools/server/actors/targets/browsing-context.js:413:5\n_windowReady@resource://devtools/server/actors/targets/browsing-context.js:1391:24\nwatch@resource://devtools/server/actors/targets/browsing-context.js:1666:25\n_onDocShellCreated/<@resource://devtools/server/actors/targets/browsing-context.js:862:32\nexports.makeInfallible/<@resource://devtools/shared/ThreadSafeDevToolsUtils.js:103:22\n", location: XPCWrappedNative_NoHelper }
ThreadSafeDevToolsUtils.js:82:13`

[...]

PR #836, opened 2022-10-29
Fix workaround for broken logout

sam365724:
Use hostname without port for setting empty cookie, avoids error in browser:  Cookie “jwt” has been rejected for invalid domain. Fixes https://github.com/LemmyNet/lemmy-ui/issues/819


dessalines (approved):
Looks good, thanks.

PR #854, opened 2022-11-17
Add Taglines support

makotech222:
See: https://github.com/LemmyNet/lemmy/pull/2548

Adds frontend support for Taglines.

Taglines are edited and displayed as Markdown. Screenshots Below:

Homepage:
image
image
image

Admin Settings:
image

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -386,3 +386,8 @@ br.big {
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
 }
+
+#tagline{
+
+  padding: 0px 0px 15px 0px;
+}
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -386,8 +386,3 @@ br.big {
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
 }
-
-#tagline{
-
-  padding: 0px 0px 15px 0px;
-}
--- a/src/shared/components/home/home.tsx
+++ b/src/shared/components/home/home.tsx
@@ -335,7 +335,10 @@ export class Home extends Component<any, HomeState> {
         {this.state.siteRes.site_view.local_site.site_setup && (
           <div className="row">
             <main role="main" className="col-12 col-md-8">
-              {this.state.tagline.isSome() && <div id="tagline" dangerouslySetInnerHTML={mdToHtml(this.state.tagline.unwrapOr(""))}></div>}
+              {this.state.tagline.match({
+                some: tagline => <div id="tagline" dangerouslySetInnerHTML={mdToHtml(tagline)}></div>,
+                none: <></>,
+              })}
               <div className="d-block d-md-none">{this.mobileView()}</div>
               {this.posts()}
             </main>
--- a/src/shared/components/home/home.tsx
+++ b/src/shared/components/home/home.tsx
@@ -53,6 +53,7 @@ import {
   getSortTypeFromProps,
   isBrowser,
   isPostBlocked,
+  mdToHtml,
   notifyPost,
   nsfwCheck,
   postToCommentSortType,
@@ -94,6 +95,7 @@ interface HomeState {
   showSidebarMobile: boolean;
   subscribedCollapsed: boolean;
   loading: boolean;
+  tagline: string;
 }
 
 interface HomeProps {
@@ -137,6 +139,7 @@ export class Home extends Component<any, HomeState> {
     dataType: getDataTypeFromProps(this.props),
     sort: getSortTypeFromProps(this.props),
     page: getPageFromProps(this.props),
+    tagline: ""
   };
 
   constructor(props: any, context: any) {
@@ -170,10 +173,12 @@ export class Home extends Component<any, HomeState> {
           wsClient.communityJoin({ community_id: 0 })
         );
       }
+      const taglines = this.state.siteRes.site_view.taglines;
       this.state = {
         ...this.state,
         trendingCommunities: trendingRes.communities,
         loading: false,
+        tagline: taglines.length == 0 ? "" : taglines[Math.floor(Math.random() * taglines.length)].content,
       };
     } else {
       this.fetchTrendingCommunities();
@@ -329,6 +334,7 @@ export class Home extends Component<any, HomeState> {
         {this.state.siteRes.site_view.local_site.site_setup && (
           <div className="row">
             <main role="main" className="col-12 col-md-8">
+              <div id="tagline" dangerouslySetInnerHTML={mdToHtml(this.state.tagline)}></div>
               <div className="d-block d-md-none">{this.mobileView()}</div>
               {this.posts()}
             </main>
--- a/src/shared/components/home/home.tsx
+++ b/src/shared/components/home/home.tsx
@@ -50,6 +50,7 @@ import {
   getDataTypeFromProps,
   getListingTypeFromProps,
   getPageFromProps,
+  getRandomFromList,
   getSortTypeFromProps,
   isBrowser,
   isPostBlocked,
@@ -95,7 +96,7 @@ interface HomeState {
   showSidebarMobile: boolean;
   subscribedCollapsed: boolean;
   loading: boolean;
-  tagline: string;
+  tagline: Option<string>;
 }
 
 interface HomeProps {
@@ -139,7 +140,7 @@ export class Home extends Component<any, HomeState> {
     dataType: getDataTypeFromProps(this.props),
     sort: getSortTypeFromProps(this.props),
     page: getPageFromProps(this.props),
-    tagline: ""
+    tagline: None
   };
 
   constructor(props: any, context: any) {
@@ -178,7 +179,7 @@ export class Home extends Component<any, HomeState> {
         ...this.state,
         trendingCommunities: trendingRes.communities,
         loading: false,
-        tagline: taglines.length == 0 ? "" : taglines[Math.floor(Math.random() * taglines.length)].content,
+        tagline: taglines.map(tls => getRandomFromList(tls).content)
       };
     } else {
       this.fetchTrendingCommunities();
@@ -334,7 +335,7 @@ export class Home extends Component<any, HomeState> {
         {this.state.siteRes.site_view.local_site.site_setup && (
           <div className="row">
             <main role="main" className="col-12 col-md-8">
-              <div id="tagline" dangerouslySetInnerHTML={mdToHtml(this.state.tagline)}></div>
+              {this.state.tagline.isSome() && <div id="tagline" dangerouslySetInnerHTML={mdToHtml(this.state.tagline.unwrapOr(""))}></div>}
               <div className="d-block d-md-none">{this.mobileView()}</div>
               {this.posts()}
             </main>
--- a/src/shared/components/home/home.tsx
+++ b/src/shared/components/home/home.tsx
@@ -174,7 +174,7 @@ export class Home extends Component<any, HomeState> {
           wsClient.communityJoin({ community_id: 0 })
         );
       }
-      const taglines = this.state.siteRes.site_view.taglines;
+      const taglines = this.state.siteRes.taglines;
       this.state = {
         ...this.state,
         trendingCommunities: trendingRes.communities,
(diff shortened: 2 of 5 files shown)


src/shared/components/home/site-form.tsx
```diff
@@ -1141,6 +1195,36 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
     this.setState(s => ((s.siteForm.legal_information = Some(val)), s));
   }
 
+  handleTaglineChange(i: SiteForm, index: number, val: string) {
+    let taglines = i.state.siteForm.taglines.unwrap();
+    taglines[index] = val;
+    i.state.siteForm.taglines = Some(taglines);
```
makotech222, line 1201: pre-emptively asking, is there a better way to modify an Option<string[]> by index beyond unwrapping and setting it directly like this?

src/assets/css/main.css
```diff
@@ -386,3 +386,8 @@ br.big {
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
 }
+
+#tagline{
+
+  padding: 0px 0px 15px 0px;
```
dessalines, line 392: This doesn't seem necessary, why not use bootstraps `padding-` classes? The less custom css the better.

src/shared/components/home/home.tsx
```diff
@@ -94,6 +95,7 @@ interface HomeState {
   showSidebarMobile: boolean;
   subscribedCollapsed: boolean;
   loading: boolean;
+  tagline: string;
```
dessalines, line 98: Should probably be an Option.

src/shared/components/home/home.tsx
```diff
@@ -137,6 +139,7 @@ export class Home extends Component<any, HomeState> {
     dataType: getDataTypeFromProps(this.props),
     sort: getSortTypeFromProps(this.props),
     page: getPageFromProps(this.props),
+    tagline: ""
```
dessalines, line 142: `tagline: None`

src/shared/components/home/home.tsx
```diff
@@ -170,10 +173,12 @@ export class Home extends Component<any, HomeState> {
           wsClient.communityJoin({ community_id: 0 })
         );
       }
+      const taglines = this.state.siteRes.site_view.taglines;
       this.state = {
         ...this.state,
         trendingCommunities: trendingRes.communities,
         loading: false,
+        tagline: taglines.length == 0 ? "" : taglines[Math.floor(Math.random() * taglines.length)].content,
```
dessalines, line 181: Add a `getRandomFromList(` function to utils.tsx .

Then if this is an `Option<Vec<Tagline>>`, you could probably do something like:

`tagline: taglines.map(tls => getRandom(tls).map(tl -> tl.content))` 

Something like that.

src/shared/components/home/site-form.tsx
```diff
@@ -159,6 +160,9 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
         blocked_instances: this.props.siteRes.federated_instances.andThen(
           f => f.blocked
         ),
+        taglines: Some(
```
dessalines, line 163: Why is Some used here, since tagline isn't required and might be missing?

src/shared/components/home/site-form.tsx
```diff
@@ -1032,6 +1036,57 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+                    .unwrapOr([])
+                    .map((cv, index) => (
+                      <tr key={cv}>
+                        <td>
+                          <MarkdownTextArea
+                            initialContent={Some(cv)}
+                            initialLanguageId={None}
+                            placeholder={None}
+                            buttonTitle={None}
+                            maxLength={None}
```
dessalines, line 1055: I'm good with not having a maxLength, its an admin thing anyway.

src/shared/components/home/site-form.tsx
```diff
@@ -1032,6 +1036,57 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+                            onContentChange={s =>
+                              this.handleTaglineChange(this, index, s)
+                            }
+                            hideNavigationWarnings
+                            allLanguages={[]}
```
dessalines, line 1060: allLanguages is coming back from `GetSiteResponse`

src/shared/components/home/site-form.tsx
dessalines: `.match` might work:

```ts
let taglines = i.state.siteForm.taglines.match({
  some: tls => { tls[index] = val; },
  none: void 0
});
i.state.siteForm.taglines = taglines;
```

If that doesn't work, at least do a `.isSome()` before doing the `unwrap()`... unwraps are always dangerous and can crash the UI on missing values.

src/shared/components/home/site-form.tsx
```diff
@@ -1141,6 +1195,36 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+    let taglines = i.state.siteForm.taglines.unwrap();
+    taglines[index] = val;
+    i.state.siteForm.taglines = Some(taglines);
+    i.setState(i.state);
+  }
+
+  handleDeleteTaglineClick(
+    i: SiteForm,
+    index: number,
+    event: InfernoMouseEvent<HTMLButtonElement>
```
dessalines, line 1208: I def agree with wanting to strictly type these events, but is MouseEvent correct? Maybe there's a ClickEvent. I'm thinking of phones and tablets that don't use mouse events.

src/shared/components/home/site-form.tsx
```diff
@@ -1141,6 +1195,36 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+  ) {
+    event.preventDefault();
+    let taglines = i.state.siteForm.taglines.unwrap();
```
dessalines, line 1211: No unwrap here, or if you do, check that it `isSome()` first.

src/shared/components/home/site-form.tsx
```diff
@@ -1141,6 +1195,36 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+    i.state.siteForm.taglines = Some(taglines);
+    i.setState(i.state);
+  }
+
+  handleAddTaglineClick(
+    i: SiteForm,
+    event: InfernoMouseEvent<HTMLButtonElement>
+  ) {
+    event.preventDefault();
+    let taglines = i.state.siteForm.taglines.unwrap();
```
dessalines, line 1222: same

src/shared/components/home/site-form.tsx
stevencherry-uu: This is what the signature says on the click event

src/shared/components/home/site-form.tsx
makotech222: In this case, the handleChange/delete events will only be applicable on an already existing tagline, so it will always have some. Ill try out the match though and let you know how it goes.

src/assets/css/main.css
makotech222: good idea, wasn't aware of available styles

src/shared/components/home/home.tsx
makotech222: Out of curiosity, wouldn't just allowing empty string be simpler? no need to match in the html on None/Some to handle displaying differently?

src/shared/components/home/site-form.tsx
makotech222: The click event signature specifies it as InfernoMouseEvent, I think it should be fine, touch will probably also include the preventDefault method

src/assets/css/main.css
dessalines: No probs.

src/shared/components/home/home.tsx
dessalines: javascript null checking is a weird nightmare, people even have to write helper functions for empty / blank string checks. Much safer to use the monad options wherever possible for missing / null values.

src/shared/components/home/site-form.tsx
dessalines: I get it, because an empty string *sort of* denotes missing data. But to me, its much more logically consistent to use `Option` when a value is missing, then front ends don't have to do array length checks specifically for array types, as compared to other types. 

I went through the process recently of replacing all javascript nullpointerexception possibilities with `Option` monads, and it eliminated a whole class of compile-time bugs.

src/shared/components/home/site-form.tsx
dessalines: So there's no `InfernoClickEvent` that would cover both cases?

makotech222:
Okay I think I fixed all the issues. Have another look when you get the chance.

src/shared/components/home/home.tsx
```diff
@@ -334,7 +335,7 @@ export class Home extends Component<any, HomeState> {
         {this.state.siteRes.site_view.local_site.site_setup && (
           <div className="row">
             <main role="main" className="col-12 col-md-8">
-              <div id="tagline" dangerouslySetInnerHTML={mdToHtml(this.state.tagline)}></div>
+              {this.state.tagline.isSome() && <div id="tagline" dangerouslySetInnerHTML={mdToHtml(this.state.tagline.unwrapOr(""))}></div>}
```
dessalines, line 338: Use `.match` here, you'll find a ton of examples in the code.

src/shared/components/home/site-form.tsx
```diff
@@ -160,9 +160,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
         blocked_instances: this.props.siteRes.federated_instances.andThen(
           f => f.blocked
         ),
-        taglines: Some(
-          this.props.siteRes.site_view.taglines.map(x => x.content)
-        ),
+        taglines: this.props.siteRes.site_view.taglines.map(x => x.map(y => y.content)),
```
dessalines, line 163: Perfect.

src/shared/components/home/site-form.tsx
```diff
@@ -1196,9 +1194,10 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
   }
 
   handleTaglineChange(i: SiteForm, index: number, val: string) {
-    let taglines = i.state.siteForm.taglines.unwrap();
-    taglines[index] = val;
-    i.state.siteForm.taglines = Some(taglines);
+    i.state.siteForm.taglines.match({
+      some: tls => { tls[index] = val; },
+      none: void 0
+    });
```
dessalines, line 1200: Make sure to test this, because I'm not totally sure that it works.

src/shared/components/home/site-form.tsx
```diff
@@ -1208,20 +1207,25 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
-    taglines.splice(index, 1);
-    i.state.siteForm.taglines = Some(taglines);
-    i.setState(i.state);
+    if (i.state.siteForm.taglines.isSome()){
+      let taglines = i.state.siteForm.taglines.unwrap();
+      taglines.splice(index, 1);
+      i.state.siteForm.taglines = None; // force rerender of table rows
+      i.setState(i.state);
+      i.state.siteForm.taglines = Some(taglines);
+      i.setState(i.state);
```
dessalines, line 1216: No need for two setStates here, unless something really weird is going on.

dessalines (commented):
Getting close. I've published `"lemmy-js-client": "0.17.0-rc.52" , with the tagline stuff added.

src/shared/components/home/site-form.tsx
makotech222: Unfortunately there is. The key for the repeater in the html is 'index', so when I delete a row, its possible that certain rows don't re-render, causing issues with displaying the markdown editor.

src/shared/components/home/site-form.tsx
makotech222: Yup, i've confirmed it working

src/shared/components/home/home.tsx
makotech222: okay fixed this one too.

src/shared/components/home/site-form.tsx
dessalines: Gotcha

src/shared/components/home/site-form.tsx
dessalines: Looks like there isn't. I'll test this on mobile anyway.

dessalines (commented):
Looks good, but you still need to update the package.json to use the new RC, so that CI will pass.

dessalines (approved):
Looks good. Once lemmy-js-client is fixed then I'll cut a new RC and merge this.

dessalines (approved):
Tested this, and it works great! Nice job.

Issue #783, opened 2022-09-19
Language tags for communities/sites


Nutomic:
- [x] allowed languages are shown in site/community sidebar (user languages are not publicly shown)
- [x] admins/mods can edit languages for sites/communities (using the same input as in user profile)
- [ ] during post/comment creation, language selection is limited to intersection between user languages and community languages. If that leaves only a single language, hide the input field (same logic as [here](https://github.com/LemmyNet/lemmy/pull/2434/files#diff-4066d2ddc688e352b1d468ce7f267b6bcc94943ae6475c72505948f5932beb22R201))

see #771

dessalines:
This is done now.

Nutomic:
Where is the code for this?

dessalines:
I'll check again today

dessalines:
I'm not exactly sure how I'm going to solve the "limit language selection when creating a post or comment" one... because `discussion_languages` only comes back from `GetCommunity`. Just too many complications to handle right now, so I'm just going to limit it to their user language, and let the backend reject incorrect community languages. Scenarios like:

- What if the persons replying from their inbox, or from someone's profile page? That'd require fetching tons of communities.
- On the CreatePost page, they have a list of communities, but it doesn't have the languages for each one. That would require an additional fetch.
- The post page doesn't have community_languages either, just the community_view. So we'd have to add that to the API.

CreatePost only runs "ListCommunities" or "SearchCommunity", which only provides the `CommunityView`s, not the set of languages for each community.

dessalines:
What currently works:

- Setting site languages
- Setting profile languages
- Setting community languages
- Create community shows all site langs

What doesn't work / is halfway broken:

- Saving the community won't live-update the community langs
- Saving the site won't live-update the site langs
- Creating a post or comment only does the intersection of the site and your langs, not the community langs. This is going to be a difficult one, as many areas of the UI are multi-community, like inbox, profile, etc. That would require selecting a community, then fetching it to get its langs.
- The language dropdown should be moved into the options bar for the markdown fields, because right now it takes up too much space.

poVoq:
Since Mastodon and Akkoma (Pleroma fork) seem to have build in translators now (using Libretranslate!) I think it might be good to see how they tag the language of a post and make this compatible.

Nutomic:
This is already implemented.

PR #868, opened 2022-11-22
Adding Community Language fixes. #783

dessalines:


src/shared/components/common/language-select.tsx
```diff
@@ -104,6 +82,40 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
...
+      this.props.showAll,
+      this.props.showSite,
+      UserService.Instance.myUserInfo
+    );
+
+    return (
+      <select
+        className={classNames("lang-select-action", {
+          "form-control custom-select": !this.props.iconVersion,
+          "btn btn-sm text-muted": this.props.iconVersion,
```
jsit, line 99: @dessalines Does this need `btn` classes here? It makes it look a little strange with the transparent border etc.:

Screenshot 2023-06-04 at 4 25 58 PM

dessalines: Probably not, I'm open to changes if something looks better.

PR #873, opened 2022-12-01
Add FeaturedPosts support

makotech222:
See https://github.com/LemmyNet/lemmy/pull/2585

Adds a new 'Pin' icon to pin to Local view

image

Marking as draft, as its still a bit unfinished. 

Also, soliciting feedback on what the Icon should look like. I'm not a fan of both 'pin to community' and 'pin to local' having the same icon. Not sure what other icons are available currently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "inferno-server": "^8.0.3",
     "isomorphic-cookie": "^1.2.4",
     "jwt-decode": "^3.1.2",
-    "lemmy-js-client": "0.17.0-rc.54",
+    "lemmy-js-client": "0.17.0-rc.56",
     "markdown-it": "^13.0.1",
     "markdown-it-container": "^3.0.0",
     "markdown-it-footnote": "^3.0.3",
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -584,7 +584,7 @@ export class Community extends Component<any, State> {
       op == UserOperation.DeletePost ||
       op == UserOperation.RemovePost ||
       op == UserOperation.LockPost ||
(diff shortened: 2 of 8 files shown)


makotech222:
One last design issue I have currently is when to display the 'pin' icon when displaying the post. We can view a Post from multiple places throughout the app, Home, Community, Search, etc. Currently, if a post is 'Featured in Community', it will show the pin icon, even when viewing from outside the community. 

Should we differentiate when viewing a post from multiple places? Should I just display the 'pin' icon if its either community or local featured? 

image

makotech222:
@dessalines tagging you here, sorry if its doubled up. for the question above.

makotech222:
@Nutomic tagging, any thoughts on the two issues above?

dessalines:
Hrm... it doesn't matter to me, but probably not text. Different color pin icons should work, text-primary and text-secondary.

makotech222:
Okay ill try different color pins. How about the other issue? On showing pinned icon?

dessalines:
It pry makes sense to show the community pin, even if you're viewing it from the front page. No need to create rules around when to show it IMO.

makotech222:
Okay, here's text-primary for community feature, text-secondary for local feature:

image

dessalines:
Looks fine to me, but maybe they should be hidden behind the 3-dot menu. Up to you tho.

makotech222:
The three-dot button is already expanded, and the pin icons only show when expanded

dessalines:
Yeah text with icon is fine.

makotech222:
Okay should be ready to go then! Thanks!

dessalines (approved):
Looks good, thx!

PR #877, opened 2022-12-11
Add Custom Emojis Support

makotech222:
This PR adds the custom emoji functionality from https://github.com/LemmyNet/lemmy/issues/2419#issuecomment-1334440588

In the frontend, I've added the following:

* Update post-listing to display a new markdown icon to open emoji picker
 
image

* Update markdown-textarea to handle emoji-picker
* Update app startup to build emoji list from siteres
* Update Tribute component to include custom emojis
image

* Tabify Site Settings page
image

* Add Custom Emoji admin tab for add/edit/delete custom emojis. Clicking on emoji in picker will scroll to entry in table.
image

Deficiencies:
* Bug on emoji-mart: (https://github.com/missive/emoji-mart/issues/756). Functionally, this means the picker can't update when emojis are being changed, unless page is refreshed. Not a show stopper, just annoying.
* Bug on emoji-mart: (https://github.com/missive/emoji-mart/issues/757) Picker blows up if emoji list is empty. Very minor and worked around

Open Questions:
* Better icon for 'save' in the table? Currently using 'edit' icon.
* Is including Image Url column necessary? Allows user to link to external images, but takes up space. I assume most people will just upload images to their server.


Nutomic:
I think a good solution would be to split the site settings page into different tabs.

dessalines:
Question: is "Category" for emoji-mart? So these all wouldn't just go in a single "custom" category, but potentially many custom categories?

> If we go this route, I think it would be prudent to add api endpoints for PUT/POST/DELETE of the custom emoji entities, rather than use Site.

Sounds good to me.

> Currently, I am just adding a <script> tag to the server's html to load the emoji-mart library. Seeing as how we don't do this for any other library, I imagine we would want to change this. 

Fairly sure you can define that custom library location in the package.json and it will download it to node_modules. If that doesn't work, you can reference the git repo, if the js file is hosted there.

makotech222:
Yeah, Category is something for emoji-mart. You can add emojis to custom categories and it will group them.

makotech222:
Okay with the latest commit, the emoji picker will place a markdown image, with its 'title' attr set to the shortcode. When the markdown renderer encounters an image, it will check if it is a custom emoji, and if so it will add the 'icon-emoji' css class, which limits the size of the image.

dessalines:
I'm just getting around to testing this now. There doesn't seem to be a save button for adding the custom emojis.

src/shared/components/home/admin-settings.tsx
```diff
@@ -112,21 +114,47 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
...
+                  className={`nav-link btn ${this.state.currentTab == "site" && "active"}`}
+                  onClick={() => { this.handleSwitchTab("site"); }}>
+                  {i18n.t("site")}
+                </button>
+              </li>
+              <li className="nav-item">
+                <button
+                  className={`nav-link btn ${this.state.currentTab == "emojis" && "active"
+                    }`}
+                  onClick={() => { this.handleSwitchTab("emojis"); }}>
```
dessalines, line 134: Use linkEvent here. Check `settings.tsx` for an example.

src/shared/components/home/admin-settings.tsx
```diff
@@ -112,21 +114,47 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
...
+          <div >
+            <HtmlTags
+                  title={this.documentTitle}
+                  path={this.context.router.route.match.url}
+                />
+            <ul className="nav nav-tabs mb-2">
+              <li className="nav-item">
+                <button
+                  className={`nav-link btn ${this.state.currentTab == "site" && "active"}`}
+                  onClick={() => { this.handleSwitchTab("site"); }}>
```
dessalines, line 126: Same, use linkEvent.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+        );
+    }
+
+    handleEmojiClick = (e: any) => {
+        if (this.props.onEmojiClick)
+            this.props.onEmojiClick(e);
+    }
+}
+
+interface EmojiPickerProps {
```
dessalines, line 44: Split these into two files, they are different components.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+                <button
+                    className="btn btn-sm text-muted"
+                    data-tippy-content={i18n.t("emoji")}
+                    aria-label={i18n.t("emoji")}
+                    onClick={this.togglePicker}
+                >
+                    <Icon icon="smile" classes="icon-inline" />
+                </button>
+
+                {this.state!.showPicker && (
```
dessalines, line 75: What's the exclamation point do here?

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    render() {
+        return (
+            <div id="emoji-picker"></div>
+        );
+    }
+
+    handleEmojiClick = (e: any) => {
```
dessalines, line 38: Use the conventions with all the other handles, no closures. Strongly type this also.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+        if (this.props.onEmojiClick)
+            this.props.onEmojiClick(e);
```
dessalines, line 40: Should be able to do `this.props.onEmojiClick?.(...)`

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    togglePicker = (e) => {
+        e.preventDefault();
+        this.setState({ showPicker: !this.state!.showPicker });
+    }
+
+    handleEmojiClick = (e: any) => {
+        if (this.props.onEmojiClick)
+            this.props.onEmojiClick(e);
```
dessalines, line 97: Same note as above.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+                            onClick={this.togglePicker}
+                            className="click-away-container"
+                        />
+                    </>
+                )}
+            </span>
+        );
+    }
+
+    togglePicker = (e) => {
```
dessalines, line 90: Functions, not closures.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
...
+        e.preventDefault();
+        this.setState({ showPicker: !this.state!.showPicker });
```
dessalines, line 92: state should be defined already.

makotech222:
Its the icon on each row next to the delete button. Its not a great icon, but couldn't find any better suiting ones.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -328,6 +331,21 @@ export class MarkdownTextArea extends Component<
     );
   }
 
+  handleEmoji(i: MarkdownTextArea, e: any) {
+    let value = e.native;
+    if (value == null){
+      let lookup = getCustomEmojiLookup()
```
dessalines, line 337: If you've already built the map, just export that value and refer to it, no need to call functions that might accidentally re-initialize things.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+import { getCustomEmojiLookup, isBrowser, myAuth, setIsoData, toast, updateEmojiDataModel, wsClient, wsSubscribe } from "../../utils";
+import { EmojiMart } from "../common/emoji-picker";
+import { HtmlTags } from "../common/html-tags";
+import { Icon } from "../common/icon";
+import { Subscription } from "rxjs";
+import { pictrsUri } from "../../env";
+
+interface EmojiFormState {
+    siteRes: GetSiteResponse;
+    custom_emojis: CustomEmojiViewForm[];
```
dessalines, line 16: camelCase

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+            id: x.custom_emoji.id,
+            category: x.custom_emoji.category,
+            shortcode: x.custom_emoji.shortcode,
+            image_url: x.custom_emoji.image_url,
+            alt_text: x.custom_emoji.alt_text,
+            keywords: x.keywords.map(x => x.keyword).join(" "),
+            changed: false,
+        }))
+    };
+    private emojiLookup = {};
```
dessalines, line 47: Get rid of this, use maps or functions to do lookups. God I hate javascript.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+            toast(JSON.stringify(data), "danger");
+        }
+        else {
+            let hash = data.files[0].file;
+            let url = `${pictrsUri}/${hash}`;
+            this.handleEmojiImageUrlChange(url, index)
+        }
+    }
+
+    configurePicker(): any {
```
dessalines, line 380: In the other example, set the picker to {}, why does it need to be configured here but not there?

src/shared/components/common/emoji-picker.tsx
```diff
@@ -0,0 +1,99 @@
+import { Component } from "inferno";
+import { i18n } from "../../i18next";
+import { getEmojiMart } from "../../utils";
+import { Icon } from "./icon";
+
+
+interface EmojiMartProps {
+    onEmojiClick?(val: any): any;
+    pickerOptions: any;
```
dessalines, line 9: Emoji-mart has typescript typings : https://www.npmjs.com/package/@types/emoji-mart

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+                    <EmojiMart onEmojiClick={this.handleEmojiClick} pickerOptions={this.configurePicker()}></EmojiMart>
+                </div>}
+                <div className="table-responsive col-12">
+                    <table
+                        id="emojis_table"
+                        className="table table-hover"
+                    >
+                        <thead className="pointer">
+                            <tr>
+                                <th style="width:70px;">{i18n.t("column_emoji")}</th>
```
dessalines, line 90: Re-do this to bootstrap columns, rather than a table. It'll make it more responsive and usable on mobile, and you won't have to hard-code widths. I need to get rid of every html table in lemmy-ui.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+                                <th style="width:150px">{i18n.t("column_shortcode")}</th>
+                                <th style="width:150px">{i18n.t("column_category")}</th>
+                                <th style="width:200px">{i18n.t("column_imageurl")}</th>
+                                <th style="width:100px">{i18n.t("column_alttext")}</th>
+                                <th>{i18n.t("column_keywords")}</th>
+                                <th style="width:107px"></th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                            {this.state!.custom_emojis
```
dessalines, line 100: I think these exclamations can cause crashes right? State should already be initialized, so I'm not sure why the IDE isn't telling you this is redundant.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+                                .map((cv, index) => (
+                                    <tr key={index} ref={e => this.scrollRef[cv.shortcode] = e}>
+                                        <td style="text-align:center;">
+                                            <label
+                                                htmlFor={index.toString()}
+                                                className="pointer text-muted small font-weight-bold"
+                                            >
+                                                {cv.image_url.length > 0 && <img style="max-width: 24px; max-height: 24px; display: inline-block;" src={cv.image_url} />}
```
dessalines, line 108: Use a style in `main.css`

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+                                            </button>
+                                        </td>
+                                    </tr>
+                                ))}
+                        </tbody>
+                    </table>
+                    <br />
+                    <button
+                        className="btn btn-sm btn-secondary mr-2"
+                        onClick={e => this.handleAddEmojiClick(e)}
```
dessalines, line 217: You have a button to add a block, but no button to save the emojis to lemmy.

dessalines:
That icon looks like an edit. Once you turn it from a table into a bootstrap form, it'll probably be better to name the button `Save` for clarity.

src/shared/components/common/emoji-picker.tsx
makotech222: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

asserts that object is not null/undefined to the compiler.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+                                        </td>
+                                        <td>
+                                            <input
+                                                type="text"
+                                                placeholder="ShortCode"
+                                                className="form-control"
+                                                disabled={cv.id > 0}
+                                                value={cv.shortcode}
+                                                onInput={(s: any) =>
+                                                    this.handleEmojiShortCodeChange(s.target.value, index)
```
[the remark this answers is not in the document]
makotech222: With linkEvent, I can't pass any parameters to the function. I can technically embed it in the html element, but that seems so much messier. I've already done same function calls in earlier PRs without linkEvent. From documentation, it just seems like its slightly faster?

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,431 @@
...
+            return;
+        }
+        else if (op == UserOperation.CreateCustomEmoji) {
+            let data = wsJsonToRes<CustomEmojiResponse>(msg);
+            const custom_emoji_view = data.custom_emoji;
+            updateEmojiDataModel(custom_emoji_view)
+            let currentEmojis = this.state!.custom_emojis;
+            let newEmojiIndex = currentEmojis.findIndex(x => x.shortcode == custom_emoji_view.custom_emoji.shortcode)
+            currentEmojis[newEmojiIndex].id = custom_emoji_view.custom_emoji.id;
+            currentEmojis[newEmojiIndex].changed = false;
```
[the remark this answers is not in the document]
makotech222: Not sure what you mean by this. both commentform and postform send websocket requests and implement parseMessage and handle the response. PostForm will notify a parent, via an event, but thats to deal with concerns outside the form. The EmojiForm doesn't need to notify any parents, and handles all its needs internally; nothing in admin-settings needs to know about what happens.

makotech222:
> That icon looks like an edit. Once you turn it from a table into a bootstrap form, it'll probably be better to name the button `Save` for clarity.

Is this okay?

image

Id prefer a better icon, as its takes up less width, but I couldn't find any in our icon set that worked properly.

src/shared/components/home/emojis-form.tsx
makotech222: The admin panel has custom picker options that remove all the built-in emojis and only shows custom emojis

src/shared/components/common/emoji-picker.tsx
makotech222: The typings are over 3 years old, unfortunately, so they are basically useless. Ill try to define my own types.

src/shared/components/common/emoji-picker.tsx
makotech222: I fixed this, but curious why? since it avoids the need to bind(this) on the function in constructor.

src/shared/components/common/emoji-picker.tsx
makotech222: noting also, i've fixed most of these anyways

src/shared/components/home/emojis-form.tsx
makotech222: I'm gonna have to push back on this one. Tables are the correct way to display this kind of data. Moving to bootstrap columns would end up wasting a ton of space, as 12 columns isn't specific enough about widths. None of the columns need to be responsive, except for the keywords column, which is set up to expand to fill any remaining width.

I have made a change to optimize it a bit better for mobile by giving the table a min-width so that keywords isn't messed up, but its as usable as its ever gonna be for a small device.

makotech222:
Among many fixes today, i added pagination to the admin panel for emojis, to deal with performance issues with large size lists.

src/shared/utils.ts
```diff
@@ -74,6 +79,9 @@ export const markdownFieldCharacterLimit = 50000;
 
 export const relTags = "noopener nofollow";
 
+let customEmojis: EmojiMartCategory[] = [];
+export let customEmojisLookup: Map<string, CustomEmojiView> = new Map<string, CustomEmojiView>();
```
dessalines, line 83: Much better. If we end up getting some weird double calls on this thing tho, we might have to set up a singleton service like was done for UserService and WebsocketService, that could hold some other globals as well.

src/shared/components/home/admin-settings.tsx
```diff
@@ -112,21 +114,49 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
...
-              <SiteForm
-                siteRes={this.state.siteRes}
-                showLocal={showLocal(this.isoData)}
-              />
-            </div>
-            <div className="col-12 col-md-6">
-              {this.admins()}
-              {this.bannedUsers()}
-            </div>
+          <div >
```
dessalines, line 117: Weird extra space, wonder how this is passing lint.

src/shared/components/home/admin-settings.tsx
```diff
@@ -179,6 +209,11 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
     );
   }
 
+  handleSwitchTab(i: AdminSettings, event: any) {
```
dessalines, line 212: Use the convention for tab switching from `settings.tsx`, line 214 .

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -226,6 +228,7 @@ export class MarkdownTextArea extends Component<
             >
               <Icon icon="link" classes="icon-inline" />
             </button>
+            <EmojiPicker onEmojiClick={(e) => this.handleEmoji(this,e)}></EmojiPicker>
```
dessalines, line 231: Use linkEvent

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,456 @@
...
+import { Paginator } from "../common/paginator";
+
+interface EmojiFormState {
+    siteRes: GetSiteResponse;
+    customEmojis: CustomEmojiViewForm[];
+    loading: boolean;
+    page: number;
+}
+
+class CustomEmojiViewForm {
```
dessalines, line 21: Why is this a class and not an interface?

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,456 @@
...
+import { Icon } from "../common/icon";
+import { Subscription } from "rxjs";
+import { pictrsUri } from "../../env";
+import { Paginator } from "../common/paginator";
+
+interface EmojiFormState {
+    siteRes: GetSiteResponse;
+    customEmojis: CustomEmojiViewForm[];
+    loading: boolean;
+    page: number;
```
dessalines, line 18: What's going on with page here? I didn't see any paging in the back end.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,456 @@
...
+        customEmojis: this.isoData.site_res.custom_emojis.map((x, index) =>
+        ({
+            id: x.custom_emoji.id,
+            category: x.custom_emoji.category,
+            shortcode: x.custom_emoji.shortcode,
+            image_url: x.custom_emoji.image_url,
+            alt_text: x.custom_emoji.alt_text,
+            keywords: x.keywords.map(x => x.keyword).join(" "),
+            changed: false,
+            page: 1 + (Math.floor(index / this.itemsPerPage))
```
dessalines, line 48: Don't hack in paging to the front end, if it needs to be added, it needs to be done properly in the back end as well.

src/shared/components/home/emojis-form.tsx
dessalines: Okay, but at the very least remove the hard-coded widths ( those aren't going to be correct for any device but the one you coded them from), and use bootstrap table directives like in `communities.tsx` . As I said though, tables are really terrible for skinny devices like phones, which is why we eventually need to get rid of that, and switch to using responsive design.

Bootstrap has responsive form examples, which is what this is, and we have a good amount of responsive forms in lemmy.

src/shared/components/home/emojis-form.tsx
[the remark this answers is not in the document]
dessalines: To pass data to linkEvent, do:

```ts
onClick = {linkEvent(
  {some_data_here},
  this.handleBleh
)}

File: src/shared/components/home/emojis-form.tsx [hunk 28]
```diff
@@ -0,0 +1,456 @@
...
+                                <th style="width:150px">{i18n.t("column_shortcode")}</th>
+                                <th style="width:150px">{i18n.t("column_category")}</th>
+                                <th style="width:200px">{i18n.t("column_imageurl")}</th>
+                                <th style="width:150px">{i18n.t("column_alttext")}</th>
+                                <th>{i18n.t("column_keywords")}</th>
+                                <th style="width:121px"></th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                            {this.state.customEmojis.slice((this.state.page - 1) * this.itemsPerPage, ((this.state.page - 1) * this.itemsPerPage) + this.itemsPerPage)
```
dessalines, line 100: Leave complicated front-end paging logic out of this, if we need to later, we can add it in the back end.

src/shared/components/home/emojis-form.tsx
```diff
@@ -0,0 +1,456 @@
...
+            return i18n.t("save");
+        else
+            return i18n.t("custom_emoji_save_validation");
+    }
+
+    handlePageChange(page: number) {
+        this.setState({ page: page });
+    }
+
+    handleEmojiClick = (e: any) => {
```
dessalines, line 249: No closures, functions only please.

src/shared/components/home/emojis-form.tsx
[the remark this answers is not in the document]
dessalines: Okay I see what you're saying, my bad.

src/shared/components/home/emojis-form.tsx
makotech222: Regarding paging: All emojis are already loaded on frontend, no need to pull pages from server. We are paging on admin settings because showing all emojis at once (especially when you get to the thousands) causes page to slow to a crawl and become unusable.

makotech222:
@dessalines Okay I fixed the comments. I updated the table to be more like communities. I do still have the last column with a set width, because the table doesn't really allow two elements in the same column without wrapping it, causing the row width the expand
desktop:
image
mobile:
image

src/shared/components/home/emojis-form.tsx
dessalines: If we're really getting to the level of crashing the front end because there are too many emojis, then we really need to take them out of `GetSite` into a paged emoji fetcher. 

Front end hacks like this have to be replicated on all lemmy front ends, which means we're not structuring things correctly. I'm okay with merging this as is, but please add an issue to the Lemmy repo for breaking emojis into its own paged endpoint.

dessalines (commented):
Once the back-end gets merged, I'll test this again, and merge it if everything works correctly.

src/shared/components/home/emojis-form.tsx
Nutomic: I would consider adding a hardcoded emoji limit in the backend then. If you hit the limit, you just need to remove an existing emoji before adding a new one.

makotech222: All users need all of the emojis available; when they open the emoji-picker, we need to know what to put in there all at once, not paged. Because of this, its kind of pointless to also have paged endpoints for emojis. I personally consider front-end paging pretty trivial; its mainly necessary once you get to the number of emojis that sites like hexbear have, which ends up with >2000 <tr> elements. Rendering all those at once is unfortunately a weakness of html and will cause page to get unresponsive.

Adding a hard coded limit would be unfortunate, as hexbear would be way over the limit of what is responsive on that page. Seems unnecessary, since the obvious way to work with it is just paginating on the front end. Alternate frontends can go without paginating, as long as they understand that it will get slower as more emojis are added.

makotech222:
@dessalines Hey, any open requests for changes here? Or just need to test?

dessalines:
Okay I deployed the new lemmy-js-client, and merged this fixing a few conflicts.

PR #927, opened 2023-02-09
Increase frame size of embedded videos, allow full screen

### LINKED ISSUE #926: Improved video embeds for the browser

At the moment, videos will be embedded as 300 px x 150 px iframes with the fullscreen option disabled. This provides a sub-optimal experience for watching the embedded videos through the browser.

image

My proposed changes are:
- Allow fullscreen
- Center the frame
- Make the frame larger

I have come up with a solution that works for me. In this solution, I have hard-coded the width and height to 560 px by 315 px (19:6), centered the frame with <center></center>, removed the frame border, and allowed full-screen. Here is an example: https://mander.xyz/post/533860

image

For this solution, the [<iframe> tag](https://github.com/LemmyNet/lemmy-ui/blob/4eb902306df7f4751cb5b7f4caa6aba48a8ddc5a/src/shared/components/post/metadata-card.tsx#L78) is modified to:

`<center><iframe allowFullscreen="allowfullscreen" src={post.embed_video_url} width="560px" height="315px" frameborder="0"></iframe></center>`

I have also tested what this does to the display on a phone screen. Unfortunately, setting the absolute height and width will cause the video to appear large in phone screens. I tried to find a simple way to create a responsive iframe, but the solutions that I have found require modifying the CSS code or running some javascript code to resize. The solution I propose is the simplest reasonable solution I could come up with. 

### Additional context

I will submit a PR in case you choose to adopt this approach. At the very least, I would suggest allowing the videos to be full-screened.

Kradyz:
A proposed change to the <iframe> of embedded videos in order to improve the browser experience. 

This change will:

- Allow full-screen
- Center the frame
- Make the frame larger

Example implementation: https://mander.xyz/post/533860

Fixes #926

### DIFF
--- a/src/shared/components/post/metadata-card.tsx
+++ b/src/shared/components/post/metadata-card.tsx
@@ -75,7 +75,7 @@ export class MetadataCard extends Component<
           </div>
         )}
         {this.state.expanded && post.embed_video_url && (
-          <iframe src={post.embed_video_url}></iframe>
+          <center><iframe allowFullscreen="allowfullscreen" src={post.embed_video_url} style="aspect-ratio:16/9;max-width:85vw" width="560px" frameborder="0"></iframe></center>
         )}
       </>
     );


hanubeki:
Your example doesn't fit in mobile screen, maybe try `max-width: 0.85vw` or something

Kradyz:
> Your example doesn't fit in mobile screen, maybe try `max-width: 0.85vw` or something

Thank you @hanubeki! I have tested and commited that change, and it is reflected by the example (https://mander.xyz/post/533860).

Do you know how to re-scale the height appropriately when max-width comes into play in order to keep the aspect ratio? The way it is now, in the phone you can see black bands on the top and bottom of the video because the height is not re-scaled. It is not as bad as before - quite an improvement already.

I have tried setting the width and height in relative rather than absolute terms, but since the phone and browser screens contain different elements, a relative size for the browser looks very small on a cellphone. That is why I have so far chosen to use the absolute pixel units.

src/shared/components/post/metadata-card.tsx
```diff
@@ -75,7 +75,7 @@ export class MetadataCard extends Component<
           </div>
         )}
         {this.state.expanded && post.embed_video_url && (
-          <iframe src={post.embed_video_url}></iframe>
+          <center><iframe allowFullscreen="allowfullscreen" src={post.embed_video_url} style="aspect-ratio:16/9;max-width:85vw" width="560px" frameborder="0"></iframe></center>
```
dessalines, line 78: Do not hardcode widths or aspect ratios like this. Use bootstrap classes made for iframes: https://getbootstrap.com/docs/5.3/helpers/ratio/#example

dessalines (changes requested):

src/shared/components/post/metadata-card.tsx
cloventt: FYI the feature you linked appears to be new in Bootstrap v5. As the UI is still using bootstrap v4 the `.ratio` helper is not available yet.

daviessm: @jsit is working on Bootstrap 5: https://github.com/LemmyNet/lemmy-ui/pull/1378

cloventt: I have opened a PR following the recommended approach #1437

Issue #965, opened 2023-03-25
Search page: can't change type, listing type, or sort when no query has been entered.


SleeplessOne1917:
Found a bug? Please fill out the sections below. 👍

For backend issues, use [lemmy](https://github.com/LemmyNet/lemmy)

### Issue Summary

When on the the search screen, if you haven't entered search text yet, you cannot change the type, listing type, or sort.

### Steps to Reproduce

1. Go to the search page on a lemmy instance using lemmy-ui for version 0.17.2 or higher.
2. Try to change any of the selectors circled in red without submitting a search term:
![buggy components](https://user-images.githubusercontent.com/28871516/227747425-238db636-792d-485e-8ddf-cdb2bb3f80a9.png)
3. Note that the URL will update, but the UI will not change

### Technical details

I already made a PR and will tag this issue in it.

dessalines:
In the future, make sure you have `Fixes #X` in the commit message.

PR #966, opened 2023-03-25
fix: Make search screen able to change type, listing type, and sort when there is no query.

SleeplessOne1917:
This is the fix for #965.


dessalines (approved):

dessalines:
Looks good, thx!

PR #812, opened 2022-10-01
Async-image-upload

sam365724:
Here is the functionality to fix https://github.com/LemmyNet/lemmy-ui/issues/777.

PR #967, opened 2023-03-26
refactor: update UI to use new client uploadImage function

SleeplessOne1917:
This PR updates the UI project to use the upload image change introduced in [the most recent lemmy-js-client](https://github.com/LemmyNet/lemmy-js-client/pull/108) update.


SleeplessOne1917:
I noticed that there's already a PR up related to uploading images (#812). Should I implement that as part of this PR? It looks like there's not much code in the change and judging by [this comment](https://github.com/LemmyNet/lemmy-ui/pull/812#discussion_r1008116190) was supposed to be in the 0.17 release but didn't make it. Also, it seems like [translations have already been added](https://github.com/LemmyNet/lemmy-translations/pull/28) for the change.

src/shared/utils.ts
```diff
@@ -1441,3 +1442,9 @@ export function selectableLanguages(
     }
   }
 }
+
+export function uploadImage(image: File): Promise<UploadImageResponse> {
+  const client = new LemmyHttp(httpBase);
```
SleeplessOne1917, line 1447: It looks like there are multiple functions that rely on a lemmy HTTP client newed up with the just the HTTP base URL (i.e. this function, `fetchCommunities`, and `fetchUsers`). It may make more sense to just make a module level variable variable of the HTTP client instead of newing it up every time one of these functions is called. Thoughts?

dessalines: I'd def be good with that. Maybe as a service-y type singleton pattern alongside `WebSocketService` and `UserService`.

The isometric back-end does have to have its own separate http client tho I think, that's in `server/index.ts`

But ya it doesn't have to be a part of this PR, but feel free to add an issue for it.

dessalines (approved):
Looks good, thanks!

Issue #969, opened 2023-03-26
Inserting markdown with button inserts undefined


SleeplessOne1917:
### Issue Summary

Inserting markdown in a markdown text area using one of the markdown buttons before any text has been entered in the text area will cause `undefined` to be added to the text area.

### Steps to Reproduce

1. Try to create a post or comment - any part of the app that uses the markdown text area
2. **Do not enter any text in the text area**
3. Click the any of the insert markdown buttons except for the image one
4. `undefined` will be inserted into the text area
![markdown text area undefined](https://user-images.githubusercontent.com/28871516/227756607-06242774-fb9b-4b69-87fa-087ea834abfb.png)

### Technical details

I am immediately making a PR for this.

KAMiKAZOW:
This may be a stupid question but in which release will this fix show up? As many, I only recently moved over and the instance is running 0.17.4 with the bug in place.

SleeplessOne1917:
@KAMiKAZOW It's fixed in 0.18.0. I'm surprised this fix didn't make it into one of the 0.17.x releases.

PR #970, opened 2023-03-26
fix: Fix undefined showing up in markdown

SleeplessOne1917:
PR for issue #969.


dessalines (approved):
Thanks again for taking these on!

PR #971, opened 2023-03-28
Multiple image upload

### LINKED ISSUE #777: Allow multiple images to be uploaded at once
Not directly. Uploading many pictures is a time-intense process with the current implementation. Each image has to be selected by hand. 

Allow the picture open dialog to select multiple images at once, and upload all of them afterwards, resulting in multiple links being inserted into the markdown body. 
Nice to have:
- Some better progress indication, showing x out of n uploaded...

Drag and drop, but I'd prefer the open dialog, it's more common.

### Additional context
I'm sure there is an existing typescript component that could be used. How difficult is this to implement?

SleeplessOne1917:
This closes #777
This does what #812 by @sam365724 does but using the updated client.

### DIFF
--- a/src/shared/components/common/emoji-picker.tsx
+++ b/src/shared/components/common/emoji-picker.tsx
@@ -5,6 +5,7 @@ import { Icon } from "./icon";
 
 interface EmojiPickerProps {
   onEmojiClick?(val: any): any;
+  disabled?: boolean;
 }
 
 interface EmojiPickerState {
@@ -15,8 +16,9 @@ export class EmojiPicker extends Component<EmojiPickerProps, EmojiPickerState> {
   private emptyState: EmojiPickerState = {
     showPicker: false,
   };
+
   state: EmojiPickerState;
-  constructor(props: any, context: any) {
+  constructor(props: EmojiPickerProps, context: any) {
     super(props, context);
     this.state = this.emptyState;
     this.handleEmojiClick = this.handleEmojiClick.bind(this);
@@ -28,6 +30,7 @@ export class EmojiPicker extends Component<EmojiPickerProps, EmojiPickerState> {
           className="btn btn-sm text-muted"
           data-tippy-content={i18n.t("emoji")}
           aria-label={i18n.t("emoji")}
+          disabled={this.props.disabled}
           onClick={linkEvent(this, this.togglePicker)}
         >
           <Icon icon="smile" classes="icon-inline" />
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -5,11 +5,14 @@ import { Language } from "lemmy-js-client";
 import { i18n } from "../../i18next";
 import { UserService } from "../../services";
 import {
+  concurrentImageUpload,
   customEmojisLookup,
   isBrowser,
   markdownFieldCharacterLimit,
   markdownHelpUrl,
+  maxUploadImages,
   mdToHtml,
+  numToSI,
   pictrsDeleteToast,
   randomStr,
   relTags,
@@ -251,6 +254,7 @@ export class MarkdownTextArea extends Component<
                 accept="image/*,video/*"
                 name="file"
                 className="d-none"
+                multiple
                 disabled={!UserService.Instance.myUserInfo}
                 onChange={linkEvent(this, this.handleImageUpload)}
               />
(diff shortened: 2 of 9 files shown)


src/shared/components/common/markdown-textarea.tsx
```diff
@@ -350,53 +354,80 @@ export class MarkdownTextArea extends Component<
...
+        const textarea: any = document.getElementById(i.id);
+        autosize.update(textarea);
+        pictrsDeleteToast(file.name, res.delete_url as string);
+      } else {
+        throw JSON.stringify(res);
+      }
+    } catch (error) {
+      i.setState({ imageLoading: false });
+      console.error(error);
+      toast(error, "danger");
```
SleeplessOne1917, line 427: Currently, if an image upload fails, a toast shows with the error of the failed upload. This made sense with single image uploads, but I'm not sure it makes the most sense if uploading multiple images. It is possible for some uploads to succeed and others to fail. @dessalines @Nutomic is there any way you would prefer this to be handled?

src/shared/utils.ts
```diff
@@ -500,15 +505,18 @@ export function toast(text: string, background = "success") {
   }
 }
 
-export function pictrsDeleteToast(
-  clickToDeleteText: string,
-  deletePictureText: string,
-  failedDeletePictureText: string,
-  deleteUrl: string
-) {
+export function pictrsDeleteToast(filename: string, deleteUrl: string) {
```
SleeplessOne1917, line 508: The 3 string arguments were the same everywhere this function was used, and there didn't seem to be a reason for non picture deletion related strings to need to be used here, so I define the strings in the function and only take the filename as an argument.

src/shared/utils.ts
```diff
@@ -500,15 +505,18 @@ export function toast(text: string, background = "success") {
...
   if (isBrowser()) {
-    let backgroundColor = `var(--light)`;
-    let toast = Toastify({
+    const clickToDeleteText = `${i18n.t(
```
SleeplessOne1917, line 510: Since these translation keys always use the filename, I think it would more sense to have the filename be an argument passed to the translation instead of using the template strings.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -350,53 +354,80 @@ export class MarkdownTextArea extends Component<
...
+  }
+
+  async uploadSingleImage(i: MarkdownTextArea, file: File) {
+    try {
+      const res = await uploadImage(file);
+      console.log("pictrs upload:");
+      console.log(res);
+      if (res.msg === "ok") {
+        const imageMarkdown = `![](${res.url})`;
+        i.setState(({ content }) => ({
```
SleeplessOne1917, line 414: I'm using a function here because if inferno is anything like react, `setState` schedules state changes instead of immediately changing them. [This article explains the concept](https://dmitripavlutin.com/how-react-updates-state/). Since the new state depends on the previous value of state, using a function makes sure we get the expected value for the prev state.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -150,6 +161,17 @@ export class MarkdownTextArea extends Component<
...
             )}
+            {this.state.imageUploadStatus &&
+              this.state.imageUploadStatus.total > 1 && (
+                <ProgressBar
+                  className="mt-2"
+                  striped
+                  animated
+                  value={this.state.imageUploadStatus.uploaded}
+                  max={this.state.imageUploadStatus.total}
+                  text={`${this.state.imageUploadStatus.uploaded}/${this.state.imageUploadStatus.total} files uploaded`}
```
SleeplessOne1917, line 172: We'll have to add a translation for this.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -15,8 +16,9 @@ export class EmojiPicker extends Component<EmojiPickerProps, EmojiPickerState> {
   private emptyState: EmojiPickerState = {
     showPicker: false,
   };
+
   state: EmojiPickerState;
-  constructor(props: any, context: any) {
+  constructor(props: EmojiPickerProps, context: any) {
```
dessalines, line 21: Oof thanks, can't believe I missed this one.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -56,12 +66,13 @@ export class MarkdownTextArea extends Component<
...
   private formId = `comment-form-${randomStr()}`;
   private tribute: any;
+
   state: MarkdownTextAreaState = {
     content: this.props.initialContent,
     languageId: this.props.initialLanguageId,
     previewMode: false,
     loading: false,
-    imageLoading: false,
+    imageUploadStatus: null,
```
dessalines, line 75: Then you can leave this off, or set to undefined.

src/shared/components/common/markdown-textarea.tsx
dessalines: Make a PR over on lemmy-translations for me to merge. Here's an example of using counts: https://github.com/LemmyNet/lemmy-translations/blob/main/translations/en.json#L148

You can grep the lemmy-ui for any of those formatted count type strings to see how to do in the UI.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -619,9 +638,16 @@ export class MarkdownTextArea extends Component<
...
   getSelectedText(): string {
-    let textarea: any = document.getElementById(this.id);
-    let start: number = textarea.selectionStart;
-    let end: number = textarea.selectionEnd;
+    const { selectionStart: start, selectionEnd: end } =
+      document.getElementById(this.id) as any;
     return start !== end ? this.state.content?.substring(start, end) ?? "" : "";
   }
+
+  getIsDisabled() {
```
dessalines, line 646: Up to you, but you can add `get getIsDisable() {...` and then access it as if it were a variable: `this.getIsDisabled` 

Usually not a bad idea when there are no function params.

src/shared/components/common/markdown-textarea.tsx
dessalines: Either way is fine.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -333,6 +285,39 @@ export class MarkdownTextArea extends Component<
     );
   }
 
+  getFormatButton(
+    type: I18nKeys,
```
dessalines, line 289: 👍

dessalines (commented):
Just a few minor things, and the one i18n string to add. After the new string gets merged by me into `lemmy-translations`, you can run:

```
git submodule update --remote                                                                                    
git add lemmy-translations

File: src/shared/components/common/markdown-textarea.tsx [hunk 10]
```diff
@@ -595,11 +614,11 @@ export class MarkdownTextArea extends Component<
   }
 
   quoteInsert() {
-    let textarea: any = document.getElementById(this.id);
-    let selectedText = window.getSelection()?.toString();
-    let content = this.state.content;
+    const textarea: any = document.getElementById(this.id);
+    const selectedText = window.getSelection()?.toString();
+    const { content } = this.state;
```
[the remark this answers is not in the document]
SleeplessOne1917: It's an alternate way of getting `content` off of `this.state` using de-structuring, although tbh I may have gotten a little carried away with the de-structuring here. The only real difference is that you get to get `content` without needing to type it twice (once for the variable name, once for the property on the state).

src/shared/components/common/progress-bar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  barColor?: ThemeColor;
+  striped?: boolean;
+  animated?: boolean;
+  min?: number;
+  max?: number;
+  value: number;
+  text?: string;
+};
+
+const ProgressBar = ({
```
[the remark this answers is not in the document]
SleeplessOne1917: I went with the lambda style function over the normal style here because it's just directly returning a result. I can understand why some prefer the normal style function over the lambda style when the function needs to do multiple statements, but this seems like the kind of situation lambda functions were made for.

SleeplessOne1917:
Woodpecker is getting a build failure:
```
+ yarn lint
yarn run v1.22.19
$ node generate_translations.js && tsc --noEmit && eslint --report-unused-disable-directives --ext .js,.ts,.tsx src && prettier --check 'src/**/*.tsx'
src/shared/components/common/markdown-textarea.tsx:311:36 - error TS2345: Argument of type 'I18nKeys' is not assignable to parameter of type 'NoOptionI18nKeys | NoOptionI18nKeys[]'.
  Type '"click_to_delete_picture"' is not assignable to type 'NoOptionI18nKeys | NoOptionI18nKeys[]'.

311         data-tippy-content={i18n.t(type)}
                                       ~~~~

src/shared/components/common/markdown-textarea.tsx:312:28 - error TS2345: Argument of type 'I18nKeys' is not assignable to parameter of type 'NoOptionI18nKeys | NoOptionI18nKeys[]'.

312         aria-label={i18n.t(type)}
                               ~~~~


Found 2 errors in the same file, starting at: src/shared/components/common/markdown-textarea.tsx:311

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I don't know why it's doing this considering the code it's complaining about in the build isn't actually in the PR. [This is what's actually in the PR's code](https://github.com/LemmyNet/lemmy-ui/blob/23a291da5712c15e7643e3be890133e6048d1797/src/shared/components/common/markdown-textarea.tsx#L288-L319).

SleeplessOne1917:
I pushed a small change an hour after the last one and the CI build worked this time. Disregard my last comment.

dessalines (approved):
Looks great, thx again!

PR #972, opened 2023-03-30
Optimize Tagline Form page

makotech222:
This work moves the tagline editing from the main Site Settings form into its own tab, like emojis. It also is optimized for larger lists, allowing hundreds of taglines without slowing down the page.

I wrote this up pretty quick, didn't want to refactor too much, just needed to get past a blocker for hexbear. Its still worthy of upstreaming it, though, as current implementation falls apart if you have lots of taglines.

image

### DIFF
--- a/src/shared/components/home/tagline-form.tsx
+++ b/src/shared/components/home/tagline-form.tsx
@@ -1,0 +1,190 @@
+import { Component, InfernoMouseEvent, linkEvent } from "inferno";
+import { EditSite, GetSiteResponse } from "lemmy-js-client";
+import { i18n } from "../../i18next";
+import { WebSocketService } from "../../services";
+import { capitalizeFirstLetter, myAuth, wsClient } from "../../utils";
+import { HtmlTags } from "../common/html-tags";
+import { Icon, Spinner } from "../common/icon";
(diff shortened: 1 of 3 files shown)


src/shared/components/home/tagline-form.tsx
```diff
@@ -0,0 +1,190 @@
...
+        <HtmlTags
+          title={this.documentTitle}
+          path={this.context.router.route.match.url}
+        />
+        <h5 className="col-12">{i18n.t("taglines")}</h5>
+        <div className="table-responsive col-12">
+          <table id="taglines_table" className="table table-sm table-hover">
+            <thead className="pointer">
+              <th></th>
+              <th style="width:121px"></th>
```
dessalines, line 54: Don't hard code widths like this. Use responsive bootstrap tables.

makotech222: This is pulled from emoji table, used for same reason, just to give a min width to the column to have two buttons on the same row.

dessalines: Okay, but I highly suggest you read over how bootstrap does responsive forms: https://getbootstrap.com/docs/5.0/forms/overview

dessalines (approved):

PR #973, opened 2023-04-01
I18 quality of life change

SleeplessOne1917:
While working on a translation issue, I thought it would be helpful for Typescript to say which parameters are needed for an i18 translation. 

Translations that don't require parameters are the same as before. However, if a translation needs parameters, Typescript will show an error if it's not getting them. In addition to that, Typescript will show which properties are needed on the parameter object. This screenshot shows an example:
translation option error

As you can see, the `cake_day_info` translation requires a `creator_name` property to work.


dessalines:
Nice!

dessalines (approved):

Issue #975, opened 2023-04-06
Links/buttons that require being logged in show up when not logged in 


SleeplessOne1917:
### Proposal
Currently in the UI, several things show up for actions that only make sense while logged in. For instance, consider the home page:

![Home Page](https://user-images.githubusercontent.com/28871516/230249564-f404a283-09aa-47c3-810d-d4147ef025a0.png)

Creating a community, creating a post, or viewing subscribed communities are things that I would think only make sense when logged in to an account. Sure enough, for the first 2, clicking on them will redirect the user to the login page (and not let them navigate back in history for some reason).

The community page is another example:

![Community Page](https://user-images.githubusercontent.com/28871516/230250169-6a54d465-43c8-443f-9638-d0178e6332a9.png)

Subscribing, creating a post, and blocking a community again don't make sense to do when not logged in and, again, they all show error toasts when you try clicking them anyway.

### Describe the solution you'd like

My proposal is to hide these UI elements when a user is not logged in. However, I saw #974 and it seems that the maintainers want to provide information for users interested in subscribing to a community from another instance. Fortunately, I have an idea that might satisfy what @Nutomic was trying to do in that issue.

The create post and create community buttons/links can still be hidden. For communities, in addition to hiding the action buttons, the buttons in the sidebar are replaced with a message like @Nutomic mentioned in #974. It could look something like this:

![community with message](https://user-images.githubusercontent.com/28871516/230252573-1af08b9a-ddef-4b6a-a947-be2f79268191.png)

Let me know if you want it implemented. I've gotten familiar enough with the UI codebase I shouldn't need any guidance. It even seems the [profile page is doing something similar](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/person/profile.tsx#L434-L479).

Nutomic:
This sounds great, I hope that you will implement it! The exact text and design will need some more fine-tuning of course. As you mention the profile page, it would also make sense to show a similar message for following the user. As this is now possible from Mastodon etc (but not from Lemmy which only lets you follow communities).

Bakcodi:
It is better to keep these buttons as-is. This is the main thing that encourages users to signup to get better experience. Otherwise people can mistake the site for a read-only blog that has a comment-box. And 90% users don't signup to a blog.

SleeplessOne1917:
[...]

PR #986, opened 2023-04-23
Handle when logged out

SleeplessOne1917:
This is for #975.

The community sidebar shows this when logged out:
logged out community sidebar

The profile page looks like this:
logged out profile

The text being used isn't final and I'll gladly change it if anyone has any better ideas. The alerts seems somewhat bland at the moment too. Maybe there should be an icon to go with them?

I have also made a version of this PR where I hide the create post and create community buttons. I will make a separate PR for that.

### DIFF
--- a/src/shared/components/community/sidebar.tsx
+++ b/src/shared/components/community/sidebar.tsx
@@ -105,10 +105,10 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
             {myUSerInfo && this.blockCommunity()}
             {!myUSerInfo && (
               <div className="alert alert-info" role="alert">
-                You are not logged in. However you can subscribe from another
-                Fediverse account, for example Lemmy or Mastodon. To do this,
-                paste the following into the search field of your instance: !
-                {name}@{hostname(actor_id)}
+                {i18n.t("community_not_logged_in_alert", {
+                  community: name,
+                  instance: hostname(actor_id),
+                })}
               </div>
             )}
           </div>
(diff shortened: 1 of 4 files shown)


src/shared/components/community/sidebar.tsx
```diff
@@ -91,15 +92,25 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
...
           <div className="card-body">
             {this.communityTitle()}
             {this.props.editable && this.adminButtons()}
-            {this.subscribe()}
+            {myUSerInfo && this.subscribe()}
             {this.canPost && this.createPost()}
-            {this.blockCommunity()}
+            {myUSerInfo && this.blockCommunity()}
+            {!myUSerInfo && (
+              <div className="alert alert-info" role="alert">
```
dessalines, line 107: Obvi make a key for this in `lemmy-translations/`. There are examples for how to include the variables in there too. Since you don't need html, you can just do:

`"not_logged_in_message":  "bunch of text... {{apubShortCode}}"`

src/shared/components/person/profile.tsx
```diff
@@ -530,6 +530,12 @@ export class Profile extends Component<
                     .format("MMM DD, YYYY")}
                 </span>
               </div>
+              {!UserService.Instance.myUserInfo && (
+                <div className="alert alert-info" role="alert">
```
dessalines, line 534: Same

SleeplessOne1917:
@dessalines Are there any other changes this needs before it's good to merge?

dessalines (approved):
Works great, thanks! I apologize it took me a while to get to this.

PR #987, opened 2023-04-23
Handle when logged out - Hide create post version

SleeplessOne1917:
This is a PR for #975 and is the alternate version of #986.

This version includes the same alerts as the other version plus this alert when not signed in on the homepage:

home page alert

The same points I made about the alerts on the other PR apply to this alert as well.


Nutomic:
I dont like this. If people want to browse without an account thats fine, no need to bother them.

dessalines:
Same, and also its better not to hide buttons like that, but redirect.

SleeplessOne1917:
Closing this as the other version is preferred.

Issue #991, opened 2023-04-27
Add content warning to modlog ala Hexbear


SleeplessOne1917:
I noticed that Hexbear has a content warning on their modlog:
![image](https://user-images.githubusercontent.com/28871516/234995010-152fb70d-ec96-44fc-9d81-f9bf0e5768a4.png)

Upon some reflection, this seems like a good addition to the upstream. Considering someone saying bigoted or otherwise disturbing things is going to be banworthy for most admin teams that aren't asleep at the wheel, and all of those potentially disturbing things will be displayed in the modlog, giving users a heads up seems sensible.

dessalines:
Agree, seems like a good idea.

PR #994, opened 2023-05-03
Add content warning to modlog and fix modlog routing bug

SleeplessOne1917:
This is the PR for #991. 

This is what it looks like on a desktop:
modlog-pc

This is what it looks like on mobile:
modlog-mobile

Translation PR will be made soon.

In addition to the content warning, I also fixed a bug I introduced in the query param PR regarding routing for community specific modlogs.

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -99,7 +99,7 @@
   border-bottom: 2px solid var(--dark);
 }
 
-.md-div table tbody+tbody {
+.md-div table tbody + tbody {
   border-top: 2px solid var(--dark);
 }
 
@@ -140,8 +140,8 @@
 }
 
 .icon-emoji-admin {
-  max-width: 24px; 
-  max-height: 24px; 
+  max-width: 24px;
+  max-height: 24px;
   display: inline-block;
 }
 
@@ -164,7 +164,7 @@
     left: 0;
   }
 
-  .emoji-picker-container>section {
+  .emoji-picker-container > section {
     width: 100% !important;
   }
 }
@@ -175,7 +175,7 @@
   left: 0;
   right: 0;
   bottom: 0;
-  background-color: rgba(0, 0, 0, .3);
+  background-color: rgba(0, 0, 0, 0.3);
   z-index: 999;
 }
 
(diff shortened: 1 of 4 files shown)


src/assets/css/main.css
```diff
@@ -140,8 +140,8 @@
 }
 
 .icon-emoji-admin {
-  max-width: 24px; 
-  max-height: 24px; 
+  max-width: 24px;
+  max-height: 24px;
```
dessalines, line 144: Thanks. We should alter prettier / to make sure it works for css files too, I'll add that now.

src/shared/components/community/sidebar.tsx
```diff
@@ -251,7 +251,7 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
         <li className="list-inline-item">
           <Link
             className="badge badge-primary"
-            to={`/modlog/community/${this.props.community_view.community.id}`}
+            to={`/modlog/${this.props.community_view.community.id}`}
```
dessalines, line 254: Oops, my bad. Thanks for catching.

src/shared/components/modlog.tsx
```diff
@@ -782,6 +782,18 @@ export class Modlog extends Component<
...
+          <div
+            className="alert alert-warning text-sm-start text-xs-center"
+            role="alert"
+          >
+            <Icon
+              icon="alert-triangle"
+              inline
+              classes="mr-sm-2 mx-auto d-sm-inline d-block"
+            />
+            <strong>CONTENT WARNING</strong>: Some deleted posts may contain
```
dessalines, line 794: Language looks good to me.

dessalines (commented):
Once you get the translation added, update this and I can merge.

src/shared/components/community/sidebar.tsx
SleeplessOne1917: TBH I think I might have introduced that bug. That route was correct until the query param change I made.

dessalines:
K that's merged.

dessalines (approved):

PR #995, opened 2023-05-03
Prettier css

dessalines:


package.json
```diff
@@ -21,6 +21,9 @@
       "prettier --write",
       "eslint --fix"
     ],
+    "*.{css, scss}": [
+      "prettier --write"
```
dessalines, line 25: eslint gets an error here if it tries to run on css files.

SleeplessOne1917 (approved):
LGTM

PR #1003, opened 2023-05-10
Using auto-generated types from ts-rs.

### LINKED ISSUE #998: Upgrade to new lemmy-js-client types
The types are auto-generated now, but the changes will affect nearly every lemmy-ui import.

dessalines:
- Fixes #998
- Added support for new `GetFederatedInstances`
- Fixed a few bugs in the process.

Notes:

- Changed a lot of `number` types to `bigint`. This is the official type from rust, so it unfortunately added some boilerplate.
- Imports aren't that clean, but IDE support should take care of that.


src/shared/components/app/footer.tsx
```diff
@@ -1,6 +1,6 @@
 import { Component } from "inferno";
 import { NavLink } from "inferno-router";
-import { GetSiteResponse } from "lemmy-js-client";
+import { GetSiteResponse } from "lemmy-js-client/dist/types/GetSiteResponse";
```
SleeplessOne1917, line 3: Can the types not be imported from just `lemmy-js-client`?

SleeplessOne1917: @dessalines I made a PR for the js client that should get rid of the need for these individual imports: https://github.com/LemmyNet/lemmy-js-client/pull/118.

dessalines: Thanks, I'll make this a draft until I fix up the imports.

SleeplessOne1917 (approved):

Issue #996, opened 2023-05-04
Make webmanifest use custom icon, name, and description


SleeplessOne1917:
Currently, whenever Lemmy is used as a PWA, the default icon, name, and description are always used.
For example, the instance lemmygrad.ml is named "Lemmygrad" and has the following icon:
![Lemmygrad icon](https://lemmygrad.ml/pictrs/image/gB8yP0oFF5.png?format=webp&thumbnail=96)

Compare this to what the icon for the Lemmygrad PWA looks like on my phone:
![PWA_icon](https://user-images.githubusercontent.com/28871516/236080000-708bd037-e316-480e-96bc-25e60da0e2bd.jpg)

Currently, the site name, icon, and description are set in the admin page and saved in the database. To implement this, I think we'll have to either have instance admins set those things in the site config, or we could generate the manifest dynamically when a page is being fetched.

dessalines:
The [webmanifest currently is just a static file](https://github.com/LemmyNet/lemmy-ui/blob/main/src/assets/manifest.webmanifest#L2), but it could built on the fly and served from the node server. It could at least get the correct name of the site and at least one size for the icon.

SleeplessOne1917:
Closing this since the PR was merged.

PR #1005, opened 2023-05-11
Add support for PWA

SleeplessOne1917:
This was originally meant to address #996, however I decided to set up the service worker so that it would work as a PWA properly on multiple platforms.


dev.dockerfile
```diff
@@ -3,9 +3,13 @@ RUN apk update && apk add curl yarn python3 build-base gcc wget git --no-cache
 
 WORKDIR /usr/src/app
 
+ENV npm_config_target_arch=x64
+ENV npm_config_target_platform=linux
+ENV npm_config_target_libc=musl
+
 # Cache deps
 COPY package.json yarn.lock ./
-RUN yarn install --ignore-scripts --prefer-offline --pure-lockfile
```
SleeplessOne1917, line 8: This and the environment variables are required to make the image processing library work. See https://github.com/lovell/sharp/issues/3663.

src/shared/env.ts
SleeplessOne1917, line 1: I was getting constants meant for the client on the server. Changing them to functions fixed it.

src/server/index.tsx
```diff
@@ -267,3 +301,63 @@ function removeParam(url: string, parameter: string): string {
...
+
+const iconSizes = [72, 96, 128, 144, 152, 192, 384, 512];
+const defaultLogoPathDirectory = path.join(
+  process.cwd(),
+  "dist",
+  "assets",
+  "icons"
+);
+
+export async function generateManifestBase64(site: Site) {
```
dessalines, line 313: We should probably re-organize some of this into separate files... not necessary for this PR tho.

src/shared/env.ts
```diff
@@ -2,49 +2,53 @@ import { isBrowser } from "./utils";
...
-export const httpBaseInternal = `http://${host}`; // Don't use secure here
-export const httpBase = `http${secure}://${host}`;
-export const wsUriBase = `ws${secure}://${wsHost}`;
-export const wsUri = `${wsUriBase}/api/v3/ws`;
-export const isHttps = secure.endsWith("s");
-
-console.log(`httpbase: ${httpBase}`);
-console.log(`wsUri: ${wsUri}`);
-console.log(`isHttps: ${isHttps}`);
+const getInternalHost = () =>
```
dessalines, line 5: Functions, not closures. Change all of these awful javascript fake functions into real ones.

src/server/index.tsx
```diff
@@ -267,3 +301,63 @@ function removeParam(url: string, parameter: string): string {
...
+
+  const manifest = {
+    name: site.name,
+    description: site.description ?? "A link aggregator for the fediverse",
+    start_url: url,
+    scope: url,
+    display: "standalone",
+    id: "/",
+    background_color: "#222222",
+    theme_color: "#222222",
```
SleeplessOne1917, line 329: I wonder if theme_color and background_color should be configurable by instance admins?

dessalines (approved):

src/server/index.tsx
dessalines: mmm.... lets not do it until someone asks for it lol. Plus I think that would require a local site setting in the database, or some image processing... pry overkill.

Issue #919, opened 2023-02-03
Create a comment in Deutsch is not allowed


xundeenergie:
Found a bug? Please fill out the sections below. 👍

For backend issues, use [lemmy](https://github.com/LemmyNet/lemmy)

### Issue Summary

When i create a comment, and the language for the comment is "Deutsch", i get the error "language not allowed" in the Webfrontend.
The "Answer"-Button changes to a spinner and is not usable anymore. It stays spinning. I have to "Cancel" and create it again, with the right language "English"

On the server i see in the logs this ERROR-Message

```
lemmy_1     | 2023-02-03T15:22:54.907502Z ERROR lemmy_server::api_routes_websocket: Error during message handling language_not_allowed: language_not_allowed                                                       
lemmy_1     |    0: lemmy_api_crud::comment::create::perform                                                                                                                                                       
lemmy_1     |            with self=CreateComment { content: "which one is it?", post_id: PostId(132386), parent_id: Some(CommentId(59722)), language_id: Some(LanguageId(0)), form_id: Some("comment-form-BWf3BJIag
OlM72MKlKvS"), auth: Sensitive }                                                                                                                                                                                   
lemmy_1     |              at crates/api_crud/src/comment/create.rs:41                                                                                                                                             
```

No message in Browser-Console.

The really strange thing is, Comments with "German" in the SAME Posting on other places are no problem.

Answering here https://lemmy.ml/comment/381491 or here https://rytter.me/objects/4c906314-1763-db68-da8c-fff695697098 from my lemmy (lemmy.schuerz.at) gives me the error.

Answering here https://lemmy.ml/comment/381812 from my lemmy in "Deutsch" is no problem.
Its the same Posting!

My language-Settings are "Deutsch" and "English" in adminsitration-section

dessalines:
Do you have Deutsch as an allowed language for both the community, and the instance?

xundeenergie:
As i told you:
Both happens in the SAME posting.
On one comment i can answer with "Deutsch" on another comment I'm not allowed to use "Deutsch"

Same community, same posting. It is a remote-followed community, not from my instance. 
It's from lemmy.ml and the posting from my examples is the announcement for lemmy 0.17.0

dessalines:
`self=CreateComment { content: "which one is it?", post_id: PostId(132386), parent_id: Some(CommentId(59722)), language_id: Some(LanguageId(0)`

[...]

PR #1008, opened 2023-05-11
Dont preselect new post language

Nutomic:
Lemmy-ui currently preselects the first language in the user settings when creating a new post or comment. This is a bad idea because this language might not actually be allowed in the community. It is better to pass the language as None if the user didnt specify it explicitly, because then the backend can smartly choose a language based on the overlap of user languages and community languages.

This fixes the problem described in
[this thread](https://lemmy.ml/post/1066608), where a user tries to post in a community that has only English allowed, with all languages enabled in user settings. In this case lemmy-ui preselects "undetermined language" as default, which is not allowed and results in an error. This PR fixes the issue because it lets the backend automatically select the correct language (English).

Edit: This probably fixes https://github.com/LemmyNet/lemmy-ui/issues/919 too.

### DIFF
--- a/src/shared/components/comment/comment-form.tsx
+++ b/src/shared/components/comment/comment-form.tsx
@@ -17,7 +17,6 @@ import { UserService, WebSocketService } from "../../services";
 import {
   capitalizeFirstLetter,
   myAuth,
-  myFirstDiscussionLanguageId,
   wsClient,
   wsSubscribe,
 } from "../../utils";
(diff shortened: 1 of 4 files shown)


SleeplessOne1917 (approved):

Nutomic:
Not sure why the CI is complaining about i18n key.

SleeplessOne1917:
@Nutomic I think it's failing because the translation PR adding the key wasn't merged yet.

dessalines:
I merged that lemmy-translation, you can run `git submodule update --remote`.

dessalines (approved):

dessalines:
Needs prettier run.

Issue #980, opened 2023-04-15
Create a better UI for blocking / allowed instances


dessalines:
It could look similar to how blocking users / communities works.

PR #1012, opened 2023-05-12
Make admin UI for allowing/blocking instances easier to work with

SleeplessOne1917:
This is the PR for #980.

This is what it looks like on desktop:
allow-block-desktop

This is what it looks like on mobile:
allow-block-mobile

It's worth noting that on the main branch when I saved the site form, navigated away from the site settings page, then navigated to the setting page again, the instances were not saved. I don't know if this is because I did something wrong entering the instances. **EDIT**: It turns out that the instances do save, but that they are not fetched when navigating by clicking a link. You need to trigger the isomorphic server side code by refreshing to actually get your saved allowed/blocked instances to show up.

One thing that I'm wondering is if it makes sense for instances to have both a block list and allow list defined. If the backend already limits the instance to using only one or the other, let me know. I'll update the UI so that only one can be used.


Nutomic:
Using both together can make sense, because with allowlist you can still see posts from non-allowed instances inside of communities. Blocklist completely prevents federation of any content from those instances.

src/shared/components/home/site-form.tsx
```diff
@@ -930,6 +909,83 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+            onClick={() => this.handleAddInstance(key)}
+            tabIndex={
+              -1 /* Making this untabble because handling enter key in text input makes keyboard support for this button redundant */
+            }
+          >
+            <Icon icon="add" classes="icon-inline text-light m-auto" />
+          </button>
+        </div>
+        {selectedInstances && selectedInstances.length > 0 && (
+          <ul className="mt-3 list-unstyled w-100 d-flex flex-column justify-content-around align-items-center">
```
dessalines, line 943: Thank you for using bootstrap and flex classes here.

src/shared/components/home/site-form.tsx
```diff
@@ -930,6 +909,83 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+          {i18n.t(key)}
+        </label>
+        <div className="d-flex justify-content-between align-items-center">
+          <input
+            type="text"
+            placeholder="instance.tld"
+            id={id}
+            className="form-control"
+            value={value}
+            onInput={event => this.handleInstanceTextChange(key, event)}
```
dessalines, line 928: With inferno, all of the *on* , or anything that deals with event, should use `linkEvent`. Check the other forms for examples.

src/shared/components/home/site-form.tsx
```diff
@@ -930,6 +909,83 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+            onKeyUp={event => this.handleInstanceEnterPress(key, event)}
+          />
+          <button
+            type="button"
+            className="btn btn-sm bg-success ml-2"
+            onClick={() => this.handleAddInstance(key)}
```
dessalines, line 934: Same here.

src/shared/components/home/site-form.tsx
```diff
@@ -930,6 +909,83 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+                className="my-1 w-100 w-md-75 d-flex align-items-center justify-content-between"
+              >
+                <label className="d-block m-0 w-100 " htmlFor={instance}>
+                  <strong>{instance}</strong>
+                </label>
+                <button
+                  id={instance}
+                  type="button"
+                  className="btn btn-sm bg-danger"
+                  onClick={() => this.handleRemoveInstance(key, instance)}
```
dessalines, line 956: Same

src/shared/components/home/site-form.tsx
```diff
@@ -986,18 +1042,37 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
-  instancesToString(opt?: string[]): string {
-    return opt ? opt.join(",") : "";
-  }
-
-  handleSiteAllowedInstances(i: SiteForm, event: any) {
-    let list = splitToList(event.target.value);
-    i.setState(s => ((s.siteForm.allowed_instances = list), s));
+  handleAddInstance(key: InstanceKey) {
+    const instance = this.state.instance_select[key].trim();
+    if (!this.state.siteForm[key]?.includes(instance)) {
```
dessalines, line 1047: Probably a good check, thx.

dessalines (commented):
Works well, thx! Just change those event handlers to `linkEvent` and we can merge.

src/shared/components/home/site-form.tsx
SleeplessOne1917: Do you know how to use 'linkEvent'  when passing in parameters other than the event and component? I wanted to use it originally, but I don't know how I'd pass the 'key' variable to the handler in that case.

SleeplessOne1917: Nvm, it seems that linkEvent can accept things besides events and components as arts.

dessalines: Here's an example: https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/person/settings.tsx#L282

Issue #982, opened 2023-04-15
Have a 500 error page for server errors


dessalines:
This currently just times out.

PR #1013, opened 2023-05-12
Improve error handling (fixes #897)

### LINKED ISSUE #897: Setup page is publicly available after instance created
It can only be used to create normal user accounts, but the route should still be disabled.

https://lemmy.ml/setup

https://ds9.lemmy.ml/setup

via https://lemmy.ml/post/717182/comment/373149

Nutomic:
A lot of routes are missing basic checks, for example /setup is publicly available even after the site is created. For pages which require auth, the check is handled separately each time when it could be done once at the top level. This is currently work in progress to see if Im on the right track. The same checks should be added to many more routes.

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -2,7 +2,7 @@ import express from "express";
 import fs from "fs";
 import { IncomingHttpHeaders } from "http";
 import { Helmet } from "inferno-helmet";
-import { matchPath, StaticRouter } from "inferno-router";
+import { StaticRouter, matchPath } from "inferno-router";
 import { renderToString } from "inferno-server";
 import IsomorphicCookie from "isomorphic-cookie";
 import { GetSite, GetSiteResponse, LemmyHttp } from "lemmy-js-client";
@@ -141,7 +141,7 @@ server.get("/*", async (req, res) => {
 
     const routeData = await Promise.all(promises);
 
-    // Redirect to the 404 if there's an API error
+    // Handle API errors
     if (routeData[0] && routeData[0].error) {
       const error = routeData[0].error;
       console.error(error);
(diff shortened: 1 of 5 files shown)


src/server/index.tsx
```diff
@@ -223,10 +223,17 @@ server.get("/*", async (req, res) => {
...
-    return res.send(`404: ${removeAuthParam(err)}`);
+    var formatted: string;
+    if (err.status && err.message) {
+      res.status(err.status);
+      formatted = `${err.status}: ${err.message}`;
+    } else {
+      res.status(500);
+      formatted = err;
+    }
+    res.send(formatted);
```
Nutomic, line 236: This could be improved further by displaying the error message as a full page with normal header and footer (if `site` was requested successfully). In case of error 401 it can also redirect to /login like some routes already do.

SleeplessOne1917: The first part was mentioned in #982. If you don't do it as part of this PR, I could whip something up. I'm wondering: if the site returns a server error but the site is still fetched, is it a good idea to list the matrix user IDs of instance admins for users to reach out to for support? That information is already on the site response, so no further requests would be required.

src/shared/components/home/admin-settings.tsx
```diff
@@ -60,6 +61,8 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
     this.parseMessage = this.parseMessage.bind(this);
     this.subscription = wsSubscribe(this.parseMessage);
 
+    check_auth();
```
SleeplessOne1917, line 64: Instead of doing this on the client, I think it would  better to handle this on the server and redirect the user before even sending the page that requires auth back.

Nutomic: To be honest Im not good at frontend dev. If you want you can take over this PR because it takes me forever just to implement the basics. So Id rather stay in Rust :)

PR #1016, opened 2023-05-13
Redirect form pages that require auth on logout

SleeplessOne1917:
I didn't make an issue for this one, but the code change does what the PR title says. Currently, if a user logs out of their account while on a page that requires auth (like /settings, /admin, /inbox, etc), they will not be navigated away from the page.

### DIFF
--- a/src/shared/services/UserService.ts
+++ b/src/shared/services/UserService.ts
@@ -48,7 +48,15 @@ export class UserService {
     this.myUserInfo = undefined;
     IsomorphicCookie.remove("jwt"); // TODO is sometimes unreliable for some reason
     document.cookie = "jwt=; Max-Age=0; path=/; domain=" + location.hostname;
-    location.reload();
+    if (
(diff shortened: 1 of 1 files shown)


src/shared/services/UserService.ts
```diff
@@ -48,7 +48,15 @@ export class UserService {
     this.myUserInfo = undefined;
     IsomorphicCookie.remove("jwt"); // TODO is sometimes unreliable for some reason
     document.cookie = "jwt=; Max-Age=0; path=/; domain=" + location.hostname;
-    location.reload();
+    if (
+      /create_.*|inbox|settings|setup|admin|reports|registration_applications/g.test(
```
SleeplessOne1917, line 52: It might be worth extracting this into a function and putting it in utils. It would help with what I called out [on another PR](https://github.com/LemmyNet/lemmy-ui/pull/1013#discussion_r1192991566).

dessalines (approved):
Seems good.

PR #1021, opened 2023-05-15
Fix strange navbar bigint issue.

dessalines:


dessalines:
Not necessary anymore.

PR #1024, opened 2023-05-16
Nicer error handling

SleeplessOne1917:
[I took up Nutomic's offer](https://github.com/LemmyNet/lemmy-ui/pull/1013#discussion_r1193031824) and implemented what was being implemented in #1013.
This is what the 404 error page looks like:
404-error-light

This is for a 500 error:
500-error-light

Besides making the error pages, I made it so that making sure an authenticated route has expected behavior with logouts and redirects only requires a change to one place. This should hopefully help make things easier to maintain.


src/shared/components/common/error-guard.tsx
```diff
@@ -0,0 +1,26 @@
+import { Component } from "inferno";
+import { ErrorPageData, setIsoData } from "../../utils";
+import { ErrorPage } from "../app/error-page";
+
+class ErrorGuard extends Component<any, any> {
```
SleeplessOne1917, line 5: I just thought of something. I could probably handle this on the server. It would probably both reduce bundle size and make my hacky use of 'routeData' unnecessary.

SleeplessOne1917:
@dessalines before you merge, I'd like to give what I mentioned [here](https://github.com/LemmyNet/lemmy-ui/pull/1024#discussion_r1195149761) a try.

SleeplessOne1917:
@dessalines I almost forgot: I'm going to have to add translations for those error messages.

Nutomic:
That looks great! But please dont link to the issue tracker, it will lead to a lot of issues opened for tech support. Instead it can link to [!lemmy_support](https://lemmy.ml/c/lemmy_support) and the [Matrix chat](https://matrix.to/#/#lemmy-space:matrix.org). That way users can help each other and devs dont have to help with every little problem.

package.json
```diff
@@ -61,6 +61,7 @@
     "isomorphic-cookie": "^1.2.4",
     "jwt-decode": "^3.1.2",
     "lemmy-js-client": "0.17.2-rc.17",
+    "lodash": "^4.17.21",
```
dessalines, line 64: lodash seems unused, and I'd def like to avoid it if possible.

src/shared/components/app/error-page.tsx
```diff
@@ -0,0 +1,71 @@
...
+  render() {
+    const { errorPageData } = this.isoData;
+
+    return (
+      <div className="container-lg text-center">
+        <h1>{errorPageData ? "Error!" : "Page Not Found"}</h1>
+        <p className="p-4">
+          {errorPageData ? (
+            <>
+              There was an error on the server. Try refreshing your browser. If
```
dessalines, line 22: This file needs i18n translations.

src/shared/components/common/error-guard.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  constructor(props: any, context: any) {
+    super(props, context);
+  }
+
+  render() {
+    const errorPageData = this.isoData.errorPageData;
+    const siteRes = this.isoData.site_res;
+    console.log("In error guard");
+    console.log(errorPageData);
```
dessalines, line 16: Needs to remove the auth data, somewhere in this file most likely.

src/shared/components/common/error-guard.tsx
dessalines: I just know that the server will return the json `{"error: "string"}`, for any failed request. Not sure how that translates for this. 

In jerboa at least, all these errors do proper exception throws, so I can catch them, read the error message, and do actions based on that.

src/shared/utils.ts
```diff
@@ -105,6 +105,11 @@ export type ThemeColor =
   | "gray"
   | "gray-dark";
 
+export interface ErrorPageData {
+  error?: string;
+  adminMatrixIds?: string[];
```
dessalines, line 110: What's this one for?

package.json
SleeplessOne1917: `lodash` is only being used in `webpack.config.js`. I originally tried having it as a dev dependency, but this caused issues when I tested the prod Dockerfile. This shouldn't be included in the bundle.

src/shared/components/app/error-page.tsx
SleeplessOne1917: Done

src/shared/components/common/error-guard.tsx
SleeplessOne1917: The error only gets added to the page data when the UI is run in development. From lines 300-304 in the server `index.ts`:
```typescript
  // Exact error should only be seen in a development environment. Users
  // in production will get a more generic message.
  if (process.env.NODE_ENV === "development") {
    errorPageData.error = error;
  }
```

However, I will remove the console logs as they are unnecessary.

src/shared/utils.ts
SleeplessOne1917: 500-error-matrix
If there's a server error but the site is fetched successfully and there are admins with matrix IDs, the error page will list those matrix IDs for users to contact.

src/shared/components/common/error-guard.tsx
SleeplessOne1917: My comment was unclear. I meant on the server side of the UI app that serves the pages, but not the actual Lemmy server.

As to doing different actions based on different error messages, that's definitely doable. We just need to defined what should be done for those error messages. At the very least there should be a way to handle displaying that a user is banned (I even think there's an open issue for that).

dessalines (approved):
Looks good to me. I'll let nutomic also look at this one.

package.json
makotech222: yeah i got this issue too when trying to run docker compose against lemmy-ui

src/shared/utils.ts
dessalines: Gotcha, cool.

src/server/index.tsx
```diff
@@ -361,3 +293,116 @@ async function fetchIconPng(iconUrl: string) {
...
 }
+
+function getErrorPageData(error: string, site?: GetSiteResponse) {
+  const errorPageData: ErrorPageData = {};
+
+  // Exact error should only be seen in a development environment. Users
+  // in production will get a more generic message.
+  if (process.env.NODE_ENV === "development") {
+    errorPageData.error = error;
+  }
```
Nutomic, line 304: Im not sure this makes sense. The reason not to show error messages is usually for security, but anyway the error messages are available if you use the API directly. 

Another reason might be to keep things simple for non-techie users. But even if they dont understand the error, there can be a text telling them to copy-paste the error message to admins/devs. That can be very helpful for debugging.

SleeplessOne1917:
@Nutomic @dessalines I made a blurb for the error code. Here is what it looks like:
image

If you like the verbiage, I'll submit a translation. On the topic of errors, [I have an idea for Lemmy errors](https://github.com/LemmyNet/lemmy/issues/2871) that I'm looking for feedback on.

Issue #1020, opened 2023-05-15
Prevent user from posting comment if language is not selected


SleeplessOne1917:
This is related to both #1015 and the selected language change that was just pushed. The UI lets a user post a comment if they don't have a language selected. This leads to a broken form like described in #1015.

SleeplessOne1917:
It turns out a language is automatically selected on the backend if a language is not selected.

PR #1025, opened 2023-05-16
Make post button disabled when no language is selected

SleeplessOne1917:
This is for #1020. I think this is better than letting users post something that is guaranteed to cause an error, but I'm worried it won't be obvious why the button is disabled to the user.  I'm open to ideas to make the fact that languages won't be auto selected anymore noticeable to the user.

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -132,7 +132,7 @@ export class MarkdownTextArea extends Component<
   }
 
   render() {
-    let languageId = this.state.languageId;
+    const languageId = this.state.languageId;
 
     return (
       <form id={this.formId} onSubmit={linkEvent(this, this.handleSubmit)}>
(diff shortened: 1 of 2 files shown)


Nutomic:
Strongly disagree with this. Posting without specifying a language is not a problem, the backend has [code to select a default language](https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/impls/actor_language.rs#L313) in that case. So if the community has only one language (excluding undetermined), that is selected automatically. Same if the user has only one language, or if the overlap between both is a single language (again excluding undetermined). That should already cover a lot of cases.

If that doesnt work, having "undetermined" as language is totally fine. Most other platforms dont support language tags, so there will be a lot of posts with undetermined language anyway. And being forced to specify it every time would be way too annoying.

About the error `language_not_allowed`, this happens if a user tries to post in a language which the community doesnt have enabled. For example a community only has English and French enabled in the community settings, but a user tries to post in Spanish. This should ideally be prevented by removing the language from the dropdown.

SleeplessOne1917:
I did a test and @Nutomic is correct: the backend will automatically pick a language if one is not selected. This story should not be needed, then.

Issue #1010, opened 2023-05-11
Language indicator for posts and comments


Nutomic:
Posts and comments can specify which language they are written in. This should be displayed in the UI.

wiki-me:
isn't this a little too excessive? is that really useful information that is worth the cognitive load? If you know the language you can recognize it and just read it, if not why should i care what language is displayed? If i want to translate it there is the "detect language" functionality on google translate.

Nutomic:
You can make the same argument about other indicators, such as those for thread creator or admin/mod. None of these are required because you can get the information in other ways. But as we already have this information, it makes sense to display it.

wiki-me:
> You can make the same argument about other indicators, such as those for thread creator or admin/mod. None of these are required because you can get the information in other ways.

In those cases i can come up with "user story" for why they could be used (So it might make sense to provide easy access for them), I can't think of a user story for this feature.

PR #1026, opened 2023-05-17
Show language on posts and comments

SleeplessOne1917:
This is implementing #1010 
Here is what it looks like for posts:
language_marker_post

Here is what it looks like for comments:
language_marker_comment


Nutomic:
I dont really like this icon that you used for languages. I would remove that and use a background instead, similar to the `creator` icon. Maybe in a slightly different color.

SleeplessOne1917:
How does this look:
Dark mode
dark
Light mode
light

src/shared/components/comment/comment-node.tsx
```diff
@@ -282,8 +282,15 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 )}
               </button>
               {this.linkBtn(true)}
+              <span className="mx-1 badge badge-secondary">
```
dessalines, line 285: I like the badge a lot for these.

dessalines (approved):

PR #1032, opened 2023-05-21
Show parent comment for comment link. Fixes #1030

### LINKED ISSUE #1030: Suggestion: Make the "Show Context" Link in the Inbox show the Parent Comment

In your inbox, it's hard to know exactly which comment someone is replying to. When clicking on the "Show context" button, you're taken to a page that effectively conveys the same information:

Referring to the following Inbox button:
image

Takes to a page like:
image

At the risk of simply "Copying Reddit", pressing the "Context" button on old.reddit.com's inbox takes you to the comment *in the message chain* so you can see the previous responses as well:

Example of what you see on Reddit:
image

Rather than https://lemmy.ml/comment/429630 linking to the comment under the OP with another button to select "Show Context", it might be best to simply show 2-3 parent comments (if they exist) on page load, and highlight the comment instead, like this:

image

(Pretend the blue highlight is part of the UI; the URL for comment 429630 shows parents as well, and has an additional highlight on comment 429360)

dessalines:


src/shared/components/comment/comment-node.tsx
```diff
@@ -1051,11 +1052,14 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
       ? i18n.t("show_context")
       : i18n.t("link");
 
+    // The context button should show the parent comment by default
+    let parentCommentId = getCommentParentId(cv.comment) ?? cv.comment.id;
```
SleeplessOne1917, line 1056: Nitpick: `const`, not `let`.

SleeplessOne1917 (approved):
LGTM

PR #1033, opened 2023-05-22
Navbar hide menu fix

SleeplessOne1917:
There isn't an issue open for this one, but currently not all options in the mobile menu cause the nav to collapse. I also removed some excess HTML.


dessalines (approved):
This worked great, thx.

PR #1036, opened 2023-05-23
Changes to language tag

Nutomic:
- Dont use bright white color because its not that important
- Hide tag in case of undetermined (doesnt add any information)

Before:
Screenshot_20230523_221859

After:
Screenshot_20230523_221840

Not sure about the specific css classes, we can also change it as long as it's gray like the other text and not bright white.


src/shared/components/comment/comment-node.tsx
```diff
@@ -283,13 +283,15 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
               </button>
               {this.linkBtn(true)}
-              <span className="mx-1 badge badge-secondary">
-                {
-                  this.props.allLanguages.find(
-                    lang => lang.id === cv.comment.language_id
-                  )?.name
-                }
-              </span>
+              {cv.comment.language_id != 0 && (
```
SleeplessOne1917, line 286: Nitpick: !== instead of !=.

src/shared/components/post/post-listing.tsx
```diff
@@ -344,13 +344,15 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
           )}
         </li>
-        <span className="mx-1 badge badge-secondary">
-          {
-            this.props.allLanguages.find(
-              lang => lang.id === post_view.post.language_id
-            )?.name
-          }
-        </span>
+        {post_view.post.language_id != 0 && (
```
SleeplessOne1917, line 347: Same as other comment.

SleeplessOne1917:
Regarding the css class badges: you can get a badge of a given color with "badge badge-(color name here)". The color name would be something from the theme, like light, dark, info, primary etc. Text gets colored similarly with class names like "text-(theme color)".

Also, how does this look in lightly?

Nutomic:
Good suggestion, `badge-muted` looks decent as its the same color used for other text in that area (like the creation time). I also changed the admin/mod badges to muted (see https://github.com/LemmyNet/lemmy-ui/issues/1037) and moved the creator badge first.

Screenshot_20230524_122115

Screenshot_20230524_122133

SleeplessOne1917 (approved):

src/shared/components/comment/comment-node.tsx
```diff
@@ -240,23 +240,23 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
+                </div>
+              )}
+              {isMod_ && (
+                <div className="badge badge-muted d-none d-sm-inline mr-2">
                   {i18n.t("mod")}
                 </div>
               )}
               {isAdmin_ && (
-                <div className="badge badge-light d-none d-sm-inline mr-2">
+                <div className="badge badge-muted d-none d-sm-inline mr-2">
```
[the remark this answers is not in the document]
Nutomic: Youre right it actually has no effect. So I removed `badge-muted`, only leaving `badge` and it still looks the same as in the screenshots.

src/shared/components/post/post-listing.tsx
```diff
@@ -345,7 +345,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
           )}
         </li>
         {post_view.post.language_id !== 0 && (
-          <span className="mx-1 badge badge-muted">
+          <span className="mx-1 badge">
```
dessalines, line 348: You're not supposed to use it without the 2nd `badge-x` class, as it might look weird on some themes. Did you try `badge-secondary` ?

SleeplessOne1917: Secondary will be orange on lightly I think. It'll bring us back to the badge being too prominent.

Nutomic: Well its working...

dessalines: If it looks good on both the litely and darkly, I have no problems merging. Might give issues on other themes tho.

PR #1040, opened 2023-05-26
Switch from websocket -> HTTP client.

dessalines:
To be updated later.


makotech222:
General Q: why are parents involved in handling api calls for their child components?

src/shared/utils.ts
```diff
@@ -911,27 +852,135 @@ export function getCommentIdFromProps(props: any): number | undefined {
...
-  if (found) {
-    found.comment.content = data.comment.content;
-    found.comment.distinguished = data.comment.distinguished;
-    found.comment.updated = data.comment.updated;
-    found.comment.removed = data.comment.removed;
-    found.comment.deleted = data.comment.deleted;
-    found.counts.upvotes = data.counts.upvotes;
-    found.counts.downvotes = data.counts.downvotes;
-    found.counts.score = data.counts.score;
+function editListImmutable<T>(fieldName: string, data: T, list: T[]): T[] {
```
SleeplessOne1917, line 855: I made a version of this with better typing:
```typescript
type FieldNames = "person_mention" | "community" | "private_message"; // etc.

function editListImmutable<T extends {[key in F]: {id: number}}, F extends FieldNames>(fieldName: F, data: T, list: T[]): T[] {
  const foundIndex = list.findIndex(c => c[fieldName].id === data[fieldName].id);
  if (foundIndex != -1) {
    const newList = list;
    newList[foundIndex] = data;
    return newList;
  } else {
    return list;
  }
}
```

src/shared/utils.ts
```diff
@@ -911,27 +852,135 @@ export function getCommentIdFromProps(props: any): number | undefined {
...
+  if (foundIndex != -1) {
+    const newList = list;
+    newList[foundIndex] = data;
+    return newList;
+  } else {
+    return list;
+  }
+}
+
+export function editComments(
```
SleeplessOne1917, line 866: If this is modifying a singular comment, the name should be singular instead of plural.

src/shared/components/app/navbar.tsx
```diff
@@ -39,88 +31,46 @@ interface NavbarProps {
...
 
   constructor(props: any, context: any) {
     super(props, context);
-
-    this.parseMessage = this.parseMessage.bind(this);
-    this.subscription = wsSubscribe(this.parseMessage);
   }
 
-  componentDidMount() {
+  async componentDidMount() {
```
SleeplessOne1917, line 56: Can lifecycle methods be async?

src/shared/services/HttpService.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * A Special type of apiWrapper, used only for the iso routes.
+ *
+ * Necessary because constructors can't be async
+ */
+export function apiWrapperIso<ResponseType>(
+  res: ResponseType
+): RequestState<ResponseType> {
+  try {
+    return {
```
SleeplessOne1917, line 63: What would throw an error inside of this try block?

src/shared/utils.ts
dessalines: Thx, I'll update this shortly.

src/shared/utils.ts
```diff
@@ -911,27 +852,135 @@ export function getCommentIdFromProps(props: any): number | undefined {
...
-    found.comment.updated = data.comment.updated;
-    found.comment.removed = data.comment.removed;
-    found.comment.deleted = data.comment.deleted;
-    found.counts.upvotes = data.counts.upvotes;
-    found.counts.downvotes = data.counts.downvotes;
-    found.counts.score = data.counts.score;
+function editListImmutable<T>(fieldName: string, data: T, list: T[]): T[] {
+  const foundIndex = list.findIndex(c => c[fieldName].id == data[fieldName].id);
+  if (foundIndex != -1) {
+    const newList = list;
```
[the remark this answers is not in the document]
dessalines: My current hack is good enough for now I suppose. 

The best long-term solution, is to not use javascript. 😆 Okay really tho we should probably use https://immutable-js.com/ , especially for all our state vars. But that can come later.

src/shared/components/app/navbar.tsx
dessalines: Ya they can.

src/shared/services/HttpService.ts
dessalines: I fixed this in a `use_http_client_2`

dessalines:
Closed in favor of the `use_http_client_2` branch.

dessalines:
@makotech222 Jetpack compose, and its signal-based reactivity, and how it handles state, has taught me how to be a better UI designer. A lot of this comes from [here](https://developer.android.com/codelabs/basic-android-kotlin-compose-viewmodel-and-state#0)

In general, you want 3 layers: 

- viewmodels (IE your data storage layer that *should* persist even after switch screens. We aren't using that here, but how I've structured it makes it really easy to do that in the future).
  - These viewmodels (think of one living with each screen, holding only immutable data) should do all your API and DB calls, and should store your app data.
- Screens / activities. This is the "skeleton", and should be the only thing that interacts with your viewmodels directly. 
- Components. These should not hold much if any state besides forms, and should pass up in `onXAction` handlers any actions performed in them to the screens. They should not make API calls, or do anything than can't be easily mocked. Keeping them modular helps with re-use and testing, and avoids spaghetti code.

src/shared/utils.ts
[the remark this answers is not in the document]
dessalines: Oh I saw you added this one in the new PR, thx!

PR #1043, opened 2023-05-30
Route data refactor

SleeplessOne1917:
I found the route data somewhat unwieldy to work with. The way it is currently done, any `fetchInitialData` returns a list of promises, which are then resolved and used as a list in components as `isoData`. This requires component constructors to make sure that the correct array index is being accessed, and it requires adding promises that do nothing to `fetchInitialData`.

This change makes it so that each route data is an object with a type. This should hopefully make any isomorphic logic easier to maintain.


dessalines:
I'm all for something like this, but its not a good time, right when I'm in the middle of the websocket -> http rewrite. All the files you changed have been reworked pretty heavily, that would make merging this really difficult. 

You *could* rebase this on top of my `use_http_client` branch, but I wouldn't recommend it yet, because I'm still in the middle of getting everything working.

Lets hold off on this and any other larger changes until after I finish the http stuff, so we don't get stuck in merge hell.

alectrocute:
Are we out of merge hell territory now? This would be a really nice refactor.

SleeplessOne1917:
@dessalines (and @alectrocute since you seem interested too), the refactor is ready to review.

src/server/index.tsx
```diff
@@ -170,25 +174,34 @@ server.get("/*", async (req, res) => {
...
+          }, {});
         }
       }
     } else if (try_site.state === "failed") {
       errorPageData = getErrorPageData(new Error(try_site.msg), site);
     }
 
+    const error = Object.values(routeData).find(
+      res => res.state === "failed"
+    ) as FailedRequestState | undefined;
```
dessalines, line 196: Seems fine.

src/server/index.tsx
```diff
@@ -170,25 +174,34 @@ server.get("/*", async (req, res) => {
         };
 
         if (activeRoute?.fetchInitialData) {
-          routeData.push(
-            ...(await Promise.all([
-              ...activeRoute.fetchInitialData(initialFetchReq),
-            ]))
+          const routeDataKeysAndVals = await Promise.all(
+            Object.entries(activeRoute.fetchInitialData(initialFetchReq)).map(
```
dessalines, line 178: All the fetchInitialData got turned async, so I really hope this is working as intended.

dessalines (approved):
Looks good, thanks for going through and typing all these.

Run prettier again, lint is complaining.

Besides that, make sure you tested all these routes on first load. If they worked, feel free to merge.

dessalines:
Lets get this one merged next.

src/server/index.tsx
SleeplessOne1917: It wasn't, but it is now.

PR #1063, opened 2023-06-03
Fix sorting help link

mezzode:
Sorting link was going to https://join-lemmy.org/docs/en/users/01-getting-started.html/docs/en/users/03-votes-and-ranking.html instead of https://join-lemmy.org/docs/en/users/03-votes-and-ranking.html


SleeplessOne1917 (approved):

novakeith:
thanks - came to suggest that fix as well!

PR #1072, opened 2023-06-04
Comment depth

### LINKED ISSUE #1069: depth of threaded comments is not communicated to a screen reader
### Issue Summary

It is difficult for screen reader users to tell when a comment is a new top-level comment, or is a reply to another comment.  

### Steps to Reproduce

1. Log into Lemmy, and open a page of threaded comments
2. Run a screen reader (NVDA or Narrator on Windows, voiceover on Mac or IOS)
3. navigate through the comments, and notice that it's difficult to know when you've descended a level in a thread, or gone back up a level

### Technical details

A good solution to this might be putting hidden text before each comment that says "level 1" or "level 2" etc; I think this is how Reddit handles it? An even better solution would be formatting comments as nested lists. This would allow screen reader users to always know where we are in a thread, and use hotkeys to skip up and down the various levels.

SleeplessOne1917:
Closes #1069. In addition to making comment threads lists to make them easier to use with screen readers, I also made it easier to track of comment depth visually per [a user suggestion](https://lemmygrad.ml/comment/409228). This is what a thread looks like now:
image


AgentElement:
User that suggested the change here -- thanks for the (very) quick PR! Hoping this gets merged soon.

src/shared/components/comment/comment-nodes.tsx
```diff
@@ -28,10 +32,20 @@ export class CommentNodes extends Component<CommentNodesProps, any> {
...
+    const maxComments = this.props.maxCommentsShown ?? this.props.nodes.length;
 
-    return (
-      <div className="comments">
+    const borderColor = this.props.depth
+      ? colorList[this.props.depth % colorList.length]
+      : colorList[0];
+
+    return this.props.nodes.length > 0 ? (
+      <ul
```
jsit, line 42: Should this be an ordered list, since it's sorted by some criterion?

SleeplessOne1917: I could make it ordered. Whichever you think would be better for accessibility. However, I don't think children of top level comments get sorted, so that's worth keeping in mind.

fastfinge:
The issue with an ordered list is that it would cause screen readers to read out a number before each comment, and that doesn't seem to make much sense. It matters that the comment is at depth 2, not that it's the fifth comment in this thread at depth 2. Perhaps someone else disagrees, but I'm struggling to think of a reason I'd ever want to innumerate the order of comments. It's the depth that matters here.

src/shared/components/comment/comment-nodes.tsx
nebeker: As @fastfinge  said, ul makes more sense from a screen reader perspective. With an ol, you'd (depending on settings) get the item number and the number of items on the list: the comment number and the total number of comments at that depth, in that thread. That seems like a lot of extra info to process.

dessalines:
The colored lines seem overkill now. Would you just make them gray / muted?

Screenshot_2023-06-05-11-56-19-832_com rubenmayayo reddit

Nutomic:
Colors seem fine but I would make them lighter or translucent.

Kommynct:
An option to click on them to collapse threads would also be good, but then you'd probably want to make them wider.

dessalines:
@Kommynct That's a separate issue. Please keep things on topic, we're juggling many things right now.

nebeker:
While we're on an accessibility vibe, passing level AA here would be excellent for readability:
https://webaim.org/resources/contrastchecker/

AgentElement:
We could have colored vs muted nesting indicators as an option in the settings, leaving them on (colored) by default. I think it's a matter of user preference.

Nutomic:
Adding new options is pretty complicated, not a great idea for something so minor.

Kommynct:
I think colored is better, and if you like uncolored lines, I can easily whip up a stylus theme that colors them however you like.

Leanna-o:
Either way, colored or uncolored is a small difference, and can be addressed later on
 For now I think the priority should be getting any variation of this PR merged, as there is a big influx of users, many of them laypeople, and whether they stay or leave depends partially on how good the UI & UX is, and this is a small but significant improvement to it.

SleeplessOne1917:
@Junebugging  Now that the big HTTP changes are finally merged into main, I can finish this up.

SleeplessOne1917:
The comments look like this now.

image

SleeplessOne1917:
@dessalines @jsit @nebeker If you don't have any more requested changes, please approve.

src/shared/components/comment/comment-nodes.tsx
```diff
@@ -74,8 +78,18 @@ export class CommentNodes extends Component<CommentNodesProps, any> {
   render() {
     const maxComments = this.props.maxCommentsShown ?? this.props.nodes.length;
 
-    return (
-      <div className="comments">
+    const borderColor = this.props.depth
+      ? colorList[this.props.depth % colorList.length]
+      : colorList[0];
+
+    return this.props.nodes.length > 0 ? (
```
dessalines, line 85: Just do &&  instead of : null

SleeplessOne1917: Done.

dessalines (approved):

PR #1073, opened 2023-06-04
Use http client 2 merge unfinished

dessalines:
@SleeplessOne1917 

I've got the http stuff almost all working, the branch is `use_lemmy_http_client_2`. You can pull it down and test it if you like.

But when trying to merge from main, I'm running into a lot of conflicts for code you worked on since I started this massive change. I've left them here, they are navbar.tsx, admin-settings.tsx, and I think one or two other components.

Could you try to fix these conflicts using `mergetool` or something similar? If its too difficult for either of us to handle, then I think we have no choice but to take my changes, and add your additions *again.*.


makotech222:
Following previous PR discussion, i disagree. React class components isn't really set up for that kind of separation, i think the only thing you are accomplishing is breaking encapsulation and requiring consumers of components to know and reimplement the process of getting data for it.

dessalines:
The alternative is spaghetti code, where you are mutating state and doing data operations within deeply nested components. 

It was jarring to me at first as well, but I trusted the way the new ui designers were building things, and it has worked out well.

SleeplessOne1917:
I'm in the process of merging now. I should be able to accept all the changes from main, but I still need to do some tweaking to get everything working right.

SleeplessOne1917:
For some reason a new PR opened when I merged and pushed. See #1081

Issue #1079, opened 2023-06-04
Language select needs label


jsit:
### Issue Summary

The language select dropdown doesn't have a label. This makes a screenreader read it as "Undetermined, action". ("action" due to the incorrect `aria-label`).

It also means that its purpose is visually unclear until you click it and see a list of languages.

<img width="573" alt="Screenshot 2023-06-04 at 4 03 54 PM" src="https://github.com/LemmyNet/lemmy-ui/assets/643417/5a61b6c2-8ac9-4ed0-91cc-aff2a6d0fe58">

### Steps to Reproduce

1. Turn on screen reader
2. Have screen reader read language dropdown
3. Hear "Undetermined, action"

PR #1088, opened 2023-06-05
Fix aria-label for language select

jwhitmarsh:
Fix https://github.com/LemmyNet/lemmy-ui/issues/1079

I've used the existing `language_select_placeholder` as I thought "Select language" is a good fit, but happy to change it simply `language` if preferred.


SleeplessOne1917 (approved):
LGTM

PR #1114, opened 2023-06-08
Fix isoData can contain user generated content

0xDEADCADE:


SleeplessOne1917 (approved):
Good catch!

0xDEADCADE:
My local version of Prettier isn't complaining about this source. I don't usually do JS development, so I don't know what exactly could be wrong here. Does anyone else know, maybe @dessalines?

dessalines (approved):
Thank you!

dessalines:
@0xDEADCADE I'll handle the formatting in one sec.

0xDEADCADE:
I would recommend doing a release with this fix relatively soon, and urging instance owners to update. The potential impact of this is high (see my Matrix DMs, or for anyone else, once this is in release, ask me at [@deadcade:deadca.de](https://matrix.to/#/@deadcade:deadca.de) on Matrix), and with the new wave of users, it would leave a very bad first impression.

PR #1129, opened 2023-06-09
Alternative way to sanitize isoData

Nutomic:
This solution isnt very clean, but it fixes the vuln in my testing and also doesnt break quotes as you can see below.

Screenshot_20230609_125419

Using [this approach](https://stackoverflow.com/a/56596217) as replaceAll is not available.


Nutomic:
Updated on suggestion from @0xDEADCADE. Use full sanitization, then restore only `>` char when rendering markdown. This way is more secure.

dessalines (approved):

PR #1145, opened 2023-06-10
Added additional keybinds

Adam-Shea:
Added additional keybinds following #1136 and based on [Github md editor binds](https://docs.github.com/en/get-started/using-github/keyboard-shortcuts#comments).
I also made sure that when creating a list while text is selected, an additional space is added to make sure it is formatted correctly. This functionality is disabled when text is not selected so as to not change the current experience

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -435,28 +435,28 @@ export class MarkdownTextArea extends Component<
   handleKeyBinds(i: MarkdownTextArea, event: KeyboardEvent) {
     if (event.ctrlKey) {
       switch (event.key) {
-        case "k": // Insert link
+        case "k":
           i.handleInsertLink(i, event);
           break;
-        case "b": // Insert bold
+        case "b":
           i.handleInsertBold(i, event);
           break;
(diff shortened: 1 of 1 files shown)


src/shared/components/common/markdown-textarea.tsx
```diff
@@ -431,13 +431,34 @@ export class MarkdownTextArea extends Component<
...
+        case "e": // Insert code
+          i.handleInsertCode(i, event);
+          break;
+        case "8": // Insert unordered list
+          i.handleInsertList(i, event);
+          break;
+        case "s": // Insert spoiler
+          i.handleInsertSpoiler(i, event);
+          break;
+        case "p": // Move to preview view
```
SleeplessOne1917, line 456: Nitpick: The comments are unnecessary because the function names already make it clear what is going on.

SleeplessOne1917 (approved):
One small nitpick, but otherwise LGTM.

src/shared/components/common/markdown-textarea.tsx
Adam-Shea: Done. I think I was being a bit overkill ha ha

Issue #1048, opened 2023-06-02
Automatically stay on an instance, when given a link to another


Kommynct:
Sometimes when I click on a link to another instance, it takes me to that instance, effectively signing me out, since lemmy is federated, this adds complexity to end users for minimal benefit

An option to automatically make links redirect to that community on your instance, instead of linking to a new instance, this should be toggleable and probably on by default, to decrease the confusion of new users.

dessalines:
I agree, we could benefit from having the `!...` autocomplete link to both the federated, and local url of the community or user being linked.

Deebster:
One problem is that this link can 404, if the local instance hasn't pulled down the community. I briefly looked at doing something automatic with the search page, but the current websockets-based querying means you can't just do `/search?q=!foo@bar`.

I suppose the real fix would be to make the local instance try to fetch the remote instead of 404ing (perhaps using a HTTP 202 instead).

elishamint:
IMO, this should be higher priority (bug), as it currently defies federation and is very confusing.
Also, URIs should be more descriptive, including the community name.

rubythulhu:
perhaps browser extensions could be made that transfer a user to their "home instance" when loading a link to a foreign instance? they could even check for the `local instance hasn't pulled down the community` first, and if not, display the original page but with a warning at the top explaining why it's not on your home instance?

Kommynct:
Why should it be a browser extension? It should really be baked into the UI, I would go so far as to say this is a necessity for new users, because the current system is confusing

rubythulhu:
> Why should it be a browser extension? It should really be baked into the UI, I would go so far as to say this is a necessity for new users, because the current system is confusing

i wasn't suggesting this as a primary solution; i agree that this is an underlying problem with the platform itself. but browser extensions are a pretty good bandaid to the problem in the short term

Kommynct:
I also want to brainstorm if there's a possible solution for external links, currently people are posting links all over reddit to the different instances, and people think you have to sign up for every single one to see content, is there any way to make sign-ins universal across lemmy? I doubt it, but it would be massively helpful for onboarding.

elishamint:
The challenge, in a nutshell:
(1) With a logged-in user, the "local" (home) instance is supposed to serve the linked page. Meaning, the link should point back at the server (self) which served the page containing the link, in the first place.
[...]

PR #1156, opened 2023-06-10
Implement instance agnostic community linking

nnhien:
Replaces the hyperlink component of the markdown link of the Tribute template with an instance agnostic way of linking to different communities.

For example, instead of autofilling !technology@beehaw.org to https://beehaw.org/c/technology, we should autofill it to /c/technology@beehaw.org. That way, we redirect users to the version of the linked community on their home instance. This makes moving around lemmy less jarring. Currently when someone whose home instance isn't beehaw clicks that link, they'll be "logged out" (even though they aren't "logged out"; they've just moved to a different website). This should be implemented because of the fact that all lemmy instances have the same UI, making what should be a pretty painless traversal kind of confusing.

I'm not sure how to spin up a local lemmy instance for testing frontend changes, so if someone could tell me how to spin one up or test this real quick for me, that would be much appreciated.

Also I'm not really acquainted with what convention the maintainers prefer for using existing types, but this seemed like the most straight forward solution. It's kind of inefficient though, since we don't use the CommunityView returned by the network response.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -700,7 +700,7 @@ export function setupTribute() {
         trigger: "!",
         selectTemplate: (item: any) => {
           const it: CommunityTribute = item.original;
-          return `[${it.key}](${it.view.community.actor_id})`;
+          return `[${it.key}](${it.key.replace("!", "/c/")})`;
         },
         values: debounce(async (text: string, cb: any) => {
           cb(await communitySearch(text));
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -700,7 +700,7 @@ export function setupTribute() {
         trigger: "!",
         selectTemplate: (item: any) => {
           const it: CommunityTribute = item.original;
-          return `[${it.key}](${it.key.replace("!", "/c/")})`;
+          return `[${it.key}](${"/c/" + it.key.substring(1)})`;
         },
         values: debounce(async (text: string, cb: any) => {
           cb(await communitySearch(text));
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -700,7 +700,7 @@ export function setupTribute() {
         trigger: "!",
         selectTemplate: (item: any) => {
           const it: CommunityTribute = item.original;
-          return `[${it.key}](${"/c/" + it.key.substring(1)})`;
+          return `${it.key}`;
         },
         values: debounce(async (text: string, cb: any) => {
           cb(await communitySearch(text));
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -1102,7 +1102,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     const words: string[] = []
     content.split(" ").forEach(word => {
       if (word.charAt(0) == "!") {
-        words.push(this.handleCommunityLinking(word))
+        words.push(this.renderCommunityLink(word))
       } else {
         words.push(word)
       }
@@ -1110,7 +1110,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     return words.join(" ")
   }
   
-  handleCommunityLinking(text: string): string {
+  renderCommunityLink(text: string): string {
     return `[${text}](/c/${text.substring(1)})`
   }
 
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -1095,13 +1095,33 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     );
   }
 
+  renderCommentContent(content: string): string {
+    /* 
+      Iterate through all the words in a comment and provide frontend support
+      for implementation-agnostic syntactic sugar standards
+    */
+    const words: string[] = []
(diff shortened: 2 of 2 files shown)


SleeplessOne1917:
I'll test this out locally when I get the chance. Otherwise, looks good.

nnhien:
Hm wait, before this is pulled there I realized there might be a slight issue with the implementation.

While Lemmy UI uses the `/c/` convention for denoting communities, KBin uses `/m/`. I'm still not super familiar with how ActivityPub works, but if we just push the raw markdown and it gets pulled, this will cause interoperability issues in regards to linking. KBin does use the same `[community_name]@[instance_domain]` format after the `/m/`, but I'm currently not quite sure what the best way forward now would be.

Since I'm not super familiar with how the standard is supposed to work, I think it's probably best to defer to the maintainers.

nnhien:
After some thought, I have a proposal

We can formally (unilaterally) establish a "Lemmy Community Linking Standard (tm)" with the format `![COMMUNITY_NAME]@[INSTANCE_DOMAIN]`. We don't manually create a hyperlink reference in the comment content itself; the comment content simply has the above format. When a federated application's frontend encounters the above format and they have good support for Lemmy posts, they can convert the community link into a hyperlink which complies with how that application handles federated content.

What are people's thoughts

novakeith:
> After some thought, I have a proposal
> 
> 
> 
> We can formally (and unilaterally) establish a "Lemmy Community Linking Standard (tm)" with the format `![COMMUNITY_NAME]@[INSTANCE_DOMAIN]
> 
> What are people's thoughts

Two thoughts: 

1. I did not see anything limiting how entities should be addressed in the ActivityPub standard. Hence why the devs probably chose to address Lemmy communities like this in the first place. If other softwares want to implement that standard, that would be great, and I think it should be encouraged for the reason outlined in my next point...

2. Kbin allows for the same community name as a user name, so you can end up with something like: 
- novakeith@kbin.tld (username)
- novakeith@kbin.tld (community / "magazine" name)

Both are considered valid endpoints in kbin and there is no way to distinguish between the two. I think using the ! to distinguish community names is important, but its adoption outside of Lemmy is up to other devs.

_Full disclosure. I am not a real software dev, I just like this project. Defer to dessalines and nutomic._

nnhien:
@dessalines @SleeplessOne1917 

Sorry for ping, but thoughts on this most recent kind of implementation?

src/shared/utils.ts
```diff
@@ -700,7 +700,7 @@ export function setupTribute() {
         trigger: "!",
         selectTemplate: (item: any) => {
           const it: CommunityTribute = item.original;
-          return `[${it.key}](${it.view.community.actor_id})`;
+          return `${it.key}`;
```
dessalines, line 703: I'm not a fan of this approach.

Instead, just have two links, the fedi one, and whatever you local instance is with the community.name.

IE something like `[it.key](community.actor_id),[^local^](this_instance/c/@{community.name}`

nnhien: My worry with having two links is that it doesn't make the linking experience transparent. In my opinion, the user shouldn't need to know about the nuances of instances and how they interact in order to consume content. They should just be able to click a link and get sent to the version of the post on their home instance; i.e. they should be able to get a single link and it should "just work".

Also, having the separate "local" link breaks interoperability with other fediverse services. The motivation behind my approach is that when a user posts they just put text in a particular format, and when that post gets fetched by other services (like Kbin), the other service can handle (or not handle) the formatted text in whichever manner they like.

nnhien: Hmm, would it be a good idea to implement a companion solution to that issue then? That is, should we make it so that if you navigate to a post on a remote instance which hasn't been federated with your home instance yet, the home instance will automatically fetch the content from the remote instance instead of 404'ing? Or was there an explicit design choice to 404 instead of fetching?

dessalines: There aren't any 404s currently, because the post and comment nodes already exist on your own instance, or you wouldn't be seeing them (so neither link will 404).

But with searched content like this, the comment with text referencing a federated community or actor, might not exist everywhere. 

> the home instance will automatically fetch the content from the remote instance instead of 404'ing?

We don't currently do this, but this is the advantage of having a local link. We could do a [resolveObject](https://join-lemmy.org/api/classes/LemmyHttp.html#resolveObject) if it 404s, and cache it to our local instance. No guarantees still, since it could be blocked, but at least its possible.

We've chosen for resolve to only work on the search page currently, and by logged-in users only, but I'm open to a discussion for changing that. Def has pros and cons.

nnhien: Another solution if we want to keep the explicit hyperlinks to different instances could be to implement a "You are leaving your home instance" screen before redirection. That way users are explicitly told that they are moving to a physically different website and they don't have to rely on vague UI clues to realize that.

Nutomic:
This is a fundamental change which will affect lots of users. It would be better to discuss this more in an issue or directly on Lemmy before implementing any specific change.

nnhien:
sgtm, found an already existing issue (#1048) so I guess discussion should be redirected there

rszyma:
> sgtm, found an already existing issue (#1048) so I guess discussion should be redirected there

While this PR is related to #1048 it doesn't completely solve the original issue that started #1048.

dessalines:
This is superceded now by #1462

PR #1158, opened 2023-06-10
1117 jump to main content

### LINKED ISSUE #1117: Accesibility: missing "jump to content" link
### Issue Summary

There's no "skip to content" link allowing users with screenreaders to skip over all the headers.

### Steps to Reproduce

Go to the main page of any lemmy site.  Click tab.  It should take you to a "jump to content" link, but none appears.

The screenshot of Reddit's interface illustrates what should happen. The "jump to content" link in the image is initally invisible, but appears when I hit a tab.

Screen Shot 2023-06-08 at 11 37 17 AM

### Technical details

https://www.useragentman.com/blog/2016/12/15/implementing-an-accessible-skip-navigation-link-requires-more-thought-than-youd-think/  is a good discussion of how to provide this functionality (although a little old, not sure if best practices have evolved since then)

skspade:
# Describe 
- Adds ability to press the "Tab" key to reveal a hidden "Jump to conent" link. Clicking or hitting enter will bring the post listings regardless of page into focus. 
- Prevents `validURL` from throwing exception into the browser console.

# Video 

https://github.com/LemmyNet/lemmy-ui/assets/43147936/203ce936-8070-4ea3-8100-0be18233f3ad

# Issue 
closes #1117

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -420,3 +420,18 @@ br.big {
 em-emoji-picker {
   width: 100%;
 }
+
+.skip-link {
+  top: -40px;
+  transition: top 0.3s ease;
+}
+
+@media (prefers-reduced-motion: reduce) {
+  .skip-link {
+    transition: none;
+  }
+}
+
+.skip-link:focus {
+  top: 0;
+}
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -327,7 +327,12 @@ export function isVideo(url: string) {
 }
 
 export function validURL(str: string) {
-  return !!new URL(str);
+  try {
+    new URL(str);
+    return true;
+  } catch (_) {
+    return false;
+  }
 }
 
 export function validInstanceTLD(str: string) {
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -330,7 +330,7 @@ export function validURL(str: string) {
   try {
     new URL(str);
     return true;
-  } catch (_) {
+  } catch {
     return false;
   }
 }
--- a/src/shared/components/app/app.tsx
+++ b/src/shared/components/app/app.tsx
@@ -1,4 +1,4 @@
-import { Component } from "inferno";
+import { Component, createRef, linkEvent, RefObject } from "inferno";
 import { Provider } from "inferno-i18next-dess";
 import { Route, Switch } from "inferno-router";
 import { i18n } from "../../i18next";
@@ -15,8 +15,15 @@ import { Theme } from "./theme";
 
 export class App extends Component<any, any> {
   private isoData: IsoDataOptionalSite = setIsoData(this.context);
+  private readonly mainContentRef: RefObject<HTMLElement>;
   constructor(props: any, context: any) {
     super(props, context);
+    this.mainContentRef = createRef();
+  }
+
+  handleJumpToContent(event) {
+    event.preventDefault();
+    this.mainContentRef.current?.focus();
   }
   render() {
     const siteRes = this.isoData.site_res;
@@ -26,6 +33,12 @@ export class App extends Component<any, any> {
       <>
         <Provider i18next={i18n}>
           <div id="app">
+            <a
+              className="skip-link bg-light text-dark p-2 text-decoration-none position-absolute start-0 z-3"
+              onClick={linkEvent(this, this.handleJumpToContent)}
+            >
+              ${i18n.t("jump_to_content", "Jump to content")}
+            </a>
             {siteView && (
               <Theme defaultTheme={siteView.local_site.default_theme} />
             )}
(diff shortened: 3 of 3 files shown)


src/assets/css/main.css
```diff
@@ -430,3 +430,19 @@ br.big {
 em-emoji-picker {
   width: 100%;
 }
+
+.skip-link {
+  position: absolute;
+  top: -40px;
+  left: 0;
+  background: #000;
```
SleeplessOne1917, line 438: Could you please use bootstrap class names for styling where possible? You could give the skip link the following classes: "bg-dark text-light p-2 text-decoration-none position-absolute start-0 z-3" The only properties you'll still need defined on this css class are top and transition.

src/server/index.tsx
```diff
@@ -393,6 +393,13 @@ async function createSsrHtml(root: string, isoData: IsoDataOptionalSite) {
   </head>
 
   <body ${helmet.bodyAttributes.toString()}>
+    <a
+    href="#main-content"
+    class="skip-link"
+    onclick="document.getElementById('main-content').focus(); return false;"
```
SleeplessOne1917, line 399: You don't have to make this function a string. You could make it:

```tsx
<a onClick={() => {
    document.getElementById('main-content')?.focus(); 
}}
>
    Jump to content
</a>
```

src/server/index.tsx
```diff
@@ -393,6 +393,13 @@ async function createSsrHtml(root: string, isoData: IsoDataOptionalSite) {
...
+    >
+      Jump to content
```
SleeplessOne1917, line 401: This is going to need a translation. [Please add one to the translations repo](https://github.com/LemmyNet/lemmy-translations).

SleeplessOne1917 (changes requested):
Please address the changes I requested in the other comments. In addition to that, this jump to content link should be able to work in more places than just the post listings. As it is now, this will only allow users to jump to content on their home screen or a community screen.

skspade:
Great! I'll get to work on your suggestions tonight. Thanks for the review

src/server/index.tsx
skspade: Are you able to use JSX syntax here? This is in the `createSsrHtml` function which does not seem to be JSX but a HTML in a template string.

SleeplessOne1917: That is a good point. I didn't realize this wasn't JSX in the small window github gave me.

skspade:
> This is going to need a translation. [Please add one to the translations repo](https://github.com/LemmyNet/lemmy-translations). 

https://github.com/LemmyNet/lemmy-translations/pull/50

src/assets/css/main.css
skspade: The bootstrap color classes are confusing me a bit. For example `bg-dark` is giving me a light color background. `text-light` is giving me dark text. The opposite is true too. Look at my screenshots below.
Screenshot from 2023-06-11 13-11-04
Screenshot from 2023-06-11 13-11-29

skspade:
I re-implemented the feature, this time I wrapped all routes at the entry point. I feel like this is effectively skips the navigation elements on every page and in the future other developers do not have to worry about wrapping new pages with an element containing the attribute `main-content`

dessalines (commented):
Externalize that one to a handler function. The convention is to have any onClick actions use linkEvent with a function.

skspade:
Sounds good, I can move it into a JSX portion of the code. I put it here since it needs to be the first focusable element on any page. But I agree consistency is important. I'll get to work on this in the evening

skspade:
Updated Video showing the functionality is the same when written in JSX
https://github.com/LemmyNet/lemmy-ui/assets/43147936/62b5f0eb-6a22-49f6-bd3c-30e718bf21f4

skspade:
https://github.com/LemmyNet/lemmy-ui/assets/43147936/f0991bfa-63d5-4ed7-9f1a-c648e525f2f2

skspade:
I see that there is conflicts now, I'll fix those tonight

SleeplessOne1917:
@skspade The only conflict is the translations submodule. Try running `git submodule update --remote --recursive` and committing that.

src/shared/utils.ts
```diff
@@ -329,7 +329,12 @@ export function isVideo(url: string) {
 }
 
 export function validURL(str: string) {
-  return !!new URL(str);
+  try {
+    new URL(str);
+    return true;
+  } catch (_) {
```
ayan4m1, line 335: The discard parameter can be omitted entirely here

```jsx
try {
  new URL(str);
  return true;
} catch {
  return false;
}
```

skspade: Good point! I'll make that change too

src/assets/css/main.css
```diff
@@ -434,3 +434,12 @@ br.big {
 em-emoji-picker {
   width: 100%;
 }
+
+.skip-link {
+  top: -40px;
+  transition: top 0.3s ease;
```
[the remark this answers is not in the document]
skspade: True! Since we do not have a user preference for this, I think Its best to just remove the transition.

skspade:
@SleeplessOne1917  I believe I got it to a good place. Let me know if you would like me to squash my commits before we merge (if you approve)

src/assets/css/main.css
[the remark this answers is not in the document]
ayan4m1: Oh, I meant using a media query: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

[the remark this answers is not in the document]
skspade: Hey thanks for that! Never knew you could query for that! 
I took up this PR to learn more about accessibility so thanks for spreading some knowledge

SleeplessOne1917:
@skspade  Squashing is fine.

SleeplessOne1917 (approved):

skspade:
@SleeplessOne1917 Thanks for the approval! Looks like woodpecker is stalling out after I squashed.  Rebased commit: ` 36033b7` passed, then I squashed to ` d3e1812 `  which doesn't seem to be detected.

PR #1170, opened 2023-06-11
Specify package manager version in package.json

MageJohn:
This setting is picked up by Corepack, and it's a way for others to concretely know which package manager they should use.


SleeplessOne1917 (approved):
LGTM

PR #1173, opened 2023-06-11
Left-align comment collapse/expand button

alectrocute:
Hi all!

Here's a quick PR to left-align the collapse/expand button on comment nodes. 

This is a pain point especially noticeable on mobile, when collapsing comments –different username lengths lead to the button being at different locations– forcing the user to hunt for the collapse button. It's very fatiguing. Having it left-aligned keeps it in the same place regardless of the other dynamic element widths.

Screenshot 2023-06-11 at 9 53 30 AM

Thanks so much!

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -240,11 +240,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 aria-label={this.expandText}
                 data-tippy-content={this.expandText}
               >
-                {this.state.collapsed ? (
-                  <Icon icon="plus-square" classes="icon-inline" />
-                ) : (
-                  <Icon icon="minus-square" classes="icon-inline" />
-                )}
+                <Icon icon={`${this.state.collapsed ? "plus" : "minus"}-square`} classes="icon-inline" />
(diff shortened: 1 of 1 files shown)


src/shared/components/comment/comment-node.tsx
```diff
@@ -234,6 +234,18 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
             })}
           >
             <div className="d-flex flex-wrap align-items-center text-muted small">
+              <button
+                className="btn btn-sm text-muted mr-2"
+                onClick={linkEvent(this, this.handleCommentCollapse)}
+                aria-label={this.expandText}
+                data-tippy-content={this.expandText}
+              >
+                {this.state.collapsed ? (
```
SleeplessOne1917, line 243: While it was like this before your PR, it could be nice to rewrite lines 243-247 as:
```jsx
<Icon icon={`${this.state.collapsed ? "plus" : "minus"}-square`} classes="icon-inline" />
```

alectrocute: Ah, yes. Good suggestion, I'll push the change.

SleeplessOne1917 (approved):

PR #1176, opened 2023-06-11
Set html lang attr to the interface language

### LINKED ISSUE #1168: [Accessibility] global HTML lang attribute must match Lemmy's UI language, be overridden on a post basis

### Issue Summary

If Lemmy is displayed in German, because the user visits from a German-speaking country, and is blind and uses a screen reader, the German words are still read with an English language synthesizer, because the global lang or xml-lang attribute is set to en-us regardless of the actual UI language.

### Steps to Reproduce

1. Set Lemmy to a UI language other than English.
2. Using the web developer tools of your choice, inspect the root HTML element and look for the lang attribute.
   * Expected: The value of the lang attribute is "de" for German, "fr" for French, etc.
   * Actual: The lang attribute's value is still en-us.

### Technical details

N/A, the setting of the attribute simply must respect the selected UI language and put in the according value accordingly. Of course, once a post has a different language, the container displaying the post should override the lang attribute with that of the post's language, so screen readers read the post in the correct language.

Note that German is just my local example, this will most likely be true fror French, Spanish, Ukrainian, Chinese, or what not as well.

MageJohn:
Fixes #1168

As there are issues with running the main branch with the current backend repo, I tested this by cherry-picking to `use_http_client_2`. 

My change fixes the linked issue by setting the `lang` attribute of the `html` tag properly in accordance to the lemmy user preferences or user-agent preferences. If the lemmy user has a UI language other than "Browser" set, then the response from the server will already have the correct `lang` set. Otherwise the attr is set on the client side according to the language settings in the browser. If any of this fails, it falls back to English.

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -346,7 +346,7 @@ async function createSsrHtml(root: string, isoData: IsoDataOptionalSite) {
 
   return `
   <!DOCTYPE html>
-  <html ${helmet.htmlAttributes.toString()} lang="en">
+  <html ${helmet.htmlAttributes.toString()}>
   <head>
   <script>window.isoData = ${sanitize(JSON.stringify(isoData))}</script>
   <script>window.lemmyConfig = ${serialize(config)}</script>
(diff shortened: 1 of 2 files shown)


SleeplessOne1917 (approved):

MageJohn:
Just rebased onto main so that the CI would pass.

PR #1179, opened 2023-06-11
Footer: Use `target="_blank"` for all external links, refactor Github-related links

alectrocute:
Hi, Lemmy devs!

Just a small UX tweak for the site footer – external links should open in a new tab. Also added pertinent links to the actual repos to hopefully get some more eyeballs on the OSS project. 👀

- Use `target="_blank"` for all external links
- Refactor Github-related links
- Format code to be consistent with other elements

Thanks.

### DIFF
--- a/src/shared/components/app/footer.tsx
+++ b/src/shared/components/app/footer.tsx
@@ -21,13 +21,13 @@ export class Footer extends Component<FooterProps, any> {
           <ul className="navbar-nav ml-auto">
             {this.props.site?.version !== VERSION && (
               <li className="nav-item">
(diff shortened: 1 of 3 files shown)


SleeplessOne1917:
@dessalines I recall seeing you mention in a Lemmy post that having links automatically open in a new tab instead of leaving it up to the user is bad for accessibility. Is this PR something that should be merged?

alectrocute:
Got it, @dessalines. I'll make sure to align with these principles in future contributions. 👍 Thanks, friends!

Issue #1169, opened 2023-06-11
Add drop shadow or in some other way distinguish voted from non voted upvote and downvote icons


davorb:
### Is your proposal related to a problem?
The colour contrast between upvoted and not upvoted (and downvoted) icons is 
too low. It is very hard to see if it's blue or gray.

### Describe the solution you'd like
Either add an outline or a shadow or change the colour.

fheft:
Maybe we could simply use the _primary color_ (orange in light theme)? The alternative would be using our _secondary color_, which is be green – that might be interpreted as 'positive' which contradicts downvotes.

What's your opinion?

![vote-color](https://github.com/LemmyNet/lemmy-ui/assets/787786/82e3a3f7-e85e-4006-a233-7b195489f8a0)

davorb:
Yes, I think this would be a good solution to this problem. 

If you guys are open to accepting it, I could even do the change myself and open a PR.

Issue #703, opened 2022-06-28
Bootstrap theme generation from scss / upgrade to bootstrap v5


dessalines:
It'd be really nice to use bootswatch scss files imported from their node repository, to generate the full css themes that lemmy-ui can then use. Then we could:
- Remove a lot of the css files from this repo.
- Make sure lemmy-ui works with the latest bootstrap versions.

PR #1191, opened 2023-06-12
Swap upvote/downvote arrow icon colors

alectrocute:
Hi all!

Quick PR here to swap upvote/downvote colors. Currently we have `text-info` for upvote and `text-danger` for downvote, when most of us are expecting quite the opposite. I propose we swap these per r*ddit conventions.

Screenshot 2023-06-11 at 11 34 30 PM

Screenshot 2023-06-11 at 11 34 09 PM

Thanks! 🙌

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -380,7 +380,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                     <>
                       <button
                         className={`btn btn-link btn-animate ${
-                          this.state.my_vote === 1 ? "text-info" : "text-muted"
+                          this.state.my_vote === 1 ? "text-danger" : "text-muted"
                         }`}
                         onClick={this.handleCommentUpvote}
                         data-tippy-content={i18n.t("upvote")}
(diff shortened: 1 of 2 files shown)


alectrocute:
I have no idea what's going on with the Prettier step on CI 🤔

Nutomic:
Lemmy is not Reddit, we dont need to follow everything that they do.

krestenlaust:
> Lemmy is not Reddit, we dont need to follow everything that they do.

This as well, but we could consider changing the colors or shade of the colors, because right now the colors are basically the same as default reddit but opposite.

This should probably be discussed further in an issue instead of a PR

SleeplessOne1917:
#1169 seems related.

alectrocute:
My thinking was/is, if we're using the same colors, we should at least use them in the same way. I'd be happy to turn this PR into one that changes these colors from orangered/blue – How should we proceed?

dessalines:
I don't understand why text-info and text-danger are inadequate, or need to be swapped. Better to just define a new theme with better colors. There's a [theming guide here](https://join-lemmy.org/docs/en/administration/theming.html?highlight=themes#theming-guide)

Or a better longer-term solution, would be #703 , IE to upgrade lemmy-ui to bootstrap v5, use its auto-generation from `scss` files, and set  colors there.

alectrocute:
I like #703. Closing.

PR #1208, opened 2023-06-12
Another try at sanitize.

dessalines:
I'll be testing this shortly.


SleeplessOne1917 (approved):

dessalines:
The solution is from [here](https://stackoverflow.com/questions/63926663/how-should-untrusted-json-be-sanitized-before-using-json-parse)

I apologize for not having a proper test setup, reddit caught us right in the middle of a large websocket -> http rework. 

I have it deployed to https://voyager.lemmy.ml : `v0.18.0-beta.4-3-g2653d65` , for testing / trying to break it.

SleeplessOne1917 (approved):

dessalines:
Adding another method, since we've verified that one also doesn't work.

PR #1241, opened 2023-06-13
Match more specific locales to supported ones

### LINKED ISSUE #1205: Lemmy is showing in a different language than my browser's and OS's

### Issue Summary

Hello, for all Lemmy instances I visit, Lemmy loads in Portuguese. While I do speak Portuguese, it is not my operating system's language neither my browser's, and I cannot find a way to change it back to English. 

Here's my Chrome's language settings:

image

Here's Windows 11's language settings:

image

And here's how I see Behaw for example:

image

I imagine playing around with the priorities would help as maybe setting a less-specific variation of english as the top language.

MageJohn:
Fixes #1205

I noticed that the linked issue could be solved by the `nonExplicitSupportedLngs` setting for i18next (see [here](https://www.i18next.com/overview/configuration-options#languages-namespaces-resources)), but the existing system for choosing the language bypassed the i18next code where that setting applied. Therefore I moved the code for getting languages into a simple `languageDetector` plugin for i18next.

For reference on these plugins, I used [the docs](https://www.i18next.com/misc/creating-own-plugins#languagedetector) and the source of the package [i18next-browser-languagedetector](https://github.com/i18next/i18next-browser-languageDetector).

I was then able to simplify some other code that references the current language by referring to the value tracked by i18next.

I've tested the code locally to confirm that the issue is resolved. I tested using a browser language order of `["en-GB", "nl", "en-US", "en"]`. The behaviour without my changes is to show the site in Dutch (it skipped `"en-GB"` and found `"nl"`). The behaviour with my changes is to show the site in English (it recognises that `"en"` strings are suitable for the `"en-GB"` locale).

### DIFF
--- a/src/shared/components/common/html-tags.tsx
+++ b/src/shared/components/common/html-tags.tsx
@@ -2,7 +2,8 @@ import { htmlToText } from "html-to-text";
 import { Component } from "inferno";
 import { Helmet } from "inferno-helmet";
 import { httpExternalPath } from "../../env";
-import { getLanguages, md } from "../../utils";
+import { i18n } from "../../i18next";
+import { md } from "../../utils";
 
 interface HtmlTagsProps {
   title: string;
@@ -17,11 +18,10 @@ export class HtmlTags extends Component<HtmlTagsProps, any> {
     const url = httpExternalPath(this.props.path);
     const desc = this.props.description;
     const image = this.props.image;
-    const lang = getLanguages()[0];
 
     return (
       <Helmet title={this.props.title}>
-        <html lang={lang == "browser" ? "en" : lang} />
+        <html lang={i18n.resolvedLanguage} />
 
         {["title", "og:title", "twitter:title"].map(t => (
           <meta key={t} property={t} content={this.props.title} />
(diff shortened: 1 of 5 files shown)


src/shared/components/common/html-tags.tsx
```diff
@@ -2,7 +2,8 @@ import { htmlToText } from "html-to-text";
 import { Component } from "inferno";
 import { Helmet } from "inferno-helmet";
 import { httpExternalPath } from "../../env";
-import { getLanguages, md } from "../../utils";
+import { i18n } from "../../i18next";
+import { md } from "../../utils";
```
SleeplessOne1917, line 6: This is newly imported, but I don't see it being used in your changes.

MageJohn: `md` isn't newly imported, it was there before. `i18n` is newly imported, and it's used on line 24.

SleeplessOne1917: Gotcha. The git changes made it look new.

MageJohn:
Rebased onto latest develop

dessalines (approved):
Looks good, thx!

SleeplessOne1917 (approved):

PR #1246, opened 2023-06-13
Generate theme css

fheft:
This adds a formal build process for the existing theme files, which should make it easier to update the themes in the future.

Themes can be built using two new npm scripts: `themes:build` and `themes:watch`. The resulting CSS files still have to be checked into git, as compilation of multiple external SASS files is a bit hard to integrate in the current build pipeline (and I want to keep this PR as non-invasive as possible). Any ideas are welcome though :slightly_smiling_face: 

There are quite a lot of non-functional changes in the compiled theme files. As far as I've seen, those are all just varying selector order or different formatting of colors (`#fff` vs `#ffffff`). Let me know if there are any manual edits or hacks that I missed.

~~[Edit] I just noticed that the `-red` themes have a few color differences (e.g. for hover states) that are hard to reproduce using the Bootstrap variable overrides. I'll have to look into it.~~ –– fixed.

I also think this is a good step towards #703, even though it still uses Bootstrap 4. Switching to Bootstrap 5 is a bigger change, as there are quite a few breaking changes that lead to a lot of changes in our tsx templates...


fheft:
Rebased onto current main

SleeplessOne1917:
@ayan4m1 I wasn't around when that decision was made, but @dessalines or @Nutomic might be able to tell you more.

fheft:
Fixed the wrong colors in `litely-red`. Now all colors should be the same as before this branch or at least match visually. 

There are some very minor deviations that previously seem to be added by manually changing the compiled CSS files. But I think for future changes it's better to rely on the existing Bootstrap color transformations instead of manually defining every single color just to be _exactly_ the same as before.

@ayan4m1 Can we use `sass-loader` to build separate CSS files for each theme during build time? I think it's beneficial to keep having separate theme CSS files instead of bundling them all together into one big `styles.css`, e.g. only loading what's actually used (smaller network footprint).

alectrocute:
This is a big one for us, IMO. Thanks for your work here, @fheft, let's keep it going. 👍

package.json
```diff
@@ -94,7 +96,7 @@
     "@types/toastify-js": "^1.11.1",
     "@typescript-eslint/eslint-plugin": "^5.59.5",
     "@typescript-eslint/parser": "^5.59.5",
-    "bootswatch": "^5.2.3",
+    "bootstrap-v4": "npm:bootstrap@^4.6.2",
```
SleeplessOne1917, line 99: The version of bootstrap listed in the prod dependencies is 5.2.3. Should we be using v4 here?

fheft: Yeah, that's a bit weird. Version 5 seems to be introduced in 15eb8326 as part of regular dependency updates. But they are not compatible and there's quite a few breaking changes (I tried compiling the themes with BS5 as part of this MR and the design was completely broken...).

But we'll have to check if any of our newer code depends on BS5 features. We use two JS components (collapse and dropdown). I think I'd rather get rid of the BS5 dependency in a separate pull request to avoid making this one even more confusing.

SleeplessOne1917 (approved):

Zetaphor:
Great work @fheft!

PR #1250, opened 2023-06-14
Add checkboxes to issue template

### LINKED ISSUE #1214: Add checkboxes to issue template

> The main thing I want to keep from those, is the checkboxes to make sure people have searched for their issue before. We have so many duplicates on all our repos, and it'll likely help.

Per: https://github.com/LemmyNet/lemmy-ui/pull/1211#pullrequestreview-1475594666

Add the checkboxes from here: https://github.com/LemmyNet/lemmy/blob/main/.github/ISSUE_TEMPLATE/BUG_REPORT.md

Docs: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#checkboxes

n/a

### Additional context

_No response_

RayBB:
Closes #1214

These are also added to https://github.com/LemmyNet/lemmy/pull/3086

You can text it here:
https://github.com/RayBB/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml&title=%5BBug%5D%3A+

Looks like this now:
image


.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -9,6 +9,19 @@ body:
...
         Thanks for taking the time to fill out this bug report!
         For backend issues, use [lemmy](https://github.com/LemmyNet/lemmy)
+  - type: checkboxes
+    attributes:
+      label: Requirements
+      description: Before you create a bug report please do the following.
+      options:
+        - label: Did you check to see if this issue already exists?
+          required: true
+        - label: Is this only a single bug. Do not put multiple bugs in one issue.
```
fheft, line 19: Question mark after the first sentence

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -9,6 +9,19 @@ body:
...
+          required: true
+        - label: Is this a question or discussion? Don't use this, use https://lemmy.ml/c/lemmy_support
```
fheft, line 21: The logic is reversed here compared to the first two checks. For me now it reads as if I had to confirm that _"this is a question or discussion"_.

I think we should either formulate it the same way as the first check, like "Is this a bug report or feature request? For questions or discussions, use ..."

Or we could get rid of the question style all along:
* This issue does not already exists.
* This contains only a single bug. Do not ...
* This is a bug report or a feature request. For ...
* This is a client-side issue (related to the UI in the browser). For backend issues ...

fheft (commented):
Just two suggestions from my side. I'm not a maintainer, so feel free to ignore :)

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
RayBB: Fixed.
PS in the future recommend using suggested changes. https://haacked.com/archive/2019/06/03/suggested-changes/

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
RayBB: I agree and thought about it a bit but decided not to change it too much to honor the spirit of just adding the boxes. However, I'll take the liberty to rework is to be like your first suggestion (rather than change them all).

Thanks :)

SleeplessOne1917 (approved):

PR #1251, opened 2023-06-14
Fixed NSFW image blur spilling outside the preview

### LINKED ISSUE #1245: Nsfw image thumbnail blur-area unintuitively large

The clickable area of blurred nsfw thumbnails is smaller than the visible blurred image. This can lead to not clicking the clickable area by mistake.

Expand the clickable area or limit the blurred area to the clickable area. Anyway the two areas should be the same size.

none

### Additional context

_No response_

fheft:
* Hide overflow of wrapping picture element to contain the blur of the inner img element.
* This aligns the visible image area to the clickable area.
* fixes #1245

Before:
before

After:
after


src/assets/css/main.css
```diff
@@ -434,3 +434,6 @@ br.big {
 em-emoji-picker {
   width: 100%;
 }
+.overflow-hidden {
```
SleeplessOne1917, line 437: I think there's already a class for this in bootstrap.

SleeplessOne1917 (changes requested):
1 small change I made in the other comment I made, but then you're good.

src/assets/css/main.css
fheft: You're absolutely right, I somehow missed that. Removed the CSS rule.

dessalines (approved):

SleeplessOne1917 (approved):

Issue #1148, opened 2023-06-10
Add long-polling for `GetUnreadCounts`


dessalines:
After the http merge is complete, we should add long-polling to refetch unread counts.

SleeplessOne1917:
Will this require a change on the backend as well?

sunaurus:
Most likely you've already thought of this, but I would like to highlight that it's probably quite important to do any polling only in active tabs, for example using https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState

dessalines:
Its on the navbar component, which has the unread counts.

sunaurus:
@dessalines Many users will have tens (in some cases even hundreds) of tabs open, so it could be extremely useful to either A) completely pause polling or B) just skip invoking the `promiseFn` in your `poll` function whenever `document.visibilityState != "visible"`

sunaurus:
If an instance has an average of 1000 users who leave 100 tabs open in the background, then not pausing background polling will generate 288000000 extra unnecessarsy http requests in a day (1000 users * 100 tabs * 2880 requests in a tab per 24h)

PR #1271, opened 2023-06-14
Add long polling to update unread counts in navbar.

dessalines:
Currently have this set to 30 seconds, either that or 1m seems good.


src/shared/utils.ts
```diff
@@ -1490,3 +1491,14 @@ export function newVote(voteType: VoteType, myVote?: number): number {
...
+
+function sleep<T>(millis: number): Promise<T> {
+  return new Promise(resolve => setTimeout(resolve, millis));
+}
+
+/**
+ * Polls / repeatedly runs a promise, every X milliseconds
+ */
+export function poll(promiseFn: any, millis: number) {
+  promiseFn().then(sleep(millis).then(() => poll(promiseFn, millis)));
```
SleeplessOne1917, line 1503: I think this would be easier to read with awaits.

SleeplessOne1917:
I saw unread messages, reports, and applications getting polling. Is there one needed for replies as well?

src/shared/utils.ts
dessalines: K fixed.

dessalines:
unreadInboxCount includes unread replies, messages, and mentions.

sunaurus:
Sorry, I realized that I was commenting on #1148 when I meant to comment here 🤦 moving my comment:

Many users will have tens (in some cases even hundreds) of tabs open, so it could be extremely useful to either A) completely pause polling or B) just skip invoking the promiseFn in your poll function whenever document.visibilityState != "visible"

For example with the 30 second polling:

If an instance has an average of 1000 users who leave 100 tabs open in the background, then not pausing background polling will generate 288 000 000 extra unnecessary http requests in a day (1000 users * 100 tabs * 2880 requests in a tab per 24h)

SleeplessOne1917:
@sunaurus I recall you linked a browser API that can be used to avoid this problem in the issue comments.

sunaurus:
Yep, it's here: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState

src/shared/utils.ts
```diff
@@ -1490,3 +1491,16 @@ export function newVote(voteType: VoteType, myVote?: number): number {
...
+
+function sleep<T>(millis: number): Promise<T> {
+  return new Promise(resolve => setTimeout(resolve, millis));
+}
+
+/**
+ * Polls / repeatedly runs a promise, every X milliseconds
+ */
+export async function poll(promiseFn: any, millis: number) {
+  await promiseFn();
```
sunaurus, line 1503: In theory, this should be enough, but I'm not able to test it myself at the moment:

```suggestion
  if (window.document.visibilityState !== "hidden") {
    await promiseFn();
  }
```

dessalines: Ohh I see so this prevents it if you're looking at other tabs?

sunaurus: Yup, exactly

dessalines: K sweet, I'll test that shortly.

src/shared/utils.ts
```diff
@@ -1490,3 +1491,16 @@ export function newVote(voteType: VoteType, myVote?: number): number {
     return myVote == -1 ? 0 : -1;
   }
 }
+
+function sleep<T>(millis: number): Promise<T> {
```
SleeplessOne1917, line 1495: The generic isn't needed here. You could get away just fine returning `Promise<void>`.

dessalines: Sry, getting to this now.

SleeplessOne1917 (approved):

Issue #1281, opened 2023-06-14
Align comment-text correctly when image is included in comment


Junebugging:
### Is your proposal related to a problem?

To demonstrate with a picture (which surprisingly is not from the marxist instance):

<img src="https://github.com/LemmyNet/lemmy-ui/assets/136490143/6539d731-0d25-4f8f-bea1-ab400d2ebc97" height="600px" />

The first image is posted with text in the comment, the last one isn't so is formatted normally
At first glance it visually looks like an abrupt jump in comment depth, as the contents of the comment appear in inconsistent places.

This seems to be a consequence of \<img> objects being placed inside \<p> objects

<img src="https://github.com/LemmyNet/lemmy-ui/assets/136490143/90793177-6964-4b9f-9f85-ceb40a125979" height="400px" />

### Describe the solution you'd like.

* Align text to the start or top somehow, so the text is placed vertically consistently
* Show text after attached images, so the image is placed horizontally consistently.

This improves the flow of reading, as you can expect the contents of the message to be placed immediately underneath the username.

Here's a quick-fix I did by making the inner-comment <p> block explicitly flex, flex-direction to column, and aligning items to start *(default is "stretch", which stretches images)*. 
![image](https://github.com/LemmyNet/lemmy-ui/assets/136490143/d6619131-6cd8-4b48-9319-33f0ff82aa12)

A rule to implement the above:
```css
.comment-node > div > div > .md-div p {
    display:          flex;
    align-items:    start;
    flex-direction: column;
}
```
Possible issues: If the comments are styled using another display-type than flex, this could mess with that (though that doesn't seem to be the case, parent objects also use flex). The CSS rule suggested above is also quite general, but it's as generic but specific as I could make it to apply to all comments.

### Describe alternatives you've considered.

Parsing the users comment input to separate \<img>'s from \<p>'s would make it easier to format them, but that requires additional parsing and additional potential surfaces for bugs.

Junebugging:
This doesn't seem to be an issue in some cases, because the \<img> is in a separate \<p>:
![image](https://github.com/LemmyNet/lemmy-ui/assets/136490143/e62c96f7-332c-47d8-a4cc-2c163cb6a388)
I assume this is the case when entering the image on a new paragraph

dessalines:
Correct, its dependent on image size, and the fact that you inlined it. You'd need to open an issue about this with markdown-it, our markdown parser.

PR #1283, opened 2023-06-14
[CSS] {Comments} Prevent text from appearing centered to the side of an <img> in the same <p> block

Junebugging:
A fix for issue #1281 

### The problem:
When an image is taller than it is wide, text will appear to the sides of an image, vertically centered.

### The fix:
Adds a simple CSS rule which enables flex on \<p> blocks within a comment, aligns them to start, and gives them flex-direction column.

Applies the style to `.comment-node > div > div > .md-div p`, and adds it to styles.css.

### Complications
This is a bit of a quick fix and doesn't look the most professional. Preferably it'd be handled by parsing comments and moving \'s out of \<p>'s, but that would take more time and as all parsing add the potential for complications. I'd prefer if this could be done with some more generic rule, or if there was a more certain way to target the \<p> blocks within comments than the above target.

There's a potential but unlikely issue, if the depth of the \<p> block varies, `.comment-node > div > div > .md-div p` might not catch it
In which case it could be made more generic, just `.comment-node .md-div p`, but that could affect other elements if `.md-div` applies to more than just comment \<p> blocks, inside `.comment-node`, which I don't know, hence why I made it as specific as possible.


src/assets/css/main.css
```diff
@@ -213,6 +213,13 @@ blockquote {
   overflow-y: auto;
 }
 
+.comment-node > div > div > .md-div p {
```
SleeplessOne1917, line 216: Please use bootstrap classes on the elements directly where possible.

Junebugging: Ah alright thanks for the feedback, I'll look into it

Junebugging: I can see in comment-node.tsx that many of the classes are as specific as individual rules, but there is no rule just for flex, so would the addition of the following be acceptable:

```css
.flex {
  display: flex !important;
}
```

The \<p> element doesn't have flex, so it would need this class. The other two rules there are already classes for

SleeplessOne1917: I just realized that this is for HTML generated from markdown, so my original suggestion isn't too great. Your original solution is closer than your revised solution. Can you get it to work with `.comment-node .md-div p`? If I'm not mistaken, selecting for multiple levels of direct child divs is unnecessary.

Junebugging: Yeah that should work just fine, but I'll have to see if there are other .md-div's that aren't the one containing the markdown text, otherwise it could screw with the formatting of other things there too

Junebugging: Okay yeah that didn't take long haha, yeah there's only one use of the class .md-div in comment-node.tsx, so that should work just fine

The next issue is where it should be placed, in main.css, or in every individual theme scss?

Junebugging:
Someone in working on an improvement of the implementation of themes, so I've kept the rule in main.css, which for the time being is the best option, but made it more general

src/assets/css/main.css
```diff
@@ -434,3 +444,8 @@ br.big {
 em-emoji-picker {
   width: 100%;
 }
+
```
dessalines, line 447: Run prettier. Or make sure your husky gets set up correctly by running yarn lint, because for some reason it didnt.

SleeplessOne1917 (approved):

dessalines:
```
Checking formatting...
[warn] src/assets/css/main.css
[warn] Code style issues found in the above file. Forgot to run Prettier?
```

dessalines (approved):

SleeplessOne1917:
@Junebugging I think this is good to merge, but it seems woodpecker does not agree with me.

jsit:
I’ll take a closer look at this

jsit:
Honestly I think this would be a good use case for adding a class to the comment body wrapper. A selector like this is so tied to the DOM that it would work better as something like `.comment-body p`

But on top of that, the same thing can be accomplished in a simpler way just by doing `display: block` on the `img` here.

Junebugging:
I agree with adding more classes like the one you used, actually that has probably already been done in the CSS styling refactoring branch, or it should be. Someone is working on moving the project more away from bootstrap, and adding more classes like the one you mentioned, and it has been approved by one of the maintainers. I don't know exactly where that is though.

Display:block on the image would be simpler, which is usually nicer, but it depends on if <p> elements are given the flex display-type, and if having a mix of flex and block elements in the same div can create issues, which I'm not sure of. I don't have the time to check that right now though, 


----

Also if you could help me with something, Dessalines mentioned
> Run prettier. Or make sure your husky gets set up correctly by running yarn lint, because for some reason it didnt.

Where are the instructions for setting up the project? The local CONTRIBUTE.MD file only links back to this repository, so I haven't been able to build the project.

Junebugging:
Also if you could help me with something, Dessalines mentioned
> Run prettier. Or make sure your husky gets set up correctly by running yarn lint, because for some reason it didnt.

Where are the instructions for setting up the project? The local CONTRIBUTE.MD file only links back to this repository, so I haven't been able to build the project.

jsit:
@Junebugging The contributing markdown file links here:

https://join-lemmy.org/docs/contributors/01-overview.html

I've had the best luck using the Cargo set up described here:

https://join-lemmy.org/docs/contributors/02-local-development.html

jsit:
@Junebugging 

> I thought img blocks were given the display type "block" by default, in which case there might be a reason they have been given a different display type

The `img` element is inline by default, `p` is block by default.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/P

jsit:
Yeah, as long as I put the image on its own line, it will display as its own block. If I put it on the same line, it will appear inline.

This seems to me like the correct behavior. 

(Don't ask why my most recent download is a photo of the Hindenburg)

Screenshot 2023-06-26 at 9 17 22 AM

Screenshot 2023-06-26 at 9 17 26 AM

Screenshot 2023-06-26 at 9 19 00 AM

Screenshot 2023-06-26 at 9 19 03 AM

SleeplessOne1917:
Using pictures inline is a pretty common thing on Hexbear.

Junebugging:
Ah I see, I still think it looks very unprofessional and odd when the text is centered next to the image though
What about aligning the text to the top of the image at least?

And the people I see using this seem to be doing it on accident, so maybe when an image is inserted it should automatically be inserted in a new paragraph? Then power-users of markdown can undo the new paragraphing if they want, to use this trick.

I don't think a power-user trick should be so easy to do accidentally, and make the website look more wack and ruin the flow of reading comments by aligning text all over the place.

jsit:
I disagree

PR #1284, opened 2023-06-14
Remove `pictrsDeleteToast` usage from `PostForm`

### LINKED ISSUE #1255: [Bug]: Clicking on "delete image" alert didnt remove image from form
### Summary

After deleting the image the image should really be removed from the form since it doesn't exist anymore.
image

Related: https://github.com/LemmyNet/lemmy-ui/issues/1254

### Steps to Reproduce

1. Upload image
2. Delete image
3. Profit?

### Technical Details

nope, all looks fine.

### Version

v0.17.3

alectrocute:
This PR stores an uploaded image's `delete_url` in state instead and makes visible a "delete" button which fetches the `delete_url` and removes the image/url. Also fixes infinite loading bug as mentioned in this PR's comments. Resolves #1255.

Future thought: It would be really nice to move away from `pictrsDeleteToast` entirely. `EmojisForm` is an easy candidate for a similar treatment to this PR. We'll need to add some more complicated functionality/UI for multiple image uploads, like in `MarkdownTextarea`, so maybe let's save that for last. 😅

Screenshot 2023-06-14 at 4 36 01 PM

Screenshot 2023-06-14 at 4 45 13 PM

Thanks, Lemmy team!


src/shared/components/post/post-form.tsx
```diff
@@ -644,18 +664,35 @@ export class PostForm extends Component<PostFormProps, PostFormState> {
...
+            imageLoading: false,
+            imageDeleteUrl: res.data.delete_url as string,
+          });
         } else {
           toast(JSON.stringify(res), "danger");
         }
       } else if (res.state === "failed") {
         console.error(res.msg);
         toast(res.msg, "danger");
+        i.setState({ imageLoading: false });
```
alectrocute, line 677: Caught an infinite loading bug here, this resolves it.

src/shared/components/post/post-form.tsx
```diff
@@ -553,7 +565,15 @@ export class PostForm extends Component<PostFormProps, PostFormState> {
...
 
   handlePostUrlChange(i: PostForm, event: any) {
-    i.setState(s => ((s.form.url = event.target.value), s));
+    const url = event.target.value;
+
+    i.setState({
+      form: {
+        url,
+      },
+      imageDeleteUrl: "",
```
alectrocute, line 574: If `form.url` is changed, we can assume user wants to abandon image – so `imageDeleteUrl` should be cleared.

src/shared/components/post/post-form.tsx
```diff
@@ -644,18 +664,35 @@ export class PostForm extends Component<PostFormProps, PostFormState> {
...
+  handleImageDelete(i: PostForm) {
+    const { imageDeleteUrl } = i.state;
+
+    fetch(imageDeleteUrl);
+
+    i.setState({
+      imageDeleteUrl: "",
+      imageLoading: false,
+      form: {
+        url: "",
```
alectrocute, line 691: Clearing the `form.url` is important here to remove the now-deleted image.

src/shared/components/post/post-form.tsx
```diff
@@ -269,6 +270,17 @@ export class PostForm extends Component<PostFormProps, PostFormState> {
             {url && isImage(url) && (
               <img src={url} className="img-fluid" alt="" />
             )}
+            {this.state.imageDeleteUrl && (
```
alectrocute, line 273: Using `this.state.imageDeleteUrl` for visibility of this button ensures it only appears alongside uploaded images.

src/shared/components/post/post-form.tsx
alectrocute: Would be really nice to add an uploaded media manager someday where all uploads are tracked in `LocalStorage`. 🤔

src/shared/components/post/post-form.tsx
SleeplessOne1917: Thank you!

SleeplessOne1917 (approved):
This is definitely easier to work with than the toast.

PR #1304, opened 2023-06-15
Fix <hr> being almost invisible

### LINKED ISSUE #533: <hr> hard to see in darkly theme.

ayan4m1:
Before:

Screenshot 2023-06-15 082518

After:

Screenshot 2023-06-15 082435

Two changes:

 * Add `border border-primary` to the `<hr>` but only if it is not the last item in the listing.
 * Add typing information to `this.posts.map()` call

Fixed #533

### DIFF
--- a/src/shared/components/post/post-listings.tsx
+++ b/src/shared/components/post/post-listings.tsx
@@ -68,7 +68,7 @@ export class PostListings extends Component<PostListingsProps, any> {
     return (
       <div>
         {this.posts.length > 0 ? (
-          this.posts.map(post_view => (
+          this.posts.map((post_view: PostView, idx: number) => (
             <>
               <PostListing
                 post_view={post_view}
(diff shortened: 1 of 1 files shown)


src/shared/components/post/post-listings.tsx
```diff
@@ -68,7 +68,7 @@ export class PostListings extends Component<PostListingsProps, any> {
     return (
       <div>
         {this.posts.length > 0 ? (
-          this.posts.map(post_view => (
+          this.posts.map((post_view: PostView, idx: number) => (
             <>
```
SleeplessOne1917, line 72: Are the explicit types necessary? Typescript should already be able to infer these types if I'm not mistaken.

ayan4m1: Good point, updated!

PR #1305, opened 2023-06-15
Fix infinite spinner for login that results in failure

ayan4m1:
Animation

Added error handling to the login form, so it doesn't just "spin infinitely" when there is an error. Error is surfaced as a toast using existing code, just calling i.setState() in the catch block.

### DIFF
--- a/src/shared/components/home/login.tsx
+++ b/src/shared/components/home/login.tsx
@@ -160,38 +160,42 @@ export class Login extends Component<any, State> {
     if (username_or_email && password) {
       i.setState({ loginRes: { state: "loading" } });
 
-      const loginRes = await HttpService.client.login({
-        username_or_email,
-        password,
-        totp_2fa_token,
-      });
-      switch (loginRes.state) {
-        case "failed": {
-          if (loginRes.msg === "missing_totp_token") {
-            i.setState({ showTotp: true });
-            toast(i18n.t("enter_two_factor_code"), "info");
+      try {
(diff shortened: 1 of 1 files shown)


SleeplessOne1917:
Your code looks good, but I'm confused as to why the exception isn't already being captured in the API wrapper and returning a result with state = failed.

ayan4m1:
Does this help at all?

Screenshot 2023-06-15 094951

SleeplessOne1917:
Here's some relevant code:
```typescript
class WrappedLemmyHttpClient {
  #client: LemmyHttp;

  constructor(client: LemmyHttp) {
    this.#client = client;

    for (const key of Object.getOwnPropertyNames(
      Object.getPrototypeOf(this.#client)
    )) {
      if (key !== "constructor") {
        WrappedLemmyHttpClient.prototype[key] = async (...args) => {
          try {
            const res = await this.#client[key](...args);

            return {
              data: res,
              state: "success",
            };
          } catch (error) {
            console.error(`API error: ${error}`);
            toast(i18n.t(error), "danger");
            return {
              state: "failed",
              msg: error,
            };
          }
        };
      }
    }
  }
}
```

Even though `console.error` is called in the catch block and an error toast is displayed, the client should still return a a `RequestState` object.

ayan4m1:
@SleeplessOne1917 I have updated the code to address your points, take a look when you can?

dessalines (approved):
Ahh thanks for this one, sry I missed it!

PR #1312, opened 2023-06-15
feat(navbar): Add isMobile util and render nav text basing on it

mr00z:


mr00z:
Screenshots (tablet and desktop view remain untouched):
localhost_1234_(iPhone SE)
localhost_1234_(iPad Air)
Zrzut ekranu 2023-06-15 o 21 11 57

src/assets/css/themes/darkly-red.css
```diff
@@ -3721,6 +3723,9 @@ input[type="submit"].btn-block {
   pointer-events: none;
   cursor: default;
 }
+.nav-link .nav-link__icon-text {
```
SleeplessOne1917, line 3726: Is nav-link__icon-text used in places other than as a descendant of a nav-link? If so, you might be able to apply it directly.

src/assets/css/themes/darkly-red.css
```diff
@@ -3936,6 +3941,9 @@ input[type="submit"].btn-block {
   .navbar-expand-md .navbar-toggler {
     display: none;
   }
+  .nav-link {
```
SleeplessOne1917, line 3944: There are styles for this further up in the file.

src/assets/css/themes/darkly-red.css
```diff
@@ -3711,6 +3711,8 @@ input[type="submit"].btn-block {
 .nav-link {
   display: block;
   padding: 0.5rem 2rem;
+  display: flex;
```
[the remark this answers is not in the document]
mr00z: oh right, didn't notice that 😅 setting margin to auto won't work I guess because what we want here is vertical alignment to be centered

src/shared/components/app/navbar.tsx
```diff
@@ -47,13 +49,18 @@ function handleLogOut(i: Navbar) {
...
+  ).userAgent;
+
   state: NavbarState = {
     unreadInboxCountRes: { state: "empty" },
     unreadReportCountRes: { state: "empty" },
     unreadApplicationCountRes: { state: "empty" },
   };
   collapseButtonRef = createRef<HTMLButtonElement>();
   mobileMenuRef = createRef<HTMLDivElement>();
+  isMobileUA: boolean = isMobile(this.userAgent);
```
SleeplessOne1917, line 63: The type should be able to be inferred correctly without the explicit boolean type.

src/assets/css/themes/darkly-red.css
mr00z: not within this media query

src/assets/css/themes/darkly-red.css
mr00z: I wanted to add proper specifity for this one, so that no one will be able to use this class outside `.nav-link`

src/shared/components/app/navbar.tsx
mr00z: right, I can remove it

alectrocute:
What's the benefit of using `UserAgent` vs. just a standard CSS media query for something like this?

mr00z:
> 



> What's the benefit of using `UserAgent` vs. just a standard CSS media query for something like this?

This way we can serve proper html from the server already, without needing to wait for browser to interpret the code

mr00z:
I accidentally commited changes to lemmy-translations submodule and cannot undo that, if that's a problem I'll try to make another PR

src/shared/components/app/navbar.tsx
```diff
@@ -47,13 +49,18 @@ function handleLogOut(i: Navbar) {
...
+  ).userAgent;
+
   state: NavbarState = {
     unreadInboxCountRes: { state: "empty" },
     unreadReportCountRes: { state: "empty" },
     unreadApplicationCountRes: { state: "empty" },
   };
   collapseButtonRef = createRef<HTMLButtonElement>();
   mobileMenuRef = createRef<HTMLDivElement>();
+  isMobileUA = isMobile(this.userAgent);
```
alectrocute, line 63: If we’re deciding to detect mobile this way –

It would be nice to have `isMobile` be a method with no parameters that can be reused in the future more easily throughout the app. That way, you can avoid having to add a `private` to this component and it would also make it much more of a utility method; It’s in `utils.ts` after all. ☺️

Also, it would be nice if we can avoid adding an additional NPM dependency.

package.json
```diff
@@ -110,6 +118,7 @@
     "style-loader": "^3.3.2",
     "terser": "^5.17.3",
     "typescript": "^5.0.4",
+    "ua-parser-js": "^1.0.35",
```
alectrocute, line 121: Any chance we can avoid adding an additional NPM dependency? A single guy operates this package.

alectrocute:
> This way we can serve proper html from the server already, without needing to wait for browser to interpret the code

It's my opinion that doing it the way demonstrated in this PR adds needless complexity for something that can be done with media queries. Any discernable performance gain of using JS vs. CSS breakpoints in Bootstrap (that we're already loading/caching in the browser) is almost definitely not worth the tradeoff.

fheft:
We should avoid doing manual changes in the theme files (`src/assets/css/themes/...`), because those files are auto-generated based on vanilla Bootstrap + variable definitions. This might not be obviously clear in the current repo state, but hopefully get's easier to see with #1246. 

Adding overrides in [main.css](https://github.com/LemmyNet/lemmy-ui/blob/main/src/assets/css/main.css) should be the way to go for all CSS changes that can't be made via theme variables (see [`_variables.scss`](https://github.com/twbs/bootstrap/blob/v4.6.2/scss/_variables.scss) in the Bootstrap project)

src/shared/components/app/navbar.tsx
mr00z: the argument here is inevitable since we need to read `this.context` where the user agent object lies which can be done only within the component

package.json
mr00z: I didn't really want to reinvent the wheel and parse user-agent header on my own. This package will not affect our bundle size as we're using it server-side only. The rest of the usages come from `@types/ua-parser-js` package which also doesn't increase bundle size since it's stripped out during build process.

src/shared/components/app/navbar.tsx
mr00z: I can rename the function though :)

src/assets/css/themes/darkly-red.css
```diff
@@ -3709,8 +3709,9 @@ input[type="submit"].btn-block {
   list-style: none;
 }
 .nav-link {
-  display: block;
   padding: 0.5rem 2rem;
+  display: flex;
```
dessalines, line 3713: Don't edit these themes directly, or their fixes won't apply anywhere else. Use `main.css`

dessalines (changes requested):
We don't care what the user agent is, we care about screen sizes. Use the bootstrap methods already in place in this codebase to render content based on screen width.

mr00z:
Closing as there is no point in changing this PR

PR #1313, opened 2023-06-15
Admin Settings: Bugfixes

alectrocute:
Hi Lemdevs!

This PR addresses some outstanding bugs in our admin settings panel –

- Tweak margin on icon & banner labels to match rest of form
- Add missing bind for `handleRemoveInstance` (ugh, I hate we have to do this!)
- Move `loading` state to `AdminSettings`, pass as prop to all children, which resolves some infinite loading issues
- Add util to properly validate allowed/banned instance TLD's
- Change admin settings header text from "Name/save your site" to "Setup/edit your site", which makes a lot more sense

Thanks!

### DIFF
--- a/src/shared/components/home/admin-settings.tsx
+++ b/src/shared/components/home/admin-settings.tsx
@@ -39,6 +39,7 @@ interface AdminSettingsState {
   instancesRes: RequestState<GetFederatedInstancesResponse>;
   bannedRes: RequestState<BannedPersonsResponse>;
   leaveAdminTeamRes: RequestState<GetSiteResponse>;
+  emojiLoading: boolean;
   loading: boolean;
   themeList: string[];
   isIsomorphic: boolean;
@@ -53,6 +54,7 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
     bannedRes: { state: "empty" },
     instancesRes: { state: "empty" },
     leaveAdminTeamRes: { state: "empty" },
+    emojiLoading: false,
     loading: false,
     themeList: [],
     isIsomorphic: false,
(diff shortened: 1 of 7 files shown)


src/shared/components/home/setup.tsx
```diff
@@ -73,6 +73,7 @@ export class Setup extends Component<any, State> {
                 onSaveSite={this.handleCreateSite}
                 siteRes={this.state.siteRes}
                 themeList={this.state.themeList}
+                loading={false}
```
alectrocute, line 76: This can always be `false` since the page is redirected upon submit of this form.

src/shared/components/home/admin-settings.tsx
```diff
@@ -198,6 +205,7 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
                     onCreate={this.handleCreateEmoji}
                     onDelete={this.handleDeleteEmoji}
                     onEdit={this.handleEditEmoji}
+                    loading={this.state.loading}
```
dessalines, line 208: Seems like this should have its own loading state, different from saving site.

src/shared/components/home/admin-settings.tsx
```diff
@@ -300,23 +312,35 @@ export class AdminSettings extends Component<any, AdminSettingsState> {
   }
 
   async handleEditEmoji(form: EditCustomEmoji) {
+    this.setState({ loading: true });
```
dessalines, line 315: Probably should be `emojiLoading`

dessalines (commented):
Looks good, but I would separate the site loading indicator from the emoji loading one.

alectrocute:
Thanks for the feedback, @dessalines! I've implemented the changes.

dessalines (approved):

PR #1314, opened 2023-06-15
Prevent non-admins from accessing `/setup`

### LINKED ISSUE #897: Setup page is publicly available after instance created

alectrocute:
Hey Lemdevs!

In this PR:

- Prevent non-admins from accessing `/setup`. This isn't much of a big concern, but non-admins really have no business seeing this form on an established site, eg: https://lemmy.ml/setup
- Add `/setup` to list of auth-required routes

Resolves #897.

Thanks so much!

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -161,6 +161,10 @@ server.get("/*", async (req, res) => {
       }
 
       if (site) {
+        if (path === "/setup" && !site.my_user?.local_user_view.person.admin) {
+          return res.redirect("/");
+        }
+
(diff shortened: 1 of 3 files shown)


SleeplessOne1917:
I'm not sure if admin users exist on the server before setup.

alectrocute:
@SleeplessOne1917 I believe the conditional is inside another conditional which checks if the site is setup.

PR #1316, opened 2023-06-15
fix "block community" functionality in sidebar

jenterkin:
It looks like a commit from yesterday broke some interactions in the sidebar. This PR only addresses blocking a community, but I can knock out the other bits tonight or tomorrow as well.

The specific issues addressed here are:
1. The community id being blocked was hard-coded to 0.
2. The loading icon would never go away since the sidebar has no way of knowing when the api request is done. I just removed that in this PR as it seems like the loading icon didn't exist before the commit yesterday and I'm not sure how to nicely implement that without some `thunk`-y mechanism.

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -647,6 +647,12 @@ export class Community extends Component<
     const blockCommunityRes = await HttpService.client.blockCommunity(form);
     if (blockCommunityRes.state == "success") {
       updateCommunityBlock(blockCommunityRes.data);
+      this.setState(s => {
+        if (s.communityRes.state == "success") {
+          s.communityRes.data.community_view.blocked =
+            blockCommunityRes.data.blocked;
+        }
+      });
     }
   }
 
--- a/src/shared/components/community/sidebar.tsx
+++ b/src/shared/components/community/sidebar.tsx
@@ -63,7 +63,6 @@ interface SidebarState {
   removeCommunityLoading: boolean;
   leaveModTeamLoading: boolean;
   followCommunityLoading: boolean;
-  blockCommunityLoading: boolean;
   purgeCommunityLoading: boolean;
 }
 
@@ -77,7 +76,6 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
     removeCommunityLoading: false,
     leaveModTeamLoading: false,
     followCommunityLoading: false,
-    blockCommunityLoading: false,
     purgeCommunityLoading: false,
   };
 
@@ -104,7 +102,6 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
         removeCommunityLoading: false,
         leaveModTeamLoading: false,
         followCommunityLoading: false,
-        blockCommunityLoading: false,
         purgeCommunityLoading: false,
       });
     }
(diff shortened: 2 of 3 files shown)


src/shared/components/community/community.tsx
```diff
@@ -647,6 +647,12 @@ export class Community extends Component<
...
     if (blockCommunityRes.state == "success") {
       updateCommunityBlock(blockCommunityRes.data);
+      this.setState(s => {
+        if (s.communityRes.state == "success") {
+          s.communityRes.data.community_view.blocked =
+            blockCommunityRes.data.blocked;
+        }
+      });
     }
   }
```
jenterkin, line 657: One thing I'd like to call out is that there is some duplication going on between here and [post.tsx](https://github.com/LemmyNet/lemmy-ui/pull/1316/files#diff-630788fba7307c5ca6e02c2ecd35085a0e801fe27e318bdfc0a5fd258b3e4e0aR730).

I'd like to pull this out in a separate PR and was wondering what the suggested way to do this is. I'm not a huge fan of large `util` libs since they can be pretty disorganized and tend to have pretty inconsistent interfaces. I have a separate branch where I pull this logic out into `UserService` that handles blocking communities and accepts an optional callback for updating component state.

```ts
  public static async followCommunity(
    form: FollowCommunity,
    updateCommunity?: (res: RequestState<CommunityResponse>) => void
  ): Promise<void> {
    /**
     * Sends a request to follow a community
     *
     * @param form - Form data for following a community
     * @callback updateCommunity - Optional callback for updating community state
     */
    const followCommunityRes = await HttpService.client.followCommunity(form);
    updateCommunity !== undefined && updateCommunity(followCommunityRes);

    // Update myuserInfo
    if (followCommunityRes.state == "success") {
      const communityId = followCommunityRes.data.community_view.community.id;
      const mui = UserService.Instance.myUserInfo;

      if (mui) {
        mui.follows = mui.follows.filter(i => i.community.id != communityId);
      }
    }
  }
```

Which would simplify these functions to something like

```ts
handleBlockCommunity(form) {
  UserService.blockCommunity(form, this.updateCommunity);
}
```

azzamsaziz:
Adding from my perspective: I'm moderating a few communities right now and needed the sidebar to be shown when a user creates content, to ensure that they've read the rules and know which community they are posting in.

src/shared/components/community/sidebar.tsx
```diff
@@ -662,10 +642,11 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
   }
 
   handleBlockCommunity(i: Sidebar) {
-    i.setState({ blockCommunityLoading: true });
+    const { community, blocked } = i.props.community_view;
+
     i.props.onBlockCommunity({
```
alectrocute, line 647: Good catch on the `0`!

jenterkin:
@dessalines @SleeplessOne1917 Looks like the ci failures are due to translations not yet being added for `number_online` and unrelated to my change. Any chance I can get some help on how to resolve or merging anyway?

dessalines:
Hrm.. it must have wrongly updated the translations somewhere. Don't worry I'll get it handled.

src/shared/components/community/community.tsx
jenterkin: Ah, that's great to hear. Inferno really hasn't aged well. Let me know if you'd like any help with the new UI. Is the idea to only migrate the frontend to leptos?

PR #1318, opened 2023-06-16
Changed required checkbox in github template

ludrol:
Changed how users should fill out git hub issues to be more in logic the the question asked.


L3v3L:
shouldn't these questions be reversed so that it is required:

Is this a UI (not related to the server side) issue? For server side issues use [Lemmy back end](https://github.com/LemmyNet/lemmy).

dessalines (approved):
Thx!

PR #1322, opened 2023-06-16
Refactor/tidy server

alectrocute:
Hi Lemdevs!

I propose we:

- Refactor our `server/index.tsx` to use a common Express.js pattern (`handlers`, `middleware` and `utils`)
- Add Express types to handler/middleware export parameters
- Make more human readable wherever possible

Screenshot 2023-06-16 at 10 47 26 AM

In future:

- Rewrite/split out `CatchAllHandler` to be simpler if possible

Thanks so much!

PR #1325, opened 2023-06-16
fix loading custom themes with a custom LEMMY_UI_EXTRA_THEMES_FOLDER

n3oney:
At least when using the `LEMMY_UI_EXTRA_THEMES_FOLDER` environment variable, themes do correctly appear in the list (nice), but when trying to use them, they unfortunately default to litely. This fixes it.

Please note: I haven't, and can't really test loading themes without a custom folder, but if [this](https://github.com/LemmyNet/lemmy-ui/blob/80e2c9602e4bad45bd9f3342a1a6114bfafdf1da/src/server/index.tsx#L107) works, then the changes I made here should still work. Do test it though.


alectrocute:
We should definitely wait until https://github.com/LemmyNet/lemmy-ui/pull/1322 is merged to implement this fix.

n3oney:
> We should definitely wait until #1322 is merged to implement this fix.

Sure, I'll just be using a patch file for now.

n3oney:
Okay I’ll rework this in a sec!

SleeplessOne1917 (approved):

PR #1327, opened 2023-06-16
Add semantic ID's and classes to improve theming/userscript experience

Zetaphor:
There is currently a major lack of semantic ID's and classes in the UI markup. This makes everything from detecting if the site is running Lemmy (for userscript) to theming more difficult than necessary.

This PR attempts to remedy that by adding semantics to the markup in a way that will simplify CSS queries and make site theming simpler. I'm aiming to take a light touch approach while still being thorough.

In the end I intend to rewrite the base CSS themes included with Lemmy to better utilize these markup enhancements. The current theme files are a bit messy and include lots of extraneous styling.

### Changes made

Changed page elements to use semantic HTML tags
   * The footer uses `<footer>`
   * Posts and comments use `article`
   * Sidebars are now using `aside` with contained `section` elements
   * Sidebar sections have distinguishing ID's

Added ID's and classes to improve CSS specificity for themes and userscripts
   * `#app` gained the `lemmy-site` class
   * ID's added to nav and footer
   * Semantic classes added to nav sections and icons
   * `MarkdownTextArea` component `id` and `formId` renamed from `comment-textarea-{randomString}` to `markdown-textarea-{randomString}` as the component is used for more than comments
   * CommentForm accepts a class string to distinguish between a top level post reply and a comment thread reply
   * Additional classes added to post listings to enable simpler CSS targeting (container, score, media)
   * Added class to the more button on comment nodes


Zetaphor:
The idea of using [semantic elements (nav, footer)](https://www.w3schools.com/html/html5_semantic_elements.asp) was brought up in a Matrix thread.

This would be an improvement for theming, but would also likely be a breaking change for any existing custom CSS and possibly even userscripts that modify layout.

I'm not opposed to this, especially this early in the lifecycle of Lemmy adoption, but I'd like to solicit feedback for the negative case (the positive is self evident).

Zetaphor:
So far nothing makes any structural changes to the frontend, I was trying to avoid that where possible by starting with the light touch of adding semantic sugar.

I do think the site could benefit from some breaking structural changes, but I think that would be outside of the scope of this task

Zetaphor:
I think this should be good enough to put up for review. It's now at a point where I feel I can start rewriting the included themes to utilize these new attributes.

src/shared/components/app/navbar.tsx
```diff
@@ -240,7 +244,7 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
               </NavLink>
             </li>
             {amAdmin() && (
-              <li className="nav-item">
+              <li className="nav-item nav-item-admin">
```
SleeplessOne1917, line 247: These admin-item-x classes seem unique. Maybe they're better off as IDs?

src/shared/components/comment/comment-node.tsx
```diff
@@ -282,7 +282,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
 
     return (
       <li className="comment" role="comment">
-        <div
+        <article
```
SleeplessOne1917, line 285: Are article tags usually used for comments in other sites using semantic tags?

src/shared/components/community/sidebar.tsx
```diff
@@ -132,8 +132,8 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
     const myUSerInfo = UserService.Instance.myUserInfo;
     const { name, actor_id } = this.props.community_view.community;
     return (
-      <div>
-        <div className="card border-secondary mb-3">
+      <div id="sidebarContainer">
```
SleeplessOne1917, line 135: Maybe aside would be better than div for this?

src/shared/components/app/navbar.tsx
Zetaphor: Fair point, I'll switch up the nav items now

src/shared/components/post/post-listing.tsx
```diff
@@ -230,13 +230,13 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   body() {
     const body = this.postView.post.body;
     return body ? (
-      <div className="col-12 card my-2 p-2">
+      <article id="postContent" className="col-12 card my-2 p-2">
```
kevinmershon, line 233: `id="postContent"` should be a class or a data-id or role etc., since ids should be unique within the page

src/shared/components/comment/comment-node.tsx
kevinmershon: > Are article tags usually used for comments in other sites using semantic tags?

`blockquote` is what's used by kbin although I don't think that's better because then quotes are also `blockquote`.

`section` might be better

src/shared/components/post/post-listing.tsx
Zetaphor: There is only one postContent in a post. This is the body text for a post:

image

src/shared/components/comment/comment-node.tsx
Zetaphor: I was following the advice I see here, thoughts?

https://www.w3schools.com/html/html5_semantic_elements.asp

image

SleeplessOne1917: @Zetaphor looks good. I was asking more because I wasn't sure the correct tag to use rather than saying you were using it wrong.

Zetaphor: I wasn't sure either, this is my first go at properly using semantic tags. If it's good with everyone I'll mark this one resolved.

src/shared/components/post/post-listing.tsx
kevinmershon: ahh this is on the post view page not on the lists page. My bad

src/shared/components/community/sidebar.tsx
Zetaphor: This is resolved in the latest commit. Here's the structure of the sidebar on the homepage:

image

And then on a post:

image

Zetaphor:
I also added a class to the more button on a comment node as per a suggestion from another user

SleeplessOne1917 (approved):

Zetaphor:
Thanks everyone! <3

Issue #1162, opened 2023-06-11
Add default profile picture


SleeplessOne1917:
This should hipefully keep things lined up visually. 
Perhaps the Lemmy lemming would be a good default pick. It could be worth letting instance admins be able to set their own default pfps, although that would require a backend change.

krestenlaust:
Related issue: #1174

SleeplessOne1917:
> Personally, I think it falls a bit short of the responsibilities of an instance, to set a default profile picture, since they (as far as I know) aren't supposed to be responsible for the theme, but rather simply focus on hosting.

I was thinking of providing a default lemmy icon so that instance admins don't need to set their own, but giving them the ability to do so if they desire.

krestenlaust:
It might be a good idea to leave it as an option to the instance admin, I agree.

It could also be possible to do it the other way around, where everyone (across instances) has the same default profile picture, but depending on what instance you're viewing as, it changes.

PR #1339, opened 2023-06-17
Add default profile picture for users who do not set one

SleeplessOne1917:
Addresses #1162. Unsure if I should open a separate issue for allowing instance admins to set their own default profile picture if they want.

Here is a post made by a user with a default pfp.
image

### DIFF
--- a/src/shared/components/person/person-details.tsx
+++ b/src/shared/components/person/person-details.tsx
@@ -107,16 +107,6 @@ export class PersonDetails extends Component<PersonDetailsProps, any> {
     setupTippy();
   }
 
-  // TODO wut?
-  // componentDidUpdate(lastProps: UserDetailsProps) {
-  //   for (const key of Object.keys(lastProps)) {
-  //     if (lastProps[key] !== this.props[key]) {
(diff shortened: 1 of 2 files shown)


src/shared/components/person/person-details.tsx
```diff
@@ -107,16 +107,6 @@ export class PersonDetails extends Component<PersonDetailsProps, any> {
     setupTippy();
   }
 
-  // TODO wut?
```
SleeplessOne1917, line 110: This was 3 years old, commented out, and I have not idea what to make of "TODO wut?" so I removed it.

dessalines (approved):

dessalines:
I haven't been a big fan of slamming in pfps where people don't set them, but if enough people want them to make things line up better, I'm not opposed.

SleeplessOne1917:
@dessalines There was a discussion about this in the dev channel. Do you think [gravatars](https://en.gravatar.com/site/implement/images/) might be a better approach?

krestenlaust:
Something like gravatars would be pretty cool to have

Issue #863, opened 2022-11-18
Better html naming schemes for themeing


drannex42:
I am tempted to start a PR on this, unless someone else more well-versed in the lemmy-ui core would prefer taking this over, but: 

We need better defining of areas in the HTML scheme, for example:

![image](https://user-images.githubusercontent.com/4322153/202792900-227064f1-f1bd-4941-b4aa-42963401b7ab.png)
This is the bottom navigation bar of a post, this should have an ID or a first-order css class that would define this as `post-meta-nav` or similar. _This does not need to have any attributes attached, just needs something to identify it easier especially for custom themes._ Because as it stands, for themes to theme it we have to target the `d-flex.justify-content-start.flex-wrap.text.muted.font-weight-bold.mb-1` css selector. This is not operationally coherent of useful, especially if any of those values changes _or_ another element with the same classifer is included. 

Comments, navigation bar, thumbnails, avatars, all have this same problem. 

For a single theme this would work! But for custom themes, not having any sort of syntactic naming IDs (`id="post-meta-nav"`) really makes theming a chore.

dessalines:
You **should not** be building themes targeting specific selectors, you should build a bootstrap v4 theme, and plug it in.. I have a theming guide here: https://join-lemmy.org/docs/en/administration/theming.html

This upcoming year, lemmy-ui will upgrade to bootstrap v5 , and do auto-theme generation from scss files: #703

drannex42:
The issue is that allowing specific targeting would allow us to fix a lot of the issues, especially in the layout issues that the current version has for our custom themes.

The current layout decisions are not great for some of themes, for example on beehaw we have the [tildes-for-lemmy](https://github.com/drannex42/tildes-for-lemmy-theme) theme since we love Lemmy, we love the design of Tildes, but we wanted the look and feel of Tildes, but the capabilities of Lemmy.

drannex42:
The lemmy-ui project is great! Especially since we could build out own front-end, but we wanted to keep up with all the great changes and additions.  If we had just simple declarative naming on elements, thus helps for both accessibility reasons, userscripts, and for theming. This isn't a call for theming, this helps a lot of different users in several cases.

dessalines:
> The issue is that allowing specific targeting would allow us to fix a lot of the issues, especially in the layout issues that the current version has for our custom themes.

[...]

PR #1340, opened 2023-06-17
fix: Remove inline styles and use Bootstrap classes

jsit:
Per conversation #863, definitely should not be using inline `style` attributes, even though the Bootstrap colors don't quite work here. Should be fixed on the theming side, not by brute-forcing inline styles.


SleeplessOne1917 (approved):
Good catch!

PR #1343, opened 2023-06-17
chore: Theme sass cleanup

jsit:
1. Remove a lot of redundancies in sass var definitions
2. Re-build theme stylesheets

Some buttons turned from "pill" to "chiclet" style, I think because the themes hadn't been rebuilt in a bit (`yarn themes:build`).

## Before

Screenshot 2023-06-17 at 1 26 39 AM

Screenshot 2023-06-17 at 1 26 29 AM

## After (should look identical)

Screenshot 2023-06-17 at 1 24 19 AM

Screenshot 2023-06-17 at 1 41 04 AM


src/assets/css/themes/_variables.darkly.scss
```diff
@@ -1,35 +1,51 @@
...
-$gray-100: #f8f9fa;
+@import "../../../../node_modules/bootstrap-v4/scss/functions";
+@import "../../../../node_modules/bootstrap-v4/scss/variables";
+
+// Colors
 $gray-200: #ebebeb;
-$gray-300: #dee2e6;
-$gray-400: #ced4da;
-$gray-500: #adb5bd;
 $gray-600: #888;
```
fheft, line 6: Overwriting variables _after_ importing the Bootstrap variables might not work, see [Theming Bootstrap: Variable defaults](https://getbootstrap.com/docs/4.1/getting-started/theming/#variable-defaults)

> Variable overrides within the same Sass file can come before or after the default variables. However, when overriding across Sass files, your overrides must come before you import Bootstrap’s Sass files.

fheft (commented):
Thanks, a clean-up is very much needed, as I omitted that in my last branch and even made the redundant variable situation worse (to get it working in the first place...).

> Some buttons turned from "pill" to "chiclet" style, I think because the themes hadn't been rebuilt in a bit (`yarn themes:build`).

There should actually be _no_ changes though (i.e. identical CSS files), because the themes were build very recently (#1246). Maybe it has something to do with the (not working) overrides?

src/assets/css/themes/litely-red.css
dessalines, line 1: These probably got deleted by accident, especially since we haven't merged the auto-gen themes yet.

src/assets/css/themes/_variables.darkly.scss
jsit: This is fine, because we are importing them again later in the main `darkly.scss` file.

We need them to set some of our theme variables to existing Bootstrap color variables; then we import Bootstrap again later.

SleeplessOne1917 (approved):
LGTM

src/assets/css/themes/_variables.darkly.scss
jsit: I'm incorrect! This is indeed causing issues. I've fixed this in the latest commit.

jsit:
@fheft 

> There should actually be no changes though (i.e. identical CSS files), because the themes were build very recently (https://github.com/LemmyNet/lemmy-ui/pull/1246). Maybe it has something to do with the (not working) overrides?

You're right -- this should be fixed in the latest commit.

Issue #1319, opened 2023-06-16
Question: Temperature check on UI reworking


djvs:
### Question

So I put some work into restyling the UI via a bunch of CSS hacks via browser extension:

![grim-20230616T022559Z 080351319](https://github.com/LemmyNet/lemmy-ui/assets/2954343/7372685a-2ef4-497b-a743-d0a2fde3daa3)

However it's really hacked together and needs to be an integrated effort.  The main thing is that the HTML actually has to be properly marked up - a lot of critical elements have no meaningful classes and only single-property utility classes.  See CSS changes here:

https://lemmy.world/post/87865

I would like to do a big PR with appropriate fixes for all this, but wanted to make sure it actually stood a chance of getting accepted.

SleeplessOne1917:
If this is to get rid of the excess whitespace on larger monitors, I was planning on taking a swing at that one.

djvs:
It's a bigger "style overhaul" deal, with a good amount of "add proper semantic HTML/classes to describe what elements are and allow future-proof styling/3rd party styles".

SleeplessOne1917:
@djvs there was a PRs merged recently that made more use of semantic tags to simplify styling. See #1327.

djvs:
Got it, this would be a furtherance of that effort (avoiding bare "divs" in general, etc.).  I suppose style updates could be a second effort (you can see some pretty large difference if you apply the user styles in the link).

djvs:
So the above PR handles the "marking up the HTML" bit of it.  The remaining bit re: this issue would basically be incorporating the styles linked in the OP here as the main theme.  I'm a fan of em (and noticed it got > 30 upvotes on lemmy.world) but up to you guys.  

You can peek at how they're basically supposed to look using Stylus/UserStyles on an update to date instance, although they're a little hacky/not polished due to current limitations.

Kache:
Surprised it's not been suggested to decouple "restyling" efforts from "marking up/structuring HTML well".

IMO the latter is more important b/c it allows users and 3rd parties to freely experiment & restyle, and b/c it's less likely to get delayed by any style bikeshedding.

djvs:
@Kache  https://github.com/LemmyNet/lemmy-ui/pull/1421 was towards that end.  Now that that's addressed I'd definitely like to tweak default styles a bit.

djvs:
[...]

PR #1361, opened 2023-06-17
Add unique classes to components

djvs:
See https://github.com/LemmyNet/lemmy-ui/issues/1319.  Want to add a lot of classes to allow proper contextualization of CSS.  

This should mostly create a migration path away from styles being tied directly to React (well, Inferno) components, which is a big anti-pattern for something that should be themeable.  

These are mostly just picked as I went, with a couple small naming conventions that should be apparent.  I'm not married to any particular change here besides the general gist of it, so if you want to edit the PR or anything like that let me know.

### DIFF
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -108,7 +108,7 @@ export class CommentReport extends Component<
           {i18n.t("reason")}: {r.comment_report.reason}
         </div>
         {r.resolver && (
-          <div className="comment-report__resolverwrap">
+          <div className="comment-report__resolver">
             {r.comment_report.resolved ? (
               <T i18nKey="resolved_by">
                 #
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -70,7 +70,7 @@ export class CommentReport extends Component<
     };
 
     return (
-      <div>
+      <div className="comment-report__root">
         <CommentNode
           node={node}
           viewType={CommentViewType.Flat}
@@ -101,14 +101,14 @@ export class CommentReport extends Component<
           onCreateComment={() => Promise.resolve({ state: "empty" })}
           onEditComment={() => Promise.resolve({ state: "empty" })}
         />
-        <div>
+        <div className="comment-report__person">
           {i18n.t("reporter")}: <PersonListing person={r.creator} />
         </div>
-        <div>
+        <div className="comment-report__reason">
           {i18n.t("reason")}: {r.comment_report.reason}
         </div>
         {r.resolver && (
-          <div>
+          <div className="comment-report__resolverwrap">
             {r.comment_report.resolved ? (
               <T i18nKey="resolved_by">
                 #
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -70,7 +70,7 @@ export class CommentReport extends Component<
     };
 
     return (
-      <div className="comment-report__root">
+      <div className="comment-report">
         <CommentNode
           node={node}
           viewType={CommentViewType.Flat}
--- a/src/shared/components/common/language-select.tsx
+++ b/src/shared/components/common/language-select.tsx
@@ -51,13 +51,17 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
     ) : (
       <div>
         {this.props.multiple && this.props.showLanguageWarning && (
-          <div className="alert alert-warning" role="alert">
+          <div
+            className="language-select__root alert alert-warning"
+            role="alert"
+          >
             {i18n.t("undetermined_language_warning")}
           </div>
         )}
         <div className="form-group row">
           <label
             className={classNames(
+              "language-select__plural",
               "col-form-label",
               `col-sm-${this.props.multiple ? 3 : 2}`
             )}
@@ -67,6 +71,7 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
           </label>
           <div
             className={classNames(
+              "language-select__deselect",
               "input-group",
               `col-sm-${this.props.multiple ? 9 : 10}`
             )}
@@ -100,7 +105,7 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
 
     return (
       <select
-        className={classNames("lang-select-action", {
+        className={classNames("language-select-action", {
           "form-control custom-select": !this.props.iconVersion,
         })}
         id={this.id}
--- a/src/shared/components/common/language-select.tsx
+++ b/src/shared/components/common/language-select.tsx
@@ -49,9 +49,9 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
     return this.props.iconVersion ? (
       this.selectBtn
     ) : (
-      <div>
+      <div className="language-select">
         {this.props.multiple && this.props.showLanguageWarning && (
-          <div className="language-select alert alert-warning" role="alert">
+          <div className="alert alert-warning" role="alert">
             {i18n.t("undetermined_language_warning")}
           </div>
         )}
--- a/src/shared/components/common/language-select.tsx
+++ b/src/shared/components/common/language-select.tsx
@@ -51,10 +51,7 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
     ) : (
       <div>
         {this.props.multiple && this.props.showLanguageWarning && (
-          <div
-            className="language-select__root alert alert-warning"
-            role="alert"
-          >
+          <div className="language-select alert alert-warning" role="alert">
             {i18n.t("undetermined_language_warning")}
           </div>
         )}
--- a/src/shared/components/common/listing-type-select.tsx
+++ b/src/shared/components/common/listing-type-select.tsx
@@ -39,11 +39,11 @@ export class ListingTypeSelect extends Component<
 
   render() {
     return (
-      <div className="btn-group btn-group-toggle flex-wrap mb-2">
+      <div className="listing-type-select__root btn-group btn-group-toggle flex-wrap mb-2">
         {this.props.showSubscribed && (
           <label
             title={i18n.t("subscribed_description")}
-            className={`btn btn-outline-secondary 
+            className={`listing-type-select__label listing-type-select__label listing-type-select__label----subscribed btn btn-outline-secondary
             ${this.state.type_ == "Subscribed" && "active"}
             ${!UserService.Instance.myUserInfo ? "disabled" : "pointer"}
           `}
@@ -62,7 +62,7 @@ export class ListingTypeSelect extends Component<
         {this.props.showLocal && (
           <label
             title={i18n.t("local_description")}
-            className={`pointer btn btn-outline-secondary ${
+            className={`listing-type-select__label listing-type-select__label listing-type-select__label----local pointer btn btn-outline-secondary ${
               this.state.type_ == "Local" && "active"
             }`}
           >
@@ -78,7 +78,7 @@ export class ListingTypeSelect extends Component<
         )}
         <label
           title={i18n.t("all_description")}
-          className={`pointer btn btn-outline-secondary ${
+          className={`listing-type-select__label listing-type-select__label listing-type-select__label----show-all pointer btn btn-outline-secondary ${
             (this.state.type_ == "All" && "active") ||
             (!this.props.showLocal && this.state.type_ == "Local" && "active")
           }`}
--- a/src/shared/components/common/listing-type-select.tsx
+++ b/src/shared/components/common/listing-type-select.tsx
@@ -43,7 +43,7 @@ export class ListingTypeSelect extends Component<
         {this.props.showSubscribed && (
           <label
             title={i18n.t("subscribed_description")}
-            className={`listing-type-select__label listing-type-select__label listing-type-select__label----subscribed btn btn-outline-secondary
+            className={`listing-type-select__label listing-type-select__label----subscribed btn btn-outline-secondary
             ${this.state.type_ == "Subscribed" && "active"}
             ${!UserService.Instance.myUserInfo ? "disabled" : "pointer"}
           `}
@@ -62,7 +62,7 @@ export class ListingTypeSelect extends Component<
         {this.props.showLocal && (
           <label
             title={i18n.t("local_description")}
-            className={`listing-type-select__label listing-type-select__label listing-type-select__label----local pointer btn btn-outline-secondary ${
+            className={`listing-type-select__label listing-type-select__label----local pointer btn btn-outline-secondary ${
               this.state.type_ == "Local" && "active"
             }`}
           >
@@ -78,7 +78,7 @@ export class ListingTypeSelect extends Component<
         )}
         <label
           title={i18n.t("all_description")}
-          className={`listing-type-select__label listing-type-select__label listing-type-select__label----show-all pointer btn btn-outline-secondary ${
+          className={`listing-type-select__label listing-type-select__label----show-all pointer btn btn-outline-secondary ${
             (this.state.type_ == "All" && "active") ||
             (!this.props.showLocal && this.state.type_ == "Local" && "active")
           }`}
--- a/src/shared/components/common/listing-type-select.tsx
+++ b/src/shared/components/common/listing-type-select.tsx
@@ -39,7 +39,7 @@ export class ListingTypeSelect extends Component<
 
   render() {
     return (
-      <div className="listing-type-select__root btn-group btn-group-toggle flex-wrap mb-2">
+      <div className="listing-type-select btn-group btn-group-toggle flex-wrap mb-2">
         {this.props.showSubscribed && (
           <label
             title={i18n.t("subscribed_description")}
--- a/src/shared/components/common/listing-type-select.tsx
+++ b/src/shared/components/common/listing-type-select.tsx
@@ -43,7 +43,7 @@ export class ListingTypeSelect extends Component<
         {this.props.showSubscribed && (
           <label
             title={i18n.t("subscribed_description")}
-            className={`listing-type-select__label listing-type-select__label----subscribed btn btn-outline-secondary
+            className={`listing-type-select__label listing-type-select__label--subscribed btn btn-outline-secondary
             ${this.state.type_ == "Subscribed" && "active"}
             ${!UserService.Instance.myUserInfo ? "disabled" : "pointer"}
           `}
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -22,7 +22,7 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
 
   render() {
     return (
-      <picture className="d-inline-block overflow-hidden">
+      <picture className="pictrs-image__root d-inline-block overflow-hidden">
         <source srcSet={this.src("webp")} type="image/webp" />
         <source srcSet={this.props.src} />
         <source srcSet={this.src("jpg")} type="image/jpeg" />
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -22,7 +22,7 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
 
   render() {
     return (
-      <picture className="pictrs-image__root d-inline-block overflow-hidden">
+      <picture className="pictrs-image d-inline-block overflow-hidden">
         <source srcSet={this.src("webp")} type="image/webp" />
         <source srcSet={this.props.src} />
         <source srcSet={this.src("jpg")} type="image/jpeg" />
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1211,7 +1211,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
                     {i18n.t("purge_user")}
                   </button>
                   <button
-                    className="post-listing__purge-purge-post btn btn-link btn-animate text-muted py-0"
+                    className="post-listing__btn post-listing__btn--purge-post btn btn-link btn-animate text-muted py-0"
                     onClick={linkEvent(this, this.handlePurgePostShow)}
                     aria-label={i18n.t("purge_post")}
                   >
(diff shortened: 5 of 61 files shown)


jsit:
@djvs Disclaimer: I'm very new to this project, so I can't speak for the code maintainers (@dessalines and @SleeplessOne1917), but here's my read on the situation:

Because the project is moving so fast and because there will be so many different people working on it, I think sticking with Bootstrap classes is an intentional decision to (a) reduce overhead for pitching in (no learning curve on class naming conventions or an extra layer in the stack) and (b) prevent inconsistencies in approach or naming.

(I have a general distaste for Bootstrap, too, but I think its use may be warranted in this case.)

Regarding the issue of elements not being easily targetable for custom styling, I think the general approach right now is, "If it can't be done with Bootstrap, don't do it (for now)." Much of Bootstrap can be customized with theming (there are other efforts going on there AFAIK).

See the conversation in #863

---

It might be a good idea to update the Contributing guide to discuss this.

Zetaphor:
I think there's a worthwhile middle ground to explore here. I understand the point that @dessalines is making in that he doesn't want to bind the base themes to custom classes because the core of lemmy-ui should be able to painlessly adopt any Bootstrap theme.

However I also agree with everyone else in the community who says that doing custom theming, which is a valid and clearly highly desirable goal, is made unnecessarily difficult by the current lack of semantics in the markup.

I don't see why it would not be a viable path forward to insist that the base Lemmy themes be only based on Bootstrap standard classes, and that any markup added or changed adhere to the Bootstrap structure and styling, _while also_ providing hooks for users who want to go beyond Bootstrap. This includes far more than just theme developers, but also people making userscripts and browser extensions.

If the goal here is to support user choice first (be it end-user or instance host), while also keeping the codebase clean and accessible, I think there's a fairly easy to reach line that accomplishes both of those goals.

We can have our stock experience that can seamlessly slot in a Bootstrap theme for those who just want to ship the defaults with maybe some light color swaps, while also enabling an ecosystem of users adding increased value and customization to either their individual or hosted experience.

ayan4m1:
Agreed, I have no idea where the naming conventions for these classes lives/came from but it needs to be much better documented if it's going to be merged in imho. The class names in this PR are confusing, sometimes abbreviated sometimes not, undocumented suffixes, does not seem to adhere to BEM or any other naming convention I'm familiar with.

Zetaphor:
> Agreed, I have no idea where the naming conventions for these classes lives/came from but it needs to be much better documented if it's going to be merged in imho. The class names in this PR are confusing, sometimes abbreviated sometimes not, undocumented suffixes, does not seem to adhere to BEM or any other naming convention I'm familiar with.

You wrote this seconds after I posted my response. I also completely agree with this post. If we are going to extend courtesy to customization it should be through sane and sensible naming conventions that align well with the existing Bootstrap standards.

djvs:
Well, in general I would advise against using Bootstrap in the first place.   That creates a baked-in opinionated approach to layout which limits themes' ability to handle responsive design.

 That aside, the current reliance on Bootstrap classes, with virtually no other classes anywhere else in the codebase, as Klobenda mentions in the other thread, is completely hostile to any third party themes.  People inevitably are forced to create 6-element-deep selectors to target specific elements which will break suddenly and non-debuggably with any change to the components.  The rule of thumb after my PR here is to try to limit necessary selectors to 2-3 levels deep if possible.

The naming convention I used is basically just "create unique names and try to namespace them".  BEM tends to be extremely verbose and unpleasant to work with, and unnecessarily confines elements to contexts that may be more harmful than useful.   All I did was basically namespace particular elements with something close to the name of the wrapping component, detail rows, buttons, li's etc per name to make them individually addressable in importance places, and the outermost component (or something very likely to be branched off) is prefaced with "cnt-".  "-w" is a similar suffix for wrapping components inside one of the containing/component-level ones.  I'm happy to change those, I imagine the "cnt-" one is least palatable.  Keep in mind that the length of class names increases payload size in general, so there's a balance here between specificity/verbosity and conciseness.

Zetaphor:
> Well, in general I would advise against using Bootstrap in the first place. That creates a baked-in opinionated approach to layout which limits themes' ability to handle responsive design.

I also agree with this, but it's not my place to drive that large of a change.

> That aside, the current reliance on Bootstrap classes, with virtually no other classes anywhere else in the codebase, as Klobenda mentions in the other thread, is completely hostile to any third party themes. People inevitably are forced to create 6-element-deep selectors to target specific elements which will break suddenly and non-debuggably with any change to the components.

This is something I tried to help with with the changes I made in #1327, but I went for a much sparser approach trying to only focus on major areas rather than tagging as thoroughly. I do hope we can reach a consensus on this with the maintainers.

>  I'm happy to change those, I imagine the "cnt-" one is least palatable.

I did read this in my head as c*nt 😅

djvs:
Well, I mean more that cnt- is a prefix, although it is done that way to make it explicit it's denoting a section.  It's short for "container".  "w" for wrap, "err" for error, etc..

jsit:
> Well, in general I would advise against using Bootstrap in the first place. That creates a baked-in opinionated approach to layout which limits themes' ability to handle responsive design.

The project is already heavily using Bootstrap, so dropping it will require quite a bit of planning. I agree that relying on Bootstrap can be problematic but it's what we've got.

Personally I think BEM can be hugely advantageous, especially in a codebase with a lot of people working on it.

Probably a conversation to be had elsewhere. Would it be useful to have GitHub Discussions set up on this project? @dessalines @SleeplessOne1917

djvs:
Well, I would use BEM here if it would get this merged.  This was quite a bit of typing and hopefully not in vain.

jsit:
> Well, I would use BEM here if it would get this merged. This was quite a bit of typing and hopefully not in vain.

I think more is needed than just converting these classes to BEM. A clear plan would probably have to be laid out.

jsit:
> This was quite a bit of typing and hopefully not in vain.

I think it would be a good idea to check with the maintainers to make sure something like this is desired before putting this much work into it.

djvs:
Resubmitted as BEM in the interest of confirming to some actual existing standard

jsit:
@djvs You seem to be using double dashes for both modifier syntax and child syntax? There are several BEM naming conventions but I'm not familiar with this one.

https://en.bem.info/methodology/naming-convention/

djvs:
Whoops, yeah I don't use BEM much these days.  Fixing.

djvs:
:+1:

jsit:
@djvs So, for something like this:

Screenshot 2023-06-17 at 8 00 40 PM

It looks like these three elements are probably styled very similar, meaning their classes should be modifiers of the same class.

So instead of:

```
className="listing-type-select__label-subscribed"
className="listing-type-select__label-local"
className="listing-type-select__label-show-all"
```

```
className="listing-type-select__label listing-type-select__label--subscribed"
className="listing-type-select__label listing-type-select__label--local"
className="listing-type-select__label listing-type-select__label--show-all"
```

jsit:
It's these kinds of things that make me think any efforts here need to be part of some larger planning, not hashed out in the comments of a big PR.

djvs:
That's a fair point, I'll address that now.

jsit:
I think if and when this project switches to semantic class names, it should be done on a component-by-component basis, not as one big PR that adds a ton of classes that aren't getting styled yet.

djvs:
I don't think being explicitly styled is a prerequisite to adding classes to a FE that's intended to support themes?  Even if so, it'd be better to at least get something down for themers to work off of.

djvs:
Anyway, probably gonna quit making changes for the moment until maintainers chime in, I feel like this approach is pretty reasonable atm.

jsit:
> it'd be better to at least get something down for themers to work off of.

I may be wrong but I don’t think user themes is a high priority right now. The UI needs a lot more work than just the way it looks. Ultimately the main benefit of using semantic class names is maintainability, not for user stylesheets to latch onto.

alectrocute:
I applaud the effort but unfortunately I think it may be a bit misguided. I’m not a fan of it. My two cents: If we want to do something like this, we should be using `id`’s. That way, our `classNames` can be separate and it can also be user style friendly.

djvs:
IDs only for unique elements (`getElementById` vs `getElementsByClassName`), keeping it broadly generic allows element reuse.  There may be a few exceptions worth isolating (e.g. `#app` that's already there), but generally very little benefit from confining them.   CSS allows both to be referenced for selectors of course.

jsit:
@djvs @alectrocute @Zetaphor I've written some more about a possible CSS plan in #1371

djvs:
These approaches are not incompatible so I don't know why you seem to be going on the offensive against this PR.  The purpose here is to solve an existing problem, the fact that the existing UI, which I doubt is going to be completely rewritten in the next week, is not themeable or addressable with user scripts.

jsit:
@djvs I just think this PR is too big without a clear vision for a larger CSS approach. If all these class names are added, we may eventually discover that there's something about them that doesn't quite work, and by the time that happens, too many styles will be tied to them and then extricating the codebase from these class names will be another difficulty on top of extricating it from Bootstrap.

On top of which, adding classnames that have no associated styles feels a lot like premature optimization to me.

It seems like this PR Is mainly designed to address custom/user themes, at the expense of possibly littering the codebase with extraneous and untested classnames that might contribute to further technical debt for the maintainers.

djvs:
There is a clear vision, component-level BEM naming for major components and sections thereof to allow external user addressability for scripts and stylesheets.  The BEM naming was *your* suggestion.  There is currently nothing to anchor styles onto besides ~6 classed elements and sequences of element names, which would be destroyed anyway by any change in UI.  It's on theme maintainers to address changes in the UI, but the UI maintainers shouldn't make that pointlessly difficult on them.

This PR was created because I authored some user styles (very painfully I would add), other users expressed that it should be added into lemmy-ui, and I already knew this was not meaningfully possible without making elements addressable.  This PR makes major components and sections thereof addressable.

djvs:
And I don't think there's a problem of "technical debt" or "class names not working".  Classes are just arbitrary strings. If you need to restructure a component, themers would need to redo styles there anyway, you can just delete them.

djvs:
If you have specific concerns with the classes added (for example the --root elements, perhaps those could be de-suffixed and consolidated), or would like to compile styles against it, feel free to contribute to the PR.

misnina:
When you're talking about how naming them now will break later when UI structural changes happen, it's going to happen already. Anything theme we make right now will break soon because the UI needs various work, but I feel that shouldn't stop people from trying to make meaningful changes now. They won't be perfect because everything is progressing at different rates, but I don't think it should hinder implementing the first steps.

I agree with @djvs about the technical debt not being a big problem, it's just names. For people that want to theme, dealing with classname and id labels that change are trivial. It's harder when the structure changes, which it will regardless of this PR or not, so themes and plugins will have to be updated anyways. The work is already here to begin changes and make access to elements easier.

The discussion here would be naming conventions, but I personally dislike BEM (but understand the reasoning) so I can't contribute to that. I will happily use any option to select specific elements because it's very painful right now, and I imagine more so for user scripts.

dessalines:
@jsit I've enabled github discussions, although I'm not the biggest fan of it, considering if we move to a new issue tracker at some point, I'm sure all of that would be lost. I'd much rather have a discussion tag and a new issue.



To outline my thoughts:

- Short term: I spose I'm fine with adding something like this, with the **very bold caveat** that we as maintainers **will not do custom theme support**, and if class names change, and themes break, that is up to themers, not us. I've had to close so many issues on this repo for people asking me to do theming support for them, which I don't have time to do.
  - Also go over these new classnames again and make sure they're consistently named.
- Long term: Completely rewrite lemmy-ui in [leptos-rs](https://github.com/leptos-rs/leptos), with DaisyUI and Tailwind for the components. Say goodbye to this isomorphic javascript frankensteins monster I've created here with inferno and bootstrap. A few of us, including me, @SleeplessOne1917 , and @alectrocute have expressed interest in this, for all the benefits of rust, native isomorphic support, as well as being able to rid ourselves of bootstrap, and switch to tailwind, which has more fine-grained control. When I wrote lemmy-ui, these frameworks didn't exist or weren't as popular as bootstrap.
  - This will probably have even less customizations, at least outside of DaisyUI's supported themes.

djvs:
OK, good to go if nobody spots anything else.

dessalines (approved):
Seems fine to me, but I'll let @SleeplessOne1917 , @jsit and @alectrocute take a pass too.

jsit:
@dessalines 

> I've enabled github discussions, although I'm not the biggest fan of it

Oh if that's you're feeling then, I don't want to add more busy work to the maintenance of this repo. I'm happy to keep discussions in `Discussion` issues.

alectrocute:
I’ll have to look this over later for actual review. My immediate thought is that adding such classes is a reasonable thing, but I really dislike the BEM practice. The four dashes really ick me out. This is my personal opinion though so take it with a grain of salt. I appreciate all contributions and work volunteering to make this software great and I feel the intention is really good here.

jsit:
Something else I kind of worry about with this is that PRs might start coming in with modifications to `main.css` that modify the styles on these classes.

src/shared/components/comment/comment-report.tsx
```diff
@@ -101,14 +101,14 @@ export class CommentReport extends Component<
...
+        <div className="comment-report__person">
           {i18n.t("reporter")}: <PersonListing person={r.creator} />
         </div>
-        <div>
+        <div className="comment-report__reason">
           {i18n.t("reason")}: {r.comment_report.reason}
         </div>
         {r.resolver && (
-          <div>
+          <div className="comment-report__resolverwrap">
```
alectrocute, line 111: `resolverwrap` can just be `resolver`.

src/shared/components/common/language-select.tsx
```diff
@@ -51,13 +51,14 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
     ) : (
       <div>
         {this.props.multiple && this.props.showLanguageWarning && (
-          <div className="alert alert-warning" role="alert">
+          <div className="language-select alert alert-warning" role="alert">
```
alectrocute, line 54: Why would this alert `div` be `language-select`? It should probably be `language-select-alert`.

src/shared/components/common/listing-type-select.tsx
```diff
@@ -39,11 +39,11 @@ export class ListingTypeSelect extends Component<
 
   render() {
     return (
-      <div className="btn-group btn-group-toggle flex-wrap mb-2">
+      <div className="listing-type-select btn-group btn-group-toggle flex-wrap mb-2">
         {this.props.showSubscribed && (
           <label
             title={i18n.t("subscribed_description")}
-            className={`btn btn-outline-secondary 
+            className={`listing-type-select__label listing-type-select__label----subscribed btn btn-outline-secondary
```
alectrocute, line 46: What's the purpose of the `----`?

src/shared/components/common/pictrs-image.tsx
```diff
@@ -22,7 +22,7 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
 
   render() {
     return (
-      <picture className="d-inline-block overflow-hidden">
+      <picture className="pictrs-image d-inline-block overflow-hidden">
```
alectrocute, line 25: This `picture` element should be easily queried, no? Why does it need an additional class? This is the only use of a `picture` element.

src/shared/components/post/post-listing.tsx
```diff
@@ -1153,20 +1153,20 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
             ) : (
               <>
                 <button
-                  className="d-inline-block mr-1 btn btn-link btn-animate text-muted py-0"
+                  className="post-listing__btn post-listing__btn--transfc-confirm d-inline-block mr-1 btn btn-link btn-animate text-muted py-0"
```
alectrocute, line 1156: `transfc` = transfer community? The class name's already so long, I see no reason to *not* just write the rest of the words.

src/shared/components/post/post-listing.tsx
```diff
@@ -1153,20 +1153,20 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
-                  className="btn btn-link btn-animate text-muted py-0 d-inline-block mr-1"
+                  className="post-listing__btn post-listing__btn--htransfc btn btn-link btn-animate text-muted py-0 d-inline-block mr-1"
                   aria-label={i18n.t("yes")}
                   onClick={linkEvent(this, this.handleTransferCommunity)}
                 >
                   {this.state.transferLoading ? <Spinner /> : i18n.t("yes")}
                 </button>
                 <button
-                  className="btn btn-link btn-animate text-muted py-0 d-inline-block"
+                  className="post-listing__btn post-listing__btn--cshtransfc btn btn-link btn-animate text-muted py-0 d-inline-block"
```
alectrocute, line 1169: `cshtransfc`?

src/shared/components/post/post-listing.tsx
jsit: This is what I mean about planning. Will this specific button really be styled much differently from other buttons? Does it need a modifier?

All these names are chosen with no intention other than making sure everything has a class name on it. These are extremely likely to change, breaking whatever user themes are created in the meantime. Then you have a bunch of user themers getting upset at the project for "breaking" their themes.

This is not what CSS is for.

jsit: @alectrocute It's extremely messy and doesn't serve the larger goals of the project, these are only serving as sledgehammer hooks for custom theming. Obviously this isn't my project, but I just think this whole thing is greatly misguided.

alectrocute (changes requested):
Objectively looked through it and called out any potential issues. I think we need to have further discussion about this.

Mixing BEM and Bootstrap is kinda an abomination. I've never seen anything like it. This will have consequences on the project. As @jsit aptly put, contributors and maintainers will have to follow this practice for every new element we add/change, and will be responsible for it in perpetuity.

src/shared/components/post/post-listing.tsx
alectrocute: I agree with you on all counts, @jsit.

dessalines: @djvs It seems like the consensus is that this is not a good idea, and theme designers are going to come complaining to us every time their themes break when we change things around.

Is it so bad to generate bootstrap-compatible themes for site customization?

djvs: That's the product of two constraints:
  * Somebody wanting to style that particular button
  * Using BEM 

And sort of additionally:
* jsit's earlier request to have generic and then modifier substrings for stuff like buttons - `post-listing__btn post-listing__btn--transfc-confirm".  Specifically, it's:

Block: post-listing

Element: btn

Modifier: transfc-confirm

The verbosity is my problem with BEM, but without it you are indeed stuck with "confirm-transfc" or whatever.  That was basically my initial approach with slightly more namespacing.

src/shared/components/post/post-listing.tsx
djvs: Named after existing method contained within -  "handle cancel show confirm transfer community". I guess that would be "hcsctransfc" or "csctransfc".  Can do "--cancel-transfc".

src/shared/components/common/pictrs-image.tsx
djvs: It may not remain that so.  It's contextualized to the component, which is the convention I used.

src/shared/components/common/listing-type-select.tsx
djvs: Typo

src/shared/components/common/language-select.tsx
djvs: Belonged on parent element, fixed.

src/shared/components/comment/comment-report.tsx
djvs: Fixed

djvs:
Well, not my decision to make, but classes/IDs are basically the only way for user styles/scripts to work, and BEM happens to be the prevailing standard for class naming (I'm not a huge fan of it either - you could commit to some modified BEM without double dashes/underscores, but that might be even more trouble).  And interest was expressed by 2+ of you in abandoning Bootstrap long-term, in which case this is basically just a step towards that.

djvs:
P.S. - the kind of bare minimum for people abiding by this standard is "just add a meaningful class name at the root of a component".  The additional ones I added were just in the interest of being thorough.

SleeplessOne1917:
Sorry for taking awhile to review this. I agree with @jsit saying this is a lot of premature optimization. I would be more for it if it wasn't for the fact that contributors will have to come up with these sorts of class names whenever something is added.

jsit:
In other words, these class names have no actual relationship to how they’re actually going to be used. They’re just there to be there.

it’s not “a start” because they’ll likely be abandoned when the codebase switches to Tailwind anyway.

**AND they will be hard to remove**. There are hundreds of new class names here, and none of them has *any* styling applied to them. To get rid of any one of them, you’d have to *know* that no styles depend on their being there.

jsit:
Adding class names without associated styles, just so that they *might* be custom-styled, is a really good way to pollute your codebase with all kinds of unnecessary markup. It’s asking for trouble.

jsit:
“Does this class do anything” shouldn’t be a hard question to answer when looking at a codebase. Here, that will be unclear in hundreds of new instances.

djvs:
OK - can we reach a compromise like "one unique class name per actual component"?   I could go either way on BEM, you guys seem split on that.

djvs:
You can use grep to check if/when a CSS class is used btw.

djvs:
There might be a bit of "missing the forest for the trees" going on here.  At the end of the day the goal is for Lemmy to be something lots of people want to use.  For it to be that, it has to look good and be intuitive.  That won't be one-size-fits-all in the first place - different visual preferences, devices, browsers, disabilities, you name it.  The goal of CSS classes in a context like that isn't simply for the convenience of the developers, it's to allow the UI to support customization.  Reddit handled this correctly for years - old.reddit.com was well marked up and styleable (allowing subreddit/community-level styling as well!), and the RES extension took full advantage of this as well.   IMO the cost of "having a few classes lying around that don't have associated styles in the main source tree" isn't even on the same order of magnitude.

djvs:
Github apparently lets me merge this but no idea what maintainer hierarchy is.  I would not like to maintain this indefinitely as a downstream branch, I'm already past the third set of merge conflict resolves.  If there's a workable solution here please let me know, I have a lot on my plate right now.

djvs:
OK, I won't be able to maintain this branch separately long-term.  Can we do a middle of the road compromise like "one meaningful class name per component"?

djvs:
Simpler version here:

https://github.com/LemmyNet/lemmy-ui/pull/1421

PR #1362, opened 2023-06-17
feat: Move advanced post menu into dropdown

jsit:
Rather than having all those "advanced"/"more" buttons for a post in a horizontally-expanding row (which causes content to move around and causes the row to wrap, making the context of the buttons confusing), I've here tucked them away into a dropdown menu.

## Before
Screenshot 2023-06-17 at 6 05 25 PM
Screenshot 2023-06-17 at 6 05 17 PM

## After

Screenshot 2023-06-17 at 6 03 33 PM
Screenshot 2023-06-17 at 6 02 35 PM

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -873,7 +873,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   get reportButton() {
     return (
       <button
-        className="btn btn-link d-flex align-items-center rounded-0 dropdown-item"
+        className="btn btn-link btn-sm d-flex align-items-center rounded-0 dropdown-item"
         onClick={linkEvent(this, this.handleShowReportDialog)}
         data-tippy-content={i18n.t("show_report_dialog")}
         aria-label={i18n.t("show_report_dialog")}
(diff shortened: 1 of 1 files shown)


Zetaphor:
Looks like the localization needs some capitalization for delete and lock

jsit:
@Zetaphor Thanks, I have that in a PR at LemmyNet/lemmy-translations#66

alectrocute:
Looks great! Is it possible to make the padding on the X axis match the padding on the Y? The hover background directly touches the side of the element with rounded borders, it’s a little wonky. Other than that, plus the capitalization mentioned by @SleeplessOne1917, this is a really good change I can’t wait to use!

jsit:
@alectrocute I have:

1. Removed the border-radius from the dropdown menu buttons
2. Vertically centered the icon and text inside the buttons

I'm just using the native padding that comes from the `.btn` class, and personally I'd rather stick with the padding Bootstrap gives us for buttons here.

In fact I think they will work better as `.btn-sm`, so I've added that class too. Let me know what you think of this:

Screenshot 2023-06-17 at 9 06 05 PM

alectrocute:
Looks great @jsit!

jsit:
Upon looking closer I see that translations are handled through Weblate, so I'm not sure that PR of mine over there will do any good. Would love to get this merged in even with the lowercase words since it shifts things around quite a bit and don't want to run into a merge conflict later.

SleeplessOne1917:
@alectrocute it was @Zetaphor who pointed out the caps thing lol.

SleeplessOne1917 (approved):

Issue #1142, opened 2023-06-10
Currently clicking on the thumbnail acts as an expando for images, but for links, it opens the link


Kommynct:
This is problematic because as i'm scrolling through, it's sometimes annoying to have the page i'm on replaced when i'm expecting an expando to expand, I think there's two possible solutions:

1. Move expandos somewhere else
2. make left clicking thumbnails open in a new tab

Thoughts?

krestenlaust:
I vote option 2.

- The pictogram used for external links indicate opening in a new tab (imo)
- I can't remember when I last encountered a popular site open external links in the same tab (somewhat breaks the expected rules of the internet)

krestenlaust:
Issue already exists #634

mwalser:
I agree that the current behavior is really confusing. While there are icons to discern an image from a link post, its easy to miss.

My suggestion would be to make two changes:
- Clicking an image always navigates to the image or article URL and never opens an expando
- For image posts, introduce a button to open an expando

That way its consistent:
- Clicking on the image always opens the source URL
- Clicking on the title always opens the comments
- Clicking on the "+" icon always opens an expando

Kommynct:
That wouldn't encourage me to participate, that would just be annoying, but if that argument can be made, it should be up to the user, or the instance.

krestenlaust:
> That wouldn't encourage me to participate, that would just be annoying, but if that argument can be made, it should be up to the user, or the instance.

Then I'd rather back up a bit, and just forget about that. I don't believe that making a new UI setting every time we don't completely agree, is the right call.

doctorrmcb:
I would use the site more if it had expandos for text and other posts. Any time you take me to another page it is really annoying. I want the primary viewing experience to be how it is in RES, where I click through expandos to get content quickly and if I'm interested in a discussion then I will open the comments. There is way too much loading when I have to open other pages to view content and going back is painful.

krestenlaust:
Agree. The way you have to navigate backwards is a real nuisance currently. I try to use the context menu or Ctrl+Left arrow, but that's just a short-term solution.

Isn't it possible to expand text as it is now anyway?

doctorrmcb:
If it is possible it isn't apparent to me how to do it. I found the image functionality easy enough to find, but after 30 minutes of using Lemmy, I got frustrated and gave up using it. Primarily due to another bug, but the inconsistency of where to click to expand things would definitely be another reason.

[...]

Issue #634, opened 2022-04-25
Toggle to always open external link in a new tab 


DMTryptamines:
I always end up leaving lemmy and losing my tabs because by default clicking an external link opens it in the same tab you're currently using. When I'm done reading articles I usually just close them right away which also closes my lemmy tab! Then I have to reopen the site and find the post again so I can comment or upvote. This would be a mega pain if I didn't just leave myself logged in on the browser. 

What I am proposing is a toggle in your personal settings (or instance config maybe?) that changes the default behaviour to always opening a new tab when clicking an external link. A step further maybe this could also do the same when clicking a post, user or community in lemmy.

Not a high urgency feature request for sure but I do think enabling customization of user workflow would go a long way in user satisfaction and making the transition easier.

dessalines:
Any reason why you can't use middle-click / right click-open in new tab?

We will never have auto-open-in-new-tab on by default, because it breaks accessibility, and because that behavior is unpredictable for most sites. 

If you'd like to submit a PR for adding this setting, I wouldn't be opposed to it.

DMTryptamines:
I'm just hearing about middle click from you and that works pretty well actually, neato :) I usually right click or hold ctrl and click which accomplishes the same but they are both 2 actions as opposed to one but your middle click suggestion actually solves that. 

As said it's pretty low priority so I'm all good parking it for now, I use this setting on Reddit so I always fumble the transition and makes the experience a little annoying.

Nutomic:
Duplicate of #282.

PR #1363, opened 2023-06-18
feat(post): Move post domain beneath post title

jsit:
Some design work here, trying to tidy up all the metadata grouped together in the post details.

I've moved the link domain and the "expand" button to beneath the post title in small text. I think this makes it easier to parse the information.

## Before

Screenshot 2023-06-18 at 10 37 24 AM
Screenshot 2023-06-18 at 10 37 19 AM

## After
Screenshot 2023-06-18 at 2 33 04 PM
Screenshot 2023-06-18 at 2 33 12 PM

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -84,10 +84,6 @@
   margin-top: -6.5px;
 }
 
-.post-title {
-  line-height: 1;
-}
-
 .post-title a:visited {
   color: var(--gray) !important;
 }
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -367,8 +367,6 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
 
   createdLine() {
     const post_view = this.postView;
-    const url = post_view.post.url;
-    const body = post_view.post.body;
     return (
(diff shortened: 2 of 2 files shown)


glibg10b:
It would be more intuitive if the preview button's behavior swapped with the thumbnail's behavior:
- The button next to the domain would link to the specific page at the domain (and have the appropriate icon)
- The post's thumbnail would then show the preview when clicked, which would make it consistent with image posts, which already have that behavior

This would solve #1142 and #634

If you don't think this should be addressed in this PR, I will start a new issue

jsit:
@glibg10b Hm yeah I think this is a change in behavior that's outside the scope of this PR. Thank you though!

Though I will say that link posts that have thumbnails have *both* (a) expandable images and (b) expandable text. This is from beehaw.org which I think runs 0.17.4:

Screenshot 2023-06-18 at 1 39 04 PM

src/assets/css/main.css
```diff
@@ -84,10 +84,6 @@
   margin-top: -6.5px;
 }
 
-.post-title {
-  line-height: 1;
-}
```
dessalines, line 89: Thanks. Every bit of custom css we can remove is a win.

dessalines (approved):
Looks much better, thanks.

PR #1366, opened 2023-06-18
Don't hide "show NSFW" checkbox on instances that have NSFW disabled

sunaurus:
Instance admins can disable NSFW submissions for users on their instances. Currently, this option also removes the "show_nsfw" setting for all users, however, this doesn't really make sense - users are still able to subscribe to NSFW communities on other instances, and being able to control whether they see the posts there should be up to each user individually.


SleeplessOne1917 (approved):
Makes sense to me.

PR #1367, opened 2023-06-18
feat: provide a reason when blocking instances

TKilFree:
This allows:
- admins to add a reason when blocking an instance;
- users to see reasons (when provided) at `/instances`

Companion MR to: https://github.com/LemmyNet/lemmy/pull/3168

Should resolve issue: https://github.com/LemmyNet/lemmy/issues/1947

New blocklist in admin interface:

New column in `/instances` table:

Notes:
- I have left in temporary types to allow everything to compile, but presumably when the backend MR is merged this can just consume the published types
- I configured the `no-unused-vars` rule to ignore variables starting with an underscore, so that it's possible to do something like discards (https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards), happy to remove that if it's not wanted though!
- Previously the allowlist/blocklist were rendered generically on the admin page (since they were doing the exact same thing), I couldn't see a nice way to keep this generic handling so explicitly separated them. Would be happy to hear an alternative though!

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -20,6 +20,14 @@
     "@typescript-eslint/no-explicit-any": 0,
     "@typescript-eslint/explicit-module-boundary-types": 0,
     "@typescript-eslint/no-empty-function": 0,
+    "@typescript-eslint/no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_",
+        "caughtErrorsIgnorePattern": "^_"
+      }
+    ],
     "arrow-body-style": 0,
     "curly": 0,
     "eol-last": 0,
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ src/shared/translations
 
 stats.json
 
+#files generated when using yalc to develop against local changes to http client
+.yalc
+yalc.lock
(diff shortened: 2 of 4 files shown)


SleeplessOne1917:
Since the current UI wasn't built with adding reasons in mind, this might need a bit of rework there as well as code. The blocked instance adding part of the UI looks a bit rough at the moment. @jsit @alectrocute @dessalines @Nutomic thoughts?

jsit:
@SleeplessOne1917 That table is restricted in width by a `col-md-6` class. Would be more readable on desktop with a `col-12`, but that wouldn't solve the issue on mobile.

The "reason" text might be ok on its own row, but that would require some robust a11y work.

Maybe a modal or popover?

TKilFree:
> Since the current UI wasn't built with adding reasons in mind, this might need a bit of rework there as well as code. The blocked instance adding part of the UI looks a bit rough at the moment.

yeah that's fair, i was trying to change the existing look & feel as little as possible.

my instinct would be to chuck the blocklist part of the admin page into its own vertical space, and probably redo it so it's a table with editable cells

TKilFree:
i've pushed an update to integrate changes to the backend, updated the UI to the one shown in that mockup and added `aria-label`s to the buttons (associated MR: https://github.com/LemmyNet/lemmy-translations/pull/70)

Nutomic:
This is a breaking change so we should wait a while to merge it, until we are ready to work on 0.19

dessalines:

PR #1376, opened 2023-06-18
fix: Fix too-intense hr color between posts

jsit:
#1304 used the `primary` color for horizontal rules, which looks subtle-ish on dark mode, but less subtle on light mode.

It also increased the border width of the `<hr>`s.

This PR makes the colors more subtle, while still being visible, and modifies the Bootstrap `hr-border-color` variable for more global `<hr>` color fixing.

Would like feedback from @ayan4m1 and @SleeplessOne1917 on this, as they were responsible for the previous fix.

## Before

Screenshot 2023-06-18 at 3 43 29 PM

Screenshot 2023-06-18 at 3 43 20 PM

## After

Screenshot 2023-06-18 at 3 37 25 PM

Screenshot 2023-06-18 at 3 37 36 PM

### DIFF
--- a/src/assets/css/themes/_variables.darkly.scss
+++ b/src/assets/css/themes/_variables.darkly.scss
@@ -92,6 +92,8 @@ $input-border-color: $body-bg;
 $input-group-addon-color: $gray-500;
 $input-group-addon-bg: $gray-700;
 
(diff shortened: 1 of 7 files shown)


SleeplessOne1917 (approved):
LGTM

PR #1378, opened 2023-06-18
feat: Bootstrap 5

jsit:
Fully remove Bootstrap 4 dependency in favor of Bootstrap 5

I should note that this changes the main green color in the Litely theme; Bootstrap 5 automatically generates foreground colors for given background colors, including with a set minimum contrast level.

I set the minimum contrast level to 3 (which passes [WCAG AA](https://webaim.org/resources/contrastchecker/) for "large" text) -- the Bootstrap 5 default is a contrast level of 4.5, which would require a much darker green to handle. (The current white-on-green is only 2.23.)

Should probably offer a high-contrast theme that gets automatically selected for the `prefers-contrast: more` media query.

cc @fheft

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -248,7 +248,7 @@ export class MarkdownTextArea extends Component<
                     />
                   )}
               </div>
-              <label className="sr-only" htmlFor={this.id}>
+              <label className="visually-hidden" htmlFor={this.id}>
                 {i18n.t("body")}
               </label>
             </div>
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -146,8 +146,8 @@ export class MarkdownTextArea extends Component<
         />
         <div className="form-group row">
           <div className="col-12">
-            <div className="rounded bg-light border border-light">
-              <div className="d-flex flex-wrap border-bottom border-light">
+            <div className="rounded bg-light border border-secondary">
+              <div className="d-flex flex-wrap border-bottom border-secondary">
                 {this.getFormatButton("bold", this.handleInsertBold)}
                 {this.getFormatButton("italic", this.handleInsertItalic)}
                 {this.getFormatButton("link", this.handleInsertLink)}
@@ -274,7 +274,7 @@ export class MarkdownTextArea extends Component<
             {this.props.buttonTitle && (
               <button
                 type="submit"
-                className="btn btn-sm btn-secondary ml-2"
+                className="btn btn-sm btn-secondary ms-2"
                 disabled={this.isDisabled}
               >
                 {this.state.loading ? (
@@ -287,7 +287,7 @@ export class MarkdownTextArea extends Component<
             {this.props.replyType && (
               <button
                 type="button"
-                className="btn btn-sm btn-secondary ml-2"
+                className="btn btn-sm btn-secondary ms-2"
                 onClick={linkEvent(this, this.handleReplyCancel)}
               >
                 {i18n.t("cancel")}
@@ -295,7 +295,7 @@ export class MarkdownTextArea extends Component<
             )}
             {this.state.content && (
               <button
-                className={`btn btn-sm btn-secondary ml-2 ${
+                className={`btn btn-sm btn-secondary ms-2 ${
                   this.state.previewMode && "active"
                 }`}
                 onClick={linkEvent(this, this.handlePreviewToggle)}
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -144,7 +144,7 @@ export class MarkdownTextArea extends Component<
             !this.state.submitted
           }
         />
-        <div className="input-group mb-3 row">
+        <div className="mb-3 row">
           <div className="col-12">
             <div className="rounded bg-light border">
               <div className="d-flex flex-wrap border-bottom">
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -146,8 +146,8 @@ export class MarkdownTextArea extends Component<
         />
         <div className="form-group row">
           <div className="col-12">
-            <div className="rounded bg-light border border-secondary">
-              <div className="d-flex flex-wrap border-bottom border-secondary">
+            <div className="rounded bg-light border">
+              <div className="d-flex flex-wrap border-bottom">
                 {this.getFormatButton("bold", this.handleInsertBold)}
                 {this.getFormatButton("italic", this.handleInsertItalic)}
                 {this.getFormatButton("link", this.handleInsertLink)}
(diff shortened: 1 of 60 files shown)


ayan4m1:
LGTM, thanks for this! 🚀

fheft:
Couldn't try it out yet, but code lgtm. Thanks for doing this! :rocket:

alectrocute:
I spent some time testing out this branch. Great work as always @jsit, just a few things that may or may not be related to these changes, but nonetheless, would be nice to have addressed –

- Error toasters have no background:

Screenshot 2023-06-19 at 8 13 28 AM

- 'Delete Account' button has no background on darkly red:

Screenshot 2023-06-19 at 8 14 33 AM

- I wish we could fit all these on the same row and avoid wrap:

Screenshot 2023-06-19 at 8 15 22 AM

- Native radio buttons on these shouldn't be visible:

Screenshot 2023-06-19 at 8 16 13 AM

- Same dealio here:

Screenshot 2023-06-19 at 8 16 34 AM

- Where thou `mr-2`:

Screenshot 2023-06-19 at 8 16 51 AM

- We can probably fit this on the same row:

Screenshot 2023-06-19 at 8 17 38 AM

alectrocute:
One more thing. When I run `yarn themes:build`, I get this error (and CSS) output:

Screenshot 2023-06-19 at 8 24 36 AM

Also, I'm wondering if we need to have `themes:build` ran upon `postinstall`? Are we committing the dist outputs instead?

jsit:
@alectrocute Thank you for the thorough review! Those issues should be addressed and I'll try to get to them this afternoon.

jsit:
@alectrocute All these issues should now be fixed.

fheft:
@jsit alectrocute and you did a great job, I haven't found any issues :slightly_smiling_face:

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -144,10 +144,10 @@ export class MarkdownTextArea extends Component<
             !this.state.submitted
           }
         />
-        <div className="form-group row">
+        <div className="input-group mb-3 row">
```
alectrocute, line 147: `input-group` has some side effects.

With `input-group`:

Screenshot 2023-06-19 at 5 50 24 PM

Without:

Screenshot 2023-06-19 at 5 50 15 PM

Note how this `MarkdownTextarea` doesn't scale to the full width of the column.

alectrocute: I also seriously wish we could get all those buttons on one line. Maybe we could put the help icon next to the label instead? 🤔

alectrocute:
One more thing – Noticed some missing right margin on various items on `Modlog`:

Screenshot 2023-06-19 at 5 55 17 PM

Anyway, pretty absurd how much better everything is looking/feeling compared to 0.17. Great work. Keep going!

src/shared/components/common/markdown-textarea.tsx
alectrocute: If we could fit this all on one row, it would be _ perfection _.

alectrocute:
Tabs have a little bit of bottom border radius wonkiness:

Screenshot 2023-06-19 at 5 58 12 PM

alectrocute:
Did this table background color change? It would be nice to have the background alternate light-dark, or have no background on the table rows at all (eg. how it's done in 0.17).

Screenshot 2023-06-19 at 6 00 22 PM

alectrocute:
At some point, our search form became a lot more squished in the corner. Would be nice to use all available space.

Screenshot 2023-06-19 at 6 03 14 PM

alectrocute:
Low priority, but the tagline items in the admin panel have a strange hover background:

https://github.com/LemmyNet/lemmy-ui/assets/35377827/e02a1878-eecd-4de2-966b-c1054d13e43b

jsit:
> Tabs have a little bit of bottom border radius wonkiness:

@alectrocute These already look this way, and are fixed in #1382.

Thank you for all the feedback! Even in `main` there are some weird visual quirks, so it would be helpful if only regressions were pointed out here.

alectrocute:
Please ignore any feedback that isn't related to/shouldn't be addressed in this PR. Thanks @jsit, good work as always!

src/shared/components/common/markdown-textarea.tsx
jsit: Ah, I think this is because I have some instances of `input-group` and `row` classes on the same element; it seems not to like that.

dessalines:
Thanks a ton for doing this! This should have some priority as the merge conflicts will start to pile up if we do too much work in other areas. 

@SleeplessOne1917 Do you think this should come before or after `0.18.0`?

SleeplessOne1917 (approved):

SleeplessOne1917:
@dessalines I think before 0.18 would be better.

jsit:
@dessalines @SleeplessOne1917 I'll work on getting these final issues sorted out. Thanks!

jsit:
@alectrocute 

> Low priority, but the tagline items in the admin panel have a strange hover background:

`main` does this too, not a regression.

> At some point, our search form became a lot more squished in the corner. Would be nice to use all available space.

Fixed.

> Tabs have a little bit of bottom border radius wonkiness:

This is in `main`; will be fixed in #1382.

> Did this table background color change?

Fixed.

> One more thing – Noticed some missing right margin on various items on Modlog:

Fixed.

> `input-group` has some side effects.

Fixed.

src/shared/components/common/markdown-textarea.tsx
jsit: > If we could fit all these icons all on one row, and have the textarea properly fit the space, it would be _ perfection 

Also remember, even if it fit, it would barely fit, so for anybody who increased their browser zoom or were on mobile, it would still wrap. It's a tight fit. I think it wraps kind of nicely, it doesn't bother me too too much.

jsit:
> @SleeplessOne1917 Do you think this should come before or after `0.18.0`?

Personally I think it might be better to wait. This PR is pretty young for its size, 0.18 is already in RC status, and I’d hate to saddle any instances with any undetected regressions this might cause until the next release.

This isn’t solving much for the end user, just some dependency maintenance.

However you’re right about the merge conflicts potentially becoming a big problem, so it should probably be merged very shortly after 0.18 is cut at least.

But if it would be more work to wait, I understand that too.

dessalines:
I'll test this out locally rn.

Once we hit `0.18.0`, we should be able to do much faster releases. These bigger PRs always tend to be blockers, and then fixing the merge conflicts in them quickly becomes a pain, so I agree that it's good to get them in ASAP. 

It will probably be a bit on `0.18.0` anyway, as we're finding some back-end issues that need fixing. So I'm good merging this if most things test out okay.

dessalines (approved):
I tested it, and did a quick pass of some of the components, and everything looks good to me. 

Yall have been making my clunky UI into something so much better :) Thanks a ton.

I agree with sleepless and merge this as is, and any other regressions we find can be fixed in smaller more disectable PRs.

PR #1382, opened 2023-06-18
fix(tabs): Fix tab semantics and a11y

jsit:
Added a lot of semantic/a11y improvements to tabs.

Works for me locally but would be good to test yourself before approving.

https://getbootstrap.com/docs/5.3/components/navs-tabs/#javascript-behavior

### DIFF
--- a/src/shared/components/common/tabs.tsx
+++ b/src/shared/components/common/tabs.tsx
@@ -1,3 +1,4 @@
+import classNames from "classnames";
 import { Component, InfernoNode, linkEvent } from "inferno";
 
 interface TabItem {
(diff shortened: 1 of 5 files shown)


src/shared/components/common/tabs.tsx
```diff
@@ -10,44 +11,42 @@ interface TabsProps {
...
+          {this.props.tabs.map(({ key, label }, index) => (
             <li key={key} className="nav-item">
               <button
                 type="button"
-                className={`nav-link btn${
-                  this.state?.currentTab === key ? " active" : ""
-                }`}
-                onClick={linkEvent({ ctx: this, tab: key }, handleSwitchTab)}
+                data-bs-toggle="tab"
+                data-bs-target={`#${key}-tab-pane`}
```
dessalines, line 28: Not going to work because we don't import bootstrap js.

dessalines: I'd prefer that we don't import that, and instead use the existing inferno handlers for changing tabs. 

I'm all for using the bootstrap styling, but would prefer using our own javascript for event handling.

jsit: Makes sense, thanks! Would still like to get that a11y markup in there so I’ll trim down this PR.

jsit: Also in that case I made a recent change to the collapsing sidebar cards using Bootstrap JS that should be reverted then 😬

dessalines: Thanks. Ya I agree that its pretty, but its a whole nother javascript dep to bring in. I think inferno has some animations now but its not worth it IMO.

SleeplessOne1917 (changes requested):
Marking as request changes until the changes @dessalines requested are made.

src/shared/components/common/tabs.tsx
jsit: Yeah definitely understandable. I've removed the Bootstrap JS dependency but kept the a11y markup.

dessalines (approved):
I just tested and it works well on both. Might be nice if the background tabnames were muted / grayed, but that also might imply they're disabled. That can always be tweaked later.

Screenshot_2023-06-21-08-25-23-159_mark via gp

Thanks!

PR #1383, opened 2023-06-19
feat(css): make vertical rhythm more consistent for post listings

ayan4m1:
Before:

Screenshot 2023-06-18 213706

After:

Screenshot 2023-06-18 213756


SleeplessOne1917:
I'm having a hard time telling the difference between before and after.

ayan4m1:
It's not a huge difference, granted, but in the before, there is slightly more bottom margin than there is top margin, which makes it look kind of imbalanced to me. Also the vertical alignment of "0 Comments" is superior in the After imo.

SleeplessOne1917 (approved):

PR #1386, opened 2023-06-19
Changed some text in feature request template

krestenlaust:
Replaced a sentence referring to a the feature request as a bug report, and changed the title to refer to the UI of lemmy instead of simply lemmy, even though it was also technically correct before.


krestenlaust:
We could consider changing the link in line 9 to a direct link to opening a new issue on LemmyNet/lemmy

SleeplessOne1917 (approved):

PR #1390, opened 2023-06-19
Move password reset form to separate route, view

alectrocute:
Hi Lemdevs!

Our password reset form has not the greatest UX at the moment. It appears like it's a link, but it's actually a button that takes the email field's value and resets the password upon click. There's also critical instructions hidden in a `title` attr. It just doesn't make much sense to me.

I propose we:

- Move the password reset form to a new route.

Screenshot 2023-06-19 at 9 02 05 AM

Screenshot 2023-06-19 at 9 02 07 AM

Translation PR to merge first:

https://github.com/LemmyNet/lemmy-translations/pull/67

Thanks all!

### DIFF
--- a/src/assets/css/themes/_variables.darkly.scss
+++ b/src/assets/css/themes/_variables.darkly.scss
@@ -32,7 +32,6 @@ $yiq-contrasted-threshold: 175;
 $font-family-sans-serif: "Lato", -apple-system, BlinkMacSystemFont, "Segoe UI",
   Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji",
   "Segoe UI Emoji", "Segoe UI Symbol";
-$font-size-base: 0.9375rem;
 $h1-font-size: 3rem;
 $h2-font-size: 2.5rem;
 $h3-font-size: 2rem;
--- a/src/assets/css/themes/_variables.scss
+++ b/src/assets/css/themes/_variables.scss
@@ -1,5 +1,6 @@
 $link-decoration: none;
 $min-contrast-ratio: 3;
+$font-size-base: 0.875rem;
 
 $container-max-widths: (
   lg: 1140px,
--- a/src/assets/css/themes/darkly-compact.css
+++ b/src/assets/css/themes/darkly-compact.css
@@ -115,7 +115,7 @@ hr.my-3 {
   --bs-font-monospace: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
   --bs-gradient: linear-gradient(180deg, rgba(255, 255, 255, 0.15), rgba(255, 255, 255, 0));
   --bs-body-font-family: var(--bs-font-sans-serif);
-  --bs-body-font-size: 0.9375rem;
+  --bs-body-font-size: 0.875rem;
   --bs-body-font-weight: 400;
   --bs-body-line-height: 1.5;
   --bs-body-color: #dee2e6;
@@ -287,20 +287,20 @@ h3, .h3 {
 }
 
 h4, .h4 {
-  font-size: calc(1.265625rem + 0.1875vw);
+  font-size: calc(1.25625rem + 0.075vw);
 }
 @media (min-width: 1200px) {
   h4, .h4 {
-    font-size: 1.40625rem;
+    font-size: 1.3125rem;
   }
 }
 
 h5, .h5 {
-  font-size: 1.171875rem;
+  font-size: 1.09375rem;
 }
 
 h6, .h6 {
-  font-size: 0.9375rem;
+  font-size: 0.875rem;
 }
 
 p {
@@ -309,11 +309,9 @@ p {
 }
 
 abbr[title] {
-  -webkit-text-decoration: underline dotted;
-          text-decoration: underline dotted;
+  text-decoration: underline dotted;
   cursor: help;
-  -webkit-text-decoration-skip-ink: none;
-          text-decoration-skip-ink: none;
+  text-decoration-skip-ink: none;
 }
 
 address {
@@ -628,7 +626,7 @@ progress {
 }
 
 .lead {
-  font-size: 1.171875rem;
+  font-size: 1.09375rem;
   font-weight: 300;
 }
 
@@ -722,7 +720,7 @@ progress {
 
 .blockquote {
   margin-bottom: 1rem;
-  font-size: 1.171875rem;
+  font-size: 1.09375rem;
 }
 .blockquote > :last-child {
   margin-bottom: 0;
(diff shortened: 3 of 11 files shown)


src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+  toast,
+  validEmail,
+} from "../../utils";
+import { HtmlTags } from "../common/html-tags";
+import { Spinner } from "../common/icon";
+
+interface State {
+  form: {
+    username_or_email: string;
+  };
```
alectrocute, line 19: D’oh! I need to make this just `email`.

alectrocute:
Made the changes. Accidentally started tracking [lemmy-translations](https://github.com/LemmyNet/lemmy-ui/pull/1390/files#diff-ad7d50e2f9ff7ef93f8fd4f3fb46795abfd26558f7247f32942382b7455942de). How do I remove this from my PR? Also, should we git ignore this?

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  render() {
+    return (
+      <div className="container-lg">
+        <HtmlTags
+          title={this.documentTitle}
+          path={this.context.router.route.match.url}
+        />
+        <div className="row">
```
SleeplessOne1917, line 62: If there's only going to be one row, why not make the parent div have a className of `container-lg row` and get rid of this element?

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+            {this.loginResetForm()}
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  loginResetForm() {
+    return (
+      <div>
```
SleeplessOne1917, line 73: Is this div necessary?

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <form onSubmit={linkEvent(this, this.handlePasswordReset)}>
+          <h5>{capitalizeFirstLetter(i18n.t("forgot_password"))}</h5>
+
+          <div className="form-group row">
+            <label className="col-form-label col-sm-10">
```
SleeplessOne1917, line 78: Does this have to be a form-group with a label as a child? I typically think of label elements being associated with input elements.

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+              {i18n.t("no_password_reset")}
+            </label>
+          </div>
+
+          <div className="form-group row">
+            <label
```
SleeplessOne1917, line 84: What would this label look like if it was a child of the same div as the input its for?

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+      const res = await HttpService.client.passwordReset({ email });
+
+      if (res.state == "success") {
+        toast(i18n.t("reset_password_mail_sent"));
+        this.context.router.history.push("/login");
+      }
+
+      i.setState({ loading: false });
+    }
+  }
```
SleeplessOne1917, line 146: These can be defined outside the component.

SleeplessOne1917 (commented):
Besides my other comments, try using the latest version of the translations by running `git submodule update --remote --recursive`.

SleeplessOne1917:
Also, this PR makes me wonder what the verify email page is for.

alectrocute:
@SleeplessOne1917 I was thinking about this, too. Should we just make it part of the login page?

SleeplessOne1917:
@dessalines @Nutomic  any input on this?

alectrocute:
> Also, this PR makes me wonder what the verify email page is for.

It’s the page linked to in the password reset email.

dessalines:
VerifyEmail is when the back-end sends you an email with a token, that links to that verifyemail page. When you load it, it should send that token back to the server, confirming that you received the email and verified it.

src/shared/components/home/login-reset.tsx
dessalines: Once the bootstrap 5 stuff gets merged shortly, it'll have to conform to the way it does form anyway, and *might* need changes.

src/shared/components/home/login-reset.tsx
dessalines: As long as it follows the bootstrap conventions, it should be okay.

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,147 @@
...
+            </div>
+          </div>
+        </form>
+      </div>
+    );
+  }
+
+  handleEmailInputChange(i: LoginReset, event: any) {
+    i.state.form.email = event.target.value.trim();
+    i.setState(i.state);
```
dessalines, line 126: To set embedded state, do:

`i.setState(s => s.form.email = ... , s)`

or

```
i.setState(s => { ...
  return s;
});
```

Do this for the others below too.

src/shared/components/home/login-reset.tsx
alectrocute: Primarily to make it match up with `/login`. I've refactored this in my upcoming changes.

src/shared/components/home/login-reset.tsx
alectrocute: It would look like this @SleeplessOne1917 

Screenshot 2023-06-20 at 3 20 42 PM

src/shared/components/home/login-reset.tsx
alectrocute: Where do you have in mind?

dessalines (approved):
LGTM

src/shared/components/home/login-reset.tsx
SleeplessOne1917: Define these functions at the module level instead of as methods on the class. Since they will be passed in `LoginReset`, they will have no need to access `this`, so there is no reason to have them as methods.

src/shared/components/home/login-reset.tsx
```diff
@@ -0,0 +1,137 @@
...
+      const res = await HttpService.client.passwordReset({ email });
+
+      if (res.state == "success") {
+        toast(i18n.t("reset_password_mail_sent"));
+        i.context.router.history.push("/login");
+      }
+
+      i.setState(s => ((s.form.loading = false), s));
+    }
+  }
```
SleeplessOne1917, line 136: These can be defined at the module level instead of as class methods since they get `this` passed in.

dessalines: I very much prefer the class methods even if `this` gets passed in, if only for organizations-sakes.

dessalines (approved):

dessalines:
Once you get the conflicts handled we can merge this.

alectrocute:
@dessalines We should be good to go.

dessalines (approved):
Sweet, thx.

PR #1400, opened 2023-06-19
feat(themes): Reorganize themes Sass and add Litely Compact

jsit:
Here's an attempt at (a) reorganizing the Sass stylesheets and (b) adding a "Litely Compact" theme (screenshots below). All I'm doing is changing the values of `$spacer` and `$font-size-base` to tighten things up, similar to what #1395 is doing.

What I'd actually like to see is each theme in its own Sass subdirectory, with the files being compiled into a flat `css` directory, but that would require a more thorough rewrite of the `themes:build` script.

## Litely

Screenshot 2023-06-19 at 5 15 20 PM

## Litely Compact (new)

Screenshot 2023-06-19 at 5 14 58 PM


dankxiaobong:
Not sure if this is the right place to ask, but how did you add that theme to the user settings? I tried taking _variables.darkly.scss and darkly.scss renamed it to _variables.darkly-compact.scss and darkly-compact.scss (updated the import) and ran "yarn run themes:build". Do I have to run an additional script?

jsit:
@dankxiaobong It's in `build-themes-list.ts`, but I think it will also offer any themes in the `extra_themes` directory.

PR #1406, opened 2023-06-19
Fix `?scrollToComments` query param feature

alectrocute:
Hi Lemdevs!

Right now, our `scrollToComments` query param feature isn't working. It's looking for a `this.state.commentSectionRef?.current` which doesn't yet exist when it's called. 

I propose:

- Add a `setTimeout` (equivalent of `nextTick`). Normally I'm not a fan of this, but it's a lot less complicated than adding some sort of watcher to wait for `this.state.commentSectionRef?.current` to `!== null` and then execute the method.

Thanks all!

### DIFF
--- a/src/shared/components/post/post.tsx
+++ b/src/shared/components/post/post.tsx
@@ -294,7 +294,9 @@ export class Post extends Component<any, PostState> {
   }
 
   scrollIntoCommentSection() {
-    this.state.commentSectionRef?.current?.scrollIntoView();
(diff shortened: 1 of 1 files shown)


SleeplessOne1917:
@alectrocute any reason you closed this?

alectrocute:
@SleeplessOne1917 My “fix” wasn’t reliable upon some self-QA testing. Back to the drawing boards.

PR #1408, opened 2023-06-20
Add Security.txt

rcmaehl:
https://securitytxt.org/


rcmaehl:
```
[warn] src/server/index.tsx
[warn] Code style issues found in the above file. Forgot to run Prettier?
```

Attempted to run a prettier using a few different online tools with no changes to the code. Not sure what it's complaining about.

src/server/handlers/security-handler.ts
```diff
@@ -0,0 +1,9 @@
+import type { Response } from "express";
+
+export default async ({ res }: { res: Response }) => {
+  res.setHeader("content-type", "text/plain; charset=utf-8");
+
+  res.send(`Contact: mailto:security@lemmy.ml
```
tgxn, line 6: Ideally this would be driven by the URL you run Lemmy on, or a configuration option.

rcmaehl: True, however this is literally the first time I've ever done anything with typescript. Although, if you have code I can add in, I'll gladly do so.

SleeplessOne1917: `process.env.LEMMY_UI_HOST` should work for the instance URL.

src/server/index.tsx
```diff
@@ -22,6 +23,7 @@ if (!process.env["LEMMY_UI_DISABLE_CSP"] && !process.env["LEMMY_UI_DEBUG"]) {
   server.use(setDefaultCsp);
 }
 
+server.get("/.well-known/security.txt", SecurityHandler);
```
dessalines, line 26: We'll need to make sure this gets handled by nginx too.

dessalines (commented):
Husky didn't get set up properly, so its failing lint. You'll need to run prettier manually.

src/server/handlers/security-handler.ts
rcmaehl: I can add it as an addition contact but I have concerns security@`process.env.LEMMY_UI_HOST` or admin@`process.env.LEMMY_UI_HOST` will reach the owner of the instance.

rcmaehl:
Not sure why it requested a review from alectrocute but okay.

src/server/handlers/security-handler.ts
rcmaehl: Added

SleeplessOne1917:
@rcmaehl The actual changes seem good to me. @dessalines should merging this be held off until the nginx configs are updated to handle it?

rcmaehl:
Ah so it requests a review every time I update the branch. New to big pull requests...

src/server/handlers/security-handler.ts
```diff
@@ -0,0 +1,11 @@
...
+  Contact: mailto:admin@` + process.env.LEMMY_UI_HOST + `
```
dessalines, line 7: `LEMMY_UI_LEMMY_EXTERNAL_HOST` is the one with the actual domain.tld . The other ones are for the internal isomorphic server.

dessalines:
@SleeplessOne1917 We can merge this before those nginx configs are done.

src/server/handlers/security-handler.ts
rcmaehl: Done

PR #1412, opened 2023-06-20
Fix redirect after successful password reset

sunaurus:
The redirect to the home page after a successful password reset is currently broken, because the function references `this.props`, which is `undefined`. This PR fixes it.

Side-note: the `i` pattern that is used a lot in this project (including in this function) seems quite error-prone and not really idiomatic. Is there an explanation anywhere as to why this is done over just regular arrow functions + `this`?

### DIFF
--- a/src/shared/components/person/password-change.tsx
+++ b/src/shared/components/person/password-change.tsx
@@ -140,7 +140,7 @@ export class PasswordChange extends Component<any, State> {
           UserService.Instance.myUserInfo = site.data.my_user;
         }
 
-        this.props.history.replace("/");
+        i.props.history.replace("/");
       }
     }
   }


SleeplessOne1917:
> Side-note: the i pattern that is used a lot in this project (including in this function) seems quite error-prone and not really idiomatic. Is there an explanation anywhere as to why this is done over just regular arrow functions + this?

[From the InfernoJS Docs](https://www.infernojs.org/docs/guides/event-handling):
> linkEvent() offers better performance than binding an event in a class constructor and using arrow functions, so use it where possible.

I couldn't tell you why `i` is used as the parameter name though. There are also some methods that aren't used with `linkEvent` that would probably be better handled using arrow functions, as that would get rid of the need to bind them in the constructor. Making that change seems like a refactor best done in its own PR, though.

SleeplessOne1917 (approved):

PR #1421, opened 2023-06-20
Component classes v2

djvs:
Much more lightweight approach than former PR:

https://github.com/LemmyNet/lemmy-ui/pull/1361

Added dev complexity here is "please add one class with a unique name to the root of a new component".  Components with a tertiary render function syntax, or multiple root elements via `<></>`, have 2 or 3.  

Using simpler hyphenated names based on (or typically identical to) component names allows reuse of existing ones and more concise syntax.   A few are prefixed with folder name or similar because otherwise they'd be too ambiguous (e.g. "setup", "details").

Line count change is down to ~16% of previous PR, but this should still be workable for theme devs.

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -415,7 +415,7 @@ br.big {
   -webkit-box-orient: vertical;
 }
 
-.emoji-picker {
+#emoji-picker {
   width: 100%;
 }
 
--- a/src/shared/components/app/error-page.tsx
+++ b/src/shared/components/app/error-page.tsx
@@ -16,7 +16,7 @@ export class ErrorPage extends Component<any, any> {
     const { errorPageData } = this.isoData;
 
     return (
-      <div className="container-lg text-center">
+      <div className="error-page container-lg text-center">
         <h1>
           {errorPageData
             ? i18n.t("error_page_title")
--- a/src/shared/components/app/footer.tsx
+++ b/src/shared/components/app/footer.tsx
@@ -16,7 +16,7 @@ export class Footer extends Component<FooterProps, any> {
 
   render() {
     return (
-      <footer className="container-lg navbar navbar-expand-md navbar-light navbar-bg p-3">
+      <footer className="app-footer container-lg navbar navbar-expand-md navbar-light navbar-bg p-3">
         <div className="navbar-collapse">
           <ul className="navbar-nav ms-auto">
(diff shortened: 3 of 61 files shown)


djvs:
Note, `.emoji-picker` appears to have been an orphaned CSS class, not sure which component that was supposed to belong to (assumed the `#emoji-picker` one).

alectrocute:
This is so much better. A lovely middle ground.

dessalines (approved):
Seems much better to me.

alectrocute (approved):

SleeplessOne1917 (approved):

djvs:
Thanks!

0xAnansi:
A bit late to the battle but wanted to check if I stumble into something that was missed or if it was done on purpose before doing anything.

In term of hierarchy, if you take the main page, you get something "logical". 
I'll use pseudo code for the sake of brevity:

```
<body>
[...]
    <main content wrapper>
        <.home>
            <main>
            <aside>
```

image
(custom class names added for the screenshot to highlight the important blocks)

This makes it easy to target .home main and .home aside to style them, resize them, etc.
Their element is logical, the placement of each element also.

This kind of logic can be found in Communities also, but looking at a single community is different.
In this layout, the above logic becomes:

```
<body>
[...]
    <main content wrapper>
        <.community>
            <row>
                <unnamed class corresponding to the previous main>
                <unnamed class corresponding to the previous aside>
                    <.community-sidebar>
```
image
(custom class names added for the screenshot to highlight the important blocks)

In this configuration, it's difficult to impossible to target the main block of the page and the sidebar to restyle them.
Since the named sidebar is a child element, sizing the main block of the page and the sidebar is made far more difficult since the elements are not at the same level. Any max-width applied to them will not be relative to each other.

I'm also pretty convinced that the element should go from div to main and aside for the 2 mentioned unnamed classes, which might be the root issue at play here.
For both cases, I'm not sure the element under the aside, which bears the custom class name, is even needed when the class name could be added to the aside element itself.

djvs:
You're trying to target the main div.col-12.col-md-8 [main content area] + div.col-md-4 [sidebar] on the community page?  It does look like that should be a main/aside like on the other pages, you could probably do a PR for that.

djvs:
That does get to a point I was trying to address in my issue linked above - lot of elements just do nothing except work as layout scaffolding.

Issue #1394, opened 2023-06-19
[Bug]: Post shows cross-posting even when it wasn't cross-posted


jsit:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [ ] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

### Summary

In 0.18.0-rc.2, a post that was posted to a single community shows that it was also cross-posted to the same community

<img width="445" alt="Screenshot 2023-06-19 at 2 25 00 PM" src="https://github.com/LemmyNet/lemmy-ui/assets/643417/ec722f74-5113-40ab-b8f4-bf367b77fbf6">

### Steps to Reproduce

1. Create a post
2. View the post
3. See that it says it was cross-posted

### Technical Details

No

### Version

0.18.0-rc.2

PR #1425, opened 2023-06-20
Fix post view crosspost display

alectrocute:
Hi Lemdevs!

Currently, the crosspost UI is displaying for all posts even if the post isn't crossposted.

I propose we:

- Filter out all crossposts with the same community ID as the current `PostView`.
- If there's a better way to do this, I propose we do that.

Resolves https://github.com/LemmyNet/lemmy-ui/issues/1394.

Thanks all!

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -591,7 +591,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   }
 
   duplicatesLine() {
-    const dupes = this.props.crossPosts;
+    const dupes = this.props.crossPosts?.filter(
+      pv => pv.community.id !== this.postView.community.id
+    );
+
     return dupes && dupes.length > 0 ? (
       <ul className="list-inline mb-1 small text-muted">
         <>


jsit:
Where is the value of `crossPosts` coming from? Shouldn't that just be an empty array to begin with? Or is it supposed to be an array of all Communities the link is posted in? Isn't this a new problem? What changed to cause it?

src/shared/components/post/post-listing.tsx
```diff
@@ -591,7 +591,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   }
 
   duplicatesLine() {
-    const dupes = this.props.crossPosts;
+    const dupes = this.props.crossPosts?.filter(
+      pv => pv.community.id !== this.postView.community.id
```
dessalines, line 595: Using the `pv.post.id` is better here, because a single post could have some historical cross-posts. For example, a github project might be posted multiple times over the course of a few years.

dessalines:
@jsit You're correct, this is also a bug in the back end. I'll get a PR out for that shortly.

jsit:
> You're correct, this is also a bug in the back end. I'll get a PR out for that shortly.

Thanks. Maybe best then, @alectrocute, to add a `TODO` line explaining that this check can be removed at some point, with a link to this comment.

dessalines:
Actually you can probably close this one, I'll have a back-end fix for it shortly anyway.

dessalines:
mmk ya I finished the back-end PR, we can close.

PR #1426, opened 2023-06-20
feat: Drop dependency for tsconfig-paths-webpack-plugin

jsit:
@alectrocute Please let me know what you think of this approach


alectrocute:
@jsit If this works, then two thumbs up! Thanks so much!

SleeplessOne1917 (approved):

PR #1427, opened 2023-06-20
`utils.ts` organization, round two

alectrocute:
Hi Lemdevs!

Here's the second round of `utils.ts` organization. No runtime code changes, just moving/exporting and changing imports.

Also broke out tippy, toast and markdown into separate files in `./shared`. I thought about making them services, but it really doesn't make sense IMO. I like the pattern established by `i18next.ts`, so decided to follow that.

In future PR:

- Move all `interfaces.ts` exports into `@utils/types`.

We're so close! 💪

Thanks!


tsconfig.json
jsit, line 1: @alectrocute Notice the new way these paths are handled here: https://github.com/LemmyNet/lemmy-ui/pull/1426/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

alectrocute:
Any help resolving these conflicting files would be greatly appreciated.

alectrocute:
Geez, this was a lot of work. CI is passing. We're ready for review!

jsit:
Looks like `lemmy-translations` got sucked into this PR? I assume that shouldn't be in there.

alectrocute:
> Looks like `lemmy-translations` got sucked into this PR? I assume that shouldn't be in there.

Yep, they shouldn’t be here. I really wish we had that ignored. I’ll try pulling the latest submodule and see if that resolves it.

src/shared/utils/app/show-avatars.ts
```diff
@@ -0,0 +1,7 @@
+import { UserService } from "../../services";
+
+export default function showAvatars(
+  myUserInfo = UserService.Instance.myUserInfo
+): boolean {
```
SleeplessOne1917, line 5: I'm curious about these functions that take myUserInfo as an arg. Do they ever use anything other than the myUserInfo from the UserService that's set as default?

alectrocute: > I'm curious about these functions that take myUserInfo as an arg. Do they ever use anything other than the myUserInfo from the UserService that's set as default?

Good point. I'd love to go through and clean up all these functions in the next PR. It's be a lot easier to clean them out now that they're all separate files.

alectrocute:
> @alectrocute Just do `git checkout main lemmy-translations` and `git submodule update`

Thanks! I believe I have it solved now.

dessalines (approved):
Good call, we should get this merged first, as its a major refactor.

src/shared/utils/app/convert-comment-sort-type.ts
```diff
@@ -0,0 +1,21 @@
...
+    sort == "TopYear"
+  ) {
+    return "Top";
+  } else if (sort == "New") {
+    return "New";
+  } else if (sort == "Hot" || sort == "Active") {
+    return "Hot";
+  } else {
+    return "Hot";
+  }
```
SleeplessOne1917, line 20: Try:

```suggestion
switch(sort) {
    case "TopAll":
    case "TopDay":
    case "TopWeek":
    case "TopMonth":
    case "TopYear": {
        return "Top";
    }
    case "New": {
         return "New";
    }
    case "Hot":
    case "Active": {
        return "Hot";
    }
    default: {
        return "Hot";
    }
}
```

SleeplessOne1917:
LGTM, but I'll hold off on merging until 0.18 is released.

alectrocute:
> LGTM, ~but I'll hold off on merging until 0.18 is released.~ Actually, @dessalines , do you think this should be merged before the release?

I vote before the release, but that's just because I'm afraid of the incoming merge conflicts. 😅

I totally understand if y'all want to wait, though.

dessalines:
@SleeplessOne1917 @alectrocute 

This is all type checked, and its only re-organization, so I'm perfectly good with doing it before the `0.18.0` release.

Also needs a prettier run.

alectrocute:
@dessalines Ran Prettier. Should be good to go!

PR #1437, opened 2023-06-21
Expand video embeds to fullwidth

### LINKED ISSUE #926: Improved video embeds for the browser

cloventt:
Use Bootstrap 5.0 `.ratio` class to wrap embedded video iframe so that it is properly handled in a responsive way. Also add `allowfullscreen` hint and `title` attribute to embedded iframe.

Fixes #926


cloventt:
TODO: the text `Embedded Video` should probably be handled as a translatable but I wasn't sure how the translation repo integrates with this one exactly.

cloventt:
I haven't been able to test this locally as I haven't been able to get a post with an embed to work properly for some reason.

cloventt:
Tested and can confirm it looks great: https://imgur.com/a/ho0nUIh

The video can also be made fullscreen.

How it renders on desktop

How it renders on mobile

src/shared/components/post/metadata-card.tsx
```diff
@@ -75,10 +75,14 @@ export class MetadataCard extends Component<
...
-          <iframe
-            className="post-metadata-iframe"
-            src={post.embed_video_url}
-          ></iframe>
+          <div className="ratio ratio-16x9">
+            <iframe
+              allowFullScreen
+              className="post-metadata-iframe"
+              src={post.embed_video_url}
+              title={"Embedded Video: " + post.embed_title}
```
SleeplessOne1917, line 83: If you add the "Embedded Video" string, there will have to be a [translation made.](https://github.com/LemmyNet/lemmy-translations). Alternatively, you could just make the title `post.embed_title`.

cloventt: Ok, for now I will remove the prefix, but it is probably worth adding in the future as screen readers might find it useful.

SleeplessOne1917 (approved):

Issue #1153, opened 2023-06-10
Setting up local lemmy-ui instance with remote backend?


alectrocute:
Hi all!

I'd like to work on contributing to the lemmy-ui project. As a mere frontend dev, I'd like to setup a local lemmy-ui instance and use a remote backend server. Can you please link me a guide or write up quick steps to make this possible?

I tried:

- Setting up the `lemmy-ui` repo locally, with submodules init'd. `yarn start` launched the frontend server successfully but couldn't connect to `ds9.lemmy.ml` as exported in my `LEMMY_UI_LEMMY_INTERNAL_HOST `, `LEMMY_UI_LEMMY_WS_HOST ` env variables. I feel this is the closest I got.
- Using `docker-compose` for both the frontend and backend locally, resulting in error:

```bash
➜  docker git:(main) docker-compose up -d        
WARNING: Compose V1 is no longer supported and will be removed from Docker Desktop in an upcoming release. See https://docs.docker.com/go/compose-v1-eol/
Creating network "docker_lemmyinternal" with driver "bridge"
Creating network "docker_lemmyexternalproxy" with the default driver
Building lemmy
[+] Building 31.3s (7/13)                                                                     [+] Building 150.8s (11/13)                                                                   
 => [internal] load build definition from Dockerfile                                     0.0ss
 => => transferring dockerfile: 1.05kB                                                   0.0ss
 => [internal] load .dockerignore                                                        0.0ss
 => => transferring context: 159B                                                        0.0ss
 => [internal] load metadata for docker.io/library/alpine:3                              2.0ss
 => [internal] load metadata for docker.io/clux/muslrust:1.67.0                          2.1ss
 => [builder 1/5] FROM docker.io/clux/muslrust:1.67.0@sha256:c92ad5296e87681efa161fdb7  90.4ss
 => => resolve docker.io/clux/muslrust:1.67.0@sha256:c92ad5296e87681efa161fdb7421bb136b  0.0ss
 => => sha256:c92ad5296e87681efa161fdb7421bb136ba5df399441a40638c40d017 2.63kB / 2.63kB  0.0ss
 => => sha256:d6f3e144d2803bee2d077a3df724c1520fe955643b268a1ba1f36e1ea 9.35kB / 9.35kB  0.0ss
 => => sha256:677076032cca0a2362d25cf3660072e738d1b96fe860409a33ce90 29.53MB / 29.53MB  14.4ss
[...]

Issue #552, opened 2022-01-26
Different view modes (compact and classic)


vinoff:
### Is your proposal related to a problem?

I hate the look of lemmy. It might be because I am used to lobste.rs, news.ycombinator or old.reddit.com. Lemmy's UI just seems to be screaming at me. Everything is too big!

### Describe the solution you'd like

I would love if lemmy had a more "compact" mode. It would be great to have 2 additional view modes actually. If you go to reddit, you have a "classic" and a "compact". Something like that would be fantastic for lemmy.

Nutomic:
Couldnt this be implemented as a theme?

dessalines:
Maybe some things could, things like font size and spacings, but the main things that are making the page bigger, are things like image previews, icon sizes, etc.

vinoff:
https://tildes.net/ could serve as a model for a compact mode. Something more towards discussion and not so much for memes/images.

dessalines:
PRs for adding this welcome.

JohanAR:
I made a quick [userstyle](https://userstyles.world/style/10446/lemmy-world-compact-view) for compact mode, in case anyone wants some inspiration for how it could look like, or just something temporary until a real solution is implemented. I'm using the Stylus plugin for Firefox, but I think there are other options too. One would have to replace lemmy.world with the domain of the instance they're using.

PR #1438, opened 2023-06-21
added litely-compact and darkly-compact - issue 552

dankxiaobong:
Added darkly-compact in relation to issue https://github.com/LemmyNet/lemmy-ui/issues/552
Took a similiar approach as here: https://github.com/LemmyNet/lemmy-ui/pull/1400 albeit due to significant markup hacky af

### Features:
- Reduced .post-listing height from ~100px+ down to ~80px per post
- 1920px Desktop Breakpoint
- Removed user and community icons (.img-icon)

### Preview:
darkly-compact
Screenshot from 2023-06-21 11-17-30
Screenshot from 2023-06-21 13-50-12

litely-compact
Screenshot from 2023-06-21 13-47-38
Screenshot from 2023-06-21 13-49-35

If someone shows me how to connect lemmy-ui to lemmy.ml https://github.com/LemmyNet/lemmy-ui/issues/1153 I'd be able to do comments too. Currently,  I'm not able to post comments on my local instance


Lloir:
Just to add a comment to this, if you have a custom theme folder set, it'll duplicate in the themes drop down AND cause some issues. Once that's disabled it's all goochi.

But I love this 10/10

SleeplessOne1917 (approved):
Looks good, although I'll wait until after 0.18 is released to merge this.

alectrocute:
I think if we’re going to make a compact theme variation, we should make it much more compact. This feels like, slightly more compact. Thoughts?

dankxiaobong:
> I think if we’re going to make a compact theme variation, we should make it much more compact. This feels like, slightly more compact. Thoughts?

I absolutely agree that we should go even more compact.

I had these concerns:
- Changing font and icon sizes alters the overall feel of the base theme. Might as well start a new theme instead of a variation
- Decreasing icons sizes and their paddings affects mobile/touchdisplays navigation negatively (@media-queries + bs = no thank)
- Doing rearrangements of elements with current markup and bs utility classes makes it really difficult and feels dirty. And will likely break 
(I read that there's plans to go with tailwind which imo doesn't make it better. @jsit's proposal https://github.com/LemmyNet/lemmy-ui/pull/1371 with a lot flatter markup would make theming a lot easier and more flexible. )

This imo is a good compromise to get something out quick in order to satisfy user demand. 
(Also I was looking for a good first contribution I'm comfortable with. A lot of the other issues on lemmy are way above my dev skills :P)

Is there anything particularly aggravating that needs fixing?

PR #1450, opened 2023-06-21
Indicate valid and invalid fields in signup form

### LINKED ISSUE #1196: Signup page fails to create account if username contains "-"  characters, but doesn't tell the user why it's not working
Username and email input validation, either fail or refuse to work if a "-" is included.

You click signup and, it just sits there, you hit refresh and fill in everything again, still nothing, it's just not doing anything when you hit signup.

It doesn't say why, it doesn't even give an error, it just doesn't do anything.

dullbananas:
Fixes #1196


src/shared/components/home/signup.tsx
```diff
@@ -144,7 +144,7 @@ export class Signup extends Component<any, State> {
   registerForm() {
     const siteView = this.state.siteRes.site_view;
     return (
-      <form onSubmit={linkEvent(this, this.handleRegisterSubmit)}>
+      <form class="was-validated" onSubmit={linkEvent(this, this.handleRegisterSubmit)}>
```
alectrocute, line 147: Huh?

dullbananas: It causes fields to show a check mark or an red exclamation mark to indicate whether or not the value is valid

https://getbootstrap.com/docs/5.0/forms/validation/

SleeplessOne1917 (approved):

dessalines (changes requested):
You didn't correctly set up husky, so you need to run prettier manually on your added files to pass lint.

PR #1456, opened 2023-06-22
Fix I18 next circular reference

SleeplessOne1917:
This should fix the I18 next circular dependency without relying on webpack weirdness. I18 will already be initialized when it's used in the `UserService`.


alectrocute (approved):
Awesome.

dessalines (approved):
Perfect.

PR #1460, opened 2023-06-22
Fix grave `ReferenceError: fetch is not defined` error

alectrocute:
Hi Lemdevs!

I'm not sure what happened, but recent changes led to the server breaking today with `ReferenceError: fetch is not defined` as an error. It appears we already have `cross-fetch` as a dependency, but weren't using it anywhere. This PR utilizes it for shared/server use cases, ensuring it's both browser & Node friendly. Resolves the issue.

Thanks!

### DIFF
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -1,5 +1,6 @@
 import { initializeSite, isAuthPath } from "@utils/app";
 import { ErrorPageData } from "@utils/types";
+import fetch from "cross-fetch";
 import type { Request, Response } from "express";
 import { StaticRouter, matchPath } from "inferno-router";
 import { renderToString } from "inferno-server";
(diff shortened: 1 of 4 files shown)


SleeplessOne1917:
I believe the js client defaults to using cross-fetch if no fetch function is passed in. You might be able to fix this by just not passing a fetch function as well.

alectrocute:
> I believe the js client defaults to using cross-fetch if no fetch function is passed in. You might be able to fix this by just not passing a fetch function as well.

That's what I tried first (and was hoping for). `fetch` is a mandatory param:

Screenshot 2023-06-21 at 11 54 55 PM

cloventt:
My IDE is telling me `headers` should be spread like so, but I don't know if I believe it:

```typescript
new LemmyHttp(getHttpBaseInternal(), { ...headers })
```

alectrocute:
Once the JS client has `fetchFunction optional`, we should remove this redundant `cross-fetch` usage. But for now, this fixes our broken app.

alectrocute:
@SleeplessOne1917 Are we good to merge this? The app is totally broken for me without these changes.

dessalines (approved):
LGTM

PR #1461, opened 2023-06-22
fix: Fix comment upvote/downvote totals not showing in some cases

jsit:
A comment count wouldn't show if the number of upvotes was equal to the total score. This meant if there was one upvote, and the total score was 1, no number would appear next to upvotes or downvotes.

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -464,13 +464,11 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                         ) : (
                           <>
                             <Icon icon="arrow-up1" classes="icon-inline" />
-                            {showScores() &&
-                              this.commentView.counts.upvotes !==
-                                this.commentView.counts.score && (
-                                <span className="ms-1">
-                                  {numToSI(this.commentView.counts.upvotes)}
-                                </span>
-                              )}
+                            {showScores() && (
+                              <span className="ms-1">
+                                {numToSI(this.commentView.counts.upvotes)}
+                              </span>
+                            )}
                           </>
                         )}
                       </button>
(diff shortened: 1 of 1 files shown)


alectrocute:
We should probably show the score always, right? Maybe there's a reason behind this year-old logic.

alectrocute (approved):

scme0:
It seems like this was intentional. If the number of upvotes is the same as the total score, showing the upvotes and downvotes is redundant (as we know it will be <score> upvotes and 0 downvotes). The inverse is true for downvotes. My guess is that they didn't want the UI to show redundant data. I do feel like this could be confusing though but I'm no UX expert. 🤷

dessalines:
Yes that's correct, its done to not show redundant data. If there are no downvotes, the score = the upvotes.

jsit:
> If there are no downvotes, the score = the upvotes, and there's no reason to show a zero for downvotes.

But this makes it so that if there is one upvote, and the score is one, it doesn't show anything for upvotes.

So for instance, here's a comment with two upvotes, and a score of 2. It doesn't show the upvotes *or* the score, because they're the same, so I can't even tell how many upvotes it has:

Screenshot 2023-06-22 at 9 16 01 AM

But then if I downvote it, the number of upvotes appears, because the number of upvotes no longer equals the score:

Screenshot 2023-06-22 at 9 15 51 AM

This is very confusing. Why did the upvotes number suddenly appear when I downvoted it? And why can't I see the number of upvotes at all if it has no downvotes?

SleeplessOne1917:
> So for instance, here's a comment with two upvotes, and a score of 2. It doesn't show the upvotes or the score, because they're the same, so I can't even tell how many upvotes it has (upvotes = 2, score = 2):

The score is on the top right next to the 10h

> This is very confusing. Why did the upvotes number suddenly appear when I downvoted it? And why can't I see the number of upvotes at all if it has no downvotes?

If there are no downvotes, the only possible thing that can contribute to the score are upvotes, since that is the only other choice out of 2 choices. A score can only be determined by a combination of upvotes and downvotes if the count of both is greater than 0.

jsit:
Oh, thanks, wow -- I wasn't even looking at that number in the upper-right.

Issue #1384, opened 2023-06-19
Multi user support and link handling


jordan1776:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [ ] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

### Is your proposal related to a problem?

I notice in some places when a Lemmy community is linked, and it is on a small instance, they would write out the url, url as seen from beehaw, url from sopuli, url from lemmy.ml, etc. If a link to a lemmy community is linked, it should redirect to the user logged in to Lemmy. But how would it know your instance? And if you have multiple accounts logged in, which one should it pick?

### Describe the solution you'd like.

There should be an account switcher in the username dropdown on the top right which allows for quick switching between accounts. Accounts should be added to this switch if when logging in, the user unchecks a "Is this a public computer/device?" question. In the switcher should be a star which marks that user as the primary account on that device (local cookie?). When a link from outside Lemmy is clicked, the link would be automatically rewritten and community discovered if need be, so the user could immediately interact with the community from their primary account.

When changing account in the account switcher, the current viewed community or post would be redirected to the new account. If the current page is the old accounts home instance, the page should be changed to the chosen accounts instance.

If no star is checked, link behavior stays the same as it is now. 

The star should be checked by default for single account users.

Possibly a user setting for aggressive redirecting when viewing communities on instances which are not logged in, but by default internal link handling should be the same as now. 

### Describe alternatives you've considered.

Rewriting links? 

### Additional context

This multi user switching might a feature that would only appeal to power users who want to see "all of Lemmy," but I feel this solution would kill 2 birds with 1 stone, since it also includes casual users who just want to go to nichecommunity@selfhostedinstance.xyz from their lemmy.ml account.

daviessm:
See also #1048 and [#2987](https://github.com/LemmyNet/lemmy/issues/2987)

Nutomic:
You can use https://lemmyverse.link/ for this.

Issue #1297, opened 2023-06-15
Convert !community@instance into a clickable link


trashhalo:
### Requirements

- [X] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a feature request? Do not put multiple feature requests in one issue.
- [X] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

### Is your proposal related to a problem?

People seem to share references to communities in comments in the format !community@instance to use it I need to
1 copy it
2 expand hamburger menu in nav
3 go to search 
4 paste it in the search box
5 click search
6 click the first result

### Describe the solution you'd like.

Let me click the text and go right to the community 

### Describe alternatives you've considered.

None

trashhalo:
A path to doing this would be to update markdown-it/linkify to recognize the !community@server text.

- Here is the code you would need to change. https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils.ts#L714
- Here is how you access linkify from markdown it. https://github.com/markdown-it/markdown-it#linkify
- Here is how you add a custom link type to linkify. https://github.com/markdown-it/linkify-it#example-2-add-twitter-mentions-handler

Options for where to link them to:
- Instance's page for community. `!selfhost@lemmy.ml` = `https://beehaw.org/c/selfhost@lemmy.ml`
- Search results for community `!selfhost@lemmy.ml` = `https://beehaw.org/search/q/!selfhost%40lemmy.ml/type/All/sort/TopAll/listing_type/All/community_id/0/creator_id/0/page/1`

trashhalo:
@dessalines are you good with the implementation plan as defined in the previous comment, and do you have a preference between the two options on where these links could go?

I ask because I want to make sure this ticket is ready to work so someone can pick it up and just build it as is.

Dakkaron:
Apparently, relative links also do the trick. So `!selfhost@lemmy.ml` could also just link to `/c/selfhost@lemmy.ml` and let the browser fill in the domain name.

Then there is no need to insert the instance domain name.

trashhalo:
Fixed by https://github.com/LemmyNet/lemmy-ui/pull/1462

Issue #1277, opened 2023-06-14
Indicate fediverse URLS


calculuschild:
### Is your proposal related to a problem?

Happily reading a post on Lemmy, I see suspicious links to [sh.itjust.works](https://the-federation.info/node/details/50296) and [remmy.dragonpsi.xyz](https://the-federation.info/node/details/30185). I do not want to click these links, because:

1) They appear highly suspicious
2) I want to stay on *Lemmy*. I have no idea where these go.

I suspect a large number of new users are also very hesitant/confused by all these seemingly random URLs until they grasp the Fediverse concept, and even then, it's not obvious which links are inside the Fediverse and not.

### Describe the solution you'd like.

Can I suggest some way of marking links appearing in posts and comments that connect to a remote Lemmy instance, such that the link is clearly "safe" place to visit and "part of the Lemmy federation"?

### Describe alternatives you've considered.

- Just clicking the risky link. Probably not good practice for internet hygiene.
- Opening https://github.com/maltfield/awesome-lemmy-instances and looking up the url for every link. Very slow.

### Additional context

Grasping the Fediverse is hard enough. Allowing any custom URL for instances that appear untrustworthy and suspicious doesn't help. Is `beehaw` a Lemmy instance? A porn page? A phishing site? Who knows! Certainly not a brand-new user.

dukethorion:
Color Lemmy instance links green, other offsite links blue.

System would have to query the remote server and look for a Lemmy instance, maybe through a port ping?  Could also draw from a database of known instances?

dwelting:
Links to other instances should be relative to the current (home) instance. Otherwise you'll be effectively signed out every time you follow a link to an other post.

Colorcoding links to differentiate between functions should be avoided in my opinion, to help with accessibility.

skspade:
People who suffer from color blindness will have trouble distinguishing and people who are vision impaired using screen readers would not be able to tell.
However usage of icons (for color blindness) and an html attribute on the HTML element for screen readers to pick could make it work!

rszyma:
This is could be solved by #1048

glibg10b:
Is there something about federated instances that makes them more trustworthy? What's stopping me from starting my own porn site and federating it?

calculuschild:
[...]

Issue #1229, opened 2023-06-13
Allow to relatively link to individual posts


db0:
### Is your proposal related to a problem?

Currently when I want to link to a specific post, I have to always make an absolute link like `https://lemmy.dbzer0.com/post/2212`

However anyone following this link from a different instance will be prevented from subscribing or commenting.

If I link to it relatively like so `/post/2212` it doesn't work, because each instance attaches its own ID to every post.

Adding the domain to the end like so `/post/2212@lemmy.dbzer0.com` also doesn't seem to work

### Describe the solution you'd like.

I want a method to allow to link to posts in other instances relatively, to allow discussions to flow natively across the fediverse. 

I am thinking of a special format, perhaps like `!post_id@lemmy.dbzer0.com` which the UI would translate to a hyperlink for the current instance if it's federated and exists, or expand to the normal absolute link  `https://lemmy.dbzer0.com/post/2212` if it's unknown.

The same functionality should also be possible in the URL links of new posts.

### Describe alternatives you've considered.

None that I could think of

### Additional context

Cousin of #1227

I am also not sure if this belongs to UI or to lemmy main repo.

Catsrules:
There is a good discussion about this issue here 

#1048

db0:
Just noticed that lemmy is perfectly capable of finding the local instances version of a comment when the full URL is posted in the search.

As such, I should be fairly easy for a link to a comment on the fediverse be automatically translated to to the local instance by the UI, if it's cached

jacklynnepenick:
We could just have it so that if a URL to a post is posted in the comments, your local instance automatically notices it and interprets it as a link to that post on your own lemmy instance.  We'd obviously need to make it so that there's some escaping mechanism, but linking to a post within your own instance seems to be the reasonable default behavior

Falmarri:
> There is a good discussion about this issue here
> 
> #1048

I think that's not the same issue. This is about comments. Which, if they don't even share the same ID could be problematic

Catsrules:
> I think that's not the same issue. This is about comments. Which, if they don't even share the same ID could be problematic

No this is about posts thus the title of this issue "Allow to relatively link to individual posts"

But comments would also have this same issue as they are not the same ID between instances. But the solution to one might also be the solution to the other.

Zetaphor:
I'm interested in working on this, but we're going to need to hash out syntax. #1462 codeifies the following formats for communities:

* `!community@server.com`
[...]

PR #1462, opened 2023-06-22
Convert Lemmy and KBin community and user references into local instance links

Zetaphor:
This PR adds a custom plugin to Markdown-It that will add local instance links to any detected community links.

It supports the following formats:

* `/c/remoteCommunity@remoteserver.com`
* `!Community@remoteserver.com`
* `/u/username@remoteserver.com`

It will rewrite the markdown to include a link to the community from the local instance. It's not actually modifying the underlying content, only the presentation of the markup at render.

**This does not handle the possibility of the local instance not having synced any data from the remote community.**

I believe resolving the 404 issue is a larger technical discussion worthy of its own PR. This PR simply adds a custom plugin to the markdown parser.

Once merged this would partly address or fully resolve the following issues:
#1048, #369, #1227, #1229, #1277, #1384, #1297, and [lemmy/#2987](https://github.com/LemmyNet/lemmy/issues/2987). There's probably more I didn't find.

Screenshots:

**Before:**
image

**After:**
image


Zetaphor:
Upon reflection I realize the `/c/localCommunity` syntax isn't going to work if the content is federated. I'll remove that tomorrow as well as look into the 404 issue. It's very late

cloventt:
Such a gnarly regex, I love it!

In terms of 404 behaviour I think it should be that hitting a 404 triggers an attempt to pull the federated community. But I think this should probably only happen if the user is logged in to the instance.

Maybe the 404 page for communities could be adjusted to say something like, "Uh oh, can't find that community on this instance! Would you like to look for it on the host instance?" and a link to the host instance - generated from the URL of the 404 page. If the community truly does not exist at all, the user could then follow that link to find determine that.

thegabriele97:
This is a comment from @dessalines on another PR:  

`There aren't any 404s currently, because the post and comment nodes already exist on your own instance, or you wouldn't be seeing them (so neither link will 404).

But with searched content like this, the comment with text referencing a federated community or actor, might not exist everywhere.

the home instance will automatically fetch the content from the remote instance instead of 404'ing?

We don't currently do this, but this is the advantage of having a local link. We could do a resolveObject if it 404s, and cache it to our local instance. No guarantees still, since it could be blocked, but at least its possible.

We've chosen for resolve to only work on the search page currently, and by logged-in users only, but I'm open to a discussion for changing that. Def has pros and cons.`

trashhalo:
Totally agree that the 404 behavior for Lemmy communities needs work. If the complexities there end up blocking this from merging I'd suggest splitting that out as its own body of work and letting this land as is.

Correctly linking communities removing what is currently a 6 step process on mobile using copy and search is a huge leap forward in user experience.

Thank you for picking this up, going to be life changing!

Undearius:
This may be out of the scope of this PR but I imagine that kbin will be doing something similar down the line with their magazines (communities). Their syntax uses `/m/` instead of `/c/`.

It might be worth it to have a simple parse replace `/m/comm@instance.tld` with `/c/comm@instance.tld`. If kbin implements this (but in reverse) it will greatly improve interoperability between the two link aggregators.

Zetaphor:
> This may be out of the scope of this PR but I imagine that kbin will be doing something similar down the line with linking their magazines (communities). Their syntax uses `/m/` instead of `/c/`.
> 
> It might be worth it to have a simple parse for `/m/comm@instance.tld` and replace it with `/c/comm@instance.tld`. If kbin implements this (but in reverse) it will greatly improve interoperability between the two link aggregators.

Great minds think alike, I was considering this as I was browsing on mobile in bed last night. I'll add that today as well.

Undearius:
Here's the modified regex that will match with `/c/community@instance.tld` and `!community@instance.tld` but not `/c/community`
`/(!\b[^@\s]+@[^@\s]+\.[^.\s]+\b)|\/c\/([^@\s]+@[^@\s]+\.[^.\s]+\b)/g`
___
Here's the regex that will match with `/m/community@instance.tld` only
`/\/m\/([^@\s]+@[^@\s]+\.[^.\s]+\b)/g`
___
Or if you want that all in one, this will match everything above that has the instance
`/(!\b[^@\s]+@[^@\s]+\.[^.\s]+\b)|(\/c\/|\/m\/)([^@\s]+@[^@\s]+\.[^.\s]+\b)/g`

Unfortunately, I don't know the language well enough to translate `/m/` to `/c/` and cannot contribute further

Zetaphor:
I've updated this with the following changes:

* Remove support for community links without a remote server (ex: /c/testing) as that would fail on federated instances
* Add support for rewriting KBin community urls, ex: /m/example@remote.com becomes /c/example@remote.com
* Add support for fully qualified users, ex: /u/username@remoteserver.com

Zetaphor:
I'm going to move this out of draft, I believe the issue of handling 404's for unsynced communties is out of scope for this PR. That warrants a larger technical discussion and this is just a plugin for the markdown parser.

src/shared/markdown.ts
```diff
@@ -74,73 +73,66 @@ const html5EmbedConfig = {
...
-              url = `${getHttpBase()}/c/${name}@${domain}.${tld}`;
+    /**
+     * Accepted formats:
+     * !community@server.com
+     * /c/community@server.com
+     * /m/community@server.com
+     * /u/username@server.com
+     */
+    const pattern =
+      /(\/[c|m|u]\/[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}|![a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})/g;
```
Undearius, line 85: Instead of having the whole regex pattern written twice, one for /_x_/ and one for !, could this be simplified to 
`/(\/[c|m|u]\/|!)[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}/g`

SleeplessOne1917:
I know we don't have unit tests yet, but a function based on a complicated regex like this is something I'm scared to merge without tests.

Zetaphor:
> I know we don't have unit tests yet, but a function based on a complicated regex like this is something I'm scared to merge without tests.

In lieu of proper unit tests, I can offer these unit tests from regex101:

https://regex101.com/r/J0L0gS/1

src/shared/markdown.ts
```diff
@@ -72,6 +72,72 @@ const html5EmbedConfig = {
...
+function localCommunityLinkParser(md) {
+  md.core.ruler.push("replace-text", state => {
+    /**
+     * Accepted formats:
+     * !community@server.com
+     * /c/community@server.com
+     * /m/community@server.com
+     * /u/username@server.com
+     */
+    const pattern =
```
alectrocute, line 84: We should export this from `config.ts` and import it here. For more visiblity.

Zetaphor:
This is also pretty easy to test locally in the frontend as it doesn't require modifying content, you can just switch back and forth between the editor and preview

src/shared/markdown.ts
Zetaphor: This is great, thank you!

Zetaphor:
Updated link with regex101 unit tests:

https://regex101.com/r/J0L0gS/3

Zetaphor:
This is now just a basic email regex with a prefix requirement

Zetaphor:
Made one last push because I accidentally removed classes. Should be all set for merge now

src/shared/config.ts
```diff
@@ -33,6 +33,6 @@ export const emDash = "\u2014";
  * /u/username@server.com
  */
 export const instanceLinkRegex =
-  /(\/[c|m|u]\/[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}|![a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})/g;
+  /(\/[c|m|u]\/|!)[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}/g;
```
Undearius, line 36: One last alteration. regex `[c|m|u]` will match the OR pipes `|` literally, so /|/community@instance.tld will match as is.
Replace `[c|m|u]` with `[cmu]`. You could also rearrange the letters for an inappropriate easter egg

Zetaphor: Good catch, updated. Decided against the easter egg lol

Zetaphor:
Unless @Undearius has any additional catches (they're clearly superior in regex-fu) this should be all set.

Latest regex101 link: https://regex101.com/r/J0L0gS/5

Zetaphor:
Going to see what I can do about these linter errors

Dakkaron:
Would it be possible to convert local community references to global links?

So say, the user test@instanceA posts `/c/community` on instanceB, it generates a link like `/c/community@instanceA` from it? So basically just take the user's instance host name and append it to the link of any reference that doesn't have an instance in the name?

Zetaphor:
> Would it be possible to convert local community references to global links?
> 
> So say, the user test@instanceA posts `/c/community` on instanceB, it generates a link like `/c/community@instanceA` from it? So basically just take the user's instance host name and append it to the link of any reference that doesn't have an instance in the name?
> 
> Thanks a ton, btw, for fixing this!

I'm not sure I fully understand your question, can you please rephrase it with a more detailed example?

I originally had a version that supported the use of the `/c/community` shorthand, but decided against that because it would break federation with other platforms. This change only modifies the rendering of the markdown when viewed through a Lemmy instance, not the underlying content. If you were to see that comment from a federated KBin or Mastodon instance you would have no way of actually following the link/reference, since it doesn't include the remote server address.

Undearius:
I believe what they're saying is this.

I write a comment from my account on lemmy.ca (undearius@lemmy.ca) and reference /c/plex. You, on your zemmy.cc instance, click the link to the community. 

Instead of it linking to `zemmy.cc/c/plex` it would take the domain from my username and link it to `zemmy.cc/c/plex@lemmy.ca`

Zetaphor:
Ahh okay, that makes sense.

I'd want to get a second opinion, but that sounds a bit too opaque from a flow perspective. The current format is more of a helper than something that expands a limited amount of text into a fully qualified route. A user would have to know ahead of time that this functionality exists, where the changes in this PR just assist with what is an already established form of addressing.

There's also still the issue of that not working outside of Lemmy. If you were to view that link on KBin, Mastodon, or another ActivityPub compatible interface you would have no way of obviously tying the intended location to the correct address without knowing that this one specific interface quirk exists on Lemmy's web UI.

Dakkaron:
Yeah, @Undearius understood me correctly.  I figured it could be working like local  dialing on the phone. If someone is not dialing a fully-qualified number, the phone system will dial "inside  their scope". But I understand your concerns, they make sense.

M4rtineski:
I may have missed something but how does this update/merge behave with links to posts and not communities? Will this work with links like that?:
image
image

Zetaphor:
@M4rtineski It won't affect those at all, it's only matching on the patterns I included above.

M4rtineski:
@Zetaphor So from my understanding this merge doesn't work for linking posts? Are you aware of any issues for linking posts without making other's leave their instance? And if does work for linking posts can you show me the syntax please?

Zetaphor:
@M4rtineski The format is in the top of the PR, but here it is again:


* `/c/remoteCommunity@remoteserver.com`
* `!Community@remoteserver.com`
* `/m/remoteCommunity@kbinserver.com`
* `/u/username@remoteserver.com`

There's an ongoing discussion at #1048

I would consider making a PR specifically for links that uses an icon similar to what exists for comments 🤔

PR #1463, opened 2023-06-22
feat: Default language dropdowns to user's interface language

jsit:
When Markdown entry fields are shown, this change will automatically set the Language dropdown to the user's selected Interface Language.

If the user hasn't selected an Interface Language, this defaults to their browser language.

Before, this dropdown defaulted to "Undetermined", which made it so that content was by default without a language. (Hence the warning about de-selecting "Undetermined" in your list of preferred languages.)

## Before

Screenshot 2023-06-23 at 5 24 36 PM

## After

Screenshot 2023-06-23 at 5 21 43 PM


snaggen:
It would be best to be able to select the default language for posts, but not sure Primary Language is a good term though since non-english speaker have another Primary Language they talk but still may post most english content. So if selection would be made, it must be clear that it is the default language for created posts, not what language the user is most fluent in.

SleeplessOne1917:
@Nutomic isn't the backend supposed to pick a language automatically if one is not specified? If so, this PR is unnecessary.

jsit:
@SleeplessOne1917 

> isn't the backend supposed to pick a language automatically if one is not specified? If so, this PR is unnecessary.

No, this is for picking the language in the language dropdown when composing a post or comment. Right now that dropdown always defaults to "Undetermined".

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -260,8 +260,12 @@ export class MarkdownTextArea extends Component<
               <LanguageSelect
                 iconVersion
                 allLanguages={this.props.allLanguages}
+                // Only set the selected language ID if it exists as an option
+                // in the dropdown; otherwise, set it to 0 (Undetermined)
                 selectedLanguageIds={
-                  languageId ? Array.of(languageId) : undefined
+                  languageId && this.props.siteLanguages.includes(languageId)
+                    ? Array.of(languageId)
```
SleeplessOne1917, line 267: Is there any reason `Array.of` is better here than just throwing it in some square brackets, or it this just to keep parity with what was there before?

jsit: Nope, just left it like that. I agree it feels better with brackets, which works fine, will commit.

SleeplessOne1917 (approved):

Nutomic:
Yes this isnt correct and shouldnt have been merged. The backend tries to select a language automatically if none was passed, and this change is going to break the automatic language selection.

jsit:
@Nutomic Thanks, my mistake, this was too aggressive. I’ll create a PR to revert this.

jsit:
@Nutomic Why does the backend dictate what language the user is writing in? This PR affects the markdown language select drop-down. I imagine the language the user is writing in won’t always be the language the server is in, no?

Will still back this out but am curious about this behavior.

snaggen:
I didn't realize the backend had this default selection of languages either, so that is a bit confusing. 
I'd rather see that selection also happening in the UI, and then having the language selection mandatory and leave it empty for the case where it cannot be auto-selected (ie. when use have more than one language selected, besides "undetermined").

Nutomic:
Making language selection mandatory would be very annoying, and would discourage users from posting. In any case there will be posts without language tag, because most other federated platforms dont support language tags.

PR #1466, opened 2023-06-22
Fix buildThemeList() function to ensure no duplicates

### LINKED ISSUE #1411: [Bug]: Custom theme item gets duplicated in user settings dropdown
### Requirements

- [x] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

### Summary

After putting a custom theme CSS file in ./extra_themes, I have run into a bug where once the theme is set via user settings and saved, each subsequent page reload will add another entry to the theme dropdown. Eg.: After selecting icy-nord and saving it, every time the page is refreshed another "icy-nord" will be added to the dropdown. Eg.: Browser Default, darkly, darkly-red, litely, litely-red, icy-nord, icy-nord, icy-nord, icy-nord, etc. 

### Steps to Reproduce

1. Change to a custom theme via user settings (I used https://github.com/promitheas17j/lemmy-ui-themes/blob/main/icy-nord.css)
2. Save the change, then refresh the page.
3. Theme dropdown will now have an extra theme for every time the page is refreshed (eg.: Browser Default, darkly, darkly-red, litely, litely-red, icy-nord, icy-nord, icy-nord, icy-nord, etc)

### Technical Details

No.

### Version

0.18.0-rc.2

### Lemmy Instance URL

_No response_

scme0:
Duplicates were created because the custom themes were added to the global `themes` property each time the function was called.

Fixes #1411

### DIFF
--- a/src/server/utils/build-themes-list.ts
+++ b/src/server/utils/build-themes-list.ts
@@ -4,15 +4,20 @@ import { readdir } from "fs/promises";
 const extraThemesFolder =
   process.env["LEMMY_UI_EXTRA_THEMES_FOLDER"] || "./extra_themes";
 
-const themes = ["darkly", "darkly-red", "litely", "litely-red"];
+const themes: ReadonlyArray<string> = [
+  "darkly",
+  "darkly-red",
+  "litely",
+  "litely-red",
+];
 
-export async function buildThemeList(): Promise<string[]> {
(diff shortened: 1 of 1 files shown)


alectrocute (approved):
LGTM!

dessalines (approved):

PR #1472, opened 2023-06-22
Remove scroll position set/get for `Community` and `Home`

alectrocute:
Hi Lemdevs!

Pretty grave bug here with our scrolling:

https://github.com/LemmyNet/lemmy-ui/assets/35377827/2e31330b-9350-45cd-afb4-6395d2a6d37b

I propose we:

- Remove scroll position set/get for `Community` and `Home`

### DIFF
--- a/src/shared/utils/browser/restore-scroll-position.ts
+++ b/src/shared/utils/browser/restore-scroll-position.ts
@@ -1,5 +1,6 @@
 export default function restoreScrollPosition(context: any) {
   const path: string = context.router.route.location.pathname;
   const y = Number(sessionStorage.getItem(`scrollPosition_${path}`));
+
   window.scrollTo(0, y);
 }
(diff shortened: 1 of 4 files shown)


src/shared/utils/browser/restore-scroll-position.ts
```diff
@@ -1,5 +1,6 @@
 export default function restoreScrollPosition(context: any) {
```
SleeplessOne1917, line 1: Are this and saveScrollPosition still used?

SleeplessOne1917 (commented):
> Is there any reason to restore scroll position on these specific routes?

Not that I know of.

src/shared/utils/browser/restore-scroll-position.ts
alectrocute: Yep, on `Profile`, `Search`, and `Post`.

SleeplessOne1917 (approved):

PR #1474, opened 2023-06-22
Enable users to zoom on mobile

### LINKED ISSUE #1473: [Bug]: Cannot zoom in on mobile
### Requirements

### Summary

As the title describes. This issue is on the latest release cut of 0.18, but not 0.17.4.

### Steps to Reproduce

1. Open https://lemmy.ml on a mobile device
2. Try to zoom in with the zoom in gesture.

### Technical Details

N/A

### Version

0.18.0-rc.6

### Lemmy Instance URL

lemmy.ml

SleeplessOne1917:
Closes #1473. Was unable to test locally as I would need to run this on a mobile device.

### DIFF
--- a/src/server/utils/create-ssr-html.tsx
+++ b/src/server/utils/create-ssr-html.tsx
@@ -73,7 +73,7 @@ export async function createSsrHtml(
     <!-- Required meta tags -->
     <meta name="Description" content="Lemmy">
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=no">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=yes">
     <link
        id="favicon"
        rel="shortcut icon"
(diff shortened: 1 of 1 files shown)


alectrocute:
This seems like a double-edged sword. Accidental zooming is something I run into frequently with 0.17.

src/server/utils/create-ssr-html.tsx
```diff
@@ -73,7 +73,7 @@ export async function createSsrHtml(
     <!-- Required meta tags -->
     <meta name="Description" content="Lemmy">
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=no">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no, user-scalable=yes">
```
alectrocute, line 76: We can also just omit `user-scalable` entirely since it defaults to `yes`.

dessalines (approved):

src/server/utils/create-ssr-html.tsx
SleeplessOne1917: Omitted.

alectrocute (approved):

SleeplessOne1917:
> This seems like a double-edged sword. Accidental zooming is something I run into frequently with 0.17.

I just saw this comment. You're probably right about not allowing zoom on mobile would mostly be a better experience, but there is one thing the zoom is really needed for: viewing inline images. Currently, it can be hard to get a bigger view of inline images. Perhaps I should open an issue for a UI feature of a display modal for viewing larger versions of images.

PR #1488, opened 2023-06-22
feat(dev): Vote components

jsit:
This PR componentizes the Votes buttons, both "regular" and "compact," in the Post Listing view and Comment Nodes.

Should be tested thoroughly, as voting is a pretty important function.


src/shared/components/common/vote-buttons.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { I18NextService } from "../../services";
+import { Icon, Spinner } from "../common/icon";
+import { PostListing } from "../post/post-listing";
+
+interface VoteButtonsProps {
+  postListing: PostListing;
+  enableDownvotes?: boolean;
+  upvoteLoading?: boolean;
+  downvoteLoading?: boolean;
+  handleUpvote: (i: PostListing) => void;
```
dessalines, line 15: What we've typically done, is pass the lemmy-js-client form (in this case `CreateCommentLike`) up to the parent. 

But here, since that could be either a post or a comment, I'd recommend passing up an id, the new vote, and an auth. Type the interface well, something like:

```
interface VoteItem {
  id: number,
  newVote: number,
  auth: ...
}
```

src/shared/components/common/vote-buttons.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { numToSI } from "@utils/helpers";
+import classNames from "classnames";
+import { Component, linkEvent } from "inferno";
+import { CommentAggregates, PostAggregates } from "lemmy-js-client";
+import { I18NextService } from "../../services";
+import { Icon, Spinner } from "../common/icon";
+import { PostListing } from "../post/post-listing";
+
+interface VoteButtonsProps {
+  postListing: PostListing;
```
dessalines, line 11: This needs to be made generic enough to work for both post and comments, so you'll need to expand out the counts, and my_vote into distinct fields.

src/shared/components/common/vote-buttons.tsx
jsit: Thanks -- are my latest edits more what you were picturing? I'm not sure where this `VoteItem` type could go.

jsit: Comments use `comment_id` for their ID, and posts use `post_id`, so I had to add a `switch` and an extra property so comments/posts can specify which type to use when submitting the click.

src/shared/components/comment/comment-node.tsx
```diff
@@ -356,29 +356,18 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
               <div className="me-lg-5 flex-grow-1 flex-lg-grow-0 unselectable pointer mx-2" />
+
               {showScores() && (
                 <>
-                  <a
-                    className={`unselectable pointer ${this.scoreColor}`}
-                    onClick={linkEvent(this, this.handleUpvote)}
-                    data-tippy-content={this.pointsTippy}
+                  <span
+                    className="me-1 font-weight-bold"
```
SleeplessOne1917, line 363: From your other PR, `font-weight-bold` will have to be changed into `fw-bold`.

src/shared/components/common/vote-buttons.tsx
```diff
@@ -0,0 +1,221 @@
...
+    count: Number(counts.upvotes),
+    formattedCount: Number(counts.upvotes),
+  });
+
+  const downvotes = I18NextService.i18n.t("number_of_downvotes", {
+    count: Number(counts.downvotes),
+    formattedCount: Number(counts.downvotes),
+  });
+
+  return `${points} • ${upvotes} • ${downvotes}`;
```
SleeplessOne1917, line 45: Maybe throw a a + in front of upvotes and a minus in front of downvotes? I don't think it would be clear which number is which otherwise.

src/shared/components/common/vote-buttons.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  switch (i.props.voteContentType) {
+    case VoteContentType.Comment:
+      i.props.onVote({
+        comment_id: i.props.id,
+        score: newVote(VoteType.Upvote, i.props.my_vote),
+        auth: myAuthRequired(),
+      });
+      break;
+    case VoteContentType.Post:
```
SleeplessOne1917, line 59: If `VoteContentType.Post` is always going to be handled the same way as default, there is no reason to make a case for it. Heck, this might be clearer as an if-else instead of a switch considering you're only choosing between 2 options.

jsit: I was just anticipating future use-cases, but happy to simplify this too if you think that's premature

src/shared/components/common/vote-buttons.tsx
jsit: It's getting something like this already:
Screenshot 2023-06-25 at 1 14 45 AM

SleeplessOne1917 (approved):

PR #1490, opened 2023-06-22
Fix `Modlog` refetch on component mount

alectrocute:
Hi Lemdevs!

Our modlog is currently broken. Unless you refresh directly on the `/modlog` route, the data is never loaded, resulting in a blank page. This PR resolves that by calling `this.refetch()` on the component mount hook, which seems to be a pretty common pattern here.

Thanks!


dessalines (approved):
Oops, thanks.

PR #1508, opened 2023-06-23
Replace momentjs with date-fns

SleeplessOne1917:
This should reduce the bundle size somewhat, although I'm not sure how much that will affect performance in production. I replaced Moment.js, a large dependency, with date-fns, which allows you to only import what you need so as to keep bundle size smaller.


src/server/index.tsx
```diff
@@ -34,7 +36,17 @@ server.listen(Number(port), hostname, () => {
...
+process.on("SIGINT", async () => {
+  const lang = I18NextService.i18n.language;
+  const locale = (
+    await import(
+      /* webpackExclude: /\.js\.flow$/ */
+      `date-fns/locale/${lang}`
+    )
+  ).default;
+  setDefaultOptions({
+    locale,
```
alectrocute, line 48: Why here?

SleeplessOne1917: I want to make sure the correct locale is used on both the server and the client.

alectrocute: But this is the process hook for when the server shuts down, yeah?

SleeplessOne1917: I am a fool and mistook this for where the server is started 🤦🏻

alectrocute: Haha, no worries. It may be better to put this in a server util file too, like `init-date-fns.ts` and import it here instead.

SleeplessOne1917:
Can't push to this for some reason. Will open another PR.

PR #1509, opened 2023-06-23
Fix server-side rendering after first load.

dudeami:
Server-side rendering was being disabled by the global variable `FirstLoadService.isFirstLoad` being set on the server side to `false` after the first page was rendered. This commit adds a check for if `!isBrowser()` so server side rendering is triggered on every page request. This was preventing Open Graph meta tags and other metadata from being generated.

### DIFF
--- a/src/shared/components/community/communities.tsx
+++ b/src/shared/components/community/communities.tsx
@@ -5,6 +5,7 @@ import {
   setIsoData,
   showLocal,
 } from "@utils/app";
+import { isBrowser } from "@utils/browser";
 import {
   getPageFromString,
   getQueryParams,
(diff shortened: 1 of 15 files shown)


src/shared/components/community/communities.tsx
```diff
@@ -67,7 +68,7 @@ export class Communities extends Component<any, CommunitiesState> {
     this.handleListingTypeChange = this.handleListingTypeChange.bind(this);
 
     // Only fetch the data if coming from another route
-    if (FirstLoadService.isFirstLoad) {
+    if (!isBrowser() || FirstLoadService.isFirstLoad) {
```
SleeplessOne1917, line 71: If everywhere `isFirstLoad` is used needs to be ORed with `!isBrowser()`, it's easier to move that logic to the `isFirstLoad` function.

dudeami: I was hesitant to change the functionality of `isFirstLoad` in the case it was used for server initialization. If `isFirstLoad` is assumed to be browser first load I'll move the check there.

SleeplessOne1917 (approved):

PR #1517, opened 2023-06-23
Replace MomentJS with Date-Fns (Retry)

SleeplessOne1917:
I messed up #1508 in a way that I couldn't push to it, so this is a new PR with changes suggested from the previous one.


src/shared/components/common/moment-time.tsx
```diff
@@ -11,22 +12,24 @@ interface MomentTimeProps {
   ignoreUpdated?: boolean;
 }
 
+function momentFormat(input: string) {
```
alectrocute, line 15: Couldn't this just be called `dateFormat`?

PR #1525, opened 2023-06-23
Fix entire app breaking if routing away from User Settings page

alectrocute:
Hi Lemdevs,

Looks like my last PR was only the tip of the iceberg: https://github.com/LemmyNet/lemmy-ui/pull/1506

https://github.com/LemmyNet/lemmy-ui/pull/1506#issuecomment-1603529191

I only fixed it for Settings -> Home, but not for all the other routes. 💀 On 0.18, it's pretty bad.

- Why are we conditionally loading data based on `!this.state.isIsomorphic`? When routing away from the User Settings page, that condition is definitely falsey.

- I added optional chaining, and added only what was needed as per console errors.

Thanks!

### DIFF
--- a/src/shared/components/community/communities.tsx
+++ b/src/shared/components/community/communities.tsx
@@ -79,9 +79,7 @@ export class Communities extends Component<any, CommunitiesState> {
   }
 
   async componentDidMount() {
-    if (!this.state.isIsomorphic) {
-      await this.refetch();
-    }
+    await this.refetch();
   }
 
   get documentTitle(): string {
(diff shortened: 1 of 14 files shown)


alectrocute:
https://github.com/LemmyNet/lemmy-ui/assets/35377827/151e17ea-e696-4bff-9509-66750443d0ae

SleeplessOne1917:
The `!isIsomorphic` check is because I only want fetches to be done on mount if content was not already fetched on the server. Perhaps a better name would be something like `didFetchOnServer`.

alectrocute:
> The `!isIsomorphic` check is because I only want fetches to be done on mount if content was not already fetched on the server. Perhaps a better name would be something like `didFetchOnServer`.

We must have something gravely broken in this logic. I feel like this PR is more of a bandaid fix. Any ideas to make it less of one?

SleeplessOne1917:
Closing in favor of #1528

Issue #1520, opened 2023-06-23
Make scrolling on homepage when returning from posts consistent


dudeami0:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Currently, when viewing a post then returning to the homepage scrolling behavior is inconsistent and unaware of updates. From what I can tell, it relies on the browser's built in scrolling functionality when moving between history items. This provides a confusing experience as users are sometimes scrolled to different positions when quickly viewing topics, and brought back to an old position with a completely different set of items when returning after a while. The scrolling is also affected by differences such as the number of subscriptions extending the document height and how quickly the contents load.

I propose scrolling to the last scroll position relative to the last viewed post on the homepage when returning from a post, or if that post is no longer available scrolling to the top. This allows users to return to where they were without scrolling, or if the content is no longer on the page to have a clear indication the feed is updated.

It would also be a good idea to extend the post listing to have the document height required to scroll without waiting for content to load. Possible solutions would be visual placeholders or temporarily setting a minimum height on the `div.post-listings` element to accommodate the saved scroll position + viewport height.

An alternative to scrolling to the top could be to load the page that comment is on, but arbitrarily determining what page a post is now on is a problem. Using a cached version of the post listings could solve this and document height issues. This would create a most consistent user experience until the user reloaded or went to a new page.

I can implement and create a PR for this if needed.

dessalines:
This has mostly been fixed I believe, since @SleeplessOne1917 implemented the scrollMixin, which has some advanced code for remembering the scroll position. 

lemmy-ui can't do this perfectly as an SPA, but it does mostly solve the problem.

PR #1526, opened 2023-06-23
Cache post listings and restore listing state when using browser back navigation buttons.

dudeami:
Possible solution to #1520, on loading post listings the `RequestState<GetPostsResponse>` is cached. When navigating away from the home page followed by using the browser navigation back button, the cache is used to restore the post listings to the state the user left it in.


src/shared/components/home/home.tsx
```diff
@@ -650,6 +661,8 @@ export class Home extends Component<any, HomeState> {
 
     if (dataType === DataType.Post) {
       switch (this.state.postsRes?.state) {
+        case "empty":
```
SleeplessOne1917, line 664: Seeing this makes me think we should default to setting the responses to `{ state: "loading" }` instead of `{ state: "empty" }`. That way, it will display the loading spinner when fetching posts instead of a big empty div.

dudeami: When loading, it should immediately switch from `"empty"` to `"loading"` state as rendering that state is near instant. The `"empty"` state is only for scrolling while the `"success"` state renders, which is fast but slow enough to stop proper scrolling and just end up where the bottom of the empty page is. If we wait for the page to render, the transition isn't as smooth with a jerky snap when the content finishes loading. It's something the user should never really see, and more of a hack to make the scrolling work as intended.

SleeplessOne1917 (approved):

jsit:
Thanks for this! This fix doesn't seem to be working for me in iOS Safari on voyager.lemmy.ml. @SleeplessOne1917 @alectrocute @dessalines

PR #1529, opened 2023-06-23
Refactor first load handling

SleeplessOne1917:
Refactor of the fix implemented in #1528. This should prevent contributors from always needing to remember to manually call `FirstLoadService.isFirstLoad` in the constructors of new route page components even if they're not using it.


alectrocute (approved):
Nice work.

Issue #1177, opened 2023-06-11
Community list sorting doesn't work; unsorted list difficult to use.


mapsedge:
Found a bug? Please fill out the sections below. 👍

For backend issues, use [lemmy](https://github.com/LemmyNet/lemmy)

### Issue Summary

On the page where communities are listed, in my case, https://lemmy.ml/communities, the column headers look like they should be clickable (and therefore sortable) but they aren't. With just a few subscriptions, it's merely annoying, but as the list grows, it will quickly become unusable.

### Steps to Reproduce

Navigate to /communities
Hover any column heading of the communities list.
Observe the mouse pointer indicates the heading is clickable.
Click.
Be unhappy.

No console messages on click. A quick inspect shows no event handlers defined.

ngosang:
I'm thinking the same. They should be sortable with the headers or with a dropdown.

machie27:
There is a site which has this functionality: check [lemmyverse](https://lemmyverse.net/communities )

jsit:
Working on this

PR #1538, opened 2023-06-23
feat: Add sorting to Community table #1177

jsit:
Completes #1177

This still needs work before review.


jsit:
@dessalines I think this might need some backend work, too; I passed `sort: "MostComments"` to the fetch and it doesn't change what order the communities are returned in.

dessalines:
Yep that's our bad, we need to add the rest of them:

Screenshot_2023-06-27-07-12-48-586_com termux

At least for now, just use the standard `SortSelect`, and eventually these will get added to the back-end.

dessalines:
#1934 correctly handles this on the back end, so this can probably be closed.

PR #1546, opened 2023-06-24
Open links in a new tab setting in UI

Dogeek:
related to https://github.com/LemmyNet/lemmy/pull/3318


src/shared/components/person/settings.tsx
```diff
@@ -802,6 +803,23 @@ export class Settings extends Component<any, SettingsState> {
...
+              <input
+                className="form-check-input"
+                id="user-open-links-in-new-tab"
+                type="checkbox"
+                checked={this.state.saveUserSettingsForm.open_links_in_new_tab}
+                onChange={linkEvent(this, this.handleOpenInNewTab)}
+              />
+              <label
+                className="form-check-label"
+                htmlFor="user-show-new-post-notifs"
```
SleeplessOne1917, line 817: Wrong htmlFor ID.

src/shared/components/person/settings.tsx
```diff
@@ -1044,6 +1062,14 @@ export class Settings extends Component<any, SettingsState> {
     );
   }
 
+  handleOpenInNewTab(i: Settings, event: any) {
```
SleeplessOne1917, line 1065: You can define this at the module level instead of as a class method.

src/shared/components/post/post-listing.tsx
```diff
@@ -759,6 +759,12 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         title={title}
         to={`/post/${post_view.post.id}?scrollToComments=true`}
         data-tippy-content={title}
+        target={
+          UserService.Instance.myUserInfo?.local_user_view.local_user
+            .open_links_in_new_tab ?? false
```
SleeplessOne1917, line 764: If it's null or undefined it will be falsy, so the coalesce is unnecessary.

SleeplessOne1917 (approved):
Some small changes and then you're good.

SleeplessOne1917 (changes requested):
Meant to request changes.

src/shared/components/person/settings.tsx
Dogeek: Can you elaborate ? All of the other settings use instance level methods to mutate the state. I could extract it outside of the class but consistency has to be better right ?

SleeplessOne1917: It's fine to leave it the way it is.

SleeplessOne1917 (approved):

dessalines (commented):
Once this gets added to the back-end, I'll do a lemmy-js-client release so you can get this working.

src/shared/components/post/post-listing.tsx
```diff
@@ -748,6 +768,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         title={title}
         to={`/post/${post_view.post.id}?scrollToComments=true`}
         data-tippy-content={title}
+        target={this.linkTarget}
```
d42ohpaz, line 771: Is there a way to add [rel="noopener"](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noopener)?

src/shared/components/post/post-listing.tsx
```diff
@@ -735,6 +744,17 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     );
   }
 
+  public get linkTarget(): string {
+    if (
+      UserService.Instance.myUserInfo?.local_user_view.local_user
+        .open_links_in_new_tab
+    ) {
+      return "_blank";
+    }
```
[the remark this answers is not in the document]
Dogeek: the `else` clause here is strictly unnecessary, it's just clutter. At the end of the day, it's the same code, just harder to read. I can add it if you feel like it's strictly necessary, but it doesn't add any value in that instance

src/shared/components/post/post-listing.tsx
Dogeek: it's implicit on target="_blank"

Dogeek:
Closing cause it's never gonna be merged anyways

krestenlaust:
@SleeplessOne1917 can we get a review on the newest changes, and an idea of whether it's good to go?

SleeplessOne1917 (approved):

src/shared/components/post/post-listing.tsx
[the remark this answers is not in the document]
dessalines: So they never added this simple change, and then closed the PR?

dessalines (changes requested):
Merge conflicts, and you didn't address PR comments.

krestenlaust:
> Merge conflicts, and you didn't address PR comments.

Are maintainer commits enabled? If that's the case, can't you address it? It looks like a simple change unrelated to the functionality, that's just because of your preference.
Otherwise, I'd have to make a PR to his fork, which then starts to get a bit messy.

PR #1547, opened 2023-06-24
fix(a11y): Close Emoji Picker on 'esc', make jump link interactive button

jsit:
Keeping this small


src/shared/components/app/app.tsx
```diff
@@ -33,12 +33,14 @@ export class App extends Component<any, any> {
       <>
         <Provider i18next={I18NextService.i18n}>
           <div id="app" className="lemmy-site">
-            <a
-              className="skip-link bg-light text-dark p-2 text-decoration-none position-absolute start-0 z-3"
+            <button
+              type="button"
+              tabIndex="0"
```
SleeplessOne1917, line 38: Don't buttons have tabIndex 0 by default?

src/shared/components/common/emoji-picker.tsx
```diff
@@ -23,6 +23,11 @@ export class EmojiPicker extends Component<EmojiPickerProps, EmojiPickerState> {
     this.state = this.emptyState;
     this.handleEmojiClick = this.handleEmojiClick.bind(this);
   }
+
+  closeEmojiMartOnEsc(i, event) {
```
SleeplessOne1917, line 27: Can be defined at module level instead of as method.

src/shared/components/common/emoji-picker.tsx
```diff
@@ -59,6 +64,10 @@ export class EmojiPicker extends Component<EmojiPickerProps, EmojiPickerState> {
   togglePicker(i: EmojiPicker, e: any) {
     e.preventDefault();
     i.setState({ showPicker: !i.state.showPicker });
+
+    i.state.showPicker
+      ? document.addEventListener("keyup", e => i.closeEmojiMartOnEsc(i, e))
+      : document.removeEventListener("keyup", e => i.closeEmojiMartOnEsc(i, e));
```
SleeplessOne1917, line 70: There should be a componentWillUnmount to remove the listener as well.

SleeplessOne1917 (changes requested):

src/shared/components/app/app.tsx
jsit: Ah of course, I was using this to debug some weird behavior but can remove it now, thanks.

SleeplessOne1917 (approved):

PR #1570, opened 2023-06-24
fix(a11y): Fixing some small a11y issues

jsit:
Fix some small accessibility issues


src/shared/components/common/searchable-select.tsx
```diff
@@ -116,17 +119,14 @@ export class SearchableSelect extends Component<
             ? `${I18NextService.i18n.t("loading")}${loadingEllipses}`
             : options[selectedIndex].label}
         </button>
-        <div
-          role="combobox"
-          aria-activedescendant={options[selectedIndex].label}
```
SleeplessOne1917, line 121: Is active descendent not needed?

jsit: Ah you're right, it is, I did some brief research on it but got it wrong, this should be there.

SleeplessOne1917 (approved):

PR #1583, opened 2023-06-25
Fix: missing semantic css classes and html elements

0xAnansi:
Now all pages have a main and aside element when a sidebar is present to facilitate custom theming. This does not impact the default behavior of the front.

~~Note: removed `ref={this.mainContentRef}` from the div element since it was flagged as an error, can't see  any impact but I don't know what it does in the first place.~~

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -22,7 +22,7 @@ import {
 } from "@utils/helpers";
 import type { QueryParams } from "@utils/types";
 import { RouteDataResponse } from "@utils/types";
-import { Component, linkEvent } from "inferno";
+import { Component, RefObject, createRef, linkEvent } from "inferno";
 import { RouteComponentProps } from "inferno-router/dist/Route";
 import {
   AddAdmin,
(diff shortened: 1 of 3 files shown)


src/shared/components/community/community.tsx
```diff
@@ -22,7 +22,7 @@ import {
 } from "@utils/helpers";
 import type { QueryParams } from "@utils/types";
 import { RouteDataResponse } from "@utils/types";
-import { Component, linkEvent } from "inferno";
+import { Component, RefObject, createRef, linkEvent } from "inferno";
```
jsit, line 25: I'm not sure why this additional `ref` stuff is necessary?

jsit (changes requested):

src/shared/components/community/community.tsx
0xAnansi: This was set in the app component for this page, so had to move it in this component to be able to link it in the "main" element that was moved from app to community.
This can only be set on main elements for some reason, or it will be considered an error.

This is used for the "skip to main content" feature.

dessalines (approved):

PR #1584, opened 2023-06-25
Large screen style

### LINKED ISSUE #1536: Add CSS styles for min-width beyond 1200px
### Requirements

I've got a browser window width spanning just over 2500 pixels.  The existing `@media (min-width: 1200px)` styles are rather limiting, considering how `.container` and its mates are forced to a `max-width` of `1140px`.  So, over half of my browser window on the front page is empty space.

Just getting rid of the `max-width` on the CSS solves part of it.  But, there's also other `.col-*` properties that could re-adjusted with their `flex` and `max-width` columns.  The right-pane needs to be shrunk, possibly even to a fix size of 300px or so, and the left-pane expanded as far as it can.

SleeplessOne1917:
Closes #1536. Here is what the pages look like now:

# Home
image

# Community Page
image

# Communities
image

# Create Post
image

# Create Community
image

# Search
image

# Admin Settings
image

# Inbox
image

# Reports
image

# Modlog
image

# Instances
image

I still have to do settings and profile. Will update the PR with screenshots when done.


jsit:
Hm it seems like some of these changes a are unrelated to screen widening?

SleeplessOne1917:
I also removed a bit of unnecessary markup.

src/shared/components/person/verify-email.tsx
```diff
@@ -53,20 +53,18 @@ export class VerifyEmail extends Component<any, State> {
...
-              </h5>
-            )}
-          </div>
+        <div className="col-12 col-lg-6 offset-lg-3 mb-4">
+          <h5>{I18NextService.i18n.t("verify_email")}</h5>
+          {this.state.verifyRes.state === "loading" && (
+            <h5>
+              <Spinner large />
+            </h5>
+          )}
```
SleeplessOne1917, line 67: All this page displays is a title and a loading spinner. @dessalines should we still keep this file?

src/assets/css/main.css
```diff
@@ -276,7 +277,10 @@ hr {
 }
 
 .mini-overlay {
-  display: block;
```
jsit, line 279: This feels like it doesn't belong in this PR?

src/assets/css/main.css
```diff
@@ -431,3 +435,15 @@ br.big {
...
+.comment,
+.markdown-textarea {
+  max-width: 50rem;
+}
+
+input[type="checkbox"],
+.form-check label,
+select,
+option {
+  cursor: pointer;
```
jsit, line 448: This feels like it doesn't belong in this PR; plus I'm not sure pointers for these things is correct.

src/shared/components/comment/comment-nodes.tsx
```diff
@@ -89,7 +89,11 @@ export class CommentNodes extends Component<CommentNodesProps, any> {
             "ms-1": !!this.props.isChild,
             "border-top border-light": !this.props.noBorder,
           })}
-          style={`border-left: 2px solid ${borderColor} !important;`}
+          style={
```
jsit, line 92: This feels separate from this PR

src/shared/components/community/communities.tsx
```diff
@@ -145,57 +145,41 @@ export class Communities extends Component<any, CommunitiesState> {
...
                           <CommunityLink community={cv.community} />
                         </td>
-                        <td className="text-right">
-                          {numToSI(cv.counts.subscribers)}
-                        </td>
-                        <td className="text-right">
-                          {numToSI(cv.counts.users_active_month)}
-                        </td>
-                        <td className="text-right d-none d-lg-table-cell">
+                        <td>{numToSI(cv.counts.subscribers)}</td>
```
jsit, line 151: Why remove `text-right`? This was intentional so numbers are lined up

jsit (changes requested):

src/shared/components/community/communities.tsx
SleeplessOne1917: I can make them right aligned again, but `text-right` isn't a valid bootstrap class in v5 regardless.

src/assets/css/main.css
SleeplessOne1917: Should selects not have the cursor be a pointer? The label? both?

src/assets/css/main.css
SleeplessOne1917: I can make it a separate PR if you think that's appropriate. Keeping it the old way makes the thumbnail icon show up incorrectly image thumbnails.

src/shared/components/comment/comment-nodes.tsx
SleeplessOne1917: I can make this a separate PR.

src/assets/css/main.css
jsit: I'm kind of a stickler in that I think the default cursor styles for form elements is handled correctly by browsers by default -- a standard `default` arrow cursor, like it is on every other form element in your operating system. The `pointer` style was designed for hyperlinks.

I know this convention has changed in recent years but I'm not sure it's necessarily the "right" thing to do.

src/shared/components/community/communities.tsx
jsit: Oh ha you're right -- I had to do this in #1538 too

src/assets/css/main.css
jsit: Ah, I remember how I handled this now. I added these classes to the thing element wrapping it:

https://github.com/LemmyNet/lemmy-ui/pull/1570/files#diff-cb34074ed5916cfbee02ee325240476c1ee8dd515e9e4559ffff1eaeec039f53R50

I think this makes more sense, to absolutely position the wrapper element (link/button) and keep the icon itself just positioned naturally within it.

jsit:
Post Listing images are now "portrait"-oriented, and are no longer top-aligned with the post title:
Screenshot 2023-06-25 at 1 11 43 PM

jsit:
Kind of throwing a wrench into things here, but I'm not sure this is even desirable in the first place? I don't know that this makes good use of the extra space gained; it will mostly just make things further apart.

I could see changing the max-width to something else (in fact it's larger in Bootstrap 5, and I overwrote the value of `$max-width` when switching to BS5 to maintain the existing look), but indefinite width feels weird to me.

Screenshot 2023-06-25 at 1 14 21 PM

jsit:
From the original issue:

> over half of my browser window on the front page is empty space.

There's still empty space; it's now just between the main and the sidebar.

I think this is this one user's personal preference and shouldn't necessarily be globally adopted.

alectrocute:
> From the original issue:
> 
> > over half of my browser window on the front page is empty space.
> 
> There's still empty space; it's now just between the main and the sidebar.
> 
> I think this is this one user's personal preference and shouldn't necessarily be globally adopted.

I agree completely.

SleeplessOne1917:
Any thoughts on how to make it better? I think there's still too much whitespace but also an improvement on what's there currently.

jsit:
1. Increase the max-width of the whole site
2. Change the sidebar to be 3 columns, not 4, on wide

SleeplessOne1917:
> Post Listing images are now "portrait"-oriented, and are no longer top-aligned with the post title:

I thought the top aligned listing images looked funny on post listings with more vertical space. I can change it back if you'd prefer.

jsit:
Here's an example of another way we might address this: #1590

src/shared/components/search.tsx
```diff
@@ -526,13 +526,8 @@ export class Search extends Component<any, SearchState> {
       creatorSearchOptions,
       searchCommunitiesLoading,
       searchCreatorLoading,
-      communitiesRes,
     } = this.state;
 
-    const hasCommunities =
```
jsit, line 532: Why remove this?

jsit:
It seems like there's a lot going on here that, while good improvements, don't really have to do with making things wider. Can these be moved to a separate chore PR?

src/shared/components/search.tsx
SleeplessOne1917: Doh. I'll put it back.

SleeplessOne1917:
What specifically do you think doesn't belong in this PR? I've gotten rid of everything not related to displaying on larger screen sizes at this point.

jsit:
Some of this looks a little wonky to me? This sidebar looks too small, and the comments aren't left-aligned with the post content

Screenshot 2023-06-25 at 3 58 30 PM

I'd kind of like to see #1590 merged in first, then some more fine-tuning done if necessary.

I also like how this PR makes some semantic fixes, but they make the diffs harder to read than they need to be for such a large PR.

It might be easier if this were broken up into smaller PRs for each page.

PR #1588, opened 2023-06-25
Add margin to new comment text

SleeplessOne1917:
Adds some space between the comment symbol and new comment symbol on a post.
image

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -741,9 +741,12 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         <Icon icon="message-square" classes="me-1" inline />
         {post_view.counts.comments}
         {this.unreadCount && (
-          <span className="text-muted fst-italic ms-1">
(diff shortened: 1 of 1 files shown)


jsit:
I think I might prefer this as a `{" "}`, so that it's an actual space character in the current font

src/shared/components/post/post-listing.tsx
```diff
@@ -741,9 +741,12 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         <Icon icon="message-square" classes="me-1" inline />
         {post_view.counts.comments}
         {this.unreadCount && (
-          <span className="text-muted fst-italic">
-            ({this.unreadCount} {I18NextService.i18n.t("new")})
-          </span>
+          <>
+            {" "}
+            <span className="text-muted fst-italic">
```
jsit, line 746: Can we safely remove the `text-muted`, since it's inside a `text-muted`?

alectrocute: @jsit By looking at the code, yes, we can safely remove.

alectrocute (approved):

PR #1591, opened 2023-06-25
feat(UI): Reduce base font size

jsit:
I think it's time. People seem to really not like how much space everything takes up. This could stand to be smaller, especially since so much of the text is UI, not body copy. The styles are responsive enough that people can zoom in and everything will work great.

## Before

Screenshot 2023-06-25 at 2 45 02 PM

## After

Screenshot 2023-06-25 at 2 45 33 PM


alectrocute (approved):
Let’s hear from the other contributors before merging, but looks good to me.

SleeplessOne1917 (approved):

dessalines (approved):

PR #1594, opened 2023-06-25
Only give child comments colored borders

SleeplessOne1917:


jsit:
Heh -- I think maybe that's what this old `hasPadding` thing was for (I forget the name of the var):

Screenshot 2023-06-25 at 3 31 46 PM

Also, I wonder about this -- it makes it a little less clear (to me) that this is a reply:

Screenshot 2023-06-25 at 3 33 23 PM

versus this:

Screenshot 2023-06-25 at 3 34 19 PM

SleeplessOne1917:
> it makes it a little less clear (to me) that this is a reply:

I think it's clearer. Top level comments have no border while replies do.

jsit (approved):

jsit:
It would be nice if the first-level children didn't have a `ms-2` on them, but not a big deal.

Issue #1595, opened 2023-06-25
Put body text behind a toggle on post listing pages


jsit:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Put body text behind a toggle on post listing pages. This will save vertical space and allow more posts to be shown.

jsit:
Fixed in #1596

PR #1596, opened 2023-06-25
feat(UI): Always put post body behind toggle on post listings

jsit:
Completes #1595

Thoughts on this? The body showing up in post listings was occupying a lot of vertical space, and from the sound of it, people don't like this.

With this PR:

1. On a Listing Page (home/community), there is an expand/collapse icon next to the title that will show both (a) the post body, and (b) the link preview card
2. On a Post Page, the link preview card always shows (if there is one). **There is no plus/minus expand/collapse icon on the post page.*

### Listing Page Unexpanded

Screenshot 2023-06-26 at 8 14 41 PM

### Listing Page Expanded

Screenshot 2023-06-26 at 8 14 54 PM

### Post Page Unexpanded

Screenshot 2023-06-26 at 8 14 30 PM

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1226,12 +1226,12 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
                 value={this.state.banReason}
                 onInput={linkEvent(this, this.handleModBanReasonChange)}
               />
-              <label className="col-form-label" htmlFor={`mod-ban-expires`}>
+              <label className="col-form-label" htmlFor="mod-ban-expires">
                 {I18NextService.i18n.t("expires")}
               </label>
               <input
                 type="number"
-                id={`mod-ban-expires`}
+                id="mod-ban-expires"
                 className="form-control me-2"
                 placeholder={I18NextService.i18n.t("number_of_days")}
                 value={this.state.banExpireDays}
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1115,7 +1115,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
             this.creatorIsMod_ &&
             (!this.state.showConfirmTransferCommunity ? (
               <button
-                className="btn btn-link btn-animate text-muted py-0 dropdown-item"
+                className="btn btn-link btn-animate text-muted py-0"
                 onClick={linkEvent(
                   this,
                   this.handleShowConfirmTransferCommunity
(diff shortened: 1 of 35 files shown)


SleeplessOne1917 (approved):
I prefer it without the body preview.

jsit:
Waiting to hear from @alectrocute @dessalines

alectrocute:
@jsit Which one’s the post body toggle? The book icon? It’s not immediately clear to me how to preview the body. Maybe that’s something that can be addressed in this PR?

Either way, I don’t like the post body previews either.

jsit:
@alectrocute Yeah, I'd like to get it right in this PR. With images, there's a plus button. It seems like there are three things that need a button:

1. Expand image
2. Expand body
3. Show link preview text

Which button should we use for each of these, and where should it be? Should they all be the same?

jsit:
@alectrocute It might be best to get rid of that "book" icon (was always confusing to me) and put everything behind a "plus/minus" icon.

alectrocute:
> @alectrocute Yeah, I'd like to get it right in this PR. With images, there's a plus button. It seems like there are three things that need a button:
> 
> 1. Expand image
> 2. Expand body
> 3. Show link preview text
> 
> Which button should we use for each of these, and where should it be? Should they all be the same?

Yep, right now it's all in `PostListing`'s `thumbnail()` return. It'd be nice to refactor that, make it a lot simpler and have it be responsible for text-only post body previews (or just expand the whole entire body, why not?) as well as image, video, embed (eg. YouTube) previews.

The book icon should probably be sunset. It doesn't even have a tooltip/label.

alectrocute:
It would give me great joy to see `PostListing` be refactored and greatly simplified.

jsit:
@alectrocute I've been focusing on it a lot, it's definitely needs to be broken up.

jsit:
@alectrocute @SleeplessOne1917 Changed the behavior here and updated the description to show that.

src/shared/components/post/post-listing.tsx
```diff
@@ -1780,6 +1755,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     setupTippy();
   }
 
+  // handleShowPreview(i: PostListing) {
```
SleeplessOne1917, line 1758: Any reason to keep this commented out function?

jsit: Ah, no -- was planning ahead. Deleted.

alectrocute (approved):

jsit:
@alectrocute @SleeplessOne1917 One final though -- on a Post page, do we *want* the "preview card" to be behind an expand click? Or could we just always show it?

dessalines:
> do we want the "preview card" to be behind an expand click? Or could we just always show it?

IMO the metadata-card should always be shown on the post page, and doesn't need a button.. For example, in github release posts (where I usually have a link, but no post body), the metadata-card automatically shows a few lines of the changelog. Metadata cards aren't usually that tall, and should be summaries. IMO we shouldn't really care too much about vertical height on the post page, because post bodies can already be really long.

This was already referenced, but yes these two buttons conflict, and its not especially clear which one does which:

Screenshot_2023-06-26-15-03-26-072_mark via gp

jsit:
@dessalines Great, thanks for the feedback. I'll work on this PR a bit to get things to work the way you described.

jsit:
@dessalines @alectrocute I've changed the behavior and updated the description. The expand/collapse button will only show on "listing" pages, and will show both the post body and the link preview card.

dessalines (approved):

PR #1597, opened 2023-06-25
feat(UI): Make post listing (icon) avatars smaller (line-height)

jsit:
Need opinions on this. This will reduce some visual noise and also tighten things up vertically a little.

## Before

Screenshot 2023-06-25 at 4 37 21 PM

## After

Screenshot 2023-06-26 at 6 06 42 PM
Screenshot 2023-06-26 at 6 09 19 PM

### DIFF
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -39,7 +39,7 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
             "img-expanded slight-radius":
               !this.props.thumbnail && !this.props.icon,
             "img-blur": this.props.thumbnail && this.props.nsfw,
-            "rounded-circle img-cover img-icon me-2": this.props.icon,
+            "img-cover img-icon me-1": this.props.icon,
             "ms-2 mb-0 rounded-circle img-cover avatar-overlay":
               this.props.iconOverlay,
             "avatar-pushup": this.props.pushup,
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -106,6 +106,7 @@ interface PostListingProps {
   siteLanguages: number[];
   showCommunity?: boolean;
   showBody?: boolean;
+  hideAvatars?: boolean;
   hideImage?: boolean;
   enableDownvotes?: boolean;
   enableNsfw?: boolean;
@@ -397,7 +398,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     const post_view = this.postView;
     return (
       <span className="small">
-        <PersonListing person={post_view.creator} />
+        <PersonListing
+          person={post_view.creator}
+          hideAvatar={this.props.hideAvatars}
+        />
         {this.creatorIsMod_ && (
           <span className="mx-1 badge text-bg-light">
             {I18NextService.i18n.t("mod")}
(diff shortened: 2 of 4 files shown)


SleeplessOne1917 (approved):
I think it looks better. Waiting for others' feedback before merge.

alectrocute (approved):
I think it looks better, too.

jsit:
@alectrocute @SleeplessOne1917 This feels like a relatively large decision, should we wait for @dessalines to weigh in?

alectrocute:
> @alectrocute @SleeplessOne1917 This feels like a relatively large decision, should we wait for @dessalines to weigh in?

Yep, let’s do.

dessalines:
I am going to dissent and say I don't like it too much. What I discovered in playing around with this in jerboa, was that the icon size was actually more important in making things look cleaner: Right now, they're wayy too big in lemmy-ui. In jerboa they're properly sized to the text, and it makes a lot of difference.

post list-view: 

Screenshot_2023-06-26-15-13-23-905_com jerboa

post card-view:

Screenshot_2023-06-26-15-16-49-006_com jerboa

And it smartly hides the community icon on the community page.

The other thing is that icons, even if they're tiny, really help with instant recognition of users and communities. When you have icons in one place but not the other, its throws that off.

jsit:
@dessalines I hear you -- glad to have gotten your feedback on this before merging. Would it be a welcome PR to make them a little smaller in the post list view?

alectrocute:
> @dessalines I hear you -- glad to have gotten your feedback on this before merging. Would it be a welcome PR to make them a little smaller in the post list view?

We should definitely at least make the avatars line-height IMO.

dessalines:
@jsit it'd be a good idea to make them match the text line height closely everywhere (except the tops of the profile and community pages).

jsit:
@alectrocute @dessalines I've restored the icon size, and made the height/width equal to one line height. Updated screenshots in description.

SleeplessOne1917 (changes requested):
> I've restored the icon size, and made the height/width equal to one line height. Updated screenshots in description.

I think you might have borked a merge, because only code changes I see in the PR currently is passing `hideAvatars` throught post listings.

jsit:
@SleeplessOne1917 Oops you're right, I forgot to push. Should be good now.

src/shared/components/common/pictrs-image.tsx
```diff
@@ -39,7 +39,7 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
             "img-expanded slight-radius":
               !this.props.thumbnail && !this.props.icon,
             "img-blur": this.props.thumbnail && this.props.nsfw,
-            "rounded-circle img-cover img-icon me-2": this.props.icon,
```
SleeplessOne1917, line 42: There`s an `me-` without a number at the end of the classnames.

src/shared/components/post/post-listing.tsx
```diff
@@ -106,6 +106,7 @@ interface PostListingProps {
   siteLanguages: number[];
   showCommunity?: boolean;
   showBody?: boolean;
+  smallAvatars?: boolean;
```
SleeplessOne1917, line 109: Do you have more changes you forgot to push? `smallAvatar` is defined as a prop but not passed in anywhere. I'm assuming there are also supposed to be changes to person and community listings that this prop gets passed down to.

SleeplessOne1917 (changes requested):

src/shared/components/common/pictrs-image.tsx
jsit: It's a `1`; the color contrast might have gotten you here

src/shared/components/post/post-listing.tsx
jsit: Missed that one, thanks. Removed and pushed.

alectrocute:
Super excited for this PR. Thanks @jsit

SleeplessOne1917 (approved):

PR #1617, opened 2023-06-26
Add `InboxService` to handle unread notification state

alectrocute:
Hi Lemdevs!

This PR:

- Add `InboxService` to handle source of truth for state related to unread notifications

Thanks!


src/shared/components/app/navbar.tsx
```diff
@@ -443,67 +429,16 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
...
   }
 
-  get unreadApplicationCount(): number {
-    if (this.state.unreadApplicationCountRes.state == "success") {
-      const data = this.state.unreadApplicationCountRes.data;
-      return data.registration_applications;
-    } else {
-      return 0;
-    }
+  get unreadApplicationCount() {
```
dessalines, line 440: You can remove these three `get` functions, and just reference the `InboxService`

dessalines (commented):
Nice, this is much cleaner.

src/shared/components/person/inbox.tsx
```diff
@@ -786,6 +791,8 @@ export class Inbox extends Component<any, InboxState> {
         mentionsRes: { state: "empty" },
         messagesRes: { state: "empty" },
       });
+
+      InboxService.fetchUnreadCounts();
```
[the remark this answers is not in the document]
alectrocute: I'm pretty sure this one's important, to get the latest unread count to the navbar when you click the "Mark all as read" button.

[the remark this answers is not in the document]
dessalines: Ah okay, for some reason I thought this was run on initial fetch.

dessalines:
@alectrocute That's because we're not re-fetching that on a mark as read action. We're only currently re-fetching that periodically every 30 seconds. We probably should tho.

alectrocute:
> @alectrocute That's because we're not re-fetching the unreads on a mark as read action. We're only currently re-fetching that periodically every 30 seconds. We probably should tho.

No, no. Sorry, let me re-explain. In the video I attached above, there *is* a call to re-fetch the counts upon clicking the mark as read button. It's updating, but the `navbar` doesn't re-render. If I click anywhere on the page, it re-renders.

dessalines:
I'm guessing because the unread counts are no longer in the navbar's state, so it has no idea that an update occurred. You might have to have these unread counts as observables, and subscribe to them in the navbar's `componentDidMount`

src/shared/services/InboxService.ts
```diff
@@ -17,11 +18,20 @@ export class InboxService {
...
     this.unreadInboxCountRes = { state: "empty" };
     this.unreadReportCountRes = { state: "empty" };
     this.unreadApplicationCountRes = { state: "empty" };
 
+    this.observable = new Observable(s => {
+      this.subscriber = s;
+    });
+
+    this.subscriber = null;
+
```
alectrocute, line 34: This feels so... wrong.

src/shared/components/app/navbar.tsx
```diff
@@ -44,6 +44,12 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
     this.handleOutsideMenuClick = this.handleOutsideMenuClick.bind(this);
   }
 
+  componentWillMount(): void {
+    InboxService.observable.subscribe(() => {
+      this.render();
```
alectrocute, line 49: Trying to get things to re-render here. It's not working still.

dessalines: Its more that on the initial `componentDidMount` , it would need to subscribe to each of those observables, and this would need a state to observe them. 

Observers are just one way of sharing state between non-parent-child components, some others are mentioned here that might be cleaner: https://stackoverflow.com/questions/38901106/how-to-make-a-shared-state-between-two-react-components

dessalines:

PR #1618, opened 2023-06-26
Remove thumbnail expand icon on `PostListing`'s `postTitleLine`

alectrocute:
Hi Lemdevs!

In this PR:

- Remove extra expand icon on `PostListing`'s `postTitleLine`
- Why would a user want to expand a thumbnail? It's a confusing button.

Screenshot 2023-06-26 at 2 33 39 PM

Thanks!

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -482,26 +482,13 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
               this.postLink
             )}
           </h5>
-          {(url && isImage(url)) ||
-            (post.thumbnail_url && (
-              <button
-                className="btn btn-sm text-monospace text-muted d-inline-block"
-                data-tippy-content={I18NextService.i18n.t("expand_here")}
-                onClick={linkEvent(this, this.handleImageExpandClick)}
-              >
-                <Icon
-                  icon={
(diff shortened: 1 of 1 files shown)


jsit:
> Why would a user want to expand a thumbnail? It's a confusing button.

Why wouldn't they? Isn't this only for image posts? Could be nice, rather than navigating to the domain where the image is hosted.

dessalines:
I agree, image expanding isn't really necessary for link posts. But it is for image posts, as that's essentially one of the best utilities of RES.

src/shared/components/post/post-listing.tsx
```diff
@@ -482,26 +482,13 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
               this.postLink
             )}
           </h5>
-          {(url && isImage(url)) ||
-            (post.thumbnail_url && (
```
dessalines, line 486: I think all that's necessary is to remove the or condition. We do want to show an expando for image posts, just not for link posts.

jsit:
Should this be handled in #1596? I sense a conflict coming.

alectrocute:
I thought clicking the thumbnail *is* the expando for image posts?

dessalines:
> I thought clicking the thumbnail is the expando for image posts?

Oh yeah I forgot. In that case yep it makes sense to remove this. That'll help @jsit 's PR as it removes the conflicting icons.

PR #1620, opened 2023-06-26
Add space between cancel and ban button

pascaldevink:
When banning a user, the modal has two buttons: cancel and ban. These buttons are so close to each other that it's hard to see they are separate buttons.

By putting each button on their own row, there is more space between them.

Before:
CleanShot 2023-06-26 at 20 55 06

After:
CleanShot 2023-06-26 at 20 53 54


src/shared/components/person/profile.tsx
```diff
@@ -693,6 +693,7 @@ export class Profile extends Component<
             >
               {I18NextService.i18n.t("cancel")}
             </button>
```
dessalines, line 695: Just add mb-3 to the button, not adding a pointless row like this.

dessalines (changes requested):

jsit:
Rather than adding an empty div, put a margin bottom and display block on the existing button.

alectrocute (approved):

dessalines (approved):

PR #1626, opened 2023-06-26
Decouple API calls from top level components.

dudeami:
This PR is demonstrates decoupling of low level components from top level components. Right now interactions bubble from the originating component to the top component for that page. The state change propagates back down from the top, triggering costly renders of large components. Handling these interactions in components closer to the originating action has the benefit of:

- Limiting update rendering to smaller components, increasing performance
- Less boilerplate code between components
- Less code in top level components (`Home`, `Community`, etc)

Disadvantages I see are API related state updates are not saved on a global level. This does currently effect how state is managed with the `HomeCacheService`, since it copies `state.postsRes`. This approach is however already problematic as state is not retained, so collapsing/expanding of elements resetting is already an issue.

I propose fixing the coupling and state issues by moving API calls closer to the components they effect, and creating a new method to create a snapshot of the listings for `HomeCacheService`.

This is just an example and request for comments. I will continue to do the rest of the API calls in this fashion unless there is some objection.

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -58,7 +58,7 @@ import {
   VoteContentType,
 } from "../../interfaces";
 import { mdToHtml, mdToHtmlNoImages } from "../../markdown";
-import { I18NextService, UserService } from "../../services";
+import { HttpService, I18NextService, UserService } from "../../services";
 import { setupTippy } from "../../tippy";
 import { Icon, PurgeWarning, Spinner } from "../common/icon";
 import { MomentTime } from "../common/moment-time";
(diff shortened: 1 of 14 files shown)


dudeami:
I agree with the separation of logical and presentation components, but this seems to be putting the logic in the incorrect component (this PR being too nested and the current approach being above the common component). The amount of duplicate code attests to this, as `Home`, `Community`, `Inbox`, `Profile`, and `Post` all need a `handleCommentVote` which does the same thing violating the DRY principle. This is repeated for every API action related to posts/comments. I would think these should be moved to a common logical component that will handle these actions, or the handlers should be abstracted in some way.

It seems I was also misattributing the reset in state to the current design. This seems to be the fault of the `componentWillReceiveProps` triggering a reset of the entire component state, triggering a render and resetting collapsed states of comments on upvote. Commenting out this code seems to fix this issue (but it's probably there for a reason).

PR #1638, opened 2023-06-27
Reduce fetchLimit (40 -> 20)

sunaurus:
This PR proposes reducing the default page size. It's certainly a change with some trade-offs, so feel free to reject it (or debate the extent of the reduction), but I will provide my argument below:

On lemm.ee, I have noticed that this small tweak has **vastly** improved the experience for our users. By reducing the page size in half, our page loads significantly faster, and server load has in general decreased slightly as well.

I will also share some actual stats. This was with `fetchLimit` set to 40:
image

And this is the same test run with `fetchLimit` reduced to 20:
image

Of course, the downside with this is that users need to load additional pages more often, but it seems like a small price to pay considering the performance benefits.

### DIFF
--- a/src/shared/config.ts
+++ b/src/shared/config.ts
@@ -21,7 +21,7 @@ export const markdownFieldCharacterLimit = 50000;
 export const maxUploadImages = 20;
 export const concurrentImageUpload = 4;
 export const updateUnreadCountsInterval = 30000;
-export const fetchLimit = 40;
+export const fetchLimit = 20;
 export const relTags = "noopener nofollow";
 export const emDash = "\u2014";
 


dessalines (approved):
I'm good with this. I know the solution to optimize our post_view queries now, but that might be a week or two yet till I can add it.

SleeplessOne1917:
This seems like a good short term solution. I think part of the problem is also that postviews are very big blobs of json, so even fetching a few is going to mean a big response from the API.

SleeplessOne1917 (approved):

alectrocute (approved):

Issue #1256, opened 2023-06-14
Blur NSFW community icons


ghost:
**Is your proposal related to a problem?**

The issue is that the pictures are blurred while the icons remain unaffected.

**Describe the solution you'd like.**

I would like either both NSFW icons and pictures to be blurred or neither of them.

**Describe alternatives you've considered.**

Users could be provided with the choice to display, blur and display on mouse over, blur, or hide all NSFW pictures and icons. This is related to #1213

**Additional context**

_No response_

dessalines:
Can re-open if someone wants to add this.

PR #1640, opened 2023-06-27
Frontend Settings - "Blur NSFW" and "Auto Expand"

tgxn:
This adds settings for "Blur NSFW" and "Auto Expand".
image

Backend Changes: https://github.com/LemmyNet/lemmy/pull/3377

Required Translations:
```json
  "blur_nsfw": "Blur NSFW content",
  "auto_expand": "Auto expand media"
```


tgxn:
Icons are now blurred. https://github.com/LemmyNet/lemmy-ui/issues/1256
image

A few things I'd also like to do/think about:
- Should NSFW content be auto-expanded? If they have "Show NSFW" on, "Blur NSFW" on, then when content is auto-expanded it should be blurred, or not expanded?
- Should we listen for scroll events, and only expand posts when they are in view? (This is how it is done by https://git.lemmynsfw.com/lemmynsfw/lemmy-ui/compare/0.18.0...0.18.0-nsfwpatch)
- Should the community icon be blurred when looking at the community? (ie. the `http://localhost:1234/c/testnsfw` page)

alectrocute:
> Icons are now blurred. #1256 image
> 
> A few things I'd also like to do/think about:
> 
> * Should NSFW content be auto-expanded? If they have "Show NSFW" on, "Blur NSFW" on, then when content is auto-expanded it should be blurred, or not expanded?
> * Should we listen for scroll events, and only expand posts when they are in view? (This is how it is done by https://git.lemmynsfw.com/lemmynsfw/lemmy-ui/compare/0.18.0...0.18.0-nsfwpatch)
> * Should the community icon be blurred when looking at the community? (ie. the `http://localhost:1234/c/testnsfw` page)

#1 and #2 are a no from me. #3 makes sense.

Also, it would be great to give our blurs hard edges (eg. `overflow: hidden;`).

Let’s let the other contributors chime in.

tgxn:
> #1 and #2 are a no from me. #3 makes sense.
For 1. I was thinking that it's a little confusing of what the expected outcome is, do you want to see NSFW, or have it hidden, or have everything auto-expanded? Maybe we can think about combining this with the spoiler feature somehow? https://github.com/LemmyNet/lemmy/issues/3185
Agreed on point 2, I'd rather just hide them or expand them, whatever the user chooses.

> Also, it would be great to give our blurs hard edges (eg. `overflow: hidden;`).
I was just looking at how to achieve that! I found reducing the radius helps, but a hard overflow makes sense.
image

> Let’s let the other contributors chime in.
100%, I'd love to see if these changes make sense to everyone.

dessalines:
These fields are included now in lemmy-js-client : `0.19.0-rc.1`

src/shared/components/common/pictrs-image.tsx
```diff
@@ -21,6 +23,14 @@ export class PictrsImage extends Component<PictrsImageProps, any> {
   }
 
   render() {
+    let user_blur_nsfw = true;
+    if (UserService.Instance.myUserInfo) {
+      user_blur_nsfw =
+        UserService.Instance.myUserInfo?.local_user_view.local_user.blur_nsfw;
+    }
+
+    const blur_image = this.props.nsfw && user_blur_nsfw;
```
SleeplessOne1917, line 32: `undefined` is falsy in JS/TS, so this can be simplified:

```suggestion
    const blur_image = this.props.nsfw && UserService.Instance.myUserInfo?.local_user_view.local_user.blur_nsfw;
```

src/shared/utils/app/setup-date-fns.ts
```diff
@@ -5,7 +5,7 @@ const EN_US = "en-US";
 
 export default async function () {
   let lang = I18NextService.i18n.language;
-  if (lang === "en") {
+  if (lang === "en" || lang === "en-AU") {
```
SleeplessOne1917, line 8: Was datefns having trouble with `en-AU`?

tgxn: Yeah, it wasn't working in my browser without this... Can remove it, it might have just been my machine... I can't seem to start `lemmy-ui` anymore either. I'll have another look tomorrow. 😖
image

tgxn: I managed to get it working by changing the build:dev command, and this `en-AU` doesn't seem to make any difference anymore, weird.  I'll revert it back 😇

src/shared/components/common/pictrs-image.tsx
tgxn: I agree this looks better, I had it default to `true`, but only logged-in users will be able to see NSFW, so if they are logged out, it's fine if the default is false 🤷‍♂️.

I think there should be some admin-settings at some point to control both logged-out NSFW visibility blur settings, but that's another story.

SleeplessOne1917 (approved):

SleeplessOne1917:
Waiting for input from @dessalines, @alectrocute, and/or @jsit before merging.

dessalines (approved):
LGTM, thanks!

Issue #1375, opened 2023-06-18
[Bug]: Lemmy.ml - profile menu had another user's username


drspod:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [ ] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

### Summary

I just loaded the front page of lemmy.ml and in the top right where my username (`drspod`) usually appears, there was another user's username (`glymfeather`).

![lemmy_bug](https://github.com/LemmyNet/lemmy-ui/assets/12265455/7d61e4de-9cd4-4844-ad9a-45053e0808c6)

The page is not using my selected theme (dark mode) and when I click the username or notification button, nothing happens, which is what I usually experience if the page has not loaded correctly.

Loading the page again in a new tab fixed the issue and showed me my front page with my theme and username in top right.

### Steps to Reproduce

1. Load the front page of https://lemmy.ml
2. The displayed page had been generated for a different user.

### Technical Details

Javascript console contains: 

    Loading failed for the <script> with source “https://lemmy.ml/static/js/client.js”. lemmy.ml:510:54
    Error: Promised response from onMessage listener went out of scope 12 init.js:75:34

in addition to some CSS errors and blocked cookie warnings.

### Version

0.17.4

### Lemmy Instance URL

lemmy.ml

drspod:
It just happened again:

![lemmy_bug2](https://github.com/LemmyNet/lemmy-ui/assets/12265455/b37efdee-fe62-44be-b040-b635faa40a73)

The posts in the feed appear to be my own subscriptions. It seems unlikely that the user whose name appears in the top-right is subscribed to the exact same communities as me. So the issue seems to be with putting incorrect user details in the top-right of the page, even though the feed was generated correctly (minus correct theme styling).

When I clicked the notifications button it took me to my own notifications page (correctly themed), so I was not able to view the other user's notifications. I don't know how this would be affected if the user had some active notifications that might create the popup toast text on the main page though.

alectrocute:
Cannot reproduce in 0.18. If this continues to be a problem, please comment or reach out and we'll address it.

rcmaehl:
Hi @alectrocute I can confirm this has been an issue since I started using Lemmy.

[...]

PR #1641, opened 2023-06-27
Set cache-control headers to reduce server load (fixes #412)

### LINKED ISSUE #412: Fix missing cache headers
https://git.radicallyopensecurity.com/nlnet/off-ngid-lemmy/-/issues/10

Nutomic:
See code comment for explanation.


SleeplessOne1917:
I don't see the code comment you're referring to in the changes.

Nutomic:
Sorry I forgot to commit some changes, check again now.

SleeplessOne1917 (approved):
There's a small change to make, but it's not consequential enough to mark as request changes.

alectrocute:
I’m not a fan of all the middleware being in the same file. We should keep it a folder, yes?

alectrocute:
Does this resolve https://github.com/LemmyNet/lemmy-ui/issues/1375?

src/server/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+  res,
+  next,
+}: {
+  res: Response;
+  next: NextFunction;
+}) {
+  const user = UserService.Instance;
+  let caching;
+  if (user.auth()) {
+    caching = "private";
```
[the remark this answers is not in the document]
pixlguru: Are these really necessary?

`unsafe-inline` and `unsafe-eval`  introduce serious security risks.

Please avoid usage or use with nonce

src/server/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+export function setDefaultCsp({
+  res,
+  next,
+}: {
+  res: Response;
+  next: NextFunction;
+}) {
+  res.setHeader(
+    "Content-Security-Policy",
+    `default-src 'self'; manifest-src *; connect-src *; img-src * data:; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; form-action 'self'; base-uri 'self'; frame-src *; media-src *`
```
pixlguru, line 13: Are these really necessary?

`unsafe-inline` and `unsafe-eval` introduce serious security risks.

Please avoid usage or use with nonce

Nutomic: Afaik they are required by the js framework, would be great if we could get rid of them.

src/server/middleware.ts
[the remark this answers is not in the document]
Nutomic: That will cause problems if you write a post, then go to the post listing which is cached, and is missing your new post. In any case the main server load is probably from unauthenticated users so this should already help a lot.

src/server/middleware.ts
dessalines: @pixlguru you can try removing these as a separate PR, but make sure you do real testing on it.

pixlguru: Maybe I could do this after 18.1 is out, if I remember correctly ,last time I took a look unsave-eval was used by websocket code, so there is a chance that removing it has no side-effects.

PR #1646, opened 2023-06-27
Display normal number up to 9999

johanlajili:
Maybe it's because i'm not american, but I find
image
to be very hard to read.

1150 takes the same amount of character and is much more readable and precise. Hence I propose to use normal numbers up to 9999.

### DIFF
--- a/src/shared/utils/helpers/num-to-si.ts
+++ b/src/shared/utils/helpers/num-to-si.ts
@@ -6,5 +6,8 @@ const SHORTNUM_SI_FORMAT = new Intl.NumberFormat("en-US", {
 });
 
 export default function numToSI(value: number): string {
+  if (value < 10000) {
+     return value.toString()
+  }
   return SHORTNUM_SI_FORMAT.format(value);
 }


src/shared/utils/helpers/num-to-si.ts
```diff
@@ -6,5 +6,8 @@ const SHORTNUM_SI_FORMAT = new Intl.NumberFormat("en-US", {
 });
 
 export default function numToSI(value: number): string {
+  if (value < 10000) {
+     return value.toString()
```
jsit, line 10: I wonder if it might be nice to use `toLocaleString()` to get the thousands separator in there?

jsit:
Thanks -- can you please post a screenshot of what the sidebar looks like when implemented, so it's easy to get a sense of how it changes things?

src/shared/utils/helpers/num-to-si.ts
```diff
@@ -6,5 +6,8 @@ const SHORTNUM_SI_FORMAT = new Intl.NumberFormat("en-US", {
 });
 
 export default function numToSI(value: number): string {
+  if (value < 10000) {
```
dessalines, line 9: I'm not a fan of this hack. I would consider using 2 instead of 3 significant figures tho when using the shortnum versions.

dessalines (changes requested):

PR #1649, opened 2023-06-27
Get rid of format button borders

SleeplessOne1917:
I noticed that the borders of the markdown format buttons will flash when submitting a post. This should stop that.


SleeplessOne1917:
Closing in favor of #1650

PR #1656, opened 2023-06-27
Move `getRoleLabelPill` to `@utils/app`

alectrocute:
In this PR:

- Move `getRoleLabelPill` to `@utils/app`

Appendum to #1623.

Thanks!

PR #1658, opened 2023-06-27
Remove hard-coded `getRoleLabelPill` text foreground color classes

alectrocute:
Appendum to #1656.

I did more testing with `lightly` and discovered we really aught to just use the default foreground colors for these pills, which means not manually specifying `text-black`/`text-white`, and just letting Bootstrap pick the default. This keeps things looking good across themes, see pics below.

Screenshot 2023-06-27 at 4 53 14 PM

Screenshot 2023-06-27 at 4 53 05 PM


jsit:
Ah, yes, the `text-bg-` classes should take care of foreground colors just fine

*In fact* I'm pretty sure Bootstrap 5 chooses the correct foreground color based on the `$min-contrast-ratio` Sass var we set (ours is set to 3, which is why that green color had to be darker in BS5)

SleeplessOne1917 (approved):

jsit (approved):

PR #1669, opened 2023-06-28
Fix search page to stop `couldnt_find_object` error

scme0:
When searching for anything that isn't a community or user link (starting with a '!' or '@') while logged in, an error appears as a toast on the search page, or if you're loading the search URL directly, it will show an error page.

Toast:
Screenshot 2023-06-28 at 23 29 50

Error page (when loading the search url with the query parameters directly):
Screenshot 2023-06-28 at 23 30 04

This is because we send a request to the server on the `resolve_object` endpoint which returns an error `couldnt_find_object` if the input is just random text and it doesn't match to any objects.

I've updated the front end to make the requests to `resolve_object` silent. This means that if there is any error, no toast is shown and it is not logged. I think this is acceptable because of the way we are using the endpoint here. we are just interested if it returns any valid results.

I'm unsure if there is a better way to implement a "silent client" though, so please let me know if you can think of something a bit more elegant.

Fixes https://github.com/LemmyNet/lemmy/issues/3296


src/shared/services/HttpService.ts
```diff
@@ -74,16 +76,23 @@ class WrappedLemmyHttpClient {
...
+  // unfortunately, this verbose cast is necessary
+  return new WrappedLemmyHttpClient(
+    client,
+    silent
+  ) as unknown as WrappedLemmyHttp;
 }
 
 export class HttpService {
   static #_instance: HttpService;
+  #silent_client: WrappedLemmyHttp;
```
scme0, line 89: Is there a nicer way to achieve this?

src/shared/components/search.tsx
```diff
@@ -332,9 +332,7 @@ export class Search extends Component<any, SearchState> {
   }
 
   async componentDidMount() {
-    if (
-      !(this.state.isIsomorphic || this.props.history.location.state?.searched)
-    ) {
+    if (!this.state.isIsomorphic) {
```
scme0, line 335: I changed this because it was causing no search results to display.

src/shared/services/HttpService.ts
SleeplessOne1917: I can't think of any.

src/shared/components/search.tsx
```diff
@@ -432,7 +430,15 @@ export class Search extends Component<any, SearchState> {
...
             auth,
           };
-          resolveObjectResponse = await client.resolveObject(resolveObjectForm);
+          resolveObjectResponse = await HttpService.silent_client.resolveObject(
+            resolveObjectForm
+          );
+
+          // If we return this object with a state of failed, the catch-all-handler will redirect
+          // to an error page, so we ignore it by covering up the error with the empty state.
+          if (resolveObjectResponse.state == "failed") {
```
SleeplessOne1917, line 439: Use triple equals instead of double equals.

SleeplessOne1917 (approved):

src/shared/components/search.tsx
```diff
@@ -1097,10 +1103,6 @@ export class Search extends Component<any, SearchState> {
       sort: sort ?? urlSort,
     };
 
-    this.props.history.push(`/search${getQueryString(queryParams)}`, {
-      searched: true,
-    });
-
-    await this.search();
+    this.props.history.push(`/search${getQueryString(queryParams)}`);
```
jsit, line 1106: Should we create a global `siteUrl` constant at some point, in case someone installs Lemmy in a subdirectory or something?

src/shared/services/HttpService.ts
```diff
@@ -61,8 +61,10 @@ class WrappedLemmyHttpClient {
               state: !(res === undefined || res === null) ? "success" : "empty",
             };
           } catch (error) {
-            console.error(`API error: ${error}`);
-            toast(I18NextService.i18n.t(error), "danger");
+            if (!silent) {
+              console.error(`API error: ${error}`);
+              toast(I18NextService.i18n.t(error), "danger");
```
jsit, line 66: The user will see a toast that says "danger"? Could this be less scary or somewhat informative?

src/shared/components/search.tsx
scme0: Good catch - will fix 👍

PR #1672, opened 2023-06-28
Add publicPath for static js

### LINKED ISSUE #1657: [Bug] - Client doesn't work if language of the web-browser is not set to one matching the server ones (and even ...)
### Requirements

### Summary

When using the client, if your browser is set to a language that is not part of the list of the instance language, content fails to load and you cannot see anything.
If your language is english, it won't accept anything else than en-US for your browser.

### Steps to Reproduce

1. Set your instance to english, french, whatever
2. Set your browser's language as English (canadian) for example
3. Check your instance's posts
4. Nothing is shown
5. Console shows that a javascript chunk does not load

```
Refused to execute script from 'https://jlai.lu/js/87360.client.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
client.js:2 Uncaught (in promise) ChunkLoadError: Loading chunk 87360 failed.
(error: https://jlai.lu/js/87360.client.js)
    at r.f.j (client.js:2:737632)
    at client.js:2:736228
    at Array.reduce (<anonymous>)
    at r.e (client.js:2:736206)
    at Array.map (<anonymous>)
    at a (client.js:2:160190)
    at client.js:2:1457844
    at p (client.js:2:547503)
    at Generator.<anonymous> (client.js:2:548858)
    at Generator.next (client.js:2:547866)
```

### Technical Details

All browsers tested show the same behavior: Safari, Firefox or Chromium

This doesn't seem to happen on instances still in 0.18

### Lemmy Instance Version

0.18.1-rc1

### Lemmy Instance URL

https://jlai.lu

sunaurus:
With the current webpack configuration, Lemmy tries to load js files from paths where there aren't any, creating errors like:

```
Unhandled Promise Rejection: ChunkLoadError: Loading chunk node_modules_date-fns_esm_locale_en-GB_index_js failed.
(missing: http://localhost:1234/js/node_modules_date-fns_esm_locale_en-GB_index_js.client.js)
```

Notice that it's loading from `/js/...` and not `/static/js/...`

This PR adds a `publicPath` config, which can be used to specify a base path for assets: https://webpack.js.org/guides/public-path/

As far as I can tell, this fixes #1657 (at least it fixes it for me locally).

Note: I also noticed that the regex for including assets for the `ServiceWorkerPlugin` was only includiding `client.js`, I changed it to include all paths in the `js/` directory.


sunaurus:
By the way, I'm not exactly sure what the context is with date-fn locale issues (is it a totally unrelated issue?), but this PR should fix loading date-fn files for locales other than en-US as well.

alectrocute (commented):
@SleeplessOne1917 Are you still planning on replacing `date-fns` with `moment.js`? If so, how does that change the proposal here? Still needed?

SleeplessOne1917:
> Are you still planning on replacing date-fns with moment.js? If so, how does that change the proposal here? Still needed?

I haven't gotten around to removing it yet, but if this PR works I'd much prefer it over bringing back  the moment.js dependency. I'll test it tonight.

alectrocute:
@SleeplessOne1917 I don't think this'll resolve our ongoing timezone issues, unfortunately.

sunaurus:
@alectrocute can I read more about the timezone issue anywhere? Maybe I can take a look at it

SleeplessOne1917:
> can I read more about the timezone issue anywhere? Maybe I can take a look at it as well

The issues you linked earlier would be the best place to look, though it seems you've already done that. 

@alectrocute since the imports are fixed, getting the timezones working might actually be possible.

SleeplessOne1917 (approved):

dessalines (approved):

PR #1682, opened 2023-06-29
feat(theme): Vaporwave

SleeplessOne1917:
Bringing back yet another old theme.

Vaporwave Light:
image
image
image
image

Vaporwave Dark:
image
image
image


jsit:
Now that all these themes are coming in it seems like a good time to mention my concern about using `secondary` as an "accent color."

In [the default Bootstrap color palette](https://getbootstrap.com/docs/5.3/utilities/colors/), this is a gray color, so in a non-themed Bootstrap setup, developers would expect `text-secondary` to be gray, not some kind of bright blue.

I think it might be better to assign things like `$teal-300` or something to the UI variables like `$border-color`, `$link-color` found in `node_modules/bootstrap/scss/_variables.scss`.

It seems to me that in many of the places where we use `-secondary` to get that green color of the default theme, instead, green should be the _primary_ color, and we should be referring to `-primary` in our classes.

So in the case of this theme, that blue would be `-primary`. This would require rewriting a couple theme vars and a bunch of classes in the JSX.

Thoughts?

SleeplessOne1917:
> Now that all these themes are coming in it seems like a good time to mention my concern about using secondary as an "accent color."

That's how I've been thinking of primary versus secondary color.

> In [the default Bootstrap color palette](https://getbootstrap.com/docs/5.3/utilities/colors/), this is a gray color, so in a non-themed Bootstrap setup, developers would expect text-secondary to be gray, not some kind of bright blue.

Are there any other official/popular bootstrap themes to reference to see what they're doing? It's hard to generalize this based on a sample size of 1. If secondary ends up being consistently used for more dull colors, then we should change our themes to accommodate that.

jsit:
No, I don’t really know any other Bootstrap themes at all, it is just a sample size of one. If it feels fine to others then it won’t bug me.

dessalines:
@SleeplessOne1917 is this taken from the most recent / bootstrap 5 version of bootswatch? IE this one: https://bootswatch.com/vapor/

SleeplessOne1917:
> is this taken from the most recent / bootstrap 5 version of bootswatch?

No. It's the theme from Lemmy 0.16.7 with some tweaks to get it working with the recent UI changes.

dessalines:
mmk, probably fine as is then.

dessalines (approved):

jsit:
@SleeplessOne1917 I wanted to point out that it looks like Bootstrap *does* consider “secondary” to specifically be a light version of “body”:

IMG_1030


https://getbootstrap.com/docs/5.3/customize/color/

PR #1684, opened 2023-06-29
Use git hash to break cache

### LINKED ISSUE #1676: Implement cache invalidation features
### Requirements

Bug report #1666 by @ruudschilders exemplifies a big issue with the current state of Lemmy upgrades: They require to clear the browser cache or things break.

According to https://stackoverflow.com/questions/1922910/force-browser-to-clear-cache#1922924 it's pretty easy to get around the issue by adding a parameter after the file name of esp. CSS and JS files. The obvious choice would be to just use the short hash of the used git commit, like `<link rel="stylesheet" type="text/css" href="/static/styles/styles.css?v=0e9faf7" />` and `<script defer src='/static/js/client.js?v=0e9faf7'></script>` for the currently latest commit 0e9faf7.

SleeplessOne1917:
Closes #1676


src/server/index.tsx
```diff
@@ -18,7 +19,7 @@ const [hostname, port] = process.env["LEMMY_UI_HOST"]
 
 server.use(express.json());
 server.use(express.urlencoded({ extended: false }));
-server.use("/static", express.static(path.resolve("./dist")));
+server.use(`/static-${COMMIT_HASH}`, express.static(path.resolve("./dist")));
```
sunaurus, line 22: I believe several instances have already set up their configurations to work based on the assumption that `/static/` is a path to cacheable files. Renaming this directory will most likely break those existing configurations. 

What do you think about adding a new subdirectory instead? So instead of `/static-af29123/` we could have `/static/af29123/`.

alectrocute (approved):
Pending @sunaurus’s feedback and the missing new line at the end of one of those files, this looks great and solves a huge bug. Fantastic!

alectrocute:
Also, epic rhyme in PR title.

alectrocute:
What’s going on with CI here? Definitely would be great to get this in the next RC.

webpack.config.js
```diff
@@ -14,6 +14,11 @@ const banner = `
   @license magnet:?xt=urn:btih:0b31508aeb0634b347b8270c7bee4d411b5d4109&dn=agpl-3.0.txt AGPL v3.0
   `;
 
+const commitHash = require("child_process")
```
dessalines, line 17: This is breaking CI, and probably shouldn't be necessary, because the docker images already set that commit hash.

alectrocute: Should we make this Docker-agnostic? What if running the frontend without Docker?

.prettierignore
```diff
@@ -2,3 +2,4 @@ src/shared/translations
 lemmy-translations
 src/assets/css/themes/*.css
 stats.json
+dist
```
alectrocute, line 5:
```suggestion
dist

```

webpack.config.js
```diff
@@ -96,15 +101,15 @@ const createClientConfig = (_env, mode) => {
     entry: "./src/client/index.tsx",
     output: {
       filename: "js/client.js",
-      publicPath: "/static/",
+      publicPath: `/static/${commitHash}`
```
sunaurus, line 104:
```suggestion
      publicPath: `/static/${commitHash}/`
```

Without the trailing slash the paths break:

```
Unhandled Promise Rejection: ChunkLoadError: Loading chunk 76292 failed.
(error: http://localhost:1234/static/3caf913614868f70475430fca3dcc40c61d613b5js/76292.client.js)
```

sunaurus:
This is totally optional, but IMO using the --short git hashes will be a bit nicer while still giving absurdly low chances of collisions.

So `git rev-parse --short HEAD`

webpack.config.js
sunaurus: The hash needs to be available during webpack build in order to write correct import paths into the generated js files. Why is it breaking CI? Maybe we can work around it

sunaurus: Ahh I see, `git` is not available in the `node:alpine` image that is used for builds

sunaurus: We could do one of the following:

1. Create a custom builder docker image which just installs git on top of node:alpine
2. Manually run `apk add git` as a first step before `yarn build` in .woodpecker.yml
3. Switch to a different base image which comes with git out of the box

Btw, option 2 is actually not very expensive at all, takes less than a second:

```
/ # time apk add git
(1/9) Installing ca-certificates (20230506-r0)
(2/9) Installing brotli-libs (1.0.9-r14)
(3/9) Installing libunistring (1.1-r1)
(4/9) Installing libidn2 (2.3.4-r1)
(5/9) Installing nghttp2-libs (1.53.0-r0)
(6/9) Installing libcurl (8.1.2-r0)
(7/9) Installing libexpat (2.5.0-r1)
(8/9) Installing pcre2 (10.42-r1)
(9/9) Installing git (2.40.1-r0)
Executing busybox-1.36.1-r0.trigger
Executing ca-certificates-20230506-r0.trigger
OK: 21 MiB in 26 packages
real	0m 0.69s
user	0m 0.27s
sys	0m 0.03s
```

alectrocute:
Sorry to complicate things, but couldn't we just do cache busting without git, eg. [webpack config](https://stackoverflow.com/questions/39238163/how-to-use-cache-busting-with-webpack)?

sunaurus:
I think the issue with webpack hashes is that we run the server and client webpack builds seperately, and I'm not sure if there's any clean way to take the resulting hash(es) from the client build, and replace them into imports in the generated server files. My assumption is that it would require a bigger refactor of the webpack configuration. But I'm no webpack expert, maybe there's some obvious way to do it that I'm missing!

alectrocute:
> 



> I think the issue with webpack hashes is that we run the server and client webpack builds seperately, and I'm not sure if there's any clean way to take the resulting hash(es) from the client build, and replace them into imports in the generated server files. My assumption is that it would require a bigger refactor of the webpack configuration. But I'm no webpack expert, maybe there's some obvious way to do it that I'm missing!

I think there's a way to export the `contenthash` that webpack generates and export it as an env variable to be used by the server, but I'm also not an expert in webpack. :/

dessalines:
@SleeplessOne1917  @sunaurus Ya that's an easy one to fix:

In the `.woodpecker.yml`, add this before the `yarn build:dev`

`- apk add git`

dessalines:
Okay actually I'm kinda changing my mind a bit:

Check out this line: https://github.com/LemmyNet/lemmy-ui/blob/main/Dockerfile#L27

It should also set the commit hash there.

Why not just let the webpack read the version / commit hash from the `version.ts` file?

SleeplessOne1917:
> I think the issue with webpack hashes is that we run the server and client webpack builds seperately, and I'm not sure if there's any clean way to take the resulting hash(es) from the client build, and replace them into imports in the generated server files. My assumption is that it would require a bigger refactor of the webpack configuration. But I'm no webpack expert, maybe there's some obvious way to do it that I'm missing!

Pretty much this. The built in hashes for webpack were the first thing I tried, but I couldn't figure out how to use them in the actual code.

@dessalines 
> Check out this line: https://github.com/LemmyNet/lemmy-ui/blob/main/Dockerfile#L27
It should also set the commit hash there.

See the most recent commit I made. Using the hash is now docker agnostic, and it's set as an environment variable.

src/shared/components/person/person-listing.tsx
```diff
@@ -88,7 +88,10 @@ export class PersonListing extends Component<PersonListingProps, any> {
           !this.props.person.banned &&
           showAvatars() && (
             <PictrsImage
-              src={avatar ?? "/static/assets/icons/icon-96x96.png"}
+              src={
+                avatar ??
+                `/static/${process.env.COMMIT_HASH}/assets/icons/icon-96x96.png`
```
dessalines, line 93: Can you make this into a `bundleFolder()` or `bundleDir()` function? That way we won't have to trace all these down again potentially in the future.

SleeplessOne1917: I'll make the change during my lunch break.

alectrocute: Made the changes.

dessalines (approved):

PR #1685, opened 2023-06-29
Communities page: make search type default to "Communities'

Raicuparta:
The search form in the Communities page currently defaults to search type "all", which isn't what users expect in that page.


SleeplessOne1917 (approved):

wiki-me:
Is it? reddit searches everything and not just communities, so does facebook iirc, I think it would better to just add "set as default" button that changes the default option (next to the button the selects the type of search), see how tildes.net does it (You will need to be registered for that, [here](https://tildes.net/register?code=IRBS7-8KALT-W7RZS) is a invite code, I can provide more if you would like).

SleeplessOne1917:
@wiki-me there is a search page for more general purpose searches. A search being done from the communities page search bar should always return communities.

alectrocute (approved):

Raicuparta:
I fixed the formatting, but the pipeline failed anyway, seems to still be using the previous commit? Not sure how this is supposed to work.

PR #1706, opened 2023-06-29
Fix `PostListing` mobile margin layout issue

alectrocute:
Hi all!

In this PR:

- Fix `PostListing` mobile margin layout issue

On mobile, the avatar of the user on the `createdLine` was touching the title of the post. It was bugging me. This fixes it.

Thanks!

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -401,10 +401,11 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     }
   }
 
-  createdLine() {
+  createdLine({ extraClasses = "" }: { extraClasses?: string }) {
     const post_view = this.postView;
+
     return (
-      <div className="small">
+      <div className={`small ${extraClasses}`}>
         <span className="me-1">
           <PersonListing person={post_view.creator} />
         </span>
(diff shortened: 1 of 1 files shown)


src/shared/components/post/post-listing.tsx
```diff
@@ -1373,7 +1373,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         <div className="d-block d-sm-none">
           <article className="row post-container">
             <div className="col-12">
-              {this.createdLine()}
+              <div className="mb-1">{this.createdLine()}</div>
```
SleeplessOne1917, line 1376: `this.createdLine` already returns a div. Put the class name on that instead of making a new div.

SleeplessOne1917 (changes requested):

src/shared/components/post/post-listing.tsx
alectrocute: But I want it to only affect mobile. Any ideas how to do that?

alectrocute:
@SleeplessOne1917 Please check out my latest commit. Thanks!

src/shared/components/post/post-listing.tsx
SleeplessOne1917: I believe you can use `mb-1 mb-md-0`. At least, I think margin classes are responsive.

alectrocute: That totally did the trick. Thanks so much.

SleeplessOne1917 (approved):

src/shared/components/post/post-listing.tsx
jsit: Always look for Bootstrap classes before adding a element or a div

PR #1708, opened 2023-06-29
Cache static data for a day

SleeplessOne1917:
This sets cache control for static content.

### DIFF
--- a/src/server/middleware.ts
+++ b/src/server/middleware.ts
@@ -1,4 +1,4 @@
-import type { NextFunction, Response } from "express";
+import type { NextFunction, Request, Response } from "express";
 import { UserService } from "../shared/services";
 
 export function setDefaultCsp({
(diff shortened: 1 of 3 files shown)


src/server/utils/create-ssr-html.tsx
```diff
@@ -24,7 +24,7 @@ export async function createSsrHtml(
 
   if (!appleTouchIcon) {
     appleTouchIcon = site?.site_view.site.icon
-      ? `data:image/png;base64,${sharp(
+      ? `data:image/png;base64,${await sharp(
```
alectrocute, line 27: Nice.

src/server/middleware.ts
```diff
@@ -22,19 +22,25 @@ export function setDefaultCsp({
...
+    req.path.includes("/css/themelist")
+  ) {
+    // Static content gets cached publicly for a day
+    caching = "public, max-age=86400";
   } else {
-    caching = "public, max-age=60";
+    if (user.auth()) {
+      caching = "private";
+    } else {
+      caching = "public, max-age=60";
```
alectrocute, line 42: Would changing `max-age` here to `5` fix/mitigate our captcha and logout issues?

SleeplessOne1917: Not sure. I left a comment on your other PR about the [Clear-Site-Data](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data) header. That would require a change to the js client instead of the UI, but it could help. The only rub is that the `cache` option of site data is not supported in Firefox.

alectrocute:
I say we change the `public, max-age` to be way less time and ship it. This will solve a lot of issues. We can revisit our cache strategy in the `0.18.2` milestone.

alectrocute (approved):

dessalines (approved):

PR #1711, opened 2023-06-29
Fix some issues

### LINKED ISSUE #1704: Visting a path containing text considered an auth path always redirects to /login, even if on a non-auth endpoint.
### Requirements

### Summary

Any path containing "admin", even in legitimate contexts (such as https://lemmy.ml/c/lemmy_admin) will be seen as the admin page, and redirected to `/login` if the user is unauthenticated.

This issue appears to have been introduced in 0.18.0, as `beehaw.org` (currently on 0.17.4) does not appear to have the issue. (https://beehaw.org/c/lemmy_admin@lemmy.ml). lemmy.world (0.17.4) appears unaffected. However, 0.18.1-rc.4 (as currently seen on https://voyager.lemmy.ml/) appears affected.

This counts for any text considered to be an "auth" path, as defined [in this file](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils/app/is-auth-path.ts)

This makes it impossible to visit any communities or users containing text that's considered an auth path when not logged in. I'm assuming any federated instances with one of these strings in its name will also be unavailable from other 0.18.0 instances (but untested).

### Steps to Reproduce

1. Create a community with a word from [this](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/utils/app/is-auth-path.ts) list
2. Visit that community from any instance (local or federated) on 0.18.0 or above
3. Get incorrectly redirected to the /login page
(The same works for users, possibly for instance domains)

### Technical Details

I am on ArchLinux with Firefox 114.0.2. The same behavior happens in ungoogled-chromium 114.0.5735.133.

### Lemmy Instance Version

0.18.0 or above

### Lemmy Instance URL

https://lemmy.ml

SleeplessOne1917:
This is a 2 in 1, which closes #1704 and fixes a problem brought up in the dev matrix channel

### DIFF
--- a/src/server/middleware.ts
+++ b/src/server/middleware.ts
@@ -29,10 +29,13 @@ export function setCacheControl(
 ) {
   const user = UserService.Instance;
   let caching: string;
+  console.log("In middleware");
+  console.log(process.env.NODE_ENV);
 
   if (
-    req.path.match(/\.(js|css|txt|manifest\.webmanifest)\/?$/) ||
-    req.path.includes("/css/themelist")
+    process.env.NODE_ENV === "production" &&
+    (req.path.match(/\.(js|css|txt|manifest\.webmanifest)\/?$/) ||
+      req.path.includes("/css/themelist"))
   ) {
     // Static content gets cached publicly for a day
     caching = "public, max-age=86400";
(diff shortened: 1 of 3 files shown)


alectrocute (approved):
Great work.

PR #1713, opened 2023-06-30
Enforce SameSite=Strict

diamondburned:
This adds additional guard into the JWT cookie so that it does not leak on cross-origin requests.

Related issue: https://github.com/LemmyNet/lemmy/pull/3301


diamondburned:
It's worth mentioning that this PR completely rips out `isomorphic-cookie` and replaces it with code that directly uses `document.cookie`. This keeps the same exact behavior as before, since the library never used the cookie in the backend anyway. All cookie-setting was done in the frontend!

SleeplessOne1917 (approved):
@alectrocute @dessalines @jsit I wonder if this will also solve the user cache busting bug?

alectrocute:
@SleeplessOne1917 Good point.

Can we use this to fix the inoperable `user.auth()` conditional on `middleware.ts` as well?

SleeplessOne1917:
> Can we use this to fix the inoperable user.auth() conditional on middleware.ts as well?

I think so. I'll do it in a separate PR when I get the chance.

alectrocute (approved):
Nice.

PR #1718, opened 2023-06-30
Fix broken `user.auth()` method on `middleware.ts`

alectrocute:
New strategy for this. Let's parse `req.headers.cookie` directly with a server helper utility. Once #1713, we should update `is-request-authenticated` to use `cookie.parse(...).jwt`.

### DIFF
--- a/src/server/middleware.ts
+++ b/src/server/middleware.ts
@@ -1,5 +1,5 @@
 import type { NextFunction, Request, Response } from "express";
-import { UserService } from "../shared/services";
+import { isRequestAuthenticated } from "./utils/is-request-authenticated";
 
 export function setDefaultCsp({
   res,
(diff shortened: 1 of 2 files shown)


alectrocute:
Maybe we can do without the `cookie` dependency in #1713 as well. Food for thought.

SleeplessOne1917 (approved):

webpack.config.js
```diff
@@ -111,9 +111,6 @@ const createClientConfig = (env, mode) => {
       new ServiceWorkerPlugin({
         enableInDevelopment: mode !== "development", // this may seem counterintuitive, but it is correct
         workbox: {
-          modifyURLPrefix: {
-            "/": `/static/${env.COMMIT_HASH}/`,
-          },
```
alectrocute, line 116: This fixes a bad path with our service worker on production.

src/server/middleware.ts
```diff
@@ -27,18 +27,20 @@ export function setCacheControl(
   res: Response,
   next: NextFunction
 ) {
-  const user = UserService.Instance;
+  if (process.env.NODE_ENV !== "production") {
+    return next();
+  }
```
alectrocute, line 32: Let's not cache anything on dev.

PR #1719, opened 2023-06-30
Fix nonexistent property `req.cookies`

alectrocute:
In this PR:

- Fix nonexistent property `req.cookie` from https://github.com/LemmyNet/lemmy-ui/pull/1713.
- Having no automated testing hurts with stuff like this. CI could have caught such a bug.

Thanks!

### DIFF
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -25,11 +25,15 @@ import { setForwardedHeaders } from "../utils/set-forwarded-headers";
 export default async (req: Request, res: Response) => {
   try {
     const activeRoute = routes.find(route => matchPath(req.path, route));
-    let auth = req.cookies ? cookie.parse(req.cookies).jwt : undefined;
+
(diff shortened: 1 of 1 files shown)


src/server/handlers/catch-all-handler.tsx
```diff
@@ -25,11 +25,15 @@ import { setForwardedHeaders } from "../utils/set-forwarded-headers";
 export default async (req: Request, res: Response) => {
   try {
     const activeRoute = routes.find(route => matchPath(req.path, route));
-    let auth = req.cookies ? cookie.parse(req.cookies).jwt : undefined;
+
+    let auth = req.headers.cookie
```
alectrocute, line 29: Actual change here.

SleeplessOne1917 (approved):

src/server/handlers/catch-all-handler.tsx
diamondburned: That's so strange! TypeScript told me it was there!

PR #1724, opened 2023-07-01
Mark permalinked comment

ftatateeta:
Right now the `link` link in each comment links to the parent. Which may be the confusing for the user (what's comment actually linked comment?). You have to look at the code to see the `link` link actually links to the parent comment which is the first comment that's the displayed. Marking the comment that's specified in the url helps make this clear. 

Reddit does this by marking the linked the comment and further allows a `context` query parameter to specify how many parent comments to display. I propose we do something similar.


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "analyze": "webpack --mode=none",
     "prebuild:dev": "yarn clean && node generate_translations.js",
-    "build:dev": "webpack --env COMMIT_HASH=$(git rev-parse --short HEAD) --mode=development",
+    "build:dev": "webpack --env COMMIT_HASH=DEV --mode=development",
```
jsit, line 11: Why this change?

alectrocute:
Please provide a description of what your PR accomplished and why it’s needed. As well as any issue tickets it may resolve. Thank you.

package.json
ftatateeta: I added that by accident I had to do that to make it build on windows.

ftatateeta:
I added a description.

alectrocute:
> I added a description.

Thanks!

yarn.lock
dessalines, line 1: Why did this file change? I don't see any deps updated.

package.json
dessalines: You need to revert that then.

src/shared/components/comment/comment-node.tsx
```diff
@@ -289,7 +290,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
           id={`comment-${cv.comment.id}`}
           className={classNames(`details comment-node py-2`, {
             "border-top border-light": !this.props.noBorder,
-            mark: this.isCommentNew || this.commentView.comment.distinguished,
+            mark: this.isCommentNew || this.commentView.comment.distinguished || this.props.markCommentId == cv.comment.id,
```
dessalines, line 293: Ah I see what its doing now. Yep this is a good idea.

src/shared/components/comment/comment-node.tsx
```diff
@@ -1209,7 +1211,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
 
     const title = this.props.showContext
       ? I18NextService.i18n.t("show_context")
-      : I18NextService.i18n.t("link");
+      : I18NextService.i18n.t("link parent");
```
dessalines, line 1214: I don't understand this one. Almost all the cases of CommentNode / CommentNodes have the show context flag, they only don't for things like the comment report. In which case the link is fine.

dessalines:

PR #1727, opened 2023-07-01
Lemmy 2515 controversial posts and comments

### LINKED ISSUE #1404: Controversal Posts and Comments (UI Side)
### Requirements

This is the UI side of the backend issue [#2515](https://github.com/LemmyNet/lemmy/issues/2515).

Buttons added for sorting by Controversial on Posts and Comments.
Maybe hide them on instances with disabled downvotes.

None

### Additional context

_No response_

iByteABit256:
Closes #1404

Added buttons to be able to sort posts and comments by controversy rank.

(Draft until lint errors are fixed and new buttons are hidden when downvotes are disabled on the Lemmy instance)


bqv:
I don't think this should be added, for ethical reasons.

At very least, instances should opt-in to having it.

iByteABit256:
> I don't think this should be added, for ethical reasons.
> 
> At very least, instances should opt-in to having it.

Can you elaborate on this?

The idea is that you can find the posts with most intense discussion, "grab the popcorn" kind of posts :P

Why is this unethical?

bqv:
Social media has become such a problem globally specifically because of such algorithms as this, that feed into peoples anger, manufacturing discord and dissent. I thought lemmy not having that was an intentional design choice, but I see @dessalines and @Nutomic already seem to support this? I just ask that it be opt-in at least, with a warning to that effect, so not all of lemmy turns into a twitter-like rage machine.

iByteABit256:
Well, it's up to the user to consciously choose to sort by Controversial, so that means that the user is specifically looking for a hot topic with conflicting opinions, the user experience will be completely unaffected if the user doesn't choose that sort option.

It's not like there's an almost sentient algorithm that serves toxic content to the users like on Twitter for example, this is determined purely by the number of upvotes/downvotes and the number of total votes.

Also, keep in mind that this will only be present if downvotes are enabled on the instance.

Still, I'm not against the idea for this to be an opt-in feature though.

ruffsl:
> the user experience will be completely unaffected if the user doesn't choose that sort option.

I think that's not taking into consideration the second or third order effects such a feature would have on a user's experience, even for those that never choose to sort by controversial. A driver could choose to decal their car in rage bating bumper stickers, but it'd be hard to argue it'd have zero effect on the driving experience for those who never use them, as they'ed still have to look at stickers to avoid fender benders in traffic, distracting one's human levels of attention.

> Still, I'm not against the idea for this to be an opt-in feature though.

It would be nice, though I'm not sure how that would shake down in practice across a federated ecosystem. Would another instance have to respect the available sorting from another instance? If a small instance hosting a news community opted out of controversial sorting to curtail argumentative pile-ons, but remote instances with the majority of the subscribers opted in, then I don't think it would be realistically enforceable. I suppose any client/server could already implement any sorting or ranking system regardless, so perhaps this is all moot anyway.

I'm not for or against, just wanted to point out possible side effects.

> Also, keep in mind that this will only be present if downvotes are enabled on the instance.

How does this currently work for federated instances, when the instance that a community is local to has down votes disabled? Do remote users only see downvotes from subscribers on the same remote instance, or does the stock release of Lemmy cordially respect the remote community's settings and similarly disables down votes originating from it's own users, and ignores federated downvotes from other instances towards the same community?

iByteABit256:
I think you're over complicating things. The opt-in would be instance wide not per user, so it would work quite like the downvote button being hidden in some instances and shown in others

ruffsl:
> I think you're over complicating things. The opt-in would be instance wide not per user, ...

Well, the premise of opt-in being per-instance, not federation-wide, was the entire complication the prior comment was pointing out. We can assume user-actions are entirely governed by their local instance, but the same can not be said for the user-experience when their instance is federated. To think - participating in a discussion where other parties are shown a more controversial sorting than your own - would not alter a users experience would be an over simplification.

> ... so it would work quite like the downvote button being hidden in some instances and shown in others

In that respect, could you still share how downvotes work when federated or propagated on a local instance with down votes enabled, when interacting with remote communities where the remote instance has down votes disabled? I've not had the chance to dive deep into both the spec and implementation just yet.

iByteABit256:
The backend serves the API which includes [LocalSite](https://join-lemmy.org/api/interfaces/LocalSite.html), the entity that contains instance configurations like `enable_downvotes` or `private_instance` and more.

The frontend (this one, Jerboa, or any other) uses this API and it's up to it to hide things in certain cases, like the downvote button I mentioned.

Specifically as you can see in my commits, the relevant Controversial Sort components are hidden depending on the value of `enable_downvotes`, so if an instance has them disabled then there's no such thing as Controversial Sort visible anywhere. You can still find the sort results through the API, but the UI simply hides this feature.

bqv:
> In that respect, could you still share how downvotes work when federated or propagated on a local instance with down votes enabled, when interacting with remote communities where the remote instance has down votes disabled? I've not had the chance to dive deep into both the spec and implementation just yet.

As I understand it, downvotes are not federated. That's why this works as a feature-flag at an instance level, whereas obviously controversial sorting has much more dire implications on the whole federal network

iByteABit256:
It's not so obvious to me, what are the dire implications? Can you give a simple example?

bqv:
As the others have vaguely noted, controversial sort is the starting pistol for a race to the bottom in terms of post quality. My lower bar is implementing an opt out because even if this PR doesnt go ahead, the luxury of open source is that someone will implement it somewhere anyway. The ideal however, is that this is not even a feature, because anywhere its implemented risks becoming a source of polarisation for the rest of the network (think, you browse the federated All feed, but with controversial sort). Im not sure what more to add without repeating my first post

iByteABit256:
Note: blocked until [backend side](https://github.com/LemmyNet/lemmy/pull/3205) is merged

SleeplessOne1917 (approved):
LGTM

iByteABit256:
@dessalines reminding this now that the backend side is merged, js client needs to be merged first too

dessalines:
@iByteABit256 Okay new `lemmy-js-client` is deployed: `0.19.0-rc.1`

iByteABit256:
Not sure why yarn lint still failed :/

src/shared/components/post/post.tsx
```diff
@@ -475,6 +475,26 @@ export class Post extends Component<any, PostState> {
           >
             {I18NextService.i18n.t("top")}
           </label>
+          {enableDownvotes(this.state.siteRes) && (
```
dessalines, line 478: Remove this enableDownvotes check, as well as the `hideControversial` props to the other two.

There are a lot of sorts that don't make sense when votes are disabled, not just controversial, so that should be done in a separate PR.

src/shared/components/common/sort-select.tsx
```diff
@@ -7,6 +7,7 @@ import { Icon } from "./icon";
 
 interface SortSelectProps {
   sort: SortType;
+  hideControversial?: boolean;
```
dessalines, line 10: Remove

src/shared/components/common/comment-sort-select.tsx
```diff
@@ -7,6 +7,7 @@ import { Icon } from "./icon";
 
 interface CommentSortSelectProps {
   sort: CommentSortType;
+  hideControversial?: boolean;
```
dessalines, line 10: Remove

dessalines (changes requested):

dessalines (approved):
Looks good, thx!

iByteABit256:
@dessalines can I do something here for the test to pass or should I wait until the appropriate version is reached?

iByteABit256:
mb I fixed it, tests pass now

iByteABit256:
@dessalines reminder, just in case

PR #1721, opened 2023-06-30
Fix date-fns locale import failing for some locales

sunaurus:
I realized that users who are still reporting issues with #1720 all have locales like `fr-FR`, the reason I could not reproduce the issue was that my browser was already setting it to `fr`.

Confirmed with two separate users that this solves the issue.

Fixes #1720

PR #1729, opened 2023-07-01
Add fallback date-fns locale import

sunaurus:
This date-fns locale bug is like a bossfight with multiple phases... hopefully this is the last phase 😅

Previously I fixed:
1. Other locales not loading at all https://github.com/LemmyNet/lemmy-ui/pull/1672
2. Locales like `xx-XX` (eg: `fr-FR`) failing to work, because for these date-fns expects just `xx` https://github.com/LemmyNet/lemmy-ui/pull/1721

These fixed the problem for most users, but turns out there are some locales out there which date-fns simply doesn't have support for. One example: `es-MX`. **For these users, the import still fails and Lemmy is still broken.**

This PR adds a fallback import of `en-US` when the user's locale is not availabe in date-fns.

### DIFF
--- a/src/shared/utils/app/setup-date-fns.ts
+++ b/src/shared/utils/app/setup-date-fns.ts
@@ -1,10 +1,12 @@
 import setDefaultOptions from "date-fns/setDefaultOptions";
 import { I18NextService } from "../../services";
 
+const EN_US = "en-US";
+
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):
Thank you for your service

PR #1730, opened 2023-07-01
Fix DM replies not working

sunaurus:
The `PrivateMessageForm` decides whether it's being used for editing a DM or replying to a DM based on whether `privateMessageView` is `undefined`.

Currently in rc7, replying to private messages is impossible, as `privateMessageView` is passed in even when replying. Clicking the reply icon opens up an edit form, and saving this edit form generates an error (because you can't edit another person's DM).

This PR fixes the bug by removing `privateMessageView` from replies.


alectrocute (approved):

SleeplessOne1917 (approved):

jsit:
@sunaurus Thank you so much for this, this was my regression.

PR #1732, opened 2023-07-01
Fix markdown editor quote bugs

sunaurus:
Currently, when using the quote functionality in markdown editors (by highlighting some text before clicking reply/comment), the editor inserts "undefined" in front of quoted text:

Screenshot 2023-07-01 at 19 44 15

This is due to a bug in the code which handles existing content in case of a quote being used. This PR fixes that bug.

(This PR also fixes the quoted text appearing appended on the end of the last line when editing existing messages)

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -702,18 +702,20 @@ export class MarkdownTextArea extends Component<
   quoteInsert() {
     const textarea: any = document.getElementById(this.id);
     const selectedText = window.getSelection()?.toString();
-    const { content } = this.state;
+    let { content } = this.state;
     if (selectedText) {
       const quotedText =
         selectedText
           .split("\n")
           .map(t => `> ${t}`)
           .join("\n") + "\n\n";
+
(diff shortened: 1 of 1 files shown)


alectrocute (approved):

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -702,18 +702,20 @@ export class MarkdownTextArea extends Component<
...
           .split("\n")
           .map(t => `> ${t}`)
           .join("\n") + "\n\n";
+
       if (!content) {
-        this.setState({ content: "" });
+        content = "";
       } else {
-        this.setState({ content: `${content}\n` });
+        content = `${content}\n\n`;
```
sunaurus, line 716: I added a second \n here, it looks much better to my eye with two lines between previously existing text and new quote.

SleeplessOne1917 (approved):

src/shared/components/common/markdown-textarea.tsx
alectrocute: Agreed. Nice work.

Issue #1723, opened 2023-07-01
Don't show "Join Lemmy" link in footer if user is signed in.


Justinzobel:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

There is no need for a Join Lemmy link if the user is already signed into an account.

Nutomic:
Its also not a problem, logged in users may want to read the docs for example.

PR #1737, opened 2023-07-01
Use Join Lemmy hostname as link text in footer

dnswrsrx:
I think it makes sense to keep a link to the Join Lemmy website even though there's a logged in user. There's plenty of other useful info on that website beyond just joining Lemmy.

Using the URL avoids the phrase "Join Lemmy" and should indicate that you're navigating to said website.

Should address issue #1723

### DIFF
--- a/src/shared/components/app/footer.tsx
+++ b/src/shared/components/app/footer.tsx
@@ -60,7 +60,7 @@ export class Footer extends Component<FooterProps, any> {
             <li className="nav-item">
               <a className="nav-link" href={joinLemmyUrl}>
                 {myAuth()
-                  ? "join-lemmy.org"
+                  ? new URL(joinLemmyUrl).hostname
                   : I18NextService.i18n.t("join_lemmy")}
               </a>
             </li>
--- a/src/shared/components/app/footer.tsx
+++ b/src/shared/components/app/footer.tsx
@@ -1,3 +1,4 @@
+import { myAuth } from "@utils/app";
 import { Component } from "inferno";
 import { NavLink } from "inferno-router";
 import { GetSiteResponse } from "lemmy-js-client";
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

alectrocute:
Please export a string from `config.ts` for the URL. Thanks!

dnswrsrx:
> Please export a string from `config.ts` for the URL. Thanks!

Ah, I should have looked a bit more to see if there was the URL hardcoded somewhere!

Using the `URL` object to return the hostname. In the future if the URL changes, we don't have to change the string in multiple instances!

alectrocute (approved):
Nice changes. Looks good to me! Thanks.

src/shared/components/app/footer.tsx
```diff
@@ -58,7 +59,9 @@ export class Footer extends Component<FooterProps, any> {
             </li>
             <li className="nav-item">
               <a className="nav-link" href={joinLemmyUrl}>
-                {I18NextService.i18n.t("join_lemmy")}
+                {myAuth()
+                  ? new URL(joinLemmyUrl).hostname
```
dessalines, line 63: Why these two different representations? This seems pointless.

dessalines: I'd rather it just be consistent. And if its useful to always show it, (I think it is), then that issue should just be closed.

dnswrsrx: > Why these two different representations? This seems pointless.

After a user has logged in, the phrase, `Join Lemmy` feels out of place since they've _joined_ Lemmy (even though we're really trying to convey the join lemmy website).

I thought the URL representation conveyed "visit this website" more than the phrase `Join Lemmy`.

If it's not that important, I'm okay to just drop this PR and close the issue!

dessalines: I'd also be fine with your host version, join-lemmy.org being on both.

dessalines (approved):

dessalines (approved):

PR #1738, opened 2023-07-01
Modify MXID regex to allow ports in hostname

### LINKED ISSUE #1726: [Bug]: Matrix user ID validation considers server names with explicit ports invalid
### Requirements

### Summary

Cross posting lemmynet/lemmy#3431 which has a UI and API component. Original credit goes to @Whodiduexpect for finding the bug

From the original Issue:

### Summary
If you set your account's Matrix user to a server that uses a port inside its server name (like @user:server.tld:443) both the validation on Lemmy UI and the Lemmy backend incorrectly state that the MXID is invalid.

### Steps to Reproduce

1. Click on logged-in user in the top-right
2. Click on settings
3. Enter an MXID into the "Matrix User" field which uses an explicit port in the server name
4. Click save button at the bottom
5. Browser prompts user to "Please match the requested format" despite it being a valid MXID.

### Technical Details

Browser: Firefox

### Lemmy Instance Version

BE 0.18.1-rc.4

### Lemmy Instance URL

vlemmy.net

Whodiduexpect:
This is the change in Lemmy UI that goes along with LemmyNet/lemmy#3441 which fixes #1726. I just updated the regex expression that validates Matrix user IDS to allow having ports in the hostname.

While this regex validates that the port is not over 5 digits, it does not validate that the port is not above 65535 while still being 5 digits, however my change on the server does validate this seperately. Since Lemmy UI accepts the subtly invalid input, but the server does not, this results in the save button seemingly not doing anything in that edge case. This is probably not optimal UX, but I don't think it matters in this case.

### DIFF
--- a/src/shared/components/person/settings.tsx
+++ b/src/shared/components/person/settings.tsx
@@ -548,7 +548,7 @@ export class Settings extends Component<any, SettingsState> {
                 placeholder="@user:example.com"
                 value={this.state.saveUserSettingsForm.matrix_user_id}
                 onInput={linkEvent(this, this.handleMatrixUserIdChange)}
-                pattern="^@[A-Za-z0-9._=-]+:[A-Za-z0-9.-]+\.[A-Za-z]{2,}$"
+                pattern="^@[A-Za-z0-9._=-]+:[A-Za-z0-9.-]+\.[A-Za-z]{2,}(:[0-9]{1,5})?$"
               />
             </div>
           </div>


SleeplessOne1917 (approved):

alectrocute (commented):
Can we please move this whole regex string into `config.ts` and export it? Great opportunity to clean things up. Thanks!

dessalines:
I can re-open if you address the comments.

PR #1741, opened 2023-07-01
Fix jwt cookie path

### LINKED ISSUE #1740: Cookies set during login are set too specific causing it to break
### Requirements

### Summary

The cookies set at the login are set specific for the current path causing the login to break in many situations.

When you are currently looking at, for example, a post and then try to log in the cookie is set for domain.com/post and not domain.com
When you then visit a different page, you are logged out again because the cookie is only valid for the specific path.

I assume this will also cause issues with API calls because the cookie is also invalid for the API path.

Screenshot 2023-07-01 at 22 17 41

### Steps to Reproduce

1. Go to https://domain.com/post/XXXX
2. Login
3. Go to the main page
4. Refresh the page
5. be logged out

### Technical Details

Lemmy: 0.18.1-RC4
Lemmy UI: 0.18.1-RC7

### Lemmy Instance Version

0.18.1-RC4

### Lemmy Instance URL

_No response_

sunaurus:
Fixes #1740

### DIFF
--- a/src/shared/services/UserService.ts
+++ b/src/shared/services/UserService.ts
@@ -38,6 +38,7 @@ export class UserService {
         secure: isHttps(),
         domain: location.hostname,
         sameSite: true,
+        path: "/",
       });
       this.#setJwtInfo();
     }


alectrocute (approved):
Wow, good catch. We’ve had a lot of bugs come from this cookie refactor. Hopefully this is the last of it.

SleeplessOne1917 (approved):

diamondburned:
Sorry, my bad...

PR #1764, opened 2023-07-02
fix: Fix vote buttons not showing spinners while registering vote #1761

### LINKED ISSUE #1761: When voting your vote gets cleared leaving the impression it did not take
### Requirements

### Summary

On an 0.18.1 instance beehaw we are seeing a behavior where you vote on a post or comment and it turns blue then goes back to the state where you have not voted. After some time it becomes blue.

### Steps to Reproduce

1. Find comment you like
2. Click up arrow
3. Notice how it's not blue or spinning. 

### Technical Details

Looking at the code I believe the issue to be here
 
https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/common/vote-buttons.tsx#L53

As far as I can tell onpost upstream is an async function. This causes the loading state to immediately turn off since nothing is awaiting the promise

### Lemmy Instance Version

0.18.1.rc.7

### Lemmy Instance URL

Https://beehaw.org

jsit:
Fixes #1761

### DIFF
--- a/src/shared/components/common/vote-buttons.tsx
+++ b/src/shared/components/common/vote-buttons.tsx
@@ -64,8 +64,6 @@ const handleUpvote = (i: VoteButtons) => {
         auth: myAuthRequired(),
       });
   }
-
-  i.setState({ upvoteLoading: false });
 };
 
 const handleDownvote = (i: VoteButtons) => {
@@ -86,7 +84,6 @@ const handleDownvote = (i: VoteButtons) => {
         auth: myAuthRequired(),
       });
   }
-  i.setState({ downvoteLoading: false });
 };
 
 export class VoteButtonsCompact extends Component<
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

dnswrsrx:
I see that `componentWillReceiveProps` has been made [legacy](https://legacy.reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops). One of their suggestion is to use `componentDidUpdate` - would this lifecycle method work here?

jsit:
@dnswrsrx 

> I see that `componentWillReceiveProps` has been made [legacy](https://legacy.reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops). One of their suggestion is to use `componentDidUpdate` - would this lifecycle method work here?

lemmy-ui uses Inferno, not React: https://www.infernojs.org/docs/guides/components

Issue #1609, opened 2023-06-26
Collapse Comment Chain When Clicking on Comment Border


guptaishan718:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

The comment border has two advantages on most sites, differentitaing between parent and child comments and being able to collapse the entire comment thread. Lemmy does not have second feature.

Right now when reading through a long chain of comments you either have to scroll past every child comment or go back to the parent comment to collapse. Having the ability to collapse a comment thread by clicking on border would make it much more user friendly.

JohanAR:
I like this, IMO the current collapse button is unfortunately placed, and it gets different horizontal positions based on the commenter's name. It feels like collapsing branches of a tree structure is part of the functionality of the tree, so the button ought to be part of the tree UI element, not off to the right on its own.

dessalines:
Not interested in this.

PR #1767, opened 2023-07-02
feat: added tap/click to collapse comment

dankxiaobong:
## UX Feature: 
Click or Tab anywhere on the comment to collapse it (Inspired by the reddit app).

## Solves this issue:
#1609 

## Info
Had to add event.stopPropagation(); to prevent bubbling.  
e.g. you click on the reply button for it not to auto collapse the comment

Please review if the A11ly implementation is correct.


src/shared/components/comment/comment-node.tsx
```diff
@@ -296,6 +296,12 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
             className={classNames({
               "ms-2": !this.props.noIndent,
             })}
+            onClick={linkEvent(this, this.handleCommentCollapse)}
+            onKeyDown={linkEvent(this, this.handleCommentCollapse)}
```
SleeplessOne1917, line 300: This will collapse on any key down?

alectrocute:
I think this deserves more discussion before implementation. This implementation doesn’t also make much sense to me.

src/shared/components/comment/comment-node.tsx
dankxiaobong: Hmm you're right! I added it because I was getting:

error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events

I removed the onKeyDown and turned the div into a button:
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/click-events-have-key-events.md

dankxiaobong:
> I think this deserves more discussion before implementation. This implementation doesn’t also make much sense to me.

It's a super convenient UX feature when browsing long comment threads. Personally, I've been using lemmy since ~mid 2020 and a feature like this where you tab anywhere on the comment to collapse it is what I've been missing from reddit most.

How would you approach the implementation differently?

src/shared/components/comment/comment-node.tsx
SleeplessOne1917: If you want to have it work with a keyboard shortcut still, you could always create a function that checks which keys are being pressed (something like ctrl or meta + something else; consult the `MarkdownTextArea` component for shortcuts that are already being used so as to avoid collision).

dankxiaobong:
> > a feature like this where you tab anywhere on the comment to collapse it is what I've been missing from reddit most.
> 
> When you say tab, you mean the tab key? I'm not familiar with that particular reddit feature, but normally I think tab is used to change which element is being focused.

Ah nop I mean tab as in tab on a touchscreen e.g. when using on mobile. I often found myself browsing in bed from the phone and touching anywhere on the comment to collapse it was always super convenient and instantly loved it when reddit introduced it some time back in their app

src/shared/components/comment/comment-node.tsx
dankxiaobong: Ok turning it into a button made it worse.. 

would this work for accessibility ?
<div onClick{() => {}} aria-hidden="true" />

SleeplessOne1917:
I see. @alectrocute @jsit @dessalines Should the entire area of the comment listen for a collapse command? Or only a certain part? Also, it would be good to get a keyboard shortcut for this, maybe `CTRL + Space` or something.

dankxiaobong:
> I see. @alectrocute @jsit @dessalines Should the entire area of the comment listen for a collapse command? Or only a certain part? Also, it would be good to get a keyboard shortcut for this, maybe `CTRL + Space` or something.

I updated it to work with CTRL + ArrowLeft for now. 
(CTRL+Space was giving me troubles)

Waiting for input from others.

alectrocute:
IMO this is something that should be a mobile-only feature.

SleeplessOne1917:
> IMO this is something that should be a mobile-only feature.

I could see collapse on touch as a mobile only feature, but I think the keyboard shortcut would be good to keep for desktop users.

src/shared/components/comment/comment-node.tsx
```diff
@@ -292,22 +292,24 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
           })}
         >
-          <button
+          <div
             className={classNames({
               "ms-2": !this.props.noIndent,
             })}
             onClick={linkEvent(this, this.handleCommentCollapse)}
-            role={"textbox"}
+            onKeyDown={linkEvent(this, this.handleKeyBinds)}
```
jsit, line 300: If this is a button, it doesn't need `onKeyDown` -- it handles `onClick` with a space or an enter.

jsit (commented):
This all feels a little inelegant.

jsit:
> > IMO this is something that should be a mobile-only feature.
> 
> I could see collapse on touch as a mobile only feature, but I think the keyboard shortcut would be good to keep for desktop users.

HItting space or enter while the collapse button is focused already is a "keyboard shortcut".

jsit:
Ctrl+Left and Option+Left aren't working for me on macOS Safari.

Keyboard users can already collapse comments by hitting enter/space on the collapse button.

SleeplessOne1917:
> Keyboard users can already collapse comments by hitting enter/space on the collapse button.

The user will still have to tab over multiple other buttons to get there. However, facilitating this would require changes outside of the scope of this PR. It could be worth devising a way to navigate comments using only the keyboard that is more comfortable than tabbing through a bunch of buttons.

src/shared/components/comment/comment-node.tsx
dankxiaobong: It's role="button" but not a <button> anymore. If I remove the onKeydown I get:

Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events

dankxiaobong:
> This all feels a little inelegant.

Could you elaborate a little please? Is it because of all the commits or because it's an onClick on a div or something else? 😅 Trying to get better at contributing here :)

dankxiaobong:
> > Keyboard users can already collapse comments by hitting enter/space on the collapse button.
> 
> The user will still have to tab over multiple other buttons to get there. However, facilitating this would require changes outside of the scope of this PR. It could be worth devising a way to navigate comments using only the keyboard that is more comfortable than tabbing through a bunch of buttons.

Is there a way to leverage the tabing + space/enter on the collapse button, but not having to navigate every single icon, in every single comment if not needed, but still having the ability to navigate to the upvote buttons with a tab (by using a hotkey)?

dessalines (commented):
There's already a `[-]` on the left of comments to collapse them. Why do we need this?

PR #1769, opened 2023-07-02
Re-add Cantarell and Noto Sans for darkly/darkly-red.

IngrownMink4:
Read more about this change here: https://github.com/LemmyNet/lemmy-ui/issues/779
I have also added **Roboto Mono** for Android users.

### DIFF
--- a/src/assets/css/themes/darkly-red.css
+++ b/src/assets/css/themes/darkly-red.css
@@ -71,8 +71,8 @@
   --bs-dark-border-subtle: #adb5bd;
   --bs-white-rgb: 255, 255, 255;
   --bs-black-rgb: 0, 0, 0;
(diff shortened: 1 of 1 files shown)


jsit:
> Read more about this change here: #779 I have also added **Roboto Mono** for Android users.

1. These CSS files are generated; you should be updating `_variables.darkly.scss` instead.
2. Why only darkly and not litely?
3. What about Linux users who aren't using GNOME? Ubuntu's system font is Ubuntu.
4. This should probably just be done with `system-ui`, not a specific font family name: https://css-tricks.com/snippets/css/system-font-stack/

IngrownMink4:
1) Okay, I'll give it a try.
2) I was thinking of doing it in two different PRs in case there was any particular regression with the combination of colors + typography. But there would be no problem in making the necessary changes in a single PR on all affected topics.
3) Not a bad idea, especially considering the popularity of Ubuntu and Linux Mint. There would be no problem with this (as long as it doesn't replace Cantarell in GNOME).
4) I don't think it's a good idea. Take a look at this article: https://infinnie.github.io/blog/2017/systemui.html

Issue #1774, opened 2023-07-03
Upload image buttons use white theme even on a dark one


lionirdeadman:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

Looks like this:
![image](https://github.com/LemmyNet/lemmy-ui/assets/39467792/1a4f2ed3-2f3a-4c04-9bfe-c950c40fda3b)

### Steps to Reproduce

1. Use a dark theme
2. Click "Create post"
3. Look

### Technical Details

Firefox 114

Likely regression from #1577 

### Lemmy Instance Version

0.18.1-rc.7-75-g90a9b6bd

### Lemmy Instance URL

https://beehaw.org

PR #1776, opened 2023-07-03
Update file input bg colors to match btn's in dark themes

dnswrsrx:
Opting to use the CSS variables for clarity instead of hardcoding the values. The name of the variables should hopefully provide context to why these are being set then.

Addresses issue #1774

### DIFF
--- a/src/assets/css/themes/darkly-compact.css
+++ b/src/assets/css/themes/darkly-compact.css
@@ -2169,11 +2169,12 @@ progress {
   opacity: 1;
 }
 .form-control::file-selector-button {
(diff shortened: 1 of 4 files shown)


jsit:
Please stop modifying the compiled CSS. The only stylesheets that should be edited are the Sass variable stylesheets.

jsit:
This should be addressed by #1782

dnswrsrx:
> Please stop modifying the compiled CSS. The only stylesheets that should be edited are the Sass variable stylesheets.

Apologies for the ignorance on my part! For the sake of learning, which files should I have modified?

Issue #1612, opened 2023-06-26
(0.18.0) Listing type on front page does not honor setting for non-logged in users


chenew:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

Since upgrading to 0.18.0, the listing type on the front page is defaulting to Local, even though the site setting is set to display All.

Set listing type to All in site setting:
<img width="139" alt="image" src="https://github.com/LemmyNet/lemmy-ui/assets/5931313/bae4a4e7-ae82-4226-be97-7c59d98f3b9a">

What a non-logged in user sees:
<img width="201" alt="image" src="https://github.com/LemmyNet/lemmy-ui/assets/5931313/b2cc84a8-76b8-41f6-abcd-b552e42b291e">

Setting the listing type to 'All' helps smaller instances that may not have a lot of posts.

### Steps to Reproduce

1. In site setting, set the Listing Type setting to 'All'
2. Visit the site as a non-logged in user
3. The listing type will default to 'Local'
4. Press F12 to bring up browser developer tools and hold the reload button to empty cache and hard reload

### Technical Details

OS: Ubuntu 22.04
Browser: Tried on latest versions of Brave, Edge, Chrome

### Lemmy Instance Version

0.18.0

randommouse:
Can confirm the same results with Firefox.

delendum:
Confirming this is still an issue in UI: 0.18.1-rc.7

bdonvr:
Yes, as a small instance admin this is quite frustrating. I imagine I (and possibly Lemmy as a whole) have lost a non-trivial amount of potential users because they thought the local feed was all that was going on Lemmy.

It seems @jcgurango solved this in an earlier version of Lemmy-UI, on a fork they made as a "hotfix". However the file he changed has been significantly updated and I don't have the skills to integrate his fix into latest. https://github.com/jcgurango/lemmy-ui/commit/a360c9a3d83e91a0a47d5633b1d4e5ecf5a964df

jcgurango:
[...]

PR #1778, opened 2023-07-03
Fix home page not using site-level listing type #1612

jcgurango:
This is a fix for #1612

### DIFF
--- a/src/shared/utils/helpers/get-query-params.ts
+++ b/src/shared/utils/helpers/get-query-params.ts
@@ -1,19 +1,17 @@
 import { isBrowser } from "@utils/browser";
-import { GetSiteResponse } from "lemmy-js-client";
 
-export default function getQueryParams<T extends Record<string, any>>(
-  processors: {
-    [K in keyof T]: (param: string) => T[K];
-  },
-  site?: GetSiteResponse
-): T {
+export default function getQueryParams<
+  T extends Record<string, any>
+>(processors: {
+  [K in keyof T]: (param: string) => T[K];
+}): T {
   if (isBrowser()) {
     const searchParams = new URLSearchParams(window.location.search);
 
     return Array.from(Object.entries(processors)).reduce(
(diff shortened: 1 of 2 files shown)


dessalines (approved):
Looks good, but @SleeplessOne1917 should take a look.

src/shared/utils/helpers/get-query-params.ts
```diff
@@ -1,17 +1,19 @@
...
+  site?: GetSiteResponse
+): T {
   if (isBrowser()) {
     const searchParams = new URLSearchParams(window.location.search);
 
     return Array.from(Object.entries(processors)).reduce(
       (acc, [key, process]) => ({
         ...acc,
-        [key]: process(searchParams.get(key)),
+        [key]: process(searchParams.get(key), site),
```
SleeplessOne1917, line 16: Why would a GetSiteResponse necessarily have properties that should be query params?

jcgurango: It's for implementing the fallback for listing type to the site-level setting set in the site admin. Would it make more sense to do something like return null in these callbacks and do the fallback on the home component?

SleeplessOne1917: > Would it make more sense to do something like return null in these callbacks and do the fallback on the home component?

Yes. You don't want any of these props in the query params since no page uses them:

```typescript
export interface GetSiteResponse {
  site_view: SiteView;
  admins: Array<PersonView>;
  version: string;
  my_user?: MyUserInfo;
  all_languages: Array<Language>;
  discussion_languages: Array<LanguageId>;
  taglines: Array<Tagline>;
  custom_emojis: Array<CustomEmojiView>;
}
```

jcgurango: Ok I've updated that.

SleeplessOne1917 (approved):

dessalines (approved):

PR #1795, opened 2023-07-04
fix: Revert smaller text size

jsit:
## Description

Would like to hear some more feedback on this one.

#1591 reduced the base font size from 16px to 14px. Users have complained that this is too small.

- https://beehaw.org/comment/559815

Furthermore, this font size reduction was done on `$font-size-base`, when it should have been done on `$font-size-root`.

This PR unsets `$font-size-base` and changes `$font-size-root` to `93.75%`. This makes body text 15px (and `small` text `13.125px`), and also (by using `font-size-root` instead of `font-size-base`) results in more Bootstrap-adherent text size proportions.

User and Community avatars (the small inline ones on posts) go from a height/width of 18.38px to 19.69px. A very small change but not nothing ¯\_(ツ)_/¯

It might be worth considering unsetting `$font-size-root` entirely, so that body text is the browser default of `16px`; but `15px` might be a happy medium.

## Screenshots

Screenshots don't really adequately convey this change; best to try it out locally.

### Before

Screenshot 2023-07-04 at 12 04 08 AM

### After

Screenshot 2023-07-04 at 12 03 21 AM

### DIFF
--- a/src/assets/css/themes/_variables.scss
+++ b/src/assets/css/themes/_variables.scss
@@ -1,3 +1,3 @@
 $link-decoration: none;
 $min-contrast-ratio: 3;
-$font-size-base: 0.875rem;
+$font-size-root: 93.75%;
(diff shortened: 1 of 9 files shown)


SleeplessOne1917:
Looks good to me, though you should get some feedback from users on this one.

dessalines (approved):
I like this better.

alectrocute (approved):

SleeplessOne1917 (approved):

PR #1804, opened 2023-07-04
Make page for federating with not yet federated instances

### LINKED ISSUE #1217: Make searching for not yet federated communities be done on its own page

Currently, the experience for federating with a community on another instance feels like finding a secret in Dark Souls: you feel clever and rewarded for doing it, but it requires picking up on several easy to miss things. Trying to think of this from someone new to Lemmy who's just wrapped their head around what federation is and wants to try federating with a community that's not available on their instance yet, there isn't an easy way to figure out what to do. The current options for finding this out are:

1. **Word of mouth.** There will likely be several posts about how to do this by devs and instance admins preparing for the reddit exodus, but those instructional posts will only be around for so long.
2. **The blurb on a community sidebar.** This blurb is easy to miss on mobile, where the sidebar is collapsed by default.

To make it clearer to users that this feature exists and they should be using it, I propose putting searching for not yet federated communities on its own page. This should only require a change on the UI because the regular search and resolve object are already separate endpoints on the API.

Putting information on how to use the feature in a hard to miss place.

### Additional context

_No response_

SleeplessOne1917:
## Description

Fixed #1217 

## Screenshots

Federate Page

### DIFF
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -90,7 +90,7 @@ export default async (req: Request, res: Response) => {
     }
 
     const error = Object.values(routeData).find(
-      res => res.state === "failed" && res.msg !== "couldnt_find_object" // TODO: find a better way of handling errors
+      res => res.state === "failed"
     ) as FailedRequestState | undefined;
 
     // Redirect to the 404 if there's an API error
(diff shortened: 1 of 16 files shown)


alectrocute:
I love that we're addressing this UX issue, but I think we need to think out the UI more. I'm still confused. :/

SleeplessOne1917:
> ...we need to think out the UI more. I'm still confused. :/

What in particular is confusing? I'm trying to make this less confusing than the current way of doing it, so the easier to understand, the better.

Nutomic:
So with this users would have to enter the community name and instance domain seperately? Whats wrong with entering `https://lemmy.ml/c/memes` as a whole? This has multiple advantages, it also works with other things like posts, users and comments. And it works with objects from other platforms which might use a different url format.

SleeplessOne1917:
The remote follow definitely seems better. I'll take a look at the spec.

PR #1809, opened 2023-07-04
Attempt to fix inability to logout from some instances (subdomains)

alectrocute:
## Description

- Attempt to fix inability to logout from some instances (subdomains)
- Turn auth cookie setting/clearing into module actions for more visiblity
- Extract cookie name into config export

## Screenshots

N/A

### DIFF
--- a/src/shared/utils/browser/clear-auth-cookie.ts
+++ b/src/shared/utils/browser/clear-auth-cookie.ts
@@ -1,5 +1,5 @@
 import * as cookie from "cookie";
-import { authCookieName } from "shared/config";
+import { authCookieName } from "../../config";
 
 export default function clearAuthCookie() {
   document.cookie = cookie.serialize(authCookieName, "", {
--- a/src/shared/utils/browser/clear-auth-cookie.ts
+++ b/src/shared/utils/browser/clear-auth-cookie.ts
@@ -3,7 +3,6 @@ import { authCookieName } from "../../config";
 
 export default function clearAuthCookie() {
   document.cookie = cookie.serialize(authCookieName, "", {
-    expires: new Date("Thu, 01 Jan 1970 00:00:01 GMT"),
     maxAge: -1,
     sameSite: true,
     path: "/",
--- a/src/shared/utils/browser/clear-auth-cookie.ts
+++ b/src/shared/utils/browser/clear-auth-cookie.ts
@@ -1,0 +1,12 @@
+import * as cookie from "cookie";
+import { authCookieName } from "shared/config";
+
+export default function clearAuthCookie() {
+  document.cookie = cookie.serialize(authCookieName, "", {
+    expires: new Date("Thu, 01 Jan 1970 00:00:01 GMT"),
+    maxAge: -1,
+    domain: location.hostname,
+    sameSite: true,
+    path: "/",
+  });
+}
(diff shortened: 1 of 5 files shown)


src/shared/utils/browser/clear-auth-cookie.ts
```diff
@@ -0,0 +1,12 @@
+import * as cookie from "cookie";
+import { authCookieName } from "shared/config";
+
+export default function clearAuthCookie() {
+  document.cookie = cookie.serialize(authCookieName, "", {
+    expires: new Date("Thu, 01 Jan 1970 00:00:01 GMT"),
```
alectrocute, line 6: Adding `expires` attribute with value way in the past, which seems to be a common pattern in other code bases.

src/shared/utils/browser/clear-auth-cookie.ts
```diff
@@ -0,0 +1,12 @@
...
+    maxAge: -1,
```
alectrocute, line 7: Changed from `0` to `-1`.

SleeplessOne1917: What is the point of having both an `expires` and a `maxAge`?

src/shared/utils/browser/set-auth-cookie.ts
```diff
@@ -0,0 +1,16 @@
...
+import { authCookieName } from "shared/config";
+
+export default function setAuthCookie(jwt: string) {
+  const expires = new Date();
+  expires.setDate(expires.getDate() + 365);
+
+  document.cookie = cookie.serialize(authCookieName, jwt, {
+    expires,
+    secure: isHttps(),
+    domain: location.hostname,
```
SleeplessOne1917, line 12: Should probably get rid of domain. Even if it's not causing the issue, I don't think we need it.

src/shared/utils/browser/clear-auth-cookie.ts
SleeplessOne1917: I found [something relevant](https://mrcoles.com/blog/cookies-max-age-vs-expires/):

> **Q. What if I set both expires and max-age in a cookie?**
A. Every browser that supports max-age will ignore the expires regardless of it’s value, and likewise, Internet Explorer will ignore the max-age and just use expires.

alectrocute: Nice, I've removed `expires`. Good call.

src/shared/utils/browser/set-auth-cookie.ts
```diff
@@ -0,0 +1,15 @@
+import { isHttps } from "@utils/env";
+import * as cookie from "cookie";
+import { authCookieName } from "../../config";
+
+export default function setAuthCookie(jwt: string) {
+  const expires = new Date();
+  expires.setDate(expires.getDate() + 365);
+
+  document.cookie = cookie.serialize(authCookieName, jwt, {
+    expires,
```
SleeplessOne1917, line 10: This still uses `expires`. Judging by the article I linked earlier, `expires` has more backward compatibility than `maxAge` (although it seems only some versions of internet explorer don't support `maxAge` - not sure if it's worth bothering). Either way, if `clearAuthCokkie` is using `maxAge`, this should as well. `maxAge` is in milliseconds.

SleeplessOne1917 (changes requested):

src/shared/utils/browser/set-auth-cookie.ts
alectrocute: Good point. Is there any way you could quickly push a commit to fix that here? I'm about to be out of office for the rest of the day.

SleeplessOne1917 (approved):

PR #1810, opened 2023-07-04
Fix data-bs-theme

SleeplessOne1917:
## Description

`data-bs-theme` wasn't being set as expected. This change should fix that.

### DIFF
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -2,7 +2,7 @@ import { initializeSite, setupDateFns } from "@utils/app";
 import { hydrate } from "inferno-hydrate";
 import { Router } from "inferno-router";
 import { App } from "../shared/components/app/app";
-import { HistoryService } from "../shared/services";
+import { HistoryService, UserService } from "../shared/services";
 
 import "bootstrap/js/dist/collapse";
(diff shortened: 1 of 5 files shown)


jsit:
Hm, it was working fine for me? Adding a prop to `<App>` feels odd to me.

jsit:
Also thanks for grabbing those extra themes in the list, I thought I pushed that change but I guess it didn’t make it in.

SleeplessOne1917:
> Hm, it was working fine for me?

When I test the latest changes locally, dark themes always have the upload button with light text on a light background.

> Adding a prop to <App> feels odd to me.

Agreed. This is a kludge. Even with this, it will only work after refreshing the page. We'll need to find a way to make sure the value that gets passed to `data-bs-theme` changes when the user changes.

jsit:
> > Hm, it was working fine for me?
> 
> When I test the latest changes locally, dark themes always have the upload button with light text on a light background.

It works fine for me on page refresh.

SleeplessOne1917:
> It works fine for me on page refresh.

@alectrocute @dessalines does it work for either of you?

dessalines (approved):

PR #1812, opened 2023-07-04
fix: Make new comment indicator easier to distinguish

jsit:
## Description

The "(2 new)" comments indicator is too hard to miss now that it's just gray. This changes that.

### Before
Screenshot 2023-07-04 at 1 12 40 PM
Screenshot 2023-07-04 at 1 12 24 PM

### After
Screenshot 2023-07-04 at 1 11 31 PM
Screenshot 2023-07-04 at 1 11 20 PM

### DIFF
--- a/src/assets/css/themes/_variables.darkly.scss
+++ b/src/assets/css/themes/_variables.darkly.scss
@@ -19,7 +19,7 @@ $cyan: #3498db;
 $primary: $green;
 $secondary: $gray-700;
 $success: $green;
-$dark: $gray-400;
+$dark: $gray-300;
 
 $body-color: $gray-300;
 $body-bg: $gray-900;
--- a/src/assets/css/themes/_variables.darkly.scss
+++ b/src/assets/css/themes/_variables.darkly.scss
@@ -19,7 +19,7 @@ $cyan: #3498db;
 $primary: $green;
 $secondary: $gray-700;
 $success: $green;
-$dark: $gray-300;
+$dark: $gray-400;
 
 $body-color: $gray-300;
 $body-bg: $gray-900;
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -760,7 +760,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         {this.unreadCount && (
           <>
             {" "}
-            <span className="fst-italic">
+            <span className="badge text-bg-info">
               ({this.unreadCount} {I18NextService.i18n.t("new")})
             </span>
           </>


src/shared/components/post/post-listing.tsx
```diff
@@ -760,7 +760,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         {this.unreadCount && (
           <>
             {" "}
-            <span className="fst-italic">
+            <span className="badge text-bg-info">
               ({this.unreadCount} {I18NextService.i18n.t("new")})
             </span>
```
SleeplessOne1917, line 765: I recall before we used the dull color, we colored the font instead of giving it a background. Maybe try that instead while keeping the italic?

src/assets/css/themes/_variables.darkly.scss
```diff
@@ -19,7 +19,7 @@ $cyan: #3498db;
 $primary: $green;
 $secondary: $gray-700;
 $success: $green;
-$dark: $gray-300;
+$dark: $gray-400;
```
jsit, line 22: Sorry, rogue change that shouldn't be in here.

src/shared/components/post/post-listing.tsx
jsit: How about this:

Screenshot 2023-07-04 at 1 26 10 PM
Screenshot 2023-07-04 at 1 26 01 PM

SleeplessOne1917: Which theme color is that? I was thinking of using the info color.

jsit: That's "danger" -- "info" feels semantically correct, but because we're using these theme colors more or less arbitrarily, it leads to some potentially confusing clashes:

Screenshot 2023-07-04 at 1 50 56 PM
Screenshot 2023-07-04 at 1 50 45 PM

SleeplessOne1917: It looks like the UI used to use text-warning:
```
<span className="small text-warning">
    ({this.unreadCount} {i18n.t("new")})
</span>
```

dessalines:
This one I'm not a fan of. We had it as a badge color, but it just looks so much better as the same muted color as the comment count. It just needs emphasized in some way (either via italics like it was), or some other font emphasis.

jsit:
@dessalines Thanks for weighing in. Two different options:

Screenshot 2023-07-04 at 6 44 57 PM


Screenshot 2023-07-04 at 6 46 50 PM

dessalines:
I'll go with the majority, but my vote would be for the original, simple italics:

jsit:
I don't feel strongly about this either. Would be fine with closing this.

dessalines:
Stale PR, but also I prefer the italics emphasis instead of using too many colors.

PR #1814, opened 2023-07-04
fix: Fix mobile thumbnail being too small

jsit:
## Description

Mobile thumbnails weren't taking up their full width; this fixes that.

### Before

Screenshot 2023-07-04 at 1 19 11 PM

### After

Screenshot 2023-07-04 at 1 18 30 PM

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -204,6 +204,11 @@ blockquote {
   height: 5rem;
 }
 
+.mobile-thumbnail-container .thumbnail {
+  width: 100%;
+  height: auto;
+}
+
 .thumbnail svg {
   height: 1.25rem;
   width: 1.25rem;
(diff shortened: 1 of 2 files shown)


SleeplessOne1917 (approved):

src/assets/css/main.css
```diff
@@ -204,6 +204,11 @@ blockquote {
   height: 5rem;
 }
 
+.mobile-thumbnail-container .thumbnail {
+  width: 100%;
```
dessalines, line 208: Bootstrap has a `w-100` and `w-auto`, could you try those?

alectrocute (approved):
Thank you so much for fixing this.

src/assets/css/main.css
jsit: Not on the `.thumbnail` element itself, which doesn’t know whether it’s in mobile or wide view.

PR #1819, opened 2023-07-05
fix: Adjust sidebar banner to wrap bellow the title

### LINKED ISSUE #1607: Sidebar flex changes banner alignment issue
### Requirements

### Summary

After upgrading to Lemmy 0.18.0 the sidebar banner and title are rendered in the same row.

Screenshot 2023-06-26 080031

A couple of changes to the component can fix the layout resulting in

Screenshot 2023-06-26 080147

Specifically:
https://github.com/LemmyNet/lemmy-ui/blob/46ab522143daedf23fb02476769650e0062fa392/src/shared/components/home/site-sidebar.tsx#L36C24-L36C24 < Add the `flex-wrap` class
 And https://github.com/LemmyNet/lemmy-ui/blob/46ab522143daedf23fb02476769650e0062fa392/src/shared/components/common/banner-icon-header.tsx#L19C68-L19C68 < Add `mx-auto` class to center the banner image

### Steps to Reproduce

1. Install Lemmy 0.18.0
2. Upload a banner image
3. Open front page

### Technical Details

This issue is visible on chromium desktop browsers

### Lemmy Instance Version

0.18.0

### Lemmy Instance URL

https://leddit.social/

oliverwhite19:
## Description
Fixes #1607 

Makes it so that the sidebar and banner don't appear in the same row.

### Before

Before

### After

Screenshot 2023-07-04 at 21 36 25

### DIFF
--- a/src/shared/components/home/site-sidebar.tsx
+++ b/src/shared/components/home/site-sidebar.tsx
@@ -33,7 +33,7 @@ export class SiteSidebar extends Component<SiteSidebarProps, SiteSidebarState> {
       <div className="site-sidebar accordion">
         <section id="sidebarInfo" className="card border-secondary mb-3">
           <header
-            className="card-header d-flex align-items-center"
+            className="card-header d-flex align-items-center flex-wrap"
             id="sidebarInfoHeader"
           >
             {this.siteName()}
--- a/src/shared/components/home/site-sidebar.tsx
+++ b/src/shared/components/home/site-sidebar.tsx
@@ -1,3 +1,4 @@
+import classNames from "classnames";
 import { Component, linkEvent } from "inferno";
 import { PersonView, Site, SiteAggregates } from "lemmy-js-client";
 import { mdToHtml } from "../../markdown";
@@ -51,7 +52,7 @@ export class SiteSidebar extends Component<SiteSidebarProps, SiteSidebarState> {
 
   siteName() {
     return (
-      <div className={!this.state.collapsed ? "mb-2" : undefined}>
+      <div className={classNames({ "mb-2": !this.state.collapsed })}>
         <h5 className="mb-0 d-inline">{this.props.site.name}</h5>
         {!this.props.isMobile && (
           <button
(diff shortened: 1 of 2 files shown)


src/shared/components/home/site-sidebar.tsx
```diff
@@ -33,7 +33,7 @@ export class SiteSidebar extends Component<SiteSidebarProps, SiteSidebarState> {
       <div className="site-sidebar accordion">
         <section id="sidebarInfo" className="card border-secondary mb-3">
           <header
-            className="card-header d-flex align-items-center"
+            className="card-header d-flex align-items-center flex-wrap"
             id="sidebarInfoHeader"
           >
             {this.siteName()}
```
SleeplessOne1917, line 39: I'd put a bit of margin bottom on the site name when the banner is displaying, but other than that you're good.

jsit:
I think rather than doing flex-wrap, these should probably just be on different rows to begin with. And I wonder if it would look better with the banner on top?

src/shared/components/home/site-sidebar.tsx
oliverwhite19: I've conditionally added some margin on the header

oliverwhite19:
> I think rather than doing flex-wrap, these should probably just be on different rows to begin with. And I wonder if it would look better with the banner on top?

I've set the elements on different rows and it does simplify things. I don't like the idea of putting the banner on top because then you move the collapse button up and down, which is bad UX imo

SleeplessOne1917 (approved):

src/shared/components/home/site-sidebar.tsx
```diff
@@ -54,7 +51,7 @@ export class SiteSidebar extends Component<SiteSidebarProps, SiteSidebarState> {
 
   siteName() {
     return (
-      <>
+      <div className={!this.state.collapsed ? "mb-2" : undefined}>
```
dessalines, line 54: If the 2nd condition is empty, just do `!this.state.collapsed && "mb-2"` , or ever better, use the `classNames` util.

PR #1822, opened 2023-07-05
Add character counter to markdown text area

### LINKED ISSUE #1671: Character Counter
### Requirements

### Problem

For long, or particularly in-depth posts, it is possible to silently exceed the character length limit, with no indication of such. Since Lemmy tends to fail silently (unrelated issue), this can cause problems where a user might write a longer post or comment, attempt to send it, and then have it silently fail, as the post/comment is too long, but with no indication of such.

### Solution/Proposal

So far, I have two variants in mind:

1. Text Entry boxes should have a counter that either shows the current amount of characters in the box, over the total amount of characters available (`343 / 10000`)

2. A counter that shows how many characters are left (`9757`). 

In this proposal, if the limit is exceeded, it might begin to count negative, showing how far beyond the character limit a given post/comment is.

A less precise alternative might be to have a graphical indicator instead of a numeric counter, that shows how close to the limit a post/comment is, with a noticeable change if a post exceeds character limits.

m-gail:
## Description

Adds a character counter to MarkdownTextArea.
The character counter only shows, if the prop `maxLength` is set.
Sets the maxLength of PostForm and CommentForm to 50000 and 10000.

Resolves #1671 

### Before

charcounterbefore

### After

charcounterafter

charcounterafterinvalid


SleeplessOne1917 (approved):

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -225,6 +230,8 @@ export class MarkdownTextArea extends Component<
                 >
                   <Icon icon="help-circle" classes="icon-inline" />
                 </a>
+                {this.props.maxLength !== undefined &&
```
jsit, line 233: Can’t this use optional chaining? `this.props?.maxLength`

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -331,6 +337,24 @@ export class MarkdownTextArea extends Component<
     );
   }
 
+  getCharacterCounter(contentLength: number, maxLength: number) {
+    return (
+      <>
+        <div className="flex-grow-1"></div>
```
jsit, line 343: Probably don’t need an empty flex-grow. Try `justify-content-end` instead.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -234,6 +241,7 @@ export class MarkdownTextArea extends Component<
                     "form-control border-0 rounded-top-0 rounded-bottom",
                     {
                       "d-none": this.state.previewMode,
+                      "is-invalid": contentLength > maxLength,
```
jsit, line 244: We might want to put an actual `invalid` prop on the textarea, too.

jsit (changes requested):

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -331,6 +337,24 @@ export class MarkdownTextArea extends Component<
...
+    return (
+      <>
+        <div className="flex-grow-1"></div>
+        <div
+          className={classNames(
+            "px-2 py-1",
+            contentLength > maxLength ? "text-danger" : "text-muted"
+          )}
+        >
+          {contentLength}
```
jsit, line 350: Might be nice to `toLocaleString` these numbers so that, for instance, comma thousands separators show up in US English.

src/shared/components/common/markdown-textarea.tsx
m-gail: `textarea` only has `aria-invalid`, which I added now. Is this what you had in mind or is there some other way I can mark it as invalid?

src/shared/components/common/markdown-textarea.tsx
m-gail: I added `ms-auto`. It behaves slightly differently, if the character count is on its own line. Before it would show at the start of the line then, now it shows at the end. Though I think it looks better with `ms-auto`

src/shared/components/common/markdown-textarea.tsx
m-gail: Props cannot be undefined, so there is no need. However, having `!== undefined` was not necessary.

alectrocute:
Any chance we can put this counter below the textarea instead of in the button bar? It's already so cramped up there.

m-gail:
> Any chance we can put this counter below the textarea instead of in the button bar? It's already so cramped up there.

I just tried out a few alternative positions:

Below left:

belowleft

Below right:

belowright

In the textarea (this one requires a bit of additional padding at the bottom of the textarea, so the text doesn't overlap with the counter):

inbox

alectrocute:
> Below left:
> 
> belowleft

Below-left has my vote. Nice work, thanks for sharing all these options.

Maybe it would also make sense to hide this counter entirely if the post has less than half the max chars? Hmm... 

I'll let others chime in.

dessalines:
I also like below left (or inside textarea)

dessalines:

Issue #1828, opened 2023-07-05
Hide Admin / Mod tags unless using "speak as moderator" (distinguishing post)


Crashdoom:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Currently, Lemmy always shows the "M" (Mod) and "A" (Admin) tags on posts and comments made by moderators and instance admins respectively which can be off-putting to some users and also discourages mods and admins from taking part in discussions for fear of their thoughts and opinions being construed as an official statement.

I'd like to see those tags hidden by default, unless the post or comment is distinguished using the existing "speak as moderator" button. Similar to how moderators and admins on Reddit can take part in individual discussions without their comments constantly being marked as "speaking officially."

I don't believe this is a core Lemmy issue as the flag to "distinguish" a post is available via the API request made by lemmy-ui, but that this change would make https://github.com/LemmyNet/lemmy/issues/3066 more relevant to help identify when a post is being made "officially" by a mod / admin.

vinoff:
I am a mod in a sub and I find the ability to toggle the Mod tag a must have! I am considering creating a secondary account just so I can post in the sub I mod without it seeing like I am spamming too much. 

I find this specially important in the current stage of Lemmy, in which it is extremely important to create engaging content. Mods are usually those who post more at the initial phase of a community. Currently, I am a bit scared of doing that with all the "Mod" tags attached.

dessalines:
> I am a mod in a sub and I find the ability to toggle the Mod tag a must have! I am considering creating a secondary account just so I can post in the sub I mod without it seeming like I am spamming too much

To me it seems not-transparent to hide the fact that you're actually a mod / admin, to others, and have extra abilities that they don't. Showing the mod / admin badge is a deliberate choice to be transparent to everyone who engages with you.

Crashdoom:
> > I am a mod in a sub and I find the ability to toggle the Mod tag a must have! I am considering creating a secondary account just so I can post in the sub I mod without it seeming like I am spamming too much
> 
[...]

PR #1833, opened 2023-07-06
Added distinguished check for isMod & isAdmin comment user badges

CBerry22:
## Description

Added a distinguished check to the isMod and isAdmin UserBadges on comments so they will only appear when speaking as a moderator.

This will fix the comment side of  #1828 

### Before
before undistinguished

### After
after undistinguished
before undistinguished

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -319,8 +319,8 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
               <UserBadges
                 classNames="ms-1"
                 isPostCreator={this.isPostCreator}
-                isMod={isMod_}
-                isAdmin={isAdmin_}
+                isMod={isMod_ && cv.comment.distinguished}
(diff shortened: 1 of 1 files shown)


dessalines:
I spose I'd be okay with this, but I personally think its useful to know which users are moderators and admins, even if they aren't speaking in an official capacity.

Kommynct:
Well, actually, because of display names, it's somewhat problematic

Multiple people can have the same display name, what's to stop someone from impersonating a mod?

Crashdoom:
> Well, actually, because of display names, it's somewhat problematic
> 
> Multiple people can have the same display name, what's to stop someone from impersonating a mod?

Valid. Hadn't thought of that. Could add a tag next to the user if they're non-local showing either their full username (`username@domain.tld`) or just `domain.tld`. That'd probably be a good idea irrespective of this change since impersonation like this would be really easy under the current system.

Nutomic:
Better leave this as is.

PR #1838, opened 2023-07-06
Improved expand/minimize buttons for the inbox for dark themes

Kradyz:
## Description

In the UI version 0.18.1-rc.11, the icon-minus-square and icon-plus-square in the Inbox window look dark, even when a dark theme is used, dramatically decreasing the contrast.

This change uses a similar class to the one that is used for the same type of button when used to expand/minimize posts. It improves the way the button looks against both light and dark themes.

### Before
image

### After
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lemmy-ui",
-  "version": "0.18.1-rc.10",
+  "version": "0.18.1-rc.11",
   "description": "An isomorphic UI for lemmy",
   "repository": "https://github.com/LemmyNet/lemmy-ui",
   "license": "AGPL-3.0",


SleeplessOne1917 (approved):

alectrocute:
I think there's some unneeded classes here. Look at the other expand buttons:

https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/comment/comment-node.tsx#L301-L311

alectrocute (changes requested):
I think there's some unneeded classes here. Look at the other expand buttons:

https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/comment/comment-node.tsx#L301-L311

PR #1840, opened 2023-07-07
Dark theme tweaks

SleeplessOne1917:
## Description

Post titles weren't showing up very well in the vaporwave dark and i386 themes. This fixes that.

### Before

image
image

### After
image
image


src/assets/css/themes/_variables.i386.scss
```diff
@@ -26,7 +26,7 @@ $danger: #aa0000;
 $info: #00aaaa;
 $warning: #aa00aa;
 $light: $gray-800;
-$dark: black;
+$dark: $gray-300;
```
SleeplessOne1917, line 29: I hate making the dark colors so light, but it's how darkly gets around the issue. This is worth revisiting some point in the future.

dessalines (approved):

PR #1846, opened 2023-07-07
Add a disclaimer about federation on sign up

sunaurus:
## Description

I have seen a lot of users confused about data deletion, what is shared with other servers, etc (especially in context of the GDPR). I think it makes sense to ask for explicit consent regarding federation on user sign up. Feedback welcome!

Depends on https://github.com/LemmyNet/lemmy-translations/pull/81

## Screenshots

image


alectrocute:
I'm tempted to suggest that we add a new page that includes info like this, and link to it from the sign up form instead.

sunaurus:
I think most big instances already have this type of thing in /legal, but the reason I thought of adding this explicit federation consent message is that I think that it's way more user friendly than the separate legal page. My fear is that 99% of users will always skip any privacy policy type links, and thus might not fully understand what they're signing up for.

alectrocute:
> I think most big instances already have this type of thing in /legal, but the reason I thought of adding this explicit federation consent message is that I think that it's way more user friendly than the separate legal page. My fear is that 99% of users will always skip any privacy policy type links, and thus might not fully understand what they're signing up for.

Sure, good point, but I’ll counter with: we don’t even have a ‘I agree to…’ checkbox or link to a TOS at all on this sign-up page.

wiki-me:
Starting to try to educate users on online privacy is a whole seperate thing IMO, maybe we could link to some other site, for example most users probably don't know services even for big and "successful" companies like google or facebook [get hacked](https://en.wikipedia.org/wiki/List_of_data_breaches), or that information about username and email from multiple hacks can be used to dox them.

alectrocute:
Where are we at with this? I would still love to see this added as an extra page and accompanying checkbox.

dessalines:
Should also add a link to the legal page: https://github.com/LemmyNet/lemmy-translations/pull/81

Nutomic:
A link to the documentation would also work. Btw the message mentions "instance" a lot but new users might have no idea what that means. So consider changing it to "server" or explaining it.

PR #1858, opened 2023-07-07
Update post listing to prefer local image when available

lemmus-org:
## Description

Currently, the backend is copying images from federated posts locally and providing the URL as `thumbnail_url`. These images are mostly unused as the UI is preferring to use the post `url`, which in the case of images are from the originating instance.

This changes the logic on the UI to prefer the `thumbnail_url` of the local instance. It should help reduce the load on some of the larger instances while ensuring a good UX even if the origin instance is having connectivity issues. The change has been applied to my instance and appears to be working as intended. That said, I may be overlooking some edgecases.

I think this is ultimately a band-aid for now until we can come up with a more comprehensive solution for image proxying/caching on the backend.

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -310,16 +310,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     const url = post.url;
     const thumbnail = post.thumbnail_url;
 
-    if (url && isImage(url)) {
-      if (url.includes("pictrs")) {
-        return url;
-      } else if (thumbnail) {
-        return thumbnail;
-      } else {
-        return url;
-      }
-    } else if (thumbnail) {
+    if (thumbnail) {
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

necropola:
Loding image resources from the originating instance allows the originating instance to kinda "track" which posts are being read on other instances. Without the user on the other instance even clicking on a link.

While this PR fixes it for post thumbnail images, inline images in posts and comments are still loaded from the originating instance without any user interaction and hence should be locally cached, too.

lemmus-org:
@necropola Correct. This was only intended as a short-term fix for some UX issues. For a long-term fix, I've started https://github.com/LemmyNet/lemmy/pull/3550 to address it more globally. Hopefully, I'll have more time soon to see it through but am happy to collaborate on it in the meantime if there's interest.

PR #1867, opened 2023-07-07
User inferno prompt instead of handrolled one

SleeplessOne1917:
## Description

Now that the newest version of inferno was released with @fheft's prompt fix, we no longer need to handroll our own prompt component. I also took the liberty of updating dependencies.


alectrocute (approved):
Nice!

PR #1874, opened 2023-07-08
Add Toast Messages for Bad Logins

rosenjcb:
## Description

Right now if you try to login and you provide an unknown username/email or a wrong password, you will get no toast message. This PR adds toast messages for those scenarios and uses the messages already found in `lemmy-translations`.

## Screenshots

image

### DIFF
--- a/src/shared/components/home/login.tsx
+++ b/src/shared/components/home/login.tsx
@@ -169,14 +169,8 @@ export class Login extends Component<any, State> {
             i.setState({ showTotp: true });
             toast(I18NextService.i18n.t("enter_two_factor_code"), "info");
           }
-          if (loginRes.msg === "couldnt_find_that_username_or_email") {
-            toast(
-              I18NextService.i18n.t("couldnt_find_that_username_or_email"),
-              "danger"
-            );
-          }
-          if (loginRes.msg === "password_incorrect") {
-            toast(I18NextService.i18n.t("password_incorrect"), "danger");
+          if (loginRes.msg === "incorrect_login") {
+            toast(I18NextService.i18n.t("incorrect_login"), "danger");
           }
 
           i.setState({ loginRes: { state: "failed", msg: loginRes.msg } });
--- a/src/shared/components/home/login.tsx
+++ b/src/shared/components/home/login.tsx
@@ -169,6 +169,15 @@ export class Login extends Component<any, State> {
             i.setState({ showTotp: true });
             toast(I18NextService.i18n.t("enter_two_factor_code"), "info");
           }
+          if (loginRes.msg === "couldnt_find_that_username_or_email") {
+            toast(
+              I18NextService.i18n.t("couldnt_find_that_username_or_email"),
+              "danger"
+            );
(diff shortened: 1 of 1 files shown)


src/shared/components/home/login.tsx
```diff
@@ -169,6 +169,15 @@ export class Login extends Component<any, State> {
...
           }
+          if (loginRes.msg === "couldnt_find_that_username_or_email") {
+            toast(
+              I18NextService.i18n.t("couldnt_find_that_username_or_email"),
+              "danger"
+            );
+          }
+          if (loginRes.msg === "password_incorrect") {
+            toast(I18NextService.i18n.t("password_incorrect"), "danger");
+          }
```
SleeplessOne1917, line 180: Should we be displaying different error messages for these? I would think letting potential attackers know which credential they got wrong is ill advised.

rosenjcb: It could be an attack vector - especially if there's no retry limit on failed password attempts. I can't seem to find a good error message that's already provided in `lemmy-translations`. Do you have a suggestion?

SleeplessOne1917: Not off the top of my head. However, worth pointing out is that even if we change the toast on the frontend, the response from the server still has distinct message for each case, so there will need to be a backend change as well.

alectrocute: If the backend response `login_res` is already spilling the beans, then it doesn’t matter much whether we use unique error toasters or not. The attack vector will still be present.

alectrocute: Ah, I see you already mentioned that @SleeplessOne1917. I need to read better before commenting.

rosenjcb: @SleeplessOne1917 This is what I can do.

1.) Make a PR on `lemmy-translations` to add a new entry `{"incorrect_login": "The login credentials are incorrect."}`
2.) Make a PR on `lemmy` to swap out the API response with this new generic entry instead of `couldnt_find_that_username_or_email` and `password_correct`. 
3.) Update this PR to  reflect the changes once they've been merged.

Would that work?

SleeplessOne1917: @rosenjcb sounds good!

rosenjcb: Great I have https://github.com/LemmyNet/lemmy/pull/3549 and https://github.com/LemmyNet/lemmy-translations/pull/83 set up. The code change isn't covered by tests though, so it's probably safe for someone to check out the change and run it locally before merging.

jsit: Thank you for the attention to security on this

rosenjcb: The two PRs above were approved and merged. I went ahead and swapped the condition check + toast response to use the generic incorrect_login string. Thanks for the feedback and support!

SleeplessOne1917 (approved):

dessalines (approved):

PR #1875, opened 2023-07-08
Remote follow

SleeplessOne1917:
## Description

Taking @Nutomic's suggestion from #1804 to make a flow for federating with communities from remote instances. So far I've only made a modal for it. I could add a route to the UI to handle these redirects, but what about Mastodon, Pleroma, Calckey, etc? Is there any standard path to target for this sort of thing?

## Screenshots

image

#### Success
remote_fetch_success

#### Loading
Screenshot_20230722_223219

#### Failed
Screenshot_20230722_223538


wiki-me:
It's a nice feature, but TBH most users will probably be served better by some browser extension that sets their "home instance" ([lemmy tools?](https://github.com/howdy-tsc/LemmyTools)), maybe suggesting that also can make people think lemmy can be less of a hassle?

Madis0:
How about showing all three options in the modal and therefore removing the blue info box from the right? 

* Log in to subscribe on instance.example
* Enter instance to subscribe on it: ________
* Search `!community@instance.example` on your instance to subscribe

Nutomic:
Based on [this comment](https://socialhub.activitypub.rocks/t/what-is-the-current-spec-for-remote-follow/2020/15) the path is available via webfinger. So it will also require some backend changes.

dessalines:
That means we will need to add a front end route for this:

`https://MY_INSTANCE/activitypub/externalInteraction?uri={uri}`

SleeplessOne1917:
@Nutomic @dessalines @jsit @alectrocute I updated this to have a page to redirect to. I could use some eyes on this.

src/shared/components/home/login.tsx
```diff
@@ -20,7 +33,69 @@ interface State {
...
+
+        break;
+      }
+    }
+  }
+}
+
+function handleLoginUsernameChange(i: Login, event: any) {
+  i.state.form.username_or_email = event.target.value.trim();
+  i.setState(i.state);
```
dessalines, line 87: Use the single line setState we do elsewhere.

Also it seems like you added these functions, but didn't remove the class ones.

src/shared/components/remote-fetch.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+    const { uri } = getRemoteFetchQueryParams();
+    const remoteCommunityName = uri ? uriToQuery(uri) : "remote community";
+
+    switch (res.state) {
+      case "success": {
+        const communityView = res.data.community as CommunityView;
+        return (
+          <>
+            <h1>Community Federated!</h1>
```
dessalines, line 142: i18n at some point.

src/shared/components/remote-fetch.tsx
```diff
@@ -0,0 +1,218 @@
...
+            </div>
+          </>
+        );
+      }
+
+      case "loading": {
+        return (
+          <>
+            <h1>
+              Fetching {remoteCommunityName}
```
dessalines, line 172: i18n

src/shared/components/remote-fetch.tsx
```diff
@@ -0,0 +1,218 @@
...
+              <Spinner large />
+            </h5>
+          </>
+        );
+      }
+
+      default: {
+        return (
+          <>
+            <h1>Could not fetch {remoteCommunityName}</h1>
```
dessalines, line 185: i18n

dessalines (commented):
The components and everything seem to work okay, but I'm struggling coming up with a way we can actually test this, besides putting it on our federation test instances.

src/shared/components/remote-fetch.tsx
```diff
@@ -0,0 +1,218 @@
...
+    resolveObjectRes: { state: "empty" },
+    isIsomorphic: false,
+    followCommunityLoading: false,
+  };
+
+  constructor(props: any, context: any) {
+    super(props, context);
+
+    if (FirstLoadService.isFirstLoad) {
+      //      const { resolveObjectRes } = this.isoData.routeData;
```
SleeplessOne1917, line 89: I'll need to uncomment this.

Nutomic:
I thought about it more and figured out how its supposed to work. The flow is like this:

- User has an account on lemmy.ml
- User browses sopuli.xyz and finds an interesting community 
- User clicks on "remote follow" button, enters identifier `@nutomic@lemmy.ml` and submits
- lemmy-ui makes a GET request to `https://lemmy.ml/.well-known/webfinger?resource=acct:nutomic@lemmy.ml`
    - Note there is no leading @
    - Can fail if identifier was entered wrong, target server is down or server doesnt support webfinger
- lemmy-ui parses the webfinger response to extract the link with `rel: http://ostatus.org/schema/1.0/subscribe` and retrieves the `template` value
    - Its possible that this value doesnt exist (remote follow not supported)
    - If target is a Lemmy instance, it might have a user and community with the same name. Webfinger doesnt support that properly, so both will be returned together in the same response. In that case, choose the one where property `https://www.w3.org/ns/activitystreams#type` has value `Person`
- Then replaces the `{url}` with community actor id (using urlencode) and redirects the user to this address
- User is now on his home instance lemmy.ml
- Handle login if necessary
- Fetch the community using resolve_object
- Display a dialog with follow button

Would be good to include this flow as a code comment for future reference.

I also updated the backend PR https://github.com/LemmyNet/lemmy/pull/3738. Here we can freely choose the route used for remote follow, it doesnt have to be `/activitypub/externalInteraction` or anything specific. It only needs to contain the `{url}` placeholder, and be consistent between backend and frontend.

poVoq:
Why the need to put the full user id?

On Mastodon etc it is sufficient to put the home instance domain only, which is much more user friendly.

Nutomic:
@poVoq You mean like `https://lemmy.ml/.well-known/webfinger?resource=acct:nutomic`? That would be possible to implement, feel free to open an issue/pr.

poVoq:
> User clicks on "remote follow" button, enters identifier @nutomic@lemmy.ml and submits

I mean this part. Why not just "lemmy.ml"?

If a user is logged in the account is automatically recognized and if they aren't you can just redirect to the login screen.

This is how other fediverse software does it.

Nutomic:
I think you misunderstand the purpose of remote follow. It is used when Im browsing an instance where I dont have an account, so I have to tell it where my account lives. You can try it [here on Peertube](https://tilvids.com/c/slopesgameroom_channel/videos) with a Mastodon account, it requires that you input `name@instance.com`.

poVoq:
No, that is exactly what I understood and I still think the name part is unnecessary and confusing.

SleeplessOne1917:
The way the flow is currently implemented in this PR is closer to what @poVoq describes than the flow laid out  [in @Nutomic's comment](https://github.com/LemmyNet/lemmy-ui/pull/1875#issuecomment-1655351733). The way it works right now is:

1. User has an account on `lemmy.ml`
2. User browses `sopuli.xyz` and finds an interesting community
3. User clicks on "remote follow" button, enters domain `lemmy.ml` and submits.
4. If user is logged in they go directly to step 5. If not, they are brought to the login page and redirected after logging in.
5. User is sent to `https://lemmy.ml/activitypub/externalInteraction?uri=https://sopuli.xyz/c/interestingCommunity` (community uri not encoded here for readability)
6. `resolveObject` request is made with user's auth and a query in the form of `!interestingCommunity@sopuli.xyz`.
7. Based on the `resolveObject` response:
    - **Success**: display the community's info in a card with a button for the user to subscribe.
    - **Loading**: Loading spinner
    - **Failure**: Failure screen displayed

Nutomic:
> User is sent to https://lemmy.ml/activitypub/externalInteraction?uri=https://sopuli.xyz/c/interestingCommunity (community uri not encoded here for readability)

This is going to break if the user is on another platform. There it needs to go to `https://mastodon.social/authorize_interaction?uri={uri}"`. Thats why the webfinger query is necessary, and for that we need the username.

poVoq:
Mastodon does not require this, see:

image

SleeplessOne1917:
I'm struggling to figure out what the right way forward on this is given the lack of consensus. @Nutomic @poVoq @dessalines

dessalines:
@Nutomic is any of this PR necessary then? https://github.com/LemmyNet/lemmy/pull/3738

If not, I'd say we should merge this, and use the federation test instances to debug / tweak it.

Nutomic:
You can already test this with the setup in Lemmy repo `docker/federation`, by manually building a lemmy-ui image from this branch and putting that in the docker-compose file.

@poVoq I think that approach only works with platforms that use the same endpoint `/authorize_interaction?uri={uri}`. I tried with peertube just now and it gives an error 404 because it uses `/remote-interaction?uri=`. So the backend PR is necessary if we want to support different platforms such as Peertube.

SleeplessOne1917:
I'm going to work on this again. @Nutomic I'm not sure what steps need to be taken to make this work with the backend change. What endpoint should I be hitting for this? It might make sense to add logic for this to the JS client as well.

Nutomic:
The Lemmy backend tells other Fediverse software that remote follow is available at `/activitypub/externalInteraction?uri={{uri}}`. Where `{{uri}}` is the actor_id of a community which can be passed to `/api/v3/site/resolve_object`. So you need to add a handler in lemmy-ui for /activitypub/externalInteraction which resolves the given community, and then shows a dialog "Do you want to follow this community yes/no?"

SleeplessOne1917:
@Nutomic the existing flow is already the same as what you described in your most recent comment. See this video:

https://github.com/LemmyNet/lemmy-ui/assets/28871516/9c8647ed-dde8-4ab5-90b4-7c862aa296e2


The only rub is that even though the remote community is resolved, the subscribed state never stops being pending. I don't know if this is due to a flaw in how I implemented it or if it's a quirk of testing locally.

Nutomic:
I tested it locally and can reproduce that problem. Its weird because the follow goes through if I fetch another community manually and press the sidebar subscribe button. Is there any difference between that and the new follow button you added for the remote follow dialog?

Im also getting various `API error: unknown` logged by lemmy-ui when viewing a remote community. Any idea what that is? Seems like the backend is returning 404 for some reason, would be good if lemmy-ui returned the actual backend error.

SleeplessOne1917:
I just tested this with the existing way of fetching remote communities by using the search page and ran into the exact same problem. I think this might be an issue with testing federation locally.

src/shared/components/common/subscribe-button.tsx
```diff
@@ -0,0 +1,208 @@
...
+function submitRemoteFollow(
+  { state: { instanceText }, props: { communityActorId } }: RemoteFetchModal,
+  event: Event,
+) {
+  event.preventDefault();
+
+  if (!validInstanceTLD(instanceText)) {
+    // TODO: Figure out appropriate way of handling invalid domainss
+    console.log("I should do something about this.");
+  }
```
SleeplessOne1917, line 112: This will need to be handled. @dessalines will a toast message about the error suffice?

dessalines: Yep: `toast(ERR_STR, "danger");`

SleeplessOne1917:
I just added the remote follow logic to the subscribe buttons on the communities page too:
image

I'm unsure if I should keep the button look here or go back to making it look like a link like it was before.

dessalines (approved):
Not a huge fan of that button, for some reason they don't look great in tables. But I leave that up to you.

Once you fix the conflicts, I'll start making some RC's, because this is going to need some playing around with on the test instances for sure.

PR #1881, opened 2023-07-08
Add polling to refresh post comments/get edits.

### LINKED ISSUE #1565: AutoUpdate setting
### Requirements

Is it possible to add an AutoUpdate setting to Lemmy? It is annoying for me to have to refresh to see new posts. In my opinion, AutoUpdate should be extended to communities and not just posts.

SleeplessOne1917:
## Description

Fixes #1565. This should update the post/comments on the post page every minute. Interestingly enough, there was already a config value for this set in the utils.

### DIFF
--- a/src/shared/components/post/post.tsx
+++ b/src/shared/components/post/post.tsx
@@ -19,7 +19,7 @@ import {
   restoreScrollPosition,
   saveScrollPosition,
 } from "@utils/browser";
-import { debounce, randomStr } from "@utils/helpers";
+import { debounce, poll, randomStr } from "@utils/helpers";
 import { isImage } from "@utils/media";
 import { RouteDataResponse } from "@utils/types";
 import autosize from "autosize";
@@ -76,7 +76,7 @@ import {
   SavePost,
   TransferCommunity,
 } from "lemmy-js-client";
-import { commentTreeMaxDepth } from "../../config";
+import { commentTreeMaxDepth, postRefetchSeconds } from "../../config";
 import {
   CommentNodeI,
   CommentViewType,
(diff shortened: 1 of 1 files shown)


alectrocute:
Hmm… What are the chances we could re-integrate websockets for stuff like this? Just an idea. Polling is probably fine, too.

alectrocute:
Also, I’m worried about the reloading causing users to lose their scroll position in the comments.

SleeplessOne1917:
> Hmm… What are the chances we could re-integrate websockets for stuff like this?

I think that would be a good idea, with the caveat that the websocket API is limited to things that actually benefit from instantaneous bilateral client-server communication. Things like replies, edits, messages, and live comments would all be well suited to this. I think the old websocket API had a really bad flaw: parity with the REST API. This lead to using websockets for things that it was a bad choice for, like fetching a post, list of posts for the home/community pages, user profiles, etc. I think that's why the pre-0.18.x versions of the UI project had so many annoying bugs, along with being more difficult to maintain.

However, I think polling is a good short to medium term solution.

> Also, I’m worried about the reloading causing users to lose their scroll position in the comments.

I could see that being an issue if we used websockets as well. There is some comfort to be had in the fact that this is only making API requests to update the UI on the client side, not reloading the whole page.

makotech222:
Most users hate the frontend refreshing while they are using the site. At least leave a user setting to disable it.

jsit:
> Most users hate the frontend refreshing while they are using the site. At least leave a user setting to disable it.

I agree. Please don’t do this without users’s consent. It’s already so easy to lose your place.

Would be nice if maybe a toast popped up that said “There is new content. Reload?”

Nutomic:
Fetching this data every second will cause an immense increase in server load, basically a builtin ddos. At least change it to 5 minutes or more. But honestly I dont see the point, if you want to reload the comments just press F5.

PR #1883, opened 2023-07-08
Use canonical URLs

### LINKED ISSUE #1418: Use canonical URLs
### Requirements

- [X] Is this a server side (not related to the UI) issue? Use the [Lemmy back end](https://github.com/LemmyNet/lemmy) repo.

Each Lemmy site duplicates all the content of every federated site. It's believe that Google and other search engines penalize for duplicated content, although this isn't known for sure. At the least, this spreads the pagerank thin and confuses searchers. It makes discoverability horrible.

Part of the great utility of reddit is that it functions as a sort of repository of knowledge. [Many people noticed how the blackout made Google so much worse](https://www.theverge.com/2023/6/13/23759942/google-reddit-subreddit-blackout-protests) because that knowledge couldn't be discovered. Lemmy should account for SEO to ensure it can help fill that role. This is very related to #1285 but with a specific solution.

Each post should have a rel=canonical meta tag back to the original post. Currently no such tag exists, which is problematic since Google says "Duplicate without user-selected canonical" on every page it indexes since it indexes both https://civilloquy.com/post/1298 and https://civilloquy.com/post/1298?scrollToComments=true. Canonical tags are needed regardless of implementation. However, rather than each Lemmy instance claiming its own mirror of content as the canonical, I propose that it link back to the original post. So https://civilloquy.com/post/1298?scrollToComments=true would list https://beehaw.org/post/495714 as its canonical URL. This would pool pagerank to the originating post, make it clear we're not trying to scam Google with duplicate content, make search results much cleaner, and give proper credit to the originating community. I know such links are already calculated for the fedilink, so hopefully it wouldn't be too hard to do either.

The alternative would be adding canonical tags to your own instances copy of each post (or giving up on them entirely), however, as described I find this likely to be horrible for SEO and a bit dishonest.

### Additional context

_No response_

mminer237:
Fixes #1418

This links each community, user profile, and post to the original instance it was made on as canonical. It also links comments to their posts as canonical, since [that seems recommended](https://proposals.codidact.com/posts/288902) and that's what reddit does. The only pages with query parameters included in the canonical URL are search pages since those are pretty inherent to the page's content and since that seems to be what other sites do.

### DIFF
--- a/src/shared/components/search.tsx
+++ b/src/shared/components/search.tsx
@@ -466,6 +466,10 @@ export class Search extends Component<any, SearchState> {
         <HtmlTags
           title={this.documentTitle}
           path={this.context.router.route.match.url}
+          canonicalPath={
+            this.context.router.route.match.url +
+            this.context.router.route.location.search
+          }
         />
         <h1 className="h4 mb-4">{I18NextService.i18n.t("search")}</h1>
         {this.selects}
--- a/src/shared/components/common/html-tags.tsx
+++ b/src/shared/components/common/html-tags.tsx
@@ -8,6 +8,7 @@ import { I18NextService } from "../../services";
 interface HtmlTagsProps {
   title: string;
   path: string;
+  canonicalPath?: string;
   description?: string;
   image?: string;
 }
@@ -16,6 +17,8 @@ interface HtmlTagsProps {
 export class HtmlTags extends Component<HtmlTagsProps, any> {
   render() {
     const url = httpExternalPath(this.props.path);
+    const canonicalUrl =
+      this.props.canonicalPath ?? httpExternalPath(this.props.path);
     const desc = this.props.description;
     const image = this.props.image;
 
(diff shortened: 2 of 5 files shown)


src/shared/components/search.tsx
```diff
@@ -466,6 +466,10 @@ export class Search extends Component<any, SearchState> {
         <HtmlTags
           title={this.documentTitle}
           path={this.context.router.route.match.url}
+          canonicalPath={
+            this.context.router.route.match.url +
+            this.context.router.route.location.search
```
SleeplessOne1917, line 471: I thought `url` already contains the query params?

mminer237: I would have thought so from the name, but apparently not:

```js
console.log(this.context.router.route);
```

```json
{
  "location": {
    "pathname": "/search",
    "search": "?q=test&type=All&listingType=All&page=1&sort=TopAll",
    "hash": "",
    "state": null,
    "key": "1hxl4e7m"
  },
  "match": {
    "isExact": true,
    "params": {},
    "path": "/search",
    "url": "/search"
  }
}
```

A separate change might want to change the OpenGraph URLs some places as it indeed always links without the query parameters.

SleeplessOne1917 (approved):
Thank you for taking care of this.

dessalines (approved):
Thanks!

PR #1887, opened 2023-07-09
Added newline button to markdown editor

bryewalks:
## Description
- Added 'newline' button to markdown editor to add line break.
- Added newline svg (left pointing arrow this can be changed if there is a better suggestion for a newline icon)

Resolves: https://github.com/LemmyNet/lemmy-ui/issues/1856
Depends on: https://github.com/LemmyNet/lemmy-translations/pull/84
### Before
Screenshot_20230709_020431-1

### After
Screenshot_20230709_014353

https://github.com/LemmyNet/lemmy-ui/assets/25917358/e9c7b03c-fa77-41a5-92c9-b709eaca7e73

### DIFF
--- a/src/assets/symbols.svg
+++ b/src/assets/symbols.svg
@@ -258,5 +258,8 @@
       <path d="M8.72046 10.6397L14.9999 7.5" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
       <path d="M8.70605 13.353L15 16.5" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
     </symbol>
+    <symbol id="icon-newline" viewBox="0 0 24 24">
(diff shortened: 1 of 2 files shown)


SleeplessOne1917:
I don't understand the benefit of this. Users can create a new line with a single press of the Enter key on both desktop and mobile.

alectrocute:
> I don't understand the benefit of this. Users can create a new line with a single press of the Enter key on both desktop and mobile.

+1. Already too many buttons as is, too.

bryewalks:
> I don't understand the benefit of this. Users can create a new line with a single press of the Enter key on both desktop and mobile.

Think the original request was  made because the enter key actually does not create a newline in the markdown. User will have to know to add two spaces. Can close if this is not needed though was looking for something easy for first task.

https://github.com/LemmyNet/lemmy-ui/assets/25917358/f8f92e0b-519e-4680-9cef-b428d67f6734

dessalines:
Same, this seems really pointless to me.

Damaskox:
I cannot see this newline button. I don't know, how to make a new line.

TonyHoyle:
+1 from me.. at the very least document this '\' thing somewhere as it's really frustrating when you write out a nicely formatted piece of text and markdown turns it into one block.

dessalines:
A newline on commonmark arkdown is just two lines. Press enter twice. That doesn't need a dedicated button unless your're using a keyboard without an enter key.

Damaskox:
> A newline on commonmark arkdown is just two lines. Press enter twice. That doesn't need a dedicated button unless your're using a keyboard without an enter key.

Yeah I did some more research after my comment and found out there are two ways to get the new line: the \ after a line and the two spaces.

Someone claimed that \doesn't work in every application though.

Damaskox:
> Nope. That leaves a blank line between the lines.
> 
> You see a lot of lemmy posts that are single lines with gaps between them because of this.

Care to share an example?

PR #1892, opened 2023-07-09
Add RES-style keyboard navigation

### LINKED ISSUE #984: RES style keyboard shortcuts
Not having the res keyboard shortcuts is making this website unusable for me, honestly.

I use j and k (to be clear, I use dvorak so that's actually v and c, which is extremely easy to press with one hand.) to go up and down and it automatically selects and expands things in res, that feature alone makes it so that i only have to push one button to navigate the entire website.

I also have it set for shift J to collapse and uncollapse the expando, and when it scrolls down to comments, i can also collapse and shrink comments, this is also extremely convenient.

Furthermore, i can push the A key to instantly open the selected posts comments, and enter for the selected posts link

I find this setup extremely convenient, scrolling and using a mouse is vastly more annoying.

Maybe this would better be relegated to an extension, but, I find this setup far too wonderful to think it isn't deserving of being native, and I think the shortcuts should be configurable per-session with the option to make it account-wide.

HamzahMansour:
## Description

Fixes #984 

~Depends on https://github.com/LemmyNet/lemmy-js-client/pull/164 and https://github.com/LemmyNet/lemmy/pull/3569~

Adds [RES](https://redditenhancementsuite.com/)-style keyboard navigation shortcuts to the homepage, community pages, and posts.

Shortcuts added:

| Shortcut | Action | Shortcut | Action |
|--------|--------|--------|--------|
| `j` | Next post/comment | `a` | Upvote post/comment (toggle) |
| `k` | Previous post/comment | `z` | Downvote post/comment (toggle) |
| `J` | Last post | `J` | Next sibling comment |
| `K` | First post | `K` | Previous sibling comment |
| `n` | Next page | `p` | Parent comment |
| `p` | Previous page | `t` | Top parent comment |
| `s` | Save post/comment (toggle) | `.` | Show advanced comment actions (toggle) |
| `r` | Reply to comment | `e` | Edit comment |
| `c` | Go to post comments in new tab | `C` | Go to post |
| `l` | Open link in new tab | `L` | Open link | 
| `u` | Go to post creator's page in new tab | `U` | Go to post creator's page |
| `g` | Go to community page in new tab | `G` | Go to community page |
| `x` | Expand post (toggle) | | |

**Known issues**:

- Collapsed comments don't stay collapsed when selecting/highlighting other comments (can only collapse one comment at a time currently, still trying to figure out why)
- When on a post with no comments, nothing is highlighted (can't interact with the post listing)
- On the homepage, comments can't be keyboard-navigated ([link](https://github.com/LemmyNet/lemmy-ui/blob/88f89cb9530cfaea0b82c8a86248d7d40a9f081d/src/shared/components/home/home.tsx#L717))
- On a person's profile, posts/comments can't be keyboard-navigated ([link](https://github.com/HamzahMansour/lemmy-ui/blob/2911cd8cdb348629a97d60239723b1925210df0a/src/shared/components/person/person-details.tsx#L238-L243))
- Highlight background colour (`.bg-body-tertiary`) not applied for logged out users in dark mode (`prefers-color-scheme: dark`). Seeing the light theme value (`--bs-tertiary-bg-rgb: 248, 249, 250;`) instead of the dark theme value (`--bs-tertiary-bg-rgb: 41, 41, 41;`). Light theme works fine
Screenshot 2023-07-27 at 13-07-54 lemmy-dev

**To-do**:

- [x] ~add account-wide preference (off by default)~
- [x] fix highlight styling + check against different themes
- [x] i18n strings
- [x] support comment navigation
- [x] add more navigation shortcuts (open to more suggestions, [RES reference](https://old.reddit.com/#res:settings/keyboardNav))
- [x] differentiation between new/distinguished and highlighted comments styles

### Before
Screenshot 2023-07-09 at 15-12-41 hmzhtest - local testing instance

### After
image

### Highlighted comment
Screenshot 2023-07-26 at 16-16-07 test img - lemmy-dev

### Highlighted new/distinguished comment
Screenshot 2023-07-26 at 16-16-18 test img - lemmy-dev

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1460,7 +1460,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   }
 
   handleKeybinds(event: KeyboardEvent) {
-    const post_view = this.postView;
+    const {
+      creator,
+      post: { id, url },
+    } = this.postView;
     if (
       this.props.enableKeyboardNav &&
       !event.ctrlKey &&
@@ -1479,22 +1482,28 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
           break;
         }
         case "c": {
-          window.open(`/post/${post_view.post.id}?scrollToComments=true`);
+          window.open(`/post/${id}?scrollToComments=true`);
           break;
         }
         case "C": {
-          this.context.router.history.push(
-            `/post/${post_view.post.id}?scrollToComments=true`
-          );
+          this.context.router.history.push(`/post/${id}?scrollToComments=true`);
+          break;
+        }
+        case "l": {
+          url && window.open(url);
+          break;
+        }
+        case "L": {
+          url && this.context.router.history.push(url);
           break;
         }
         case "u": {
-          const [, link] = getPersonDetails(post_view.creator);
+          const [, link] = getPersonDetails(creator);
           window.open(link);
           break;
         }
         case "U": {
-          const [, link] = getPersonDetails(post_view.creator);
+          const [, link] = getPersonDetails(creator);
           this.context.router.history.push(link);
           break;
         }
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1373,8 +1373,8 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         })}
         onKeyPress={e => this.handleKeybinds(e)}
         onClick={() => this.props.handleHighlight?.(this.props.idx)}
-        role="menuitem"
-        tabIndex={0}
+        role="row"
+        tabIndex={-1}
       >
         {/* The mobile view*/}
         <div className="d-block d-sm-none">
@@ -1395,11 +1395,6 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         {/* The larger view*/}
         <div className="d-none d-sm-block">
           <article className="row post-container">
-            {this.props.idx != undefined && (
-              <div className="col flex-grow-0 d-inline-flex align-items-center">
-                <div className="row px-1">{this.props.idx + 1}</div>
-              </div>
-            )}
             {!this.props.viewOnly && (
               <div className="col flex-grow-0">
                 <VoteButtons
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1367,11 +1367,11 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     return (
       <div
         id={`post-listing-${this.props.idx}}`}
-        className={
+        className={`my-1 p-2 row post-container ${
           this.props.enableKeyboardNav && this.props.isHighlighted
-            ? "row post-container bg-light"
-            : "row post-container"
-        }
+            ? "rounded bg-body-tertiary"
+            : ""
+        }`}
         onKeyDown={e => this.handleKeybinds(e)}
         onClick={() => this.props.handleHighlight?.(this.props.idx)}
         role="menuitem"
@@ -1469,6 +1469,8 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     ) {
       switch (event.key) {
         case "x": {
+          this.img &&
+            this.setState({ imageExpanded: !this.state.imageExpanded });
           this.handleShowBody(this);
           break;
         }
@@ -1477,26 +1479,31 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
           break;
         }
         case "c": {
+          window.open(`/post/${post_view.post.id}?scrollToComments=true`);
+          break;
+        }
+        case "C": {
           this.context.router.history.push(
             `/post/${post_view.post.id}?scrollToComments=true`
           );
           break;
         }
-        case "C": {
-          window.open(`/post/${post_view.post.id}?scrollToComments=true`);
-          break;
-        }
         case "u": {
-          const [, link] = getPersonDetails(post_view.creator);
-          this.context.router.history.push(link);
-          break;
-        }
-        case "U": {
           const [, link] = getPersonDetails(post_view.creator);
           window.open(link);
           break;
         }
+        case "U": {
+          const [, link] = getPersonDetails(post_view.creator);
+          this.context.router.history.push(link);
+          break;
+        }
         case "g": {
+          const [, , link] = getCommunityDetails(this.postView.community);
+          window.open(link);
+          break;
+        }
+        case "G": {
           const [, , link] = getCommunityDetails(this.postView.community);
           this.context.router.history.push(link);
           break;
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -1355,7 +1355,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
       <button
         type="button"
         className="btn btn-sm btn-link link-dark link-opacity-75 link-opacity-100-hover py-0 align-baseline"
-        onClick={linkEvent(this, this.handleShowBody)}
+        onClick={linkEvent(this, this.handleBodyExpandClick)}
       >
         <Icon
           icon={!this.state.showBody ? "plus-square" : "minus-square"}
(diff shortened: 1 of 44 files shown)


Kommynct:
In a new tab should be the without shift option, else you run the risk of accidentally replacing your current tab.

SkaveRat:
RES uses Shift-C ("C") to open a post in a new tab, and just "c" to open it in the current tab.

If the idea is to keep the layout, that should be kept

vmavromatis:
@HamzahMansour I wrote a [userscript](https://github.com/vmavromatis/Lemmy-keyboard-navigation) for pretty much the same purpose, feel free to take a look. I am just using arrow keys instead because I LOVE this setup and that's how I had it in RES. Hopefully this feature will make keys customisable. Thanks again and keep it up!

One last thing: Jump to Next/Previous same-level comment is super useful and RES had it so please make sure this get included 🙏

Kommynct:
> RES uses Shift-C ("C") to open a post in a new tab, and just "c" to open it in the current tab.
> 
> If the idea is to keep the layout, that should be kept

Res is an external extension that expects you to be a power user, somebody might accidentally do a keystroke and replace their page, res also allows you to change this to whatever you want, I think we should go for a safe default rather than sticking to what RES does exactly.

HamzahMansour:
> I think we should go for a safe default rather than sticking to what RES does exactly.

Yeah, I agree with having the safer options as defaults, and carrying over most other shortcuts from RES unchanged. At the very least I'll be adding a toggle for what `shift` does, current vs. new tab, for all actions but the end goal is to have all shortcuts be configurable (would need a separate settings tab probably).

src/shared/utils/app/enable-keyboard-navigation.ts
```diff
@@ -0,0 +1,9 @@
+import { UserService } from "../../services";
+
+export default function enableKeyboardNavigation(
```
alectrocute, line 3: A name like `isKeyboardNavigationEnabled` would be more apt.

src/shared/utils/helpers/get-community-details.ts
```diff
@@ -0,0 +1,19 @@
...
+    title = community.title;
+    link = `/c/${community.name}`;
+  } else {
+    const domain = hostname(community.actor_id);
+    name_ = `${community.name}@${domain}`;
+    title = `${community.title}@${domain}`;
+    link = !this.props.realLink ? `/c/${name_}` : community.actor_id;
+  }
+
+  return [name_, title, link];
```
alectrocute, line 18: This should have a return type.

src/shared/utils/helpers/get-person-details.ts
```diff
@@ -0,0 +1,20 @@
...
+    link = `/u/${person.name}`;
+  } else {
+    const domain = hostname(person.actor_id);
+    apubName = `@${person.name}@${domain}`;
+    link = !this.props.realLink
+      ? `/u/${person.name}@${domain}`
+      : person.actor_id;
+  }
+
+  return [apubName, link] as const;
```
alectrocute, line 19: A return type for this would be nice too.

dessalines:
I tested these, and I couldn't get any of them to work from my android tablet + keyboard. I'm guessing its a highlight related issue.

dessalines:
Tested, and it mostly works. A few things:

- Remove those left-aligned post numbers: Screenshot_2023-07-19-11-30-19-265_mark via gp
- The `r` reply to comment also types an `r` character.
- Comment saving could also be implemented in the same way you've done post saving.
- The `c` going to post comments should do a navigation, not a fresh page load. You've done this correctly with loading a creators page, but for some reason the comments seems to do a fresh page load.

HamzahMansour:
> Tested, and it mostly works. A few things:

* the `r` reply issue is only happening in chrome/safari, so I guess `onKeyPress` is only fully supported in firefox. switching to `onKeyUp` works, but also means that holding down a nav key (j or k) won't scroll quickly through comments on key repeat. not a big deal though.
* not sure how to check for or solve the fresh page load issue though, but I'm doing the same thing for both: `this.context.router.history.push(url)`

dessalines (approved):
Seems good to me, but I'd like some other UI devs here to test before merging.

src/shared/components/post/post-listings.tsx
```diff
@@ -113,6 +139,49 @@ export class PostListings extends Component<PostListingsProps, any> {
     );
   }
 
+  handleHighlight(postIndex: number | undefined) {
+    postIndex != undefined && this.setState({ highlightedPost: postIndex });
+  }
+
+  toggleExpand() {
+    this.setState({ isExpanded: !this.state.isExpanded });
```
SleeplessOne1917, line 147: State updates that depend on previous state should use functions to update.

```suggestion
    this.setState(prev => ({
        ...prev,
        isExpanded: !prev.isExpanded
    }));
```

src/shared/components/post/post-listings.tsx
```diff
@@ -113,6 +139,49 @@ export class PostListings extends Component<PostListingsProps, any> {
...
+  }
+
+  handleKeybinds(i: PostListings, event: KeyboardEvent) {
+    const idx = i.state.highlightedPost;
+    if (
+      !i.props.viewOnly &&
+      !event.ctrlKey &&
+      !event.metaKey &&
+      !event.altKey
```
SleeplessOne1917, line 156: If you want to avoid needing a not in front of everything, you could apply [DeMorgan's law](https://betterprogramming.pub/de-morgans-laws-in-software-engineering-b08feb897f75).

src/shared/components/post/post-listing.tsx
```diff
@@ -1410,6 +1434,115 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
+  }
+
+  handleKeybinds(i: PostListing, event: KeyboardEvent) {
+    const { creator, community, my_vote, post } = i.postView;
+    const { id, url } = post;
+    if (
+      !i.props.viewOnly &&
+      !event.ctrlKey &&
+      !event.metaKey &&
+      !event.altKey
```
SleeplessOne1917, line 1473: DeMorgan here as well.

src/shared/components/post/post.tsx
```diff
@@ -279,13 +283,33 @@ export class Post extends Component<any, PostState> {
...
+    this.setState({
+      nodes: document.querySelectorAll(".post-listing, .comment-node"),
+    });
+
+    if (this.state.nodes && this.state.nodes?.length > 0)
+      this.setState({
+        highlightedNode: this.checkScrollIntoCommentsParam
+          ? this.state.nodes[1]
+          : this.state.nodes[0],
+      });
```
SleeplessOne1917, line 296:
```suggestion
      this.setState(prev => ({
        highlightedNode: prev.nodes[this.checkScrollIntoCommentsParam ? 1 : 0],
      }));
```

src/shared/utils/helpers/get-community-details.ts
```diff
@@ -0,0 +1,21 @@
...
```
SleeplessOne1917, line 20: Nitpick. Could be worth annotating the function return type as `[string, string, string]` since the return is meant to be a tuple instead of a list of strings that could vary in length.

src/shared/utils/helpers/get-person-details.ts
```diff
@@ -0,0 +1,20 @@
+import { Person } from "lemmy-js-client";
+import hostname from "./hostname";
+
+export default function getPersonDetails(person: Person): [string, string] {
```
SleeplessOne1917, line 4: Same as [here](https://github.com/LemmyNet/lemmy-ui/pull/1892/files#r1275484567).

src/shared/utils/helpers/get-community-details.ts
HamzahMansour: do you mean a change in the function return type `getCommunityDetails(...): [string, string, string] {`

or this return value being cast here, something like

`return [name_, title, link] as [string, string, string];` or

`return [name_, title, link] as const` (I think both work)

SleeplessOne1917: > do you mean a change in the function return type getCommunityDetails(...): [string, string, string] {

That one.

Kommynct:
This seems to have stalled along with #1367 awaiting review from jsit, and it seems he hasn't been active in lemmy for a while (i checked recent MR's and didn't see any he's approved, although I only went around 10 down), and he hasn't done anything on his profile related to lemmy since july, is it possible for someone else to review this code?

Sorry, I don't mean to be rude, this just seems completely ready and I'm excited for it.

> > Tested, and it mostly works. A few things:
> 
>     * the `r` reply issue is only happening in chrome/safari, so I guess `onKeyPress` is only fully supported in firefox. switching to `onKeyUp` works, but also means that holding down a nav key (j or k) won't scroll quickly through comments on key repeat. not a big deal though.

Also, couldn't you just make it onkeyup for replies/anything that opens a text box?

dessalines:
@HamzahMansour needs to update this, and solve the merge conflicts still.

PR #1894, opened 2023-07-09
Use the 'node' user instead of root while running in a container

camporter:
## Description
Rather than running lemmy-ui as the root user in the container, run it as the `node` user provided by the node runner image.

## Screenshots
N/A

### Before
The `node` process runs as `root`, which should be avoided if possible.

### After
The `node` lemmy-ui process runs as the `node` user.


Nutomic (approved):

SleeplessOne1917 (approved):

dessalines (approved):

makotech222:
This one killed my 'docker compose up' run on our server, just fyi. dunno why, just removed it.

PR #1906, opened 2023-07-10
Sanitize html in emojis with potentially malicious alt text.

### LINKED ISSUE #1895: Possible XSS attack
### Requirements

### Summary

~~The sidebar [dangerously sets HTML](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/community/sidebar.tsx#L306C36-L306C36) but does not configure [the Markdown render](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/markdown.ts#L22) to strip HTML codes. This enables simple XSS attacks like ``. It seems like an attempt is made to [create a markdown renderer with HTML disabled](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/markdown.ts#L157), however.~~

It now seems that this attack is done [via custom emojis](https://github.com/LemmyNet/lemmy-ui/pull/1897/files).

### Technical Details

`markdown-it` does some extremely simple guarding, but they don't claim to prevent XSS. Custom HTML should be removed in favor of plugins.

### Lemmy Instance Version

0.18.1

### Lemmy Instance URL

_No response_

FireMasterK:
## Description

Fixes #1895

### DIFF
--- a/src/shared/markdown.ts
+++ b/src/shared/markdown.ts
@@ -15,6 +15,7 @@ import markdown_it_sub from "markdown-it-sub";
 import markdown_it_sup from "markdown-it-sup";
 import Renderer from "markdown-it/lib/renderer";
 import Token from "markdown-it/lib/token";
+import sanitizeHtml from "sanitize-html";
 import { instanceLinkRegex } from "./config";
 
 export let Tribute: any;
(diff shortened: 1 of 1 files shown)


dessalines:
Already been fixed by https://github.com/LemmyNet/lemmy-ui/pull/1897

FireMasterK:
> Already been fixed by #1897

But isn't the fix ineffective in cases where the custom emoji itself has a malicious alt text?

db0:
The custom emoji alt text can only be modified by the admin. And if the admin is malicious, then they can add a ton of other exploits just as well.

tsweet64:
Alt text on a custom emoji should never be able to execute JavaScript. Period. I think it's still worth fixing any known instance of this vulnerability.

Daniel15:
The actual fix is to stop building HTML using string concatenation. This is an outdated and dangerous practice. Also, sanitization like this should really be done on the server-side rather than the client-side.

phoenixeliot:
Very much agreed with @Daniel15. Using string concatenation is always begging for trouble. Use the standard JS API for creating elements and setting their attributes instead.

PR #1907, opened 2023-07-10
Add nonce-based CSP header

sunaurus:
## Description

As discussed in #1900, let's keep data serialized as before, but use a nonce to protect against malicious scripts.

Also removes unused websocket config related script tag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "analyze": "webpack --mode=none",
     "prebuild:dev": "yarn clean && node generate_translations.js",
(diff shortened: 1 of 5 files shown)


sunaurus:
It's working with both `build:prod` and `build:dev` for me (dev works by disabling CSP headers)

dessalines (approved):
Thanks!

I just deployed this to https://ds9.lemmy.ml , and everything seems to be working correctly.

SleeplessOne1917 (approved):

PR #1911, opened 2023-07-11
Remove lodash.merge dependency

tiziodcaio:
## Description

Remove lodash.merge dependency from webpack.config.js file, replacing it with using [spread sintax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)


SleeplessOne1917:
Have you tested this? Sometimes merge is needed because the spread syntax is only shallow merge but you need a deep merge.

SleeplessOne1917 (approved):

PR #1929, opened 2023-07-13
Set person_id to myId in handleLeaveModTeam

lionirdeadman:
## Description

This fixes leaving the mod team. It previously always removed the user with person_id 92.

NOTE: This has been tested :heavy_check_mark: :


SleeplessOne1917 (approved):
LGTM.

PR #1931, opened 2023-07-13
Fix start_url and scope in web app manifest

randfur:
## Description

Lemmy's manifest is currently broken.
https://lemmy.world/manifest.webmanifest serves `"start_url":"http://lemmy.world","scope":"http://lemmy.world"` which doesn't match the scheme and is rejected by the browser.
This change makes it use "/" to keep things simple.


SleeplessOne1917 (approved):

randfur:
That's weird, the check doesn't seem to see my second commit.

randfur:
I think I need someone to press the merge button.

PR #1934, opened 2023-07-13
Allow sorting on /communities

googleben:
## Description

Fixes [#3500 mistakenly filed on the main repo](https://github.com/LemmyNet/lemmy-ui/issues/1941) by simply adding a sort dropdown to /communities using the existing `SortSelect` element. The right-justification of the search bar is preserved and the default sort is still `TopMonth`. The one issue is that the question mark should link to an explanation of community sorting, but the linked page on join-lemmy ([03 Votes and Ranking](https://join-lemmy.org/docs/users/03-votes-and-ranking.html)) only talks about post and comment sorting.

The functionality for sorting communities is already present in the API, so this feels like a no-brainer once join-lemmy is updated, which I'm happy to do as well if it needs doing.

Also, just a minor note, this also fixes a small inaccuracy where the bottom margin for the selector row (`.mb-2`) was applied to the search bar instead of the entire row, which only acted correctly because the containing row was set to justify. The screenshots were taken before this fix so there's a few things off by a few pixels, but nothing really noticeable.

## Screenshots

Examples using lemmy.world (mobile screenshots taken using emulated 335x667 display on Firefox):

### Before
image
image

### After
image
image
image
image

### DIFF
--- a/src/shared/components/community/communities.tsx
+++ b/src/shared/components/community/communities.tsx
@@ -20,6 +20,7 @@ import {
   ListCommunities,
   ListCommunitiesResponse,
   ListingType,
+  SortType,
 } from "lemmy-js-client";
 import { InitialFetchRequest } from "../../interfaces";
 import { FirstLoadService, I18NextService } from "../../services";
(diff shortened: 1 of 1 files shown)


dessalines (approved):
Looks good, thanks!

SleeplessOne1917 (approved):

PR #1963, opened 2023-07-20
Add RTL/BIDI to post text

### LINKED ISSUE #347: Add bidi / RTL support

Considering the very nature of social media platforms (like Lemmy), people with different languages would create content. It becomes problematic when one creates content in RTL languages (like Persian, Arabic, Urdu,...). Lemmy is not ready to render RTL (or even mixed RTL/LTR) text properly.

There are two approaches in general. One is through  HTML modification which can be done by add `dir="auto"` attribute to elements containing user generated text or by wrapping it in `<bdi>` tags. The other approach would be through CSS by adding `unicode-bidi: plaintext`. In either of approaches, there should be avoidance in using absolute positioning or spacing. For example instead of `text-align: left` we should use `text-align: start` or instead of `padding-left` we should use `padding-inline-start` unless there be some strong UX reason.

--

### Additional context

Just keep in mind that bidi (bidirectional text) is different from using RTL or LTR in general based on the language user choose for the interface. 

I have opened this issue here not on back-end because I think this can be managed on front-end.

hamidrezakp:
## Description
This PR will add Bi Directional text and RTL support to posts text.
However it will not fix comments text direction because the plugin I'm using `markdown-it-bidi` by @ahangarha doesn't support lists yet.

Partially fixes #347

### Before
before

### After
after

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "markdown-it-ruby": "^0.1.1",
     "markdown-it-sub": "^1.0.0",
     "markdown-it-sup": "^1.0.0",
+    "markdown-it-bidi": "^0.1.0",
     "mini-css-extract-plugin": "^2.7.5",
     "register-service-worker": "^1.7.2",
     "run-node-webpack-plugin": "^1.3.0",
--- a/src/shared/markdown.ts
+++ b/src/shared/markdown.ts
@@ -8,6 +8,7 @@ import { CustomEmojiView } from "lemmy-js-client";
 import { default as MarkdownIt } from "markdown-it";
 import markdown_it_container from "markdown-it-container";
 // import markdown_it_emoji from "markdown-it-emoji/bare";
+import markdown_it_bidi from "markdown-it-bidi";
 import markdown_it_footnote from "markdown-it-footnote";
 import markdown_it_html5_embed from "markdown-it-html5-embed";
 import markdown_it_ruby from "markdown-it-ruby";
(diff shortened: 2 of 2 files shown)


ahangarha:
Make it WIP PR. Next week we can work together to fix the rest. For example, the comment form needs `dir="auto"` to the `textarea`.

We should fix the comments issue as well. It shouldn't be a big issue.

hamidrezakp:
> A great step forward 👍🏾
> 
> Please make it WIP PR. Next week we can work together to fix the rest. For example, the comment form needs `dir="auto"` to the `textarea`.
> 
> We should fix the comments issue as well. It shouldn't be a big issue.

Thanks, sure, can't wait to have this feature merged.

SleeplessOne1917 (dismissed):

ahangarha:
Let's merge this PR as a partial fix. With the feedback we receive, we can move forward with more confidence.

SleeplessOne1917 (approved):
LGTM

dessalines (approved):

ahangarha:
Thanks @hamidrezakp for taking the first step for adding bidi support.
I hope to see this on the public instances soon.

hamidrezakp:
> Thanks @hamidrezakp for taking the first step for adding bidi support. I hope to see this on the public instances soon.

Thanks, yes i would love to see Lemmy  fully bidirectional and available for all users.
And Thank you for writing `markdown-it-bidi` plugin. looking forward to see it support lists.

Issue #1974, opened 2023-07-23
Federated user home instance should still be visible even with a display name set


Gyrfalcon5:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

When moderating a community, it is sometimes helpful to get a sense of whether a user you are interacting with is a member of your own instance, or if they are someone interacting with your community through federation. This allows moderators to make more effective corrections to users, as well as determine when an issue may have gone beyond gentle correction and requires a temporary or permanent ban. 

Currently, this is possible for many users, but if a user sets a display name, then their home instance is hidden. 

It is still possible to view their instance by mousing over their name or clicking on their profile, but this makes moderation somewhat more difficult, especially on mobile.

Ideally it should be possible, whether by default or as an option, to have all users' home instances displayed by their username, even if they have a display name set.

Nutomic:
Home instance is always visible for remote users now.

PR #1975, opened 2023-07-24
Use showApubName in comment and post listings 

lionirdeadman:
## Description

This helps distinguish people from other instances which use a Display name and can help prevent impersonations which we've seen in the past.

Fixes https://github.com/LemmyNet/lemmy-ui/issues/1974

### Before
image

### After
image

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -306,7 +306,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 />
               </button>
 
-              <PersonListing person={cv.creator} />
+              <PersonListing person={cv.creator} showApubName />
 
               {cv.comment.distinguished && (
                 <Icon icon="shield" inline classes="text-danger ms-1" />
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -401,7 +401,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
 
     return (
       <div className="small mb-1 mb-md-0">
-        <PersonListing person={post_view.creator} />
+        <PersonListing person={post_view.creator} showApubName />
         <UserBadges
           classNames="ms-1"
           isMod={this.creatorIsMod_}


SleeplessOne1917 (approved):

dessalines (commented):
Very much against this. Bloating the username lengths massively everywhere. The lack of a `@` tells you that its a display name, and that if you want to see the actual identifier, you can either click and hold, or go to their profile page.

lionirdeadman:
Then there is no way to differentiate from a local user and someone who is outside of the instance.

Now, if you want to propose Displayname@instance then it causes another problem because it can be made to look like someone else's user@instance.

I will note that other federated services like Mastodon, Firefish do show both:

image
image

creesch:
> The lack of a `@` tells you that its a display name. 

This is a really easy to overlook thing that a lot of people will not easily pick up on. Even if that is the case it no longer makes it clear if a user is from your own instance or from a different instance.

SleeplessOne1917:
I'm with @lionirdeadman and @creesch on this one. The "@" and instance name will not show up when a display name is set regardless of whether a user is on the same instance as you or a different one.

dessalines:
Having really long names is terrible for mobile. Having display names like this: , or even just `A pretty long display name@instance.TLD` will cause a ton of line-wraps and noise on mobile.

> It no longer makes it clear if a user is from your own instance or from a different instance.

You can easily click-and-hold. I followed matrix, which is to stay mobile-friendly, and have the display name being an actual display name. If people want to know your specific instance, and full identifier, they can click and hold, or go to your profile page.

I tested fully qualified names on mobile before both in lemmy-ui, and jerboa, and they cause a ton of incredibly ugly line wraps.

lionirdeadman:
Comparing a chat app and a forum app is not really a fair comparison because for a chat app, the members in the room are a lot less likely to use the same display name.

That said, if two people have the same display name on matrix - here's how that looks :
image

Unless we can somehow know if two people have the same display name and act accordingly, it seems best to simply do it.

SleeplessOne1917:
We could also do an approach similar to the Mastodon screenshot and display the apub info on a separate line under the display name in a less prominent color. That could make apub names easier to see while also avoiding overlong lines and line wraps.

ruffsl:
> We could also do an approach similar to the Mastodon screenshot and display the apub info on a separate line under the display name in a less prominent color.

This could give more room in the future for things like user flares as well, like when a user is flared as verified domain expert in a Q&A community, or marked as frequent contributer. If these multiple line could match the height of the avatar icons, then it might not be too bulky, while still leaving space for other info later.

makotech222:
Just noting that for hexbear, upon federation we almost immediate changed the frontend to display 'display_name@domain' to aid in moderation efforts and to make it clear if someone is outside our instance. It would be shorter than 'display_name (username@domain)' so that might be better for upstream.

dessalines:
I'll go back against what I said previously, and am willing to use the larger apub name. But it needs some testing and before / after pics, because its going to cause a ton of wrapping on mobile.

dessalines:
Closed in favor of #2064

PR #1977, opened 2023-07-24
Allow limited set of markdown in title inline md rendering, fixes #1976

### LINKED ISSUE #1232: [Bug]: You can post images into titles
### Summary

Might be opening pandora's box for a moment until it's fixed. Not a bug, but unintended side effect. As cute as it is, you can just put huge images into the titles of your posts that will show on the feed. I would think it would be hard to make exceptions in the markdown, it's possible for UI just to hide/remove/or shrink images in that particular field?

screenshot of an inline image in a post title

### Steps to Reproduce

1. Create new post in a community.
2. In your title add the markdown for an image `screenshot of an inline image in a post title` of any image (preferably large) into the title 
EDIT: This is not the original image I used, I have linked the wrong one, if that matters. `cain`
3. Maybe do this in a contained space that it wouldn't disrupt the feed while testing.

### Technical Details

No errors.

### Version

v0.17.3

### LINKED ISSUE #1976: Inline rendering of MD in titles should not be a thing or heavily limited.
### Requirements

Earlier today I noticed a post in my feed that had a green color as title, making it look like an admin/meta post. However the post was not created by an admin, in fact it was posted by someone from a different instance. 

What did happen is that the person making the post included a markdown link as the title. And, apparantly markdown in titles is fully rendered.  This is an explicit feature as can be seen here: https://github.com/LemmyNet/lemmy-ui/pull/860 so initially I was going to make a bug report.

However, I believe that the current implementation is far too lenient and should be toned down as it leaves the door wide open for abuse: 

- Links in titles directly go to wherever they link. This is different from how titles normally work as they usually go to comments. As they are also greenly colored they leave the door open for someone impersonating an admin post that leads to somewhere else entirely.  I can see some really potential for phising here. 
- Images are also fully rendered, leaving the door wide open for incredibly spammy posts and also other abuse vectors. For example someone can include an image that looks like interface elements. 

The screenshot below should give a good impression of why full markdown rendering in titles is a bad idea. 

image

At the very least, to prevent abuse: 

- Images should not be rendered in titles. 
- Markdown links should not render in titles.

It probably is a good idea to limit markdown rendering to just italic, bold and strike through. Anything else does make very little sense. 

Additionally: 

- Instance admins probably should have the option to disable title markdown entirely.

creesch:
## Description
Fixes #1976 by only allowing a very limited set of markdown rules for inline markdown rendering. 
Also fixes #1232

### Before
image

### After
image

### DIFF
--- a/src/shared/markdown.ts
+++ b/src/shared/markdown.ts
@@ -23,6 +23,14 @@ export let md: MarkdownIt = new MarkdownIt();
 
 export let mdNoImages: MarkdownIt = new MarkdownIt();
 
+// Zero disables all rules.
+// Only explicitly allow a limited set of rules safe for use in post titles.
+export const mdLimited: MarkdownIt = new MarkdownIt("zero").enable([
+  "emphasis",
+  "backticks",
+  "strikethrough",
+]);
+
 export const customEmojis: EmojiMartCategory[] = [];
 
 export let customEmojisLookup: Map<string, CustomEmojiView> = new Map<
@@ -43,7 +51,7 @@ export function mdToHtmlNoImages(text: string) {
 }
 
 export function mdToHtmlInline(text: string) {
-  return { __html: md.renderInline(text) };
+  return { __html: mdLimited.renderInline(text) };
 }
 
 const spoilerConfig = {


dessalines (approved):
Looks good, thx.

dessalines (approved):

PR #1135, opened 2023-06-09
Feature/mark post as read button

ItsEcholot:
Fixes #384

Issue #384, opened 2021-08-31
A post action button to hide posts


Nirose:
Lemmy does have "Show/Hide read posts" but there are posts that you don't want to engage with at all. So, could it be possible to add a button to hide them from the UI so they don't appear on return visit? 

This will help user keep their front-page clean and might allow users to view more posts that interest them. As the clutter is be hidden.

May be a "Ignore" or "Hide" icon as a post action.

dessalines:
We don't have the concept of post hiding... but would a button that has `Mark as read` work?

Nirose:
I think that would work, is there a section I can go to view the already read posts? I couldn't find it.

dessalines:
Hrm, no we don't have a read / unread filter for posts.

dessalines:
We'd need to add a `MarkPostAsRead`, similar to `MarkCommentAsRead` in the API.

krestenlaust:
Isn't 'mark post as read' a (wildly) different feature than 'hide post'?

To me, 'Mark post as read' implies that every post you click into, is going to disappear once you click out of it again.
While a 'hide post'-feature sounds more like simply useful for hiding pinned posts

joelthelion:
To elaborate on what @krestenlaust said, I personally don't like hiding read posts. I enjoy going back to posts I have already read.
I do, however, like to occasionally hide posts I have no interest in. This is why I'd like a "hide" feature. Definitely not the same as "marking as read" from my perspective.

ericjmorey:
I would prefer to have the hide feature have a time element to it. You choose the length of time you want to hide the post from a menu. 

15 minutes
1 hour
6 hours
1 day
3 days
1 week
1 month
Infinite 

Maybe even have a custom option.

A similar concept could be applied to blocking accounts communities and instances.

krestenlaust:
What would the use of hiding a post with a timer be. This site is designed around the concept of posts decaying naturally. A post that is hidden in more than a day or so, would be gone anyway.

ericjmorey:
> A post that is hidden in more than a day or so, would be gone anyway.

If you're only using the front page with default settings, perhaps. But there are smaller instances and communities where this would come in handy, especially in posts that stay active in the comments for months.

TailyFair:
From what I've seen on various posts and comments, this is a real pain point for a lot of users.

ColorfanXP:
Personally, I like the action of voting on a post causes it to become hidden.

[...]

PR #1978, opened 2023-07-24
Mark post as read when clicking "Expand here" on the preview image on the post listing page (#1600)

### LINKED ISSUE #1600: Profile option to mark post as read on preview
### Requirements

Not sure how possible this is, but it'd be useful for an option to mark posts as read when previewing content from the feed screen.

This would be especially useful for image-only posts so, if I click the image thumbnail to expand it, the post is marked as read, rather than having to go into an empty post to achieve that.

milandamen:
## Description

Fixes #1600 
When the image preview is clicked, and the post has not been "read" before, a request is made to the lemmy server API to mark the post as read. If the user is not logged in, no attempt is made to mark the post as read. This uses functionality that has already been implemented in lemmy-js-client and lemmy server.

The reason for this change is explained in #1600, but basically it would be logical for a post to have been "read" if you have previewed the image. This is useful in combination with turning off "Show Read Posts" in the user profile settings, so that you only see posts in your feed that you have not viewed before. This already works well with non-image posts, but for image posts you don't want to open the whole post to check out the image.

I have not worked with modern javascript development before, so any feedback is appreciated. I looked at #1135 / #384 for inspiration for the code.

## Screenshots

Graphics did not change.

## Before

When the image preview is clicked, the post was not marked as "read". It was only marked as "read" if you opened the post, or liked the post.

## After

When the image preview is clicked, the post is marked as read. The other ways of marking as read remain unchanged.


src/shared/components/post/post-listing.tsx
```diff
@@ -1723,6 +1725,20 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     event.preventDefault();
     i.setState({ imageExpanded: !i.state.imageExpanded });
     setupTippy();
+
+    const auth = myAuth();
+    if (
+      auth &&
+      (typeof i.props.post_view.read === "undefined" ||
+        i.props.post_view.read === false)
```
SleeplessOne1917, line 1733: Is the check on the type necessary? If the type of `post_view.read` is `boolean | undefined`, `!i.props.post_view.read` should suffice.

milandamen: Thank you for the feedback. I have fixed it.

src/shared/components/post/post-listing.tsx
```diff
@@ -1723,6 +1725,16 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
     setupTippy();
+
+    const auth = myAuth();
+    if (auth && !i.props.post_view.read) {
+      const form: MarkPostAsRead = {
+        post_id: i.props.post_view.post.id,
+        read: true,
+        auth: auth,
+      };
+      HttpService.client.markPostAsRead(form);
```
dessalines, line 1736: Don't do HTTPService calls here. Follow the convention with the rest of the actions, and add `onMarkPostAsRead` to the props.

Check out `onSavePost` for how that works.

milandamen: As a professional software engineer it smells weird to me when I need to duplicate so much code to add a new feature (+56 lines changed), while the original had no duplicate code (+13 lines changed) and didn't need the reader to know 10 different files.

If different places in the code want to do something when the post is marked as read, they should add an event listener on the "read" state of the post (or a more general state change), in my opinion.

I am used to programming using the Qooxdoo framework, where you have a domain object (something like Post in this case), which would have properties like "read: Boolean". Whenever you click to expand the image, you would do the HTTP request, and if that succeeded, you do a `setRead(true)`, and Qooxdoo would fire all event listeners that are listening to that property of that object. So if for example the Post is shown on the home page, the home page code would `post.addListener('changeRead', this.onChangeRead)` to register the listener, so that any changes to that property causes Home#onChangeRead() to be called.

Nonetheless, I have made a commit that does what you asked.

SleeplessOne1917 (approved):

dessalines (approved):
Thanks!

PR #1980, opened 2023-07-25
change dockerfile build to use a standard glibc image, LTS-Debian (and others)

Security-Chief-Odo:
Increasing performance and compatibility, changed entrypoint to dumb-init which allows the container to respect signals sent to kill/sighup/restart and moved the `chown` action to the copy action which reduces build time substantially. Use Docker buildkit when building `DOCKER_BUILDKIT=1 docker build`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-FROM node:20.2-alpine as builder
-RUN apk update && apk add curl yarn python3 build-base gcc wget git --no-cache
-RUN curl -sf https://gobinaries.com/tj/node-prune | sh
+FROM node:lts-bullseye-slim as builder
+RUN apt-get update && apt-get install -y build-essential ca-certificates apt-utils nodejs npm curl wget git yarn python3 dumb-init --no-install-recommends
+RUN curl -k -sf https://gobinaries.com/tj/node-prune | sh
 
 WORKDIR /usr/src/app
 
 ENV npm_config_target_arch=x64
 ENV npm_config_target_platform=linux
-ENV npm_config_target_libc=musl
+ENV npm_config_target_libc=glibc
+ENV NODE_ENV=production
 
 # Cache deps
 COPY package.json yarn.lock ./
@@ -24,7 +25,7 @@ COPY src src
 COPY .git .git
 
 # Set UI version 
-RUN echo "export const VERSION = '$(git describe --tag)';" > "src/shared/version.ts"
+RUN echo "export const VERSION = \"$(git describe --tag | awk -F'-' '{print $1"-"$2}')\";" > "src/shared/version.ts"
 
 RUN yarn --production --prefer-offline
 RUN NODE_OPTIONS="--max-old-space-size=8192" yarn build:prod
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,12 +1,13 @@
...
+RUN apt-get update && apt-get install -y build-essential ca-certificates apt-utils nodejs npm curl wget git yarn python3 dumb-init --no-install-recommends
+RUN curl -k -sf https://gobinaries.com/tj/node-prune | sh
 
 WORKDIR /usr/src/app
 
 ENV npm_config_target_arch=x64
 ENV npm_config_target_platform=linux
-ENV npm_config_target_libc=musl
+ENV npm_config_target_libc=glibc
+ENV NODE_ENV=production
```
SleeplessOne1917, line 10: `NODE_ENV` is set in the webpack config.

Security-Chief-Odo: Difference in where it is built, and where it matters.

Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM node:20.2-alpine as builder
-RUN apk update && apk add curl yarn python3 build-base gcc wget git --no-cache
-RUN curl -sf https://gobinaries.com/tj/node-prune | sh
+FROM node:lts-bullseye-slim as builder
```
dessalines, line 1: Why is this necessary? alpine works in docker just fine, and is going to create smaller images. If it works in docker, it works anywhere.

dessalines (changes requested):

Dockerfile
Security-Chief-Odo: Not sure the size matters when the node JS application crashes so often using alpine. This isn't a _needed_ change, but it's a performance and stability one. Especially with the `dumb-init` addition. I am experiencing less headaches and issues by using this custom docker image than using your alpine built one. I thought contributing back to the community a useful change would be beneficial to everyone. 


As for _creates a smaller image_,  I see the following sizes for the images on disk:

```
dessalines/lemmy-ui            latest          a008bbfd24e1       480MB
```

```
custom/lemmy-ui                0.18.3          a87d192120c9      452MB
```

This one is _smaller_, but again not by much not sure why 30MB matter when the image is already nearly half a GB.

dessalines: You should always try to move from debian/ubuntu to alpine if possible for images, musl has better performance than glibc.

Security-Chief-Odo: > musl has better performance than glibc.

No it does not; it makes a smaller environment but it does not have better performance. musl libc is built for correctness and size over performance. This is especially obvious with the implementation of malloc, and rust, leading to a few memory leak issues seen in this project. It also has known issues with regards to DNS stack and often causes `Unknown host` errors from the container.

dessalines: That's debatable, but I've never seen anyone move away from alpine and musl back to ubuntu / glibc. If there were the issues you're talking about, none of the production servers running lemmy-ui would work at all.

dessalines:
We have some very large lemmy instances, both the UI and back-end, all running on alpine and musl images. Moving away from those isn't something I wanna do for the production dockerfiles. You're of course free to have your own custom dockerfiles.

Dockerfile
sunaurus: I realize this is a closed issue, but if anybody else ends up here at some point, I just want to point out that glibc is a totally valid option for Lemmy - lemm.ee has been on glibc from the beginning.


I personally also have had a few very bad experiences with Alpine, strictly because of musl, so I avoid it whenever I can. But this is of course completely anecdotal.

pallebone:
Just fyi as I saw this. 

Unfortunately the use of alpine has not proved stable for some in live production use. This has meant that alternative debian images have been created. 

For example I use on my amd64 machine these images:

lemmy:
    image: ghcr.io/ubergeek77/lemmy:0.18.4
  
  and 

lemmy-ui:
    image: ghcr.io/ubergeek77/lemmy-ui:0.18.4
    
These are built on top of debian. I recommend them if anyone is having issues.

PR #1989, opened 2023-07-26
Fixing comment report showing dot.

dessalines:
## Description

The comment report showed a dot because it was incorrectly a `<li>`

### Before

Screenshot_2023-07-26-14-36-40-082_mark via gp

### After

Screenshot_2023-07-26-14-46-49-802_mark via gp


SleeplessOne1917:
I believe comment nodes were made lists/list items for accessibility purposes. You could probably preserve the list items while still fixing the issue by setting the the list decoration to none.

dessalines:
Ah good call, found a better way to do it.

SleeplessOne1917 (approved):

PR #1993, opened 2023-07-27
Add moderator view button

### LINKED ISSUE #1294: A moderated communities post feed
### Requirements

- [ ] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [ ] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.

No

On the main page where you can select “All, Local, Subscriptions” it would be awesome if there was an additional option for those who are moderators to be able to also have a “Moderated” option so we can scroll through all posts of communities we moderate

The only other option would be to only subscribe to communities we moderate but that’s pretty inconvenient 

### Additional context

_No response_

biosfood:
## Description

Fixes  #1294.

This PR will add a moderator view button to the home page and uses the newly added moderator view parameter from https://github.com/LemmyNet/lemmy/pull/3176 to give moderators easy access to all the communities they moderate.

The moderator view button should only be visible for logged in users that moderate at least one community.

### Before

TODO

### After

TODO


src/shared/components/home/home.tsx
```diff
@@ -179,12 +181,17 @@ function getSortTypeFromQuery(type?: string): SortType {
   return (type ? (type as SortType) : mySortType) ?? "Active";
 }
 
+function getModeratorViewFromQuery(moderatorView?: string): boolean {
+  return moderatorView?.toLowerCase() == "true";
```
SleeplessOne1917, line 185: Triple equals, not double.

SleeplessOne1917:
Should moderator view be it's own parameter instead of one of the listing types? I can see `Local` and `All` making sense with it, but what would it mean to have moderator view and `Subscribed` at the same time? I would think that a moderator of a community would always be following that community, so having both selected seems redundant.

biosfood:
> Should moderator view be it's own parameter instead of one of the listing types? I can see `Local` and `All` making sense with it, but what would it mean to have moderator view and `Subscribed` at the same time? I would think that a moderator of a community would always be following that community, so having both selected seems redundant.

You're right it would probably make more sense if moderator view was just another listing type (in appearance for the user). Listing type "all" would be applicable at all times but it will make for some ugly code in the listing type selector and home page. Besides, someone might actually want to differentiate between moderator view for local and all communities and unsubscribe from communities they consider themselves not to be very involved in moderating.

A compromise might be automatically seeing the listing type to all when moderator view of selected but I'm not sure that wouldn't cause frustrations for some users. Let me know which solution you prefer.

Another problem I just noticed now: the moderator view button should probably be hidden when only viewing comments, right?

SleeplessOne1917:
> Another problem I just noticed now: the moderator view button should probably be hidden when only viewing comments, right?

Not necessarily. Comments are content that will need moderation just as much as posts will. As to the rest of your response, I'd like to get some more input on this one.

cc @dessalines @Nutomic @alectrocute @jsit @lionirdeadman

dessalines:
Its up to you whether its a part of the listing_type button group, or separate. Its technically possible to be a mod of non-local communities, so maybe separate is better, but either way works for me.

> Another problem I just noticed now: the moderator view button should probably be hidden when only viewing comments, right?

Unfortunately we didn't add `moderator_view` to `comment_view.rs`, but @SleeplessOne1917 is correct, and it should be added, since comments also need a moderation view. We don't need to wait on it for this PR, but @biosfood could you do a PR to the lemmy back-end to add that?

src/shared/components/common/moderator-view-toggle.tsx
dessalines, line 1: Great job. For some reason people seem to try to code these things from scratch rather than using the already existing examples.

src/shared/components/home/home.tsx
```diff
@@ -771,6 +793,14 @@ export class Home extends Component<any, HomeState> {
             onChange={this.handleListingTypeChange}
           />
         </div>
+        <div className="col-auto">
+          {(UserService.Instance.myUserInfo?.moderates?.length ?? 0) > 0 && (
```
dessalines, line 797: This check could be moved above to the div.

biosfood:
I can certainly add moderator view support for comments to the Lemmy backend. I should be able to open a PR sometime next week but don't count on it for this one here.

Since the moderation view currently is implemented as an additional parameter in the get posts method, I'd say it should be exposed to the user as such with its own toggle switch as I implemented it right now.

lionirdeadman:
I think it should be a listing type because if I click that button, I would want it to be all the communities I moderate regardless of local vs all vs subscribed. It's ultimately a way to filter the feed, not a different type of option.

Comments should be a possible view with moderation actions possible of course.

> Since the moderation view currently is implemented as an additional parameter in the get posts method, I'd say it should be exposed to the user as such with its own toggle switch as I implemented it right now.

I actually think that the API should be using the same param as the other listing types.. making it its own thing just complicates the API needlessly as the response is the same as any other view. 

IMO the API in general is not a good indication of how the UI should be made and it is often too inflexible - often giving too much information or too little - to actually experiment with the UI.

biosfood:
As the consesus seems to be that moderator view should just be a listing type, I will open a PR for the lemmy backend and js client to implement those changes. Keeping this PR as draft until then.

biosfood:
With the backend changes merged, this is now what the selector looks like:

new ModeratorView select

The ModeratorView option is only available, if the user moderates at least one community to avoid confusion.

biosfood:
https://github.com/LemmyNet/lemmy-js-client/commit/973b255c829e78a0667f472db0ebd864c1ad7aa3 has added some more backend api changes that have built up over some time. The tests won't run clean before those issues are fixed (happens in yarn lint, mostly type problems).

Things that would need to be added:
- When deleting an account, ask the user weather all their content should be removed as well (src/shared/components/person/settings.tsx:1289:71). Pass false as a default for now? I could open another pull request for that or should I do this here?
- When adding a new Admin, a local user Id is required but the contexts where a button for adding an admin exist (comment view, person view) only provide a user id. What is the "right" way to get a local user id in this context?

src/shared/components/comment/comment-form.tsx
```diff
@@ -84,15 +84,14 @@ export class CommentForm extends Component<CommentFormProps, any> {
       : capitalizeFirstLetter(I18NextService.i18n.t("reply"));
   }
 
-  handleCommentSubmit(content: string, form_id: string, language_id?: number) {
+  handleCommentSubmit(content: string, _form_id: string, language_id?: number) {
```
SleeplessOne1917, line 87: Is there any reason this form ID argument wasn't removed entirely?

SleeplessOne1917 (approved):

src/shared/components/comment/comment-form.tsx
biosfood: MarkdownTextArea wants the function signature to contain 3 parameters in that order. language_id is needed to pass to the server.

dessalines: You can probably just remove that from markdowntextarea too, since I don't think its necessary anymore.

dessalines (commented):
Just that one comment above, then the linting errors, and we can get this merged. Thx!

biosfood:
@dessalines I have now fixed 3 of the linting errors I knew how to approach. There are 2 more I need some guidance on fixing though, both of them stemming from https://github.com/LemmyNet/lemmy/commit/6047257bfced900c805be4cb965d81caaa7fab4f:

src/shared/components/comment/comment-node.tsx:1494:7 - error TS2345: Argument of type '{ person_id: number; added: boolean; auth: string; }' is not assignable to parameter of type 'AddAdmin'.
  Object literal may only specify known properties, and 'person_id' does not exist in type 'AddAdmin'.
1494       person_id: i.commentView.creator.id,

AND:

src/shared/components/post/post-listing.tsx:1690:7 - error TS2345: Argument of type '{ person_id: number; added: boolean; auth: string; }' is not assignable to parameter of type 'AddAdmin'.
  Object literal may only specify known properties, and 'person_id' does not exist in type 'AddAdmin'.
1690       person_id: i.postView.creator.id,

Both of those are buttons to add another admin to the instance. The new api needs a Local user id to do this.
__IF__ the local user id is equal to the person id of a user, one can just replace the key to be "local_user_id: i.????View.creator.id". Would that be ok or will it break everything?

dessalines:
Thx!

PR #2001, opened 2023-07-28
ES-Lint tweak

SleeplessOne1917:
Triple equals and `const` instead of `let` are my two most common nitpicks. I figure an es-lint rule should help prevent that from being necessary in the future.


dessalines (approved):
Nice, thx.

PR #2022, opened 2023-08-04
Fix comment context bug

### LINKED ISSUE #1999: Show context button does not work
### Requirements

### Summary

When clicking the "Show context" button above a comment, the same comment is loaded instead of the parent comment.

Additionally, I noticed that parent comments are not shown anymore which was introduced earlier in c16488153f01fd355d36513396e30d817c7df4cb

This appears since the upgrade to 0.18.3 and is not present in 0.18.2

### Steps to Reproduce

1. Go to a comment that has a parent comment (e.g. [this one](https://lemmy.ml/comment/2074795))
2. Click "Show context"
3. Page refreshes but parent is not shown

### Technical Details

OS: macOS, Linux
Browser: Safari, Firefox

### Lemmy Instance Version

0.18.3

### Lemmy Instance URL

- https://lemmy.ml
- https://feddit.de

SleeplessOne1917:
## Description

Fixes #1999

### DIFF
--- a/src/shared/routes.ts
+++ b/src/shared/routes.ts
@@ -67,12 +67,12 @@ export const routes: IRoutePropsWithFetch<Record<string, any>>[] = [
     fetchInitialData: Communities.fetchInitialData,
   },
   {
-    path: `/post/:post_id`,
+    path: `/post/:post_id/comment/:comment_id`,
     component: Post,
     fetchInitialData: Post.fetchInitialData,
   },
   {
-    path: `/comment/:comment_id`,
+    path: `/post/:post_id`,
     component: Post,
     fetchInitialData: Post.fetchInitialData,
   },
--- a/generate_translations.js
+++ b/generate_translations.js
@@ -8,12 +8,12 @@ fs.readdir(translationDir, (_err, files) => {
     const lang = filename.split(".")[0];
     try {
       const json = JSON.parse(
-        fs.readFileSync(translationDir + filename, "utf8")
+        fs.readFileSync(translationDir + filename, "utf8"),
       );
       let data = `export const ${lang} = {\n  translation: {`;
       for (const key in json) {
         if (key in json) {
-          const value = json[key].replace(/"/g, '\\"');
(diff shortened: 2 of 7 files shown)


src/shared/routes.ts
```diff
@@ -67,12 +67,12 @@ export const routes: IRoutePropsWithFetch<Record<string, any>>[] = [
     fetchInitialData: Communities.fetchInitialData,
   },
   {
-    path: `/post/:post_id`,
+    path: `/post/:post_id/comment/:comment_id`,
     component: Post,
     fetchInitialData: Post.fetchInitialData,
   },
   {
-    path: `/comment/:comment_id`,
```
dessalines, line 75: Removing this would break all previous comment links. Plus the GetComment endpoint should correctly get the post and context.

I'll take a crack at fixing this, I'm not sure where this bug got introduced, because it worked well before `0.18.3`

SleeplessOne1917: [Someone discussing the issue pointed out the problem.](https://github.com/LemmyNet/lemmy-ui/issues/1999#issuecomment-1664724892) It turns out the issue was because the post ID was not being passed to `getComments`. Hopefully there is a way to fix this without breaking links, but the URL change was motivated by the need to have both the post ID and the comment ID.

I may be able to accomplish this without changing the path if I utilize history state instead.

dessalines: Sure, go right ahead.

dessalines: Ugh, I can't even replicate the issue locally. Its probably because the root issue is this: https://github.com/LemmyNet/lemmy/pull/3717

Comments are now coming back in a random order, and our tree building code everywhere isn't resilient enough to handle unordered comments.

dessalines: Okay I have a PR out for that now.

dessalines: I've found a way to test this locally with my back-end PR, so this may not be necessary anymore. I'll come back in a few minutes to verify.

dessalines:
Okay, I've verified that my backend PR fixes this issue. I apologize for that regression.

PR #2032, opened 2023-08-04
Add open links in new tab

dessalines:
## Description

Supercedes #1546


SleeplessOne1917 (approved):

DetroitLibertyPenguin:
Am I correct in understanding that this adds an option for individual users' to toggle the ability in a SETTINGS menu whether or not they'd like external hyperlinks to be owned in a new tab/window by default?

dessalines:
Yes

projektilski:
In my case this option does not do anything. Links are still opened in the same tab.

dessalines:
Probably a browser or plugin issue. You'd need to open an issue with how to reproduce.

PR #2063, opened 2023-08-15
Federate emoji sizing

makotech222:
## Description

Instances federated with others (mostly hexbear) are finding that custom emojis on federated instance aren't rendered at the same size as locally. This change embeds a 'emoji' string in the markdown title to signify that it is an image that is intended to be rendered the same size as a local custom emoji.


SleeplessOne1917 (approved):
The days of giant pig poop balls are coming to a close! I'm approving this under the assumption you tested this.

makotech222:
Actually i would like some additional testing if you have time, just given the history around emojis being security issue. I'm quite sure its safe, but extra testing and extra set of eyes is always good

PR #2064, opened 2023-08-16
Make alternate way of displaying user names

SleeplessOne1917:
## Description

As discussed in #1975, this is an attempt at displaying apub names alongside display names in a way as to not take up too much horizontal space.

### Before
image

### After
image

### DIFF
--- a/src/shared/components/person/person-listing.tsx
+++ b/src/shared/components/person/person-listing.tsx
@@ -1,8 +1,7 @@
 import { showAvatars } from "@utils/app";
 import { getStaticDir } from "@utils/env";
 import { hostname, isCakeDay } from "@utils/helpers";
-import classNames from "classnames";
-import { Component } from "inferno";
+import { Component, InfernoNode } from "inferno";
 import { Link } from "inferno-router";
 import { Person } from "lemmy-js-client";
 import { relTags } from "../../config";
@@ -11,11 +10,7 @@ import { CakeDay } from "./cake-day";
 
 interface PersonListingProps {
   person: Person;
-  realLink?: boolean;
-  useApubName?: boolean;
-  muted?: boolean;
-  hideAvatar?: boolean;
-  showApubName?: boolean;
+  profile?: boolean;
 }
 
 export class PersonListing extends Component<PersonListingProps, any> {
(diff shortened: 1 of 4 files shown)


src/shared/components/person/person-listing.tsx
```diff
@@ -11,11 +10,7 @@ import { CakeDay } from "./cake-day";
 
 interface PersonListingProps {
   person: Person;
-  realLink?: boolean;
-  useApubName?: boolean;
-  muted?: boolean;
-  hideAvatar?: boolean;
```
SleeplessOne1917, line 17: These 4 props were only used on the profile page, so they were all rolled into the `profile` prop.

src/shared/components/person/person-listing.tsx
```diff
@@ -24,76 +19,62 @@ export class PersonListing extends Component<PersonListingProps, any> {
...
-    let displayName = this.props.useApubName
-      ? apubName
-      : person.display_name ?? apubName;
+    function linkify(node: InfernoNode, isApubName?: boolean) {
+      const className = `text-${display_name || profile ? "muted" : "info"}`;
 
-    if (this.props.showApubName && !local && person.display_name) {
-      displayName = `${displayName} (${apubName})`;
+      if (local && profile) {
+        return isApubName ? <div className={className}>{node}</div> : node;
```
SleeplessOne1917, line 42: This logic wasn't here before, but I figured that there's no point in making it a link when viewing the profile locally because it would just bring the user to the page that they're already on.

lemmy-translations
SleeplessOne1917, line 1: I took the liberty of pulling the latest translations.

dessalines (commented):
Not a big fan of the two line approach either, doesn't look too great.

What about using a marquee / bouncing text? It's used a few places in jerboa and it doesn't look too bad.

https://www.html.am/html-codes/marquees/bouncing-text.cfm

SleeplessOne1917:
> What about using a marquee / bouncing text? It's used a few places in jerboa and it doesn't look too bad.

Do you mean that if a name is too long, it will "bounce" horizontally like a stock ticker to show the full length of the name? I'll take a look at Jerboa and see if that helps me understand.

dessalines:
Yep. Check out the header bar of a federated communitin jerboa.

Another possibility, that I see mastodon android using, is to just use ellipses for overflowing names:



That would force the comment and post header bar to one row.

SleeplessOne1917:
Perhaps both can be displayed and, instead of overflowing with ellipses, the apub username can have the bouncing text affect.

> A note tho, that masto is user-focused like twitter, so they do the two-line usernames, which I don't want here, and which no reddit-like app uses.

While that is true, reddit didn't have to deal with:
- Display names that are different than account names
- Different users on different instances using the same username and/or display name.

Since twitter has "dual names" so to speak, its UI choices for that might be worth emulating more than reddits.

dessalines:
I don't think there's really a good way to show all three pieces of info (username, display name, instance), without cluttering up the UI a lot, without going to the profile.

Matrix, and discord only show display name, and make you go to the profile page just like lemmy-ui does to get the other info. This by far looks the best. 

In the cases I do want to which instance someone is coming from, I can either click the link, or click-and-hold the link to see that info.

Mastodon and twitter show all 3, and scale it by sticking a large user thumbnail next to two lines. IMO this looks pretty ugly, and doesn't work too well for us since we often need to put info on either side of the username. 

We could show `displayname/username@instance` in all cases (or for non-local only), which I would be okay with, as long as its `overflow: ellipses`. It would clutter up the UI a lot tho.

Nutomic:
If you ask me we could get rid of displaynames entirely, they arent that important. But that would be a breaking change and lots of people would complain, so its probably not a good option. 

How about keeping the display like now (show displayname if set, otherwise fallback to user@instance), but show the displayname in a slightly different color to make it obvious that its something else? If displayname is shown, we could even include a button like "show user identifier".

SleeplessOne1917:
Considering we all seem to have different ideas on how this should look, do you think a post should be made about this on lemmy.ml to ask for end user input?

dessalines:
Display names are very useful, I def don't want to get rid of them.

> How about keeping the display like now (show displayname if set, otherwise fallback to user@instance), but show the displayname in a slightly different color to make it obvious that its something else? If displayname is shown, we could even include a button like "show user identifier".

I already use an `@username` to differentiate usernames from displaynames.

@SleeplessOne1917 I have a feeling ppl will vote for the most verbose and least-mobile-friendly option. It also gets complicated because local vs. non-local should be displayed differently. Some options I can think of, in terms of preference for me. 

- {{displayname OR @username}}
- {{displayname OR @username}}@instance.tld
- displayname (username@instance.tld) OR @username@instance.tld *I really don't like this one*.

Nutomic:
> Considering we all seem to have different ideas on how this should look, do you think a post should be made about this on lemmy.ml to ask for end user input?

Sure it cant hurt, maybe there will be some new ideas that we didnt think of.

UInt2048:
Maybe I’m overthinking this, but why couldn’t we just have an additional setting for users to toggle between displaying like Hexbear does (always show `@lemmy.ml` or whatever other instance for federated users after display name) and the current behavior on other instances (don’t show instance name when a display name is set)?

SleeplessOne1917:
@UInt2048 I like that idea, although I would still want to keep the style of displaying instances like the one in the image in the original PR post when the option is enabled. That way, less horizontal space is taken and it's harder for someone to impersonate another user on the same instance.

dessalines:
I'd be *okay* with adding a back-end setting, but I like a sane default better. Plus front-ends are likely to ignore that and display however they want to.

I spose I'd be okay with:

type | display
--- | ---
local | `{{displayname OR @username}}`
federated | `{{displayname OR @username}}@instance.tld`

That's simple, fits on a single line, makes it clear, and shows you where someone came from. 

It has the downside of potentially really long names like `My Long UserName@lemmy.dbzero.com`, which is why I used to be against it, but at least its clearer.

UInt2048:
> That's simple, fits on a single line, makes it clear, and shows you where someone came from.

I think that’s a great solution, but just pointing out the downsides and edge cases for completeness’ sake.

This solution wouldn’t be completely impersonation-proof (I could just go change my display name and profile picture to match Sleepless One’s and impersonate them on the same instance). It would still solve the vast majority of the problem nonetheless, because impersonation would basically be limited to same-instance users, which instance admins could address if it became a problem.

Nor would it always fit on a single line (e.g. “馬鹿 [Ke/Kim/Un bot]@lemmy.fediverse.0.30000000000000004.com”). Not that this would ever be a real Lemmy instance, but there’s not much you’d be able to do if an instance name were to be that long anyways. A double line solution would only be marginally better in allowing a few more characters, but probably not enough to justify the increased space it would take up.

dessalines:
Closed in favor of #2425

PR #2069, opened 2023-08-21
Only show mod/admin identifiers if explictly distinguished for UserBadges on comments

Security-Chief-Odo:
## Description

Hide comment mod/admin identifiers for comments on local instance unless specifically set to _distinguished_. 

Partially Fixes [#1828](https://github.com/LemmyNet/lemmy-ui/issues/1828)

## Screenshots
https://postimg.cc/gallery/7zPFGLR

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -315,8 +315,8 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
               <UserBadges
                 classNames="ms-1"
                 isPostCreator={this.isPostCreator}
-                isMod={isMod_}
-                isAdmin={isAdmin_}
+                isMod={isMod_ && cv.comment.distinguished}
+                isAdmin={isAdmin_ && cv.comment.distinguished}
                 isBot={cv.creator.bot_account}
               />
 


SleeplessOne1917:
I believe showing admin badges at all time was a deliberate decision. @dessalines @Nutomic seeking input here.

dessalines:
Distinguish means (speak officially), not just that you're engaging with someone, and you also happen to be a mod / admin.

To me its not transparent or honest to hide the fact to others that you're really a moderator. Clutter isn't an issue here since its a small icon.

If people want to hide the fact that they are really moderators / admins, maybe they should make other accounts. But that would also be potentially dishonest to the people they're engaging with, who don't realize they're actually moderators.

Security-Chief-Odo:
Why does it matter if someone is engaging with a mod or administrator or just another user of the instance? They should not be entitled to be treated or approached any differently than any other user _unless peaking in the official capacity as such_. 

There is a big difference between me as a regular user saying:

> You aren't following the rules, better do that or be banned

And me the ADMIN saying

> [](https://postimg.cc/1g224k5N) You aren't following the rules, better do that or be banned.

As it stands right now, there _is no difference_ between those two comments on Lemmy. Sometimes admins and moderators just want to discuss things with the group, not be elevated above.

dessalines:
> As it stands right now, there is no difference between those two comments on Lemmy.

A `distinguished` comment looks different from regular comments by adding the `mark` class. [Here's where that is done](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/comment/comment-node.tsx#L292).

> Sometimes admins and moderators just want to discuss things with the group, not be elevated above.

You're looking at it from the mods point of view, but what about the user who's interacting with them? In reality, that person *really does* have some extra abilities, and IMO its not fair to hide that.

Nutomic:
I agree with this change. When I make a joke comment on some funny meme I dont want people to think about the fact that Im an admin, or wonder if it was meant as an official statement. Thats exactly why we have the distinguish feature. I would also hide admin/mod status on posts unless they are stickied. Reddit also does it like this btw.

dessalines:
The only way to resolve this would be some kind of poll, with the question: "Should mod / admin status be hidden from users unless you go to their profile?"

alectrocute (approved):
I support this change. Changes look good, too.

SleeplessOne1917:
Did we ever reach a consensus over whether we even want this change?

dessalines:
No, and also for some reason this is going to the release branch, and not main. I care more about dev consensus tho, so if both yall prefer this, then I'd go with the majority. 

But it still needs to go to main tho.

SleeplessOne1917:
I'm for showing admin badges at all times and showing mod badges for posts/comments in a community a user moderates. Basically the way it is currently.

dessalines:
Ah k. In that case I'll close this one.

Issue #2065, opened 2023-08-18
Linked Instances and Blocked Instances in the same column when they should be in separate columns


DraconicNEO:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

In the Instances section on Lemmy the Blocked and Linked Instances sections used to be separated into separate collumns but in one of the recent updates for some reason it doesn't seem to be working right and the Blocked instances are within the same column but just below the linked Instances.

### Steps to Reproduce

1. Go to an Up to date Instance i.e. lemmy.ml
2. Go to the Instances page
3. The Instances used to be separated into two separate columns, linked and blocked but they now aren't

### Technical Details

Firefox Browser 116 on Ubuntu 22.04 LTS Linux on a PC. It doesn't appear to give any errors it's just no longer separated into two columns like it used to be.

### Lemmy Instance Version

BE: 0.18.4

### Lemmy Instance URL

https://lemmy.ml

PR #2071, opened 2023-08-23
Move allowed and blocked instance tables to the column right of linked instance table

dnswrsrx:
## Description

Addresses #2065 

I think for most instances, the linked instance list will be longer than the allowed and blocked list, combined or not. So it made sense to join the allowed and blocked instance tables into the same column, to the right of the linked instance table.

If there were blocked instances, they would be shown in another table below `Allowed Instances`.

### Before
before
### After
After

### DIFF
--- a/src/shared/components/home/instances.tsx
+++ b/src/shared/components/home/instances.tsx
@@ -108,7 +108,7 @@ export class Instances extends Component<any, InstancesState> {
             />
           </>
         ) : (
-          <></>
+          <h5>No linked instance</h5>
         );
       }
     }
--- a/src/shared/components/home/instances.tsx
+++ b/src/shared/components/home/instances.tsx
@@ -6,12 +6,14 @@ import {
   GetSiteResponse,
   Instance,
 } from "lemmy-js-client";
+import classNames from "classnames";
 import { relTags } from "../../config";
 import { InitialFetchRequest } from "../../interfaces";
 import { FirstLoadService, I18NextService } from "../../services";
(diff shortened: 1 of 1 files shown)


camporter:
Looks better. Thoughts on tabs between the linked, blocked, and allowed instances to make it easier to get to the blocked ones as well?

dnswrsrx:
Hmm, tabs are not a bad option. I'll try it out and post a screenshot of it a bit later on. Or if you (@camporter) wanted to you could to!

SleeplessOne1917:
There's a `Tabs` component that's used in some other parts of the UI you can use if you're trying to add tabs.

dnswrsrx:
A screenshot of the instances organised into tabs. I'm only showing the status tab if there are instances with that status. In this example, there are no blocked instances, so the `Blocked Instances` tab is not displayed.

Instances organised in tabs

SleeplessOne1917:
I think a max width for the table on screen sizes medium and above would be a good addition. The table looks a bit too spaced out in the screenshot.

dnswrsrx:
How about something like this? I wrapped the tabs and table in a `col-lg-8`. On medium screen width and smaller, it'll take the full width. Otherwise it'll be restricted as per that column class.

[Live version with these changes to try resizing](https://lemmy.dnswrsrx.com/instances)

Screenshot 2023-08-29 at 12 11 28

SleeplessOne1917 (approved):

PR #2073, opened 2023-08-23
always show community name with instance for easy copy-pasting to friends

Die4Ever:
## Description

I don't like the inconsistency of community names currently, the UI shouldn't be presenting things differently because of technical reasons that are hard for users to understand. This makes it easier to share communities with your friends!

### Before

image

### After

image

### DIFF
--- a/src/shared/components/community/community-link.tsx
+++ b/src/shared/components/community/community-link.tsx
@@ -23,18 +23,18 @@ export class CommunityLink extends Component<CommunityLinkProps, any> {
     const community = this.props.community;
     let name_: string, title: string, link: string;
     const local = community.local === null ? true : community.local;
+    const domain = hostname(community.actor_id);
     if (local) {
       name_ = community.name;
       title = community.title;
       link = `/c/${community.name}`;
     } else {
-      const domain = hostname(community.actor_id);
       name_ = `${community.name}@${domain}`;
       title = `${community.title}@${domain}`;
       link = !this.props.realLink ? `/c/${name_}` : community.actor_id;
     }
 
-    const apubName = `!${name_}`;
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):
LGTM. Will wait for approval from the maintainers before I merge.

src/shared/components/community/community-link.tsx
```diff
@@ -23,18 +23,18 @@ export class CommunityLink extends Component<CommunityLinkProps, any> {
     const community = this.props.community;
     let name_: string, title: string, link: string;
     const local = community.local === null ? true : community.local;
+    const domain = hostname(community.actor_id);
     if (local) {
       name_ = community.name;
```
camporter, line 28: `_name` isn't really used now.

Die4Ever: thanks, fixed

src/shared/components/community/community-link.tsx
```diff
@@ -21,20 +21,19 @@ export class CommunityLink extends Component<CommunityLinkProps, any> {
...
     } else {
-      const domain = hostname(community.actor_id);
-      name_ = `${community.name}@${domain}`;
+      const name_ = `${community.name}@${domain}`;
       title = `${community.title}@${domain}`;
       link = !this.props.realLink ? `/c/${name_}` : community.actor_id;
     }
 
-    const apubName = `!${name_}`;
+    const apubName = `!${community.name}@${domain}`;
```
dessalines, line 36: This *should* be okay, because it should only use the apubName on community pages, hopefully not in most community links.

dessalines (approved):

PR #2083, opened 2023-08-26
Added error handling to format-past-date to prevent timeouts on modlog

Security-Chief-Odo:
## Description

Improper date strings such as those very far into the future, or typo'd dates cause format-past-date to fail. This causes `modlog` page to load/spin forever instead of displaying any information as expected. This single file change adds error handling to the parse date function to skip improperly formatted dates, and log such to the console.

### DIFF
--- a/src/shared/utils/helpers/format-past-date.ts
+++ b/src/shared/utils/helpers/format-past-date.ts
@@ -2,13 +2,13 @@ import formatDistanceStrict from "date-fns/formatDistanceStrict";
 import parseISO from "date-fns/parseISO";
 
 export default function (dateString?: string) {
-  if (!dateString || dateString === undefined) {
+  if (!dateString) {
     console.error(err);
     return "DATE ERROR";
   }
 
   try {
-    const parsed = parseISO((dateString ?? Date.now().toString()) + "Z");
+    const parsed = parseISO(Date.now().toString() + "Z");
     return formatDistanceStrict(parsed, new Date(), {
       addSuffix: true,
     });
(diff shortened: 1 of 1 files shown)


src/shared/utils/helpers/format-past-date.ts
```diff
@@ -2,8 +2,23 @@ import formatDistanceStrict from "date-fns/formatDistanceStrict";
 import parseISO from "date-fns/parseISO";
 
 export default function (dateString?: string) {
-  const parsed = parseISO((dateString ?? Date.now().toString()) + "Z");
-  return formatDistanceStrict(parsed, new Date(), {
-    addSuffix: true,
-  });
+  if (!dateString || dateString === undefined) {
```
SleeplessOne1917, line 5: The `|| dateString === undefined` part shouldn't be necessary. `undefined` is falsey in JS, thus `!dateString` will evaluate to true if it is undefined.
```suggestion
  if (!dateString) {
```

src/shared/utils/helpers/format-past-date.ts
```diff
@@ -2,8 +2,23 @@ import formatDistanceStrict from "date-fns/formatDistanceStrict";
...
+    console.error(err);
+    return "DATE ERROR";
+  }
+
+  try {
+    const parsed = parseISO((dateString ?? Date.now().toString()) + "Z");
```
SleeplessOne1917, line 11: Since you return an error string if `dateString` is nullish, the null coalesce operator shouldn't be needed anymore:
```suggestion
    const parsed = parseISO(Date.now().toString() + "Z");
```

SleeplessOne1917 (changes requested):
Some minor changes, then you're good to go.

Security-Chief-Odo (commented):
Applied changes from maintainer.

SleeplessOne1917 (approved):

PR #2101, opened 2023-08-30
Improve handling/escaping of a few  html entities for posts titles

Security-Chief-Odo:
## Does not fix backend, only frontend display

Removed a mistakenly included 'helper function' file and index.ts

Modified/removed broken indexes helper functions to good working state. Removed unnecessary 'unescapeHTML' on a few areas that do not require such.

Add better error message to first formatPastDate check


src/shared/utils/helpers/format-past-date.ts
```diff
@@ -2,13 +2,13 @@ import formatDistanceStrict from "date-fns/formatDistanceStrict";
 import parseISO from "date-fns/parseISO";
 
 export default function (dateString?: string) {
-  if (!dateString) {
-    console.error(err);
-    return "DATE ERROR";
+  if (!dateString || dateString === undefined) {
```
SleeplessOne1917, line 5: `undefined` is already falsey. The or is not needed.

src/shared/utils/helpers/index.ts
```diff
@@ -18,7 +18,6 @@ import numToSI from "./num-to-si";
 import poll from "./poll";
 import randomStr from "./random-str";
 import removeAuthParam from "./remove-auth-param";
-import returnStringFromString from "./return-str";
```
SleeplessOne1917, line 21: What is this util? Is it even used? If not, the file it's from should be deleted.

src/shared/components/post/post-listing.tsx
```diff
@@ -286,8 +286,8 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
           <iframe
             allowFullScreen
             className="post-metadata-iframe"
-            src={unescapeHTML(post.embed_video_url)}
-            title={unescapeHTML(post.embed_title)}
+            src={post.embed_video_url}
+            title={post.embed_title}
```
SleeplessOne1917, line 290: I would think a title is one place where you want the string to be unescaped.

SleeplessOne1917 (changes requested):

src/shared/utils/helpers/format-past-date.ts
Security-Chief-Odo: It is needed I believe due to the function parameter _for some reason_ being optional? 

`dateString?:`

I'll remove the `?` from the parameter and the right side `or` check.

SleeplessOne1917: The arg being optional in not a problem. Refer to this [small demo](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABMOcAUBDATgcwPwBciAzlFjGDgJSIDeAUIk4jMImgITbV2PP8QExOABsApgDoRcHGgDkAOTiJucqnyYBfRGJHExvfgKGjJ02dyoBuDYk3179FOmtA). You will see that when an optional arg is not passed (and is thus equal to `undefined`), `!arg` evaluates to `true`. This is because `undefined` is falsey.

Security-Chief-Odo (commented):
Changed as requested

SleeplessOne1917 (approved):

SleeplessOne1917:
Fix the linting errors and it's good to go.

Security-Chief-Odo:
@dessalines When making potential changes and requesting PRs which should I base them against then? I was bashing off the `v0.18.4` tag instead of the `release 0.18` . Would that be more proper?

dessalines:
We use the [stable mainline](https://www.bitsnbites.eu/a-stable-mainline-branching-model-for-git/) branching strategy, so all PRs should go to main. When we do a bug-fix release or another release, we will cherry-pick commits from main.

Right now, your PR will be completely lost / forgotten on the next release.

src/shared/components/post/post-listing.tsx
```diff
@@ -207,9 +207,9 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
           <>
             {this.listing()}
             {this.state.imageExpanded && !this.props.hideImage && this.img}
-            {this.showBody &&
-              unescapeHTML(post.url) &&
-              unescapeHTML(post.embed_title) && <MetadataCard post={post} />}
+            {this.showBody && post.url && post.embed_title && (
```
dessalines, line 210: Asf or the reason for the failure on L290, its because you removed a check for `unescape(post.embed_title)`

Security-Chief-Odo:
> When we do a bug-fix release or another release, we will **cherry-pick commits from main**.

> Right now, your PR will be completely lost / forgotten on the next release.

Then I do not see how those two work together or prevent the issue you say. You cherry pick commits from main, and if the one I pushed to main wasn't cherry picked, it will indeed be _lost/forgotten_ on the next release.

src/shared/components/post/post-listing.tsx
Security-Chief-Odo: Can you help me understand what is causing the error here so I can prevent it in the future? Looking at the [file history - 2fa7c11](https://github.com/LemmyNet/lemmy-ui/blob/2fa7c11666455d28c395d7b8b830faf1c94b2158/src/shared/components/post/post-listing.tsx#L209), it has `{this.showBody && post.url && post.embed_title &&` for that line, instead of separated out to ` unescapeHTML` for each. 

I think what you are saying is that on L209 for the `post.`'s to _not have the `unescapeHTML`_, but instead only  needed on L290. Is that correct?

dessalines: The error is telling you that unescapeHtml needs a `string`, not a `string | undefined`. So the easiest way to fix it, is in L283, do add `&& post.embed_title` to the check.

dessalines:
Before we do releases, we cherry-pick the bug fix / non-breaking commits from `main`. That's something that me and nutomic handle. All devs should work against main, never against release branches.

Read this again: https://www.bitsnbites.eu/a-stable-mainline-branching-model-for-git/

PR #2109, opened 2023-09-02
Display more information why has the login failed

noskla:
## Description
Before this change the user was not aware why the login failed unless they looked into the browser console, this simple change brings more errors to show up in a toast making it more clear and verbose what went wrong. Translations seem to be already in place. (https://github.com/LemmyNet/lemmy-translations/blob/38bb32f6898b227aaad06a48d8bf69a5b48416d6/translations/en.json#L433)

## Screenshots
image

### DIFF
--- a/src/shared/components/home/login.tsx
+++ b/src/shared/components/home/login.tsx
@@ -155,7 +155,13 @@ export class Login extends Component<any, State> {
             i.setState({ showTotp: true });
             toast(I18NextService.i18n.t("enter_two_factor_code"), "info");
           }
-          if (["incorrect_login", "email_not_verified", "registration_application_pending"].includes(loginRes.msg)) {
+          if (
+            [
+              "incorrect_login",
+              "email_not_verified",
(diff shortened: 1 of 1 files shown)


src/shared/components/home/login.tsx
```diff
@@ -155,8 +155,14 @@ export class Login extends Component<any, State> {
             i.setState({ showTotp: true });
             toast(I18NextService.i18n.t("enter_two_factor_code"), "info");
           }
-          if (loginRes.msg === "incorrect_login") {
-            toast(I18NextService.i18n.t("incorrect_login"), "danger");
+          if (
```
dessalines, line 158: You could probably just remove these extra checks, and do an `else { toast(...`

noskla: Yea, that sounds better, in that case even when the localization for the key doesn't exist then it will just display the key itself I think. I've pushed the changes.

SleeplessOne1917 (approved):

PR #2114, opened 2023-09-04
Disable "Next" button in Paginator when the next page is empty

noskla:
## Description
It may cause confusion to the user why pages in the community list are empty when there are no more communities to display. This small change fixes this issue by disabling the Next button if the number of fetched communities is less than the limit.

Additionally, I have moved the communityLimit constant to the config file.

## Possible issues with this approach
The Next button will be still enabled, if there are exactly _{communityLimit}_ communities on the instance, to mitigate this the API would have to tell the client if currently viewed page is last, or the client could fetch next page in advance.

### DIFF
--- a/src/shared/components/common/paginator.tsx
+++ b/src/shared/components/common/paginator.tsx
@@ -4,6 +4,7 @@ import { I18NextService } from "../../services";
 interface PaginatorProps {
   page: number;
   onChange(val: number): any;
+  nextDisabled?: boolean;
 }
 
 export class Paginator extends Component<PaginatorProps, any> {
@@ -23,6 +24,7 @@ export class Paginator extends Component<PaginatorProps, any> {
         <button
           className="btn btn-secondary"
           onClick={linkEvent(this, this.handleNext)}
+          disabled={props.nextDisabled || false}
         >
           {I18NextService.i18n.t("next")}
         </button>
--- a/src/shared/components/common/paginator.tsx
+++ b/src/shared/components/common/paginator.tsx
@@ -4,7 +4,7 @@ import { I18NextService } from "../../services";
 interface PaginatorProps {
   page: number;
   onChange(val: number): any;
-  nextDisabled?: boolean;
+  nextDisabled: boolean;
 }
 
 export class Paginator extends Component<PaginatorProps, any> {
--- a/src/shared/components/common/paginator.tsx
+++ b/src/shared/components/common/paginator.tsx
@@ -24,7 +24,7 @@ export class Paginator extends Component<PaginatorProps, any> {
         <button
           className="btn btn-secondary"
           onClick={linkEvent(this, this.handleNext)}
-          disabled={props.nextDisabled || false}
+          disabled={this.props.nextDisabled || false}
         >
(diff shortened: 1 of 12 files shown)


SleeplessOne1917 (approved):
This is a good fix. It's probably worth doing something like this everywhere the paginator is used.

src/shared/components/community/communities.tsx
```diff
@@ -221,7 +221,14 @@ export class Communities extends Component<any, CommunitiesState> {
                 </tbody>
               </table>
             </div>
-            <Paginator page={page} onChange={this.handlePageChange} />
+            <Paginator
+              page={page}
+              onChange={this.handlePageChange}
+              nextDisabled={
+                communityLimit >
+                this.state.listCommunitiesResponse.data.communities.length
```
dessalines, line 229: Looks good. Could you add this logic also everywhere else that its feasible? (IE Home.tsx, Community.tsx especially) 

The pages with multiple request types (like `inbox.tsx`, `modlog.tsx` , it won't work, so you can ignore those ones)

dessalines (approved):
Looks good, thx!

src/shared/components/common/paginator.tsx
```diff
@@ -23,6 +24,7 @@ export class Paginator extends Component<PaginatorProps, any> {
         <button
           className="btn btn-secondary"
           onClick={linkEvent(this, this.handleNext)}
+          disabled={this.props.nextDisabled || false}
```
SleeplessOne1917, line 27: Is the `|| false` necessary here?

src/shared/components/home/emojis-form.tsx
```diff
@@ -267,7 +267,11 @@ export class EmojiForm extends Component<EmojiFormProps, EmojiFormState> {
             {I18NextService.i18n.t("add_custom_emoji")}
           </button>
 
-          <Paginator page={this.state.page} onChange={this.handlePageChange} />
+          <Paginator
+            page={this.state.page}
+            onChange={this.handlePageChange}
+            nextDisabled={false}
```
SleeplessOne1917, line 273: Is there never a time that this needs to be disabled?

src/shared/components/modlog.tsx
```diff
@@ -860,7 +860,11 @@ export class Modlog extends Component<
               </thead>
               {this.combined}
             </table>
-            <Paginator page={page} onChange={this.handlePageChange} />
+            <Paginator
+              page={page}
+              onChange={this.handlePageChange}
+              nextDisabled={false}
```
SleeplessOne1917, line 866: Same here.

src/shared/components/common/paginator.tsx
noskla: Good catch, it's a residue after nextDisabled was optional to not pass "undefined" to the property, I'll fix it

src/shared/components/home/emojis-form.tsx
dessalines: It needs to be disabled when there are multiple lists (because the API doesn't combine them)

SleeplessOne1917 (approved):

Issue #2115, opened 2023-09-04
Allow for Spoiler Text in Markdown in addition to the collapsed sections currently supported


DraconicNEO:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

I think it would be good if Lemmy-ui supported spoiler text in addition to the spoiler tags since collapsed spoiler sections don't work well if you only want to censor specific words, names, or phrases.

I think a good solution would be to add a simple spoiler text MD implementation such as `>!spoiler text!<` where the spoiler portion is began by typing `>!` before the spoilered section and terminated by typing `!<` at the end of it and unlike the current spoiler implementation which uses a drop-down to reveal it, this one would just be blocks over the text that would reveal the text when clicked on and since the syntax is simpler it can be used to block out words or phrases without disrupting the structure of a sentence.

DraconicNEO:
> This has been implemented in #2126

Awesome, does it use the same `>!spoiler text!<` implementation or is it initiated using different characters?

Also if you don't mind could I see a preview of what it looks like?

DraconicNEO:
> > This has been implemented in #2126
> 
> If I understand the issue descriptions correctly, then no, I don't think this has been implemented. That PR is about supporting markdown styling within spoiler block headings. This issue is about implementing an inline spoiler syntax.

Well I do know that some people have features as part of unrelated pull requests so I wasn't sure if he did that or if he's just confused and thought that I meant something different.

>This issue may be a duplicate of https://github.com/LemmyNet/lemmy-ui/issues/687.

I believe it is, that is exactly what I'm describing.

PR #2126, opened 2023-09-11
allow limited markdown in spoiler summary

biosfood:
## Description

Parse markdown in spoiler summaries (the text you see before expanding it)
Only the limited markdown parser with only emphasis, backticks and strikethrough allowed to prevent issues with unexpected content there.

## Screenshots

spoiler

### DIFF
--- a/src/shared/markdown.ts
+++ b/src/shared/markdown.ts
@@ -61,10 +61,11 @@ const spoilerConfig = {
 
   render: (tokens: any, idx: any) => {
     var m = tokens[idx].info.trim().match(/^spoiler\s+(.*)$/);
+    var summary = mdToHtmlInline(md.utils.escapeHtml(m[1])).__html;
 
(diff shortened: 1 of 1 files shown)


biosfood:
fixes https://github.com/LemmyNet/lemmy-ui/issues/2115

src/shared/markdown.ts
```diff
@@ -61,10 +61,11 @@ const spoilerConfig = {
 
   render: (tokens: any, idx: any) => {
     var m = tokens[idx].info.trim().match(/^spoiler\s+(.*)$/);
+    var summary = mdToHtmlInline(md.utils.escapeHtml(m[1])).__html;
```
dessalines, line 64: Seems fine. We should really be using const for all those tho, but nbd.

dessalines (approved):

SleeplessOne1917 (approved):

Issue #1744, opened 2023-07-02
Notification bell icon does not update the unread counter


Azzurite:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

When marking all notifications as read, the little bell icon still shows the previous number of notifications. Only after a full page refresh does the little number next to the bell icon disappear.

### Steps to Reproduce

1. Get some notifications
2. Mark them as "read"
3. Bell icon still shows the same number of notifications

### Technical Details

Windows, Firefox 115.0b9

### Lemmy Instance Version

UI: 0.18.1-rc.7, BE: 0.18.1-rc.4

### Lemmy Instance URL

https://lemm.ee

PR #2132, opened 2023-09-13
Improve notifications

biosfood:
## Description

Streamline the user experience by making the inbox counter more responsive.

Fixes https://github.com/LemmyNet/lemmy-ui/issues/1744

Right now, lemmy sends a cache header for some frequently queried data, like api/v3/user/unread_count or /api/v3/private_message/list of 60s.

When updating a message to be displayed as read, I would ideally like to just refresh the notification counter, but get old cached data from the browser.

@Nutomic @dessalines Is this caching behaviour intended or should it be changed on the backend? If a client is acting correctly, it should just avoid sending a request more than every minute but shortly after making changes, I think it would be wise to allow it to update.

An alternative might be to just update the view on the client and purely hope the server registers the changes but a user might find it useful to get immediate feedback of the new data that is stored on the server (to avoid desync of client and server).

### DIFF
--- a/src/shared/components/app/navbar.tsx
+++ b/src/shared/components/app/navbar.tsx
@@ -51,12 +51,16 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
     unreadApplicationCount: 0,
   };
 
-  updateUnreads = (service: UnreadCounterService) => {
+  updateUnreads = ({
+    unreadInboxCount,
+    unreadReportCount,
+    unreadApplicationCount,
+  }: UnreadCounterService) => {
     this.setState({
       ...this.state,
-      unreadInboxCount: service.unreadInboxCount,
-      unreadReportCount: service.unreadReportCount,
-      unreadApplicationCount: service.unreadApplicationCount,
+      unreadInboxCount,
+      unreadReportCount,
+      unreadApplicationCount,
     });
   };
 
--- a/src/shared/components/app/navbar.tsx
+++ b/src/shared/components/app/navbar.tsx
@@ -14,6 +14,7 @@ import {
 import { toast } from "../../toast";
 import { Icon } from "../common/icon";
 import { PictrsImage } from "../common/pictrs-image";
+import { Subscription } from "rxjs";
 
 interface NavbarProps {
   siteRes?: GetSiteResponse;
@@ -44,24 +45,14 @@ function handleLogOut(i: Navbar) {
 export class Navbar extends Component<NavbarProps, NavbarState> {
   collapseButtonRef = createRef<HTMLButtonElement>();
   mobileMenuRef = createRef<HTMLDivElement>();
+  unreadInboxCountSubscription: Subscription;
+  unreadReportCountSubscription: Subscription;
+  unreadApplicationCountSubscription: Subscription;
 
   state: NavbarState = {
     unreadInboxCount: 0,
     unreadReportCount: 0,
     unreadApplicationCount: 0,
-  };
-
-  updateUnreads = ({
-    unreadInboxCount,
-    unreadReportCount,
-    unreadApplicationCount,
-  }: UnreadCounterService) => {
-    this.setState({
-      ...this.state,
-      unreadInboxCount,
-      unreadReportCount,
-      unreadApplicationCount,
-    });
   };
 
   constructor(props: any, context: any) {
@@ -75,7 +66,18 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
     if (isBrowser()) {
       // On the first load, check the unreads
       this.requestNotificationPermission();
-      UnreadCounterService.Instance.subscribe(this.updateUnreads);
+      this.unreadInboxCountSubscription =
+        UnreadCounterService.Instance.unreadInboxCountSubject.subscribe(
+          unreadInboxCount => this.setState({ unreadInboxCount }),
+        );
+      this.unreadReportCountSubscription =
+        UnreadCounterService.Instance.unreadReportCountSubject.subscribe(
+          unreadReportCount => this.setState({ unreadReportCount }),
+        );
+      this.unreadApplicationCountSubscription =
+        UnreadCounterService.Instance.unreadApplicationCountSubject.subscribe(
+          unreadApplicationCount => this.setState({ unreadApplicationCount }),
+        );
       this.requestNotificationPermission();
 
       document.addEventListener("mouseup", this.handleOutsideMenuClick);
@@ -84,7 +86,9 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
 
   componentWillUnmount() {
     document.removeEventListener("mouseup", this.handleOutsideMenuClick);
-    UnreadCounterService.Instance.unsubscribe(this.updateUnreads);
+    this.unreadInboxCountSubscription.unsubscribe();
+    this.unreadReportCountSubscription.unsubscribe();
+    this.unreadApplicationCountSubscription.unsubscribe();
   }
 
   // TODO class active corresponding to current pages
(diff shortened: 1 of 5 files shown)


src/shared/services/UnreadCounterService.ts
```diff
@@ -0,0 +1,95 @@
...
+  // fetched by HttpService.getUnreadRegistrationApplicationCount, appear in registration application page
+  public unreadApplicationCount = 0;
+
+  static #instance: UnreadCounterService;
+  private subscriptions = new Set<(service: UnreadCounterService) => any>();
+  private fetching = false;
+
+  constructor() {
+    if (isBrowser()) {
+      poll(this.update, updateUnreadCountsInterval);
```
biosfood, line 33: I am pretty sure this poll function just leaks memory over time due to the recursion (need a new stack frame every time)

src/shared/services/UnreadCounterService.ts
```diff
@@ -0,0 +1,95 @@
...
+  commentReportCount = 0;
+  postReportCount = 0;
+  messageReportCount = 0;
+  public unreadReportCount = 0;
+
+  // fetched by HttpService.getUnreadRegistrationApplicationCount, appear in registration application page
+  public unreadApplicationCount = 0;
+
+  static #instance: UnreadCounterService;
+  private subscriptions = new Set<(service: UnreadCounterService) => any>();
```
biosfood, line 28: A bit of a hack right now.
A more "react" way to do this would be to use React.createContext and React.useContext for a global state.
This here has the advantage of not requiring to be called from a react component

dessalines (commented):
A **much** simpler solution would be to just refetch those counts after a message is marked as read / unread.

Those routes shouldn't be using a 60s cache, I thought that was only in use for public / non-logged-in users?

biosfood:
> A **much** simpler solution would be to just refetch those counts after a message is marked as read / unread.

That is exactly what I was trying to do. However, the response headers for api/v3/user/unread_count and also the private message query and some more have this:

cache-control: public, max-age=60

Because the client already fetches the unread count every 30s, the cache present and I get back the old data.

dessalines:
> It seems the whole authentication system is being reworked right now in https://github.com/LemmyNet/lemmy/pull/3946, so I'll just wait with this until that auth system is done. From what I have seen of it so far, the issue should be gone by then (auth just passed around with a header, right?)

Yep, that's entirely correct. Once we update lemmy-ui to account for the new auth method, hopefully that will fix the caching issue.

biosfood:
@SleeplessOne1917 Thanks for adopting the improved authentication. Things are working pretty well now. When marking a message as read, the counter will immediately update. Let me know if any more changes are needed.

src/shared/components/app/navbar.tsx
```diff
@@ -44,14 +42,24 @@ function handleLogOut(i: Navbar) {
...
+    unreadApplicationCount: 0,
+  };
+
+  updateUnreads = (service: UnreadCounterService) => {
+    this.setState({
+      ...this.state,
+      unreadInboxCount: service.unreadInboxCount,
+      unreadReportCount: service.unreadReportCount,
+      unreadApplicationCount: service.unreadApplicationCount,
+    });
```
SleeplessOne1917, line 60: No need to change if you don't want to, but this can be written more concisely like so:
```suggestion
  updateUnreads = ({
      unreadInboxCount,
      unreadReportCount,
      unreadApplicationCount
  }: UnreadCounterService) => {
    this.setState({
      ...this.state,
      unreadInboxCount,
      unreadReportCount,
      unreadApplicationCount,
    });
```

src/shared/services/UnreadCounterService.ts
SleeplessOne1917: You could try using an interval, although that may require a change to `poll` that is outside of the context of this PR.

src/shared/services/UnreadCounterService.ts
SleeplessOne1917: Inferno does have support for contexts, but they are undocumented and have the same API as contexts for react class components. The only point I have is on the type: will the functions in the set actually return anything? If not, you could swap the `any` for `void`. If it does have to return something, you could try:
```typescript
<T>(service: UnreadCounterService) => T
```

src/shared/components/app/navbar.tsx
```diff
@@ -44,14 +42,28 @@ function handleLogOut(i: Navbar) {
...
 export class Navbar extends Component<NavbarProps, NavbarState> {
-  state: NavbarState = {
-    unreadInboxCountRes: { state: "empty" },
-    unreadReportCountRes: { state: "empty" },
-    unreadApplicationCountRes: { state: "empty" },
-  };
   collapseButtonRef = createRef<HTMLButtonElement>();
   mobileMenuRef = createRef<HTMLDivElement>();
 
+  state: NavbarState = {
```
dessalines, line 48: Why is all this global state necessary now tho? Why not just re-fetch the unread counts after an update, that'd be much cleaner and more reliable (esp since two ppl could be doing reports at the same time)

biosfood: To avoid requerying the counts almost every few seconds to make things responsive, I need to trigger the reloading of the updates from other components (especially the inbox, but also when marking reports as read etc.). All other options I have considered seem not really suitable (very frequent queries, having a reference to the navbar in the inbox component, adding a global variable to notify the navbar that it (probably) needs to update its counts).

When this is merged, I would plan to update the other "inboxes" (reports and applications) to work the same way the inbox does now. So the whole thing should be updatable from any component.

dessalines: You don't need to re-query them every few seconds tho, only after an action (like mark read, etc). And those counts are very inexpensive for a server.

biosfood: What I was trying to say is that I don't see another option than to introduce a global object that can be notified by any component that it probably has to update itself. The component model in inferno generally does not allow one component to call a method on another component, because they don't know of each other (except their children).

I can't just call something like `Navbar.notifyChange()` from the Inbox component, because there _might_ be multiple navbars or the only one is currently not being rendered. This is why I introduced this global service and state, to get around this issue and not have do do a lot of API calls, only when necessary.

SleeplessOne1917: If you're trying to keep state in one place that can be accessed by several components, you could try a context like you mentioned in the other comment. The context API in inferno should work the same way as described in these [old React docs](https://legacy.reactjs.org/docs/context.html) (inferno makes its API similar to pre-hooks react, which is why the docs I'm suggesting are old).

biosfood: That is of course something I can do, just tell me weather you would prefer a proper context over what is currently present. Might take a day or two though as I'm quite at the moment

SleeplessOne1917: I think the context is the way to go since we're dealing with state that can be changed from several different places in the app.

dessalines (commented):
Run `git show 2b1af707c3df6126b3e6890106c03c60ad49b1be` , and look at `navbar.tsx` and `UserService.ts`

That was when we used proper rxjs observables to update the navbar, with `UserService` being the main source of truth:

`public unreadReportCountSub: BehaviorSubject<number> = new BehaviorSubject<number>(0);` . 

Then any component like the navbar could subscribe to it to get real-time changes:

`this.unreadInboxCountSub = UserService.Instance.unreadInboxCountSub.subscribe(res => {this.setState({ unreadInboxCount: res });});`


 You can update it like: 

`UserService.Instance.unreadInboxCountSub.next(DATA)`

dessalines (approved):
Much better, thx!

SleeplessOne1917 (approved):

PR #2141, opened 2023-09-24
Smart url parsing

### LINKED ISSUE #2116: Smart link formatting
### Requirements

Rather than needing to manually type out hyperlinks, this would allow seamlessly creating links on the fly.

The ideal workflow would be:

1. Copy the desired URL
2. Select the text you want to make into a link
3. Paste the URL. The system would then automatically turn the selected text into a hyperlink pointing to the pasted URL.

The idea is to listen to paste events, check if the pasted text is a URL and if there's selected text substitute it with the pasted text.

Something like this:

```js
// Store selected text 
let selectedText = '';

document.addEventListener('copy', e => {
  selectedText = window.getSelection().toString();
})

document.addEventListener('paste', e => {
  const pasted = e.clipboardData.getData('text');
  if(isValidUrl(pasted)) { 
    e.preventDefault();
    insertLink(selectedText, pasted);
  }
});

function insertLink(text, url) {
  document.execCommand('insertHTML', false, `[${text}](${url})`);
}
```

There could also be a user setting to choose the preferred link format between

1. Inline Link:

```markdown
[Link](http://a.com)
```

2. Reference Link:

```markdown
[Link][1]
⋮
[1]: http://b.org
```

Related to [...](https://github.com/LemmyNet/lemmy-ui/issues/735)

Trombach:
## Description

This PR resolves #2116 

Update: The behaviour is now a bit more similar to how Github handles this. I have also updated the video to show the new behaviour

Adds logic to detect if a paste event contains a URL. If true the URL will be formatted with the appropriate markdown. Open to suggestions on how to handle text selection after the paste event. The current implementation is demonstrated in the videos below.

It might also be better to implement this in the same way github does this, only adding formatting if text is already selected and otherwise just pasting the plain link, which might make things easier to read in the editor if no alternative link text is desired.

### Before
[Screencast from 2023-09-24 17-30-03.webm](https://github.com/LemmyNet/lemmy-ui/assets/19306765/9b175520-77fa-42af-880c-fe16012c092f)

### After
[Screencast from 2023-09-26 21-59-39.webm](https://github.com/LemmyNet/lemmy-ui/assets/19306765/55bcdac4-c705-4301-adb6-c10a9487c266)

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -233,7 +233,7 @@ export class MarkdownTextArea extends Component<
                   )}
                   value={this.state.content}
                   onInput={linkEvent(this, this.handleContentChange)}
-                  onPaste={linkEvent(this, this.handleImageUploadPaste)}
+                  onPaste={linkEvent(this, this.handlePaste)}
                   onKeyDown={linkEvent(this, this.handleKeyBinds)}
                   required
                   disabled={this.isDisabled}
@@ -374,10 +374,55 @@ export class MarkdownTextArea extends Component<
     autosize.update(textarea);
   }
 
-  handleImageUploadPaste(i: MarkdownTextArea, event: any) {
+  handlePaste(i: MarkdownTextArea, event: ClipboardEvent) {
+    if (!event.clipboardData) return;
+
+    // check clipboard files
     const image = event.clipboardData.files[0];
     if (image) {
       i.handleImageUpload(i, image);
+      return;
+    }
+
+    // check clipboard url
+    const url = event.clipboardData.getData("text");
+    if (i.isValidUrl(url)) {
+      i.handleUrlPaste(url, i, event);
+      return;
+    }
+  }
+
+  handleUrlPaste(url: string, i: MarkdownTextArea, event: ClipboardEvent) {
+    // query textarea element
+    const textarea = document.getElementById(i.id);
+
+    if (textarea instanceof HTMLTextAreaElement) {
+      event.preventDefault();
+      const { selectionStart, selectionEnd } = textarea;
+
+      // update textarea content
+      i.setState({
+        content: `${
+          i.state.content ? i.state.content.substring(0, selectionStart) : ""
+        }[${i.getSelectedText()}](${url})${
+          i.state.content ? i.state.content.substring(selectionEnd) : ""
+        }`,
+      });
+      i.contentChange();
+
+      // shift selection 1 to the right
+      textarea.setSelectionRange(selectionStart + 1, selectionEnd + 1);
+    }
+  }
+
+  isValidUrl(value: string): boolean {
+    if (!value) return false;
+
+    try {
+      new URL(value);
+      return true;
+    } catch {
+      return false;
     }
   }
 
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -21,6 +21,7 @@ import { EmojiPicker } from "./emoji-picker";
 import { Icon, Spinner } from "./icon";
 import { LanguageSelect } from "./language-select";
 import ProgressBar from "./progress-bar";
+import validUrl from "@utils/helpers/valid-url";
 interface MarkdownTextAreaProps {
   /**
    * Initial content inside the textarea
@@ -385,14 +386,13 @@ export class MarkdownTextArea extends Component<
     }
 
     // check clipboard url
-    const url = i.isValidUrl(event.clipboardData.getData("text"));
-    if (url) {
+    const url = event.clipboardData.getData("text");
+    if (validUrl(url)) {
       i.handleUrlPaste(url, i, event);
-      return;
     }
   }
 
-  handleUrlPaste(url: URL, i: MarkdownTextArea, event: ClipboardEvent) {
+  handleUrlPaste(url: string, i: MarkdownTextArea, event: ClipboardEvent) {
     // query textarea element
     const textarea = document.getElementById(i.id);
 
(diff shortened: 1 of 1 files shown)


SleeplessOne1917:
> It might also be better to implement this in the same way github does this, only adding formatting if text is already selected and otherwise just pasting the plain link, which might make things easier to read in the editor if no alternative link text is desired.

This is the better approach in my opinion. I'll wait for others to chime in on how it should be done before checking out the code.

dessalines:
I agree with sleepless, it'd be better to detect selected text first, then a paste event can turn it into a markdown link.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -374,10 +374,59 @@ export class MarkdownTextArea extends Component<
...
+      const selectedText = i.getSelectedText() || url.toString();
+
+      // update textarea content
+      i.setState({
+        content: `${
+          i.state.content ? i.state.content.substring(0, selectionStart) : ""
+        }[${selectedText}](${url.toString()})${
+          i.state.content ? i.state.content.substring(selectionEnd) : ""
+        }`,
+      });
```
SleeplessOne1917, line 411: I'm pretty sure inferno has similar issues to react when trying to set state using a previous value of that state. Try this:
```suggestion
     i.setState(({content}) => ({
         content: `${content?.substring(0, selectionStart) ?? ""}[${selectedText}](${url.toString()})${content?.substring(selectionEnd) ?? ""}`
     }));
```

You will probably need to format this with prettier still.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -374,10 +374,59 @@ export class MarkdownTextArea extends Component<
...
+  }
+
+  handleUrlPaste(url: URL, i: MarkdownTextArea, event: ClipboardEvent) {
+    // query textarea element
+    const textarea = document.getElementById(i.id);
+
+    if (textarea instanceof HTMLTextAreaElement) {
+      event.preventDefault();
+      const { selectionStart, selectionEnd } = textarea;
+      const selectedText = i.getSelectedText() || url.toString();
```
SleeplessOne1917, line 402: If the return value of `getSelectedText` can be null or undefined, it is slightly clearer to use `??` instead of `||`.

Trombach: I actually used `||` specifically since `getSelectedText` can't return `null` or `undefined` but could return `""`. I wanted to fallback to the url in that case, but since the consensus seems to be that the smart formatting should not be applied if no text is selected this fallback will become obsolete.

Trombach:
@SleeplessOne1917 @dessalines Thanks for the initial feedback. I have updated the PR and brought the behaviour closer to what Github does. I have updated the video as well

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -374,10 +374,63 @@ export class MarkdownTextArea extends Component<
...
+
+      // shift selection 1 to the right
+      textarea.setSelectionRange(
+        selectionStart + 1,
+        selectionStart + 1 + selectedText.length,
+      );
+    }
+  }
+
+  isValidUrl(value: string) {
```
SleeplessOne1917, line 426: This function should either be renamed or inlined. Seeing a url returned from what the name makes seem should be a boolean threw me through a loop for a second.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -374,10 +374,63 @@ export class MarkdownTextArea extends Component<
...
     if (image) {
       i.handleImageUpload(i, image);
+      return;
+    }
+
+    // check clipboard url
+    const url = i.isValidUrl(event.clipboardData.getData("text"));
+    if (url) {
+      i.handleUrlPaste(url, i, event);
+      return;
```
SleeplessOne1917, line 391: This return statement is unnecessary since it's at the end of the function anyway.

SleeplessOne1917 (changes requested):

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -374,10 +374,63 @@ export class MarkdownTextArea extends Component<
...
+
+  handleUrlPaste(url: URL, i: MarkdownTextArea, event: ClipboardEvent) {
+    // query textarea element
+    const textarea = document.getElementById(i.id);
+
+    if (textarea instanceof HTMLTextAreaElement) {
+      const { selectionStart, selectionEnd } = textarea;
+
+      // if no selection, just insert url
+      if (selectionStart === selectionEnd) return;
```
dessalines, line 403: How is it doing anything here?

src/shared/components/common/markdown-textarea.tsx
dessalines: Also, make sure the return type is explicitly named: `: bool`

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -374,10 +374,63 @@ export class MarkdownTextArea extends Component<
...
+      // shift selection 1 to the right
+      textarea.setSelectionRange(
+        selectionStart + 1,
+        selectionStart + 1 + selectedText.length,
+      );
+    }
+  }
+
+  isValidUrl(value: string) {
+    if (!value) return false;
```
dessalines, line 427: This isn't necessary, you aren't passing in `string | undefined`, so we can assume its there.

dessalines (commented):
I can't wait to move all this to rust so I can make everyone stop doing early returns ;)

IE `return if (condition) { } else if (condition) { } else { }`

src/shared/components/common/markdown-textarea.tsx
Trombach: The comment explains it. If no text is selected (i.e. the curser is in one position and therefore `selectionStart === selectionEnd`) we don't want to apply formatting

src/shared/components/common/markdown-textarea.tsx
Trombach: Due to the beauty of javascript and `""` being falsy, it does something :) It actually prevents a constructor call to `URL` if you pass in an empty string which would never be a valid URL. I'm moving to the `validUrl` helper function, so it won't be relevant.

SleeplessOne1917 (approved):

src/shared/components/common/markdown-textarea.tsx
dessalines: Gotcha, it has to do with event.preventDefault() not being called.

PR #2144, opened 2023-09-26
feat: Block instance

SleeplessOne1917:
## Description
This makes the UI able to take advantage of the [block instance feature](https://github.com/LemmyNet/lemmy/pull/3869). 

### Desktop
image

### Mobile
image

### DIFF
--- a/src/shared/components/person/settings.tsx
+++ b/src/shared/components/person/settings.tsx
@@ -10,6 +10,7 @@ import {
   setTheme,
   showLocal,
   updateCommunityBlock,
+  updateInstanceBlock,
   updatePersonBlock,
 } from "@utils/app";
 import { capitalizeFirstLetter, debounce } from "@utils/helpers";
@@ -19,6 +20,7 @@ import { NoOptionI18nKeys } from "i18next";
 import { Component, linkEvent } from "inferno";
 import {
   BlockCommunityResponse,
+  BlockInstanceResponse,
   BlockPersonResponse,
   CommunityBlockView,
   DeleteAccountResponse,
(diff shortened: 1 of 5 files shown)


src/shared/components/person/settings.tsx
```diff
@@ -45,11 +51,17 @@ import { SortSelect } from "../common/sort-select";
 import Tabs from "../common/tabs";
 import { CommunityLink } from "../community/community-link";
 import { PersonListing } from "./person-listing";
+import { InitialFetchRequest } from "shared/interfaces";
```
dessalines, line 54: Even tho this compiles, I'm like 80% sure this will fail. The build system does not liken these style imports.

src/shared/components/person/settings.tsx
```diff
@@ -45,11 +51,17 @@ import { SortSelect } from "../common/sort-select";
...
+
+type SettingsData = RouteDataResponse<{
```
dessalines, line 56: Your call, but seems weird to mix type and interface, and probably better to always use interface wherever possible.

src/shared/routes.ts
```diff
@@ -95,6 +95,7 @@ export const routes: IRoutePropsWithFetch<Record<string, any>>[] = [
   {
     path: `/settings`,
     component: Settings,
+    fetchInitialData: Settings.fetchInitialData,
```
dessalines, line 98: Thx for adding.

src/shared/components/person/settings.tsx
```diff
@@ -987,6 +1084,27 @@ export class Settings extends Component<any, SettingsState> {
     });
   });
 
+  handleInstanceSearch = debounce(async (text: string) => {
```
dessalines, line 1087: With the massive number of federated instances we have now, we really need to add instance searching to `/search`, as well as do paging for `getFederatedInstances` . Obvi not rn, but ideally would be using the search endpoint like the others.

src/shared/components/person/settings.tsx
SleeplessOne1917: An interface won't work here. Interfaces can be used to defined the shape of a type, but they cannot be used to alias types like is being done here.

dessalines (approved):
👍

PR #2173, opened 2023-10-05
Adding cursor pagination. Fixes #2155

### LINKED ISSUE #2155: Add cursor-based pagination
### Requirements

The current paging method will eventually be deprecated, and should now use the `paging_cursor` going forward.

https://github.com/LemmyNet/lemmy/pull/3872

dessalines:
## Description

Adds cursor pagination by adding a `pageCursor` in place of `page` in the home and community urls.

Since the API doesn't return a `previous_cursor`, I had to use the back button for a previous, but it works well.

This breaks next page comment pagination (the first page shows fine), but we should probably do away with those at some point anyway.

### DIFF
--- a/src/shared/components/common/paginator-cursor.tsx
+++ b/src/shared/components/common/paginator-cursor.tsx
@@ -1,0 +1,46 @@
+import { Component, linkEvent } from "inferno";
+import { I18NextService } from "../../services";
+import { PaginationCursor } from "lemmy-js-client";
+
+interface PaginatorCursorProps {
+  prevPage?: PaginationCursor;
+  nextPage?: PaginationCursor;
+  onNext(val: PaginationCursor): any;
+  onPrev(): any;
+}
+
+export class PaginatorCursor extends Component<PaginatorCursorProps, any> {
+  constructor(props: any, context: any) {
+    super(props, context);
+  }
+  render() {
+    return (
+      <div className="paginator my-2">
+        <button
+          className="btn btn-secondary me-2"
+          disabled={!this.props.prevPage}
+          onClick={linkEvent(this, this.handlePrev)}
+        >
+          {I18NextService.i18n.t("prev")}
+        </button>
+        <button
+          className="btn btn-secondary"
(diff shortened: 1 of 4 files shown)


src/shared/components/common/paginator-cursor.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  handlePrev(i: PaginatorCursor) {
+    i.props.onPrev();
+  }
+
+  handleNext(i: PaginatorCursor) {
+    if (i.props.nextPage) {
+      i.props.onNext(i.props.nextPage);
+    }
+  }
```
SleeplessOne1917, line 45: Nitpick: these can be declared as regular functions instead of class methods since you are using `linkEvent`.

src/shared/components/common/paginator-cursor.tsx
```diff
@@ -0,0 +1,46 @@
+import { Component, linkEvent } from "inferno";
+import { I18NextService } from "../../services";
+import { PaginationCursor } from "lemmy-js-client";
+
+interface PaginatorCursorProps {
+  prevPage?: PaginationCursor;
+  nextPage?: PaginationCursor;
+  onNext(val: PaginationCursor): any;
+  onPrev(): any;
```
SleeplessOne1917, line 9: We don't use what gets returned from this (if anything), so it could just return `void`.

src/shared/components/common/paginator-cursor.tsx
```diff
@@ -0,0 +1,46 @@
+import { Component, linkEvent } from "inferno";
+import { I18NextService } from "../../services";
+import { PaginationCursor } from "lemmy-js-client";
+
+interface PaginatorCursorProps {
+  prevPage?: PaginationCursor;
+  nextPage?: PaginationCursor;
+  onNext(val: PaginationCursor): any;
```
SleeplessOne1917, line 8: Do we use the return value for this? If not, we could make this return `void` as well.

SleeplessOne1917 (approved):

PR #2179, opened 2023-10-11
Tweak 2fa

SleeplessOne1917:
A small tweak to the 2fa modal. This should hopefully work better with virtual keyboards on mobile phones, however I am not sure how to test that locally.

This is what it's like on voyager currently:

https://github.com/LemmyNet/lemmy-ui/assets/28871516/be612172-9b48-4328-aa92-0db7158db545


dessalines (approved):
👍

Nutomic:
In the video it looks like your keyboard closes and opens after every digit, seems like a bug.

PR #2185, opened 2023-10-18
Fix bug where up/down votes would discard comment draft

### LINKED ISSUE #1968: Voting on a comment discards the comment you are typing
### Requirements

### Summary

Voting on posts closes the comment form, losing whatever you've typed.

### Steps to Reproduce

1. Start typing a reply to a comment. Don't submit it yet.
2. Upvote a comment while your reply is insubmitted. It doesn't seem to matter if I upvote the comment I was reply to or a different one.
3. The reply I was typing is now lost. The text area was discarded and reopening the reply form has the text area open again, forcing me to rewrite my entire comment.

### Technical Details

Noticed this in mobile Chrome (114.0.5735.196) on lemm.ee with Android 13, Samsung s23 ultra. Haven't checked if anything else is affected.

### Lemmy Instance Version

0.18.2 (backend, couldn't tell where it says the front-end version)

### Lemmy Instance URL

lemme.ee

SleeplessOne1917:
Fixes #1968

While this fixes the issue, I believe further refactoring on this would be good. @dessalines I know that you said that pulling out the API calls to higher level components was a design pattern, but in this case there was a lot of state being handled in higher level components that didn't need to know that state. This made the components leaky so to speak. I believe it was the attempt at managing this leakiness that introduced the bug.

Currently on main, when a POST request that changes the state of a given post or comment returns successfully to the browser, it gets handles by using the `editListImmutable` function with the existing list of posts/comments in on the page and the newly returned post/comment from the update. It returns the existing list, but with the item that got updated replaced with a completely different object. This forces a rerender on the entire comment instead of just the part that got changed. That's my suspicion at least.


dessalines (commented):
I'm gonna have to look at this, because fixing this bug shouldn't need to change 20 files, and move API calls into components.

dessalines:
Okay I figured it out. Was actually a really simple issue having to do with vote / save / non-create or edit actions setting the `finished` map. 

https://github.com/LemmyNet/lemmy-ui/pull/2189

PR #2190, opened 2023-10-19
use improved notification system for all notification types

biosfood:
## Description
Refresh the notification counts when marking a report or application as done as well.

Also prevent unneccesary refetching of unchanged notification counts (don't need to refetch reports when marking a PM as read)

### DIFF
--- a/src/shared/components/person/inbox.tsx
+++ b/src/shared/components/person/inbox.tsx
@@ -795,7 +795,7 @@ export class Inbox extends Component<any, InboxState> {
         limit,
       }),
     });
-    UnreadCounterService.Instance.updatePrivateMessages();
+    UnreadCounterService.Instance.updateInboxCounts();
   }
 
   async handleSortChange(val: CommentSortType) {
--- a/src/shared/components/person/inbox.tsx
+++ b/src/shared/components/person/inbox.tsx
@@ -795,7 +795,7 @@ export class Inbox extends Component<any, InboxState> {
         limit,
       }),
     });
-    UnreadCounterService.Instance.update();
+    UnreadCounterService.Instance.updatePrivateMessages();
   }
 
   async handleSortChange(val: CommentSortType) {
--- a/src/shared/components/person/reports.tsx
+++ b/src/shared/components/person/reports.tsx
@@ -46,6 +46,7 @@ import { Spinner } from "../common/icon";
 import { Paginator } from "../common/paginator";
 import { PostReport } from "../post/post-report";
 import { PrivateMessageReport } from "../private_message/private-message-report";
+import { UnreadCounterService } from "../../services";
 
 enum UnreadOrAll {
   Unread,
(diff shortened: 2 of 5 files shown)


src/shared/services/UnreadCounterService.ts
```diff
@@ -32,27 +32,36 @@ export class UnreadCounterService {
...
-      this.unreadPrivateMessages = unreadCountRes.data.private_messages;
-      this.unreadReplies = unreadCountRes.data.replies;
-      this.unreadMentions = unreadCountRes.data.mentions;
-      this.unreadInboxCountSubject.next(
-        this.unreadPrivateMessages + this.unreadReplies + this.unreadMentions,
-      );
+    return true;
+  }
+
+  public updatePrivateMessages = async () => {
```
dessalines, line 49: Change all these to async functions rather than closures if you would.

src/shared/components/person/inbox.tsx
```diff
@@ -795,7 +795,7 @@ export class Inbox extends Component<any, InboxState> {
         limit,
       }),
     });
-    UnreadCounterService.Instance.update();
+    UnreadCounterService.Instance.updatePrivateMessages();
```
dessalines, line 798: Why only update private messages here?

src/shared/components/person/reports.tsx
```diff
@@ -610,16 +611,28 @@ export class Reports extends Component<any, ReportsState> {
   async handleResolveCommentReport(form: ResolveCommentReport) {
     const res = await HttpService.client.resolveCommentReport(form);
     this.findAndUpdateCommentReport(res);
+    if (this.state.unreadOrAll === UnreadOrAll.Unread) {
+      this.refetch();
```
dessalines, line 615: Wouldn't this need to await?

src/shared/components/person/reports.tsx
```diff
@@ -610,16 +611,28 @@ export class Reports extends Component<any, ReportsState> {
...
+      UnreadCounterService.Instance.updateReports();
+    }
   }
 
   async handleResolvePostReport(form: ResolvePostReport) {
     const res = await HttpService.client.resolvePostReport(form);
     this.findAndUpdatePostReport(res);
+    if (this.state.unreadOrAll === UnreadOrAll.Unread) {
+      this.refetch();
```
dessalines, line 624: Same

src/shared/components/person/reports.tsx
biosfood: I see no problem in refetching and rendering the reports at the same time as the unread counts are refreshed. The resolveComment call has already awaited so the data will be fresh for both calls and the UnreadCounterService will sooner be refreshed and doesn't need to wait for the rest of the UI to update

src/shared/components/person/inbox.tsx
biosfood: Only "private messages" (the counts of which are returned by the UnreadCounts API call) are displayed in the inbox so I can be certain that the function wasn't called because for instance a report was dismissed. Might be a bad name but the "updatePrivateMessages" function will update the counts for PMs, mentions and replies for the number next to the inbox button.

Would "updateInboxCounts" (doing the same thing) be better?

dessalines: Yes please change that name, private messages are only one of the three types of unreads for normal users.

Maybe `updatePersonUnreads`

src/shared/components/person/reports.tsx
dessalines: Ah okay.

src/shared/services/UnreadCounterService.ts
```diff
@@ -75,6 +87,12 @@ export class UnreadCounterService {
         );
       }
     }
+  }
+
+  public updateAll = async () => {
```
dessalines, line 92: This one also.

biosfood: This still needs to be a closure so it can be properly used in `poll(this.updateAll, updateUnreadCountsInterval);`.

SleeplessOne1917: You could bind it in the constructor.

dessalines: Ah okay. That's fine then.

dessalines (approved):

SleeplessOne1917 (approved):

PR #2199, opened 2023-10-23
Make 2FA input play nicer with phone keyboards

SleeplessOne1917:
While the 2FA modal doesn't seem as fancy with 1 text input as opposed to 6 small text inputs, this should prevent the janky behavior with phone keyboards.


dessalines (approved):
👍

PR #2200, opened 2023-10-25
Fixed markdown parsing error particularly with the spoiler tag.

jgrim:
## Description
The sidebar will fail to render when a spoiler tag is used. This fixes the markdown parsing to no longer crash the front-end.


SleeplessOne1917:
Thanks!

PR #2208, opened 2023-10-31
Use `auth` cookie set by backend instead of `jwt` (fixes #2193)

### LINKED ISSUE #2193: [Bug] Logging in displays: auth_cookie_insecure
### Requirements

### Summary

In Lemmy 0.19.0-beta.8 when you sign in for the first time, it works fine. But as soon as you logout and try to signin for a second time, it displays the message:

`auth_cookie_insecure`

This error does not go away until you clear your cache/cookies. 

As a workaround I always need to use an incognito tab now.

### Steps to Reproduce

1. Sign in
2. Sign out
3. Try to sign in
4. auth_cookie_insecure error

### Technical Details

Samsung Internet browser on Android. But also with Edge on Win11.

### Lemmy Instance Version

0.19.0-beta.8

### Lemmy Instance URL

_No response_

Nutomic:
I think #2193 happens because lemmy-ui still uses the `jwt` cookie while the backend automatically sets `auth` on login. So by switching lemmy-ui to the same cookie it should prevent the login error. Unfortunately I cant reproduce the problem locally so Im not certain.

Requires https://github.com/LemmyNet/lemmy-js-client/pull/208


dessalines:
You can use `yalc publish --push` to test lemmy-js-client changes locally, but I'll test this now.

Nutomic:
Im not setting any path, seems like the library or browser is doing it automatically. Should I change it to a different one?

Btw if there is an existing `jwt` cookie, lemmy-ui should change that to `auth` (with secure and httponly) so that users dont get logged out when their instance is upgraded. But not sure how to implement that.

dessalines:
I've got this accounted for in #2210 

Unfortunately we still have to use the manual set method for various reasons:

- The cookie from the server is a short-term session one, so we'd have to manually overwrite it anyway
- Tons of domain name problems with development if we try to read or set that cookie from the response headers, rather than the LoginResponse JSON.

Nutomic:
We could also change the cookie that the backend sets.

dessalines:
I think session cookies are intened to expire right after you close that tab / app. But its probably better for clients / apps to decide their own expirations anyway, and if we do have any expiration, have it built into the JWT.

Nutomic:
I didnt know that there are so many different options for cookies. Should we get rid of the code that sets cookies in backend then?

dessalines:
Tough call... I'd bring it up in one of the dev matrix chats. I do think it'd be much simpler tho to only support the Authorization Bearer header.

Nutomic:
I would still read the cookie for request auth besides the header, but only if it was explicitly set by the client.

PR #2220, opened 2023-11-14
Fix back button from posts going back twice instead of once

SleeplessOne1917:
## Description
Does as the title says. I'm unsure if there's already an issue that exists for this.


dessalines (approved):
Works much better, thx.

PR #2223, opened 2023-11-15
Settings Import/export

SleeplessOne1917:
## Description
Adds UI for the import and export settings feature.

## Screenshots
Screenshot_20231115_143442

### Loading state:

https://github.com/LemmyNet/lemmy-ui/assets/28871516/2fcdee1a-b1a1-48ad-845e-07036b2ca9cf

### DIFF
--- a/src/shared/components/person/settings.tsx
+++ b/src/shared/components/person/settings.tsx
@@ -257,6 +257,7 @@ export class Settings extends Component<any, SettingsState> {
           show_read_posts,
           send_notifications_to_email,
           email,
+          open_links_in_new_tab,
         },
         person: {
           avatar,
@@ -294,6 +295,7 @@ export class Settings extends Component<any, SettingsState> {
           bio,
           send_notifications_to_email,
           matrix_user_id,
+          open_links_in_new_tab,
         },
       };
     }
@@ -641,6 +643,7 @@ export class Settings extends Component<any, SettingsState> {
                 className="btn btn-secondary w-100 mt-3"
                 onClick={linkEvent(this, this.handleImportSettings)}
                 type="button"
+                disabled={!this.state.settingsFile}
               >
                 Import
               </button>
@@ -648,11 +651,13 @@ export class Settings extends Component<any, SettingsState> {
           </>
         ) : (
           <div>
-            {this.state.importOrExportLoading === "exporting"
-              ? "Exporting"
-              : "Importing"}
-            <LoadingEllipses />
-            <Spinner />
+            <div className="text-center">
+              {this.state.importOrExportLoading === "exporting"
+                ? "Exporting"
+                : "Importing"}
+              <LoadingEllipses />
+            </div>
+            <Spinner large />
           </div>
         )}
       </>
@@ -1160,6 +1165,7 @@ export class Settings extends Component<any, SettingsState> {
       this.setState({ show2faModal: false });
 
       const siteRes = await HttpService.client.getSite();
+
       UserService.Instance.myUserInfo!.local_user_view.local_user.totp_2fa_enabled =
         enabled;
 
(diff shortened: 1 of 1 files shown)


src/shared/components/person/settings.tsx
```diff
@@ -332,8 +340,18 @@ export class Settings extends Component<any, SettingsState> {
   }
 
   render() {
+    /* eslint-disable jsx-a11y/anchor-has-content, jsx-a11y/anchor-is-valid */
     return (
       <div className="person-settings container-lg">
+        <a
+          ref={this.exportSettingsLink}
+          download={`lemmy-settings-${UserService.Instance.myUserInfo?.local_user_view.person.name}.json`}
```
SleeplessOne1917, line 348: Please suggest a better default file name if you have one.

src/shared/components/person/settings.tsx
```diff
@@ -595,6 +617,53 @@ export class Settings extends Component<any, SettingsState> {
     );
   }
 
+  importExport() {
+    return (
+      <>
+        <h2 className="h5">Import/Export</h2>
+        <p>Import and export your account settings as JSON.</p>
```
dessalines, line 624: i18n all these strings.

src/shared/components/person/settings.tsx
dessalines: Looking at another backup program, I'd say do something like:

`lemmy_user_settings.${(new Date()).toISOString().replace('/:/g','').replace('/-/g','')}.json`

So it'll look like `lemmy_user_settings.20231119T000Z.json`

Removing colons and dashes are good to make iso8601 date strings file-system safe everywhere. Might be good to extract that to its own function too.

src/shared/components/person/settings.tsx
```diff
@@ -332,8 +340,18 @@ export class Settings extends Component<any, SettingsState> {
...
 
   render() {
+    /* eslint-disable jsx-a11y/anchor-has-content, jsx-a11y/anchor-is-valid */
     return (
       <div className="person-settings container-lg">
+        <a
+          ref={this.exportSettingsLink}
+          download={`lemmy-settings-${UserService.Instance.myUserInfo?.local_user_view.person.name}.json`}
+          style={{
+            display: "none",
```
dessalines, line 350: boostrap class `d-none`

src/shared/components/person/settings.tsx
```diff
@@ -1474,6 +1544,120 @@ export class Settings extends Component<any, SettingsState> {
...
+
+  async handleImportSettings(i: Settings) {
+    i.setState({ importOrExportLoading: "importing" });
+
+    const res = await HttpService.client.importSettings(
+      JSON.parse(await i.state.settingsFile!.text()),
+    );
+
+    if (res.state === "success") {
+      toast("Settings successfully imported!", "success");
```
dessalines, line 1580: Its probably a good idea to just reload the page a few seconds after the successful import, rather than doing this below. Its a pretty rare operation anyway. If this works below tho, its pry fine.

src/shared/components/person/settings.tsx
```diff
@@ -119,6 +120,8 @@ interface SettingsState {
   searchInstanceOptions: Choice[];
   isIsomorphic: boolean;
   show2faModal: boolean;
+  importOrExportLoading: "importing" | "exporting" | null;
```
dessalines, line 123: Why not have this as an `importAccountRes` and `exportAccountRes` like the other calls above, which already has a loading state. Seems more consistent.

src/shared/components/person/settings.tsx
SleeplessOne1917: Of course. Just want to make sure the text works in context before creating a PR for the translations.

src/shared/components/person/settings.tsx
SleeplessOne1917: Using a response object is somewhat awkward here as the import and export responses don't have types in the js client.

dessalines: `RequestState<any>` still has the loading and success states, no need to make custom ones then.

dessalines (approved):
Looks good, thx.

PR #2228, opened 2023-11-22
Adding creator_is_admin and creator_is_mod, removing pointless functions

dessalines:
## Description

Adds the new `creator_is_moderator` and `creator_is_admin` properties. Removes some pointless functions.

One other unfortunate thing: I still need to fill the list of moderators and admins into each post and comment listing. This is because the `canMod` functions still need the hierarchies of admins then mods, to make sure you can't do actions on mods / admins above you. Just knowing the creator is a moderator is unfortunately not enough info to know if you can do actions on them.

### DIFF
--- a/src/shared/utils/roles/am-mod.ts
+++ b/src/shared/utils/roles/am-mod.ts
@@ -1,10 +1,11 @@
-import { isMod } from "@utils/roles";
-import { CommunityModeratorView } from "lemmy-js-client";
+import { CommunityId } from "lemmy-js-client";
 import { UserService } from "../../services";
 
 export default function amMod(
-  mods?: CommunityModeratorView[],
+  communityId: CommunityId,
   myUserInfo = UserService.Instance.myUserInfo,
 ): boolean {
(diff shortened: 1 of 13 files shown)


src/shared/utils/roles/am-mod.ts
```diff
@@ -1,10 +1,11 @@
...
 import { UserService } from "../../services";
 
 export default function amMod(
-  mods?: CommunityModeratorView[],
+  communityId: CommunityId,
   myUserInfo = UserService.Instance.myUserInfo,
 ): boolean {
-  return myUserInfo ? isMod(myUserInfo.local_user_view.person.id, mods) : false;
+  return myUserInfo
+    ? myUserInfo.moderates.map(cmv => cmv.community.id).includes(communityId)
```
dessalines, line 9: Another change is here. Use the communities you moderate from your `myUserInfo`, rather than a list of mods from the `GetCommunityResponse`

dessalines:
@SleeplessOne1917 I know you're working on a refactor. I'll handle any conflicts with main, or just remove my changes here.

Issue #2106, opened 2023-09-01
ARM builds for Docker


kroese:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

The `lemmy` docker images finally support ARM cpu's since recently, but that is kinda useless as long as the `lemmy-ui` images do not.

So my request is for `lemmy-ui` to support ARM also.

Issue #1566, opened 2023-06-24
lemmy-ui (0.18.0) docker image fails to run on CPUs not supporting SSE4


thomasdouwes:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

Trying to run the docker lemmy-ui 0.18.0 on a CPU not supporting the SSE4 instruction set such as the intel atom N2800 will fail with Illegal instruction.  
From outside this appears as the container continuously restarting and only printing `Inferno is in development mode.`

### Steps to Reproduce

1. run lemmy-ui:0.18.0 docker container with /bin/sh as entrypoint
2. run node dist/js/server.js
3. node exits with Illegal instruction

### Technical Details

OS: debian 11 (bullseye)
CPU: intel atom N2800
Server: K1 kimsufi from OVH  

The exception is in the function `aom_init_plugin()` of `/app/node_modules/sharp/build/Release/../.././vendor/8.14.2/linuxmusl-x64/lib/libvips-cpp.so.42` and is at `0x00007fa846b7e343 <+339>:   pinsrq $0x1,%rax,%xmm0` from gdb. although I don't think that is too useful.  

I don't really know where this issue should go, It's not caused directly by lemmy-ui but possibly changing the Dockerfile might be the only way to fix this. Or if older CPUs are unsupported then feel free to close this issue.

### Lemmy Instance Version

0.18.0

mikeesto:
This error stems from the sharp library and its dependency on libvips. You would need to use a custom version of libvips to get it to run on that CPU: https://sharp.pixelplumbing.com/install#custom-libvips

No change to the Dockerfile is going to help here

dessalines:
We've had to disable arm builds for lemmy-ui because of a libvips issue issue: https://woodpecker.join-lemmy.org/repos/135/pipeline/8497/3

PR #2235, opened 2023-11-26
Multi-platform Dockerfile

kroese:
## Description

The backend supports ARM now, but that is not very useful as long as the frontend does not.

So this PR lets the frontend Dockerfile be build for ARM64 (and also x86 CPU's missing the SSE4 instruction set).

It is based on the work of [ubergeek77](https://github.com/ubergeek77/), all credits to him.

Fixes: https://github.com/LemmyNet/lemmy-ui/issues/1566
Fixes: https://github.com/LemmyNet/lemmy-ui/issues/2106

It also adds the propers labels, adds a healthcheck to monitor the container, and switches Inferno to production-mode which should increase performance.


Dockerfile
```diff
@@ -1,16 +1,41 @@
+FROM golang as go-builder
```
SleeplessOne1917, line 1: What is golang needed for?

Dockerfile
```diff
@@ -1,16 +1,41 @@
...
 
-ENV npm_config_target_arch=x64
 ENV npm_config_target_platform=linux
 ENV npm_config_target_libc=musl
 
 # Cache deps
 COPY package.json yarn.lock ./
-RUN yarn --production --prefer-offline --pure-lockfile
+
+RUN sed -i 's|"sharp": "^0.32.4"|"sharp": "^0.32.6"|g' package.json
```
SleeplessOne1917, line 36: Should the repo just update to a newer version of `sharp`?

SleeplessOne1917: Made a PR for the dependency upgrade.

dessalines:
Actually I can't because for some reason edits by maintainers aren't being allowed.

dessalines:
Yeah, I think it might just be a checkmark somewhere on the pull request specifically. https://stackoverflow.com/questions/63341296/github-pull-request-allow-edits-by-maintainers#63341866

dessalines:
Question, node-prune the only thing that requires go correct? Because if that's the case I'm in favor of just removing it, since development is dead on it. I'm sure there are better ways to minimize node docker images now.

dessalines:
I fixed some issues. Another build started.

dessalines (approved):

dessalines:
I increased the timeout from 1h to 2h and restarted it. Not sure why these ones take so long, not even the CI for the lemmy back-end docker builds are this slow.

dessalines:
It passed : https://hub.docker.com/layers/dessalines/lemmy-ui/0.19.0-alpha.2/images/sha256-200e55a928dce5b9460f674d584b1124a761d2e700beed8c11d20b7495af76ee?context=explore

PR #2239, opened 2023-11-27
Fix shutdown behaviour

kroese:
## Description

The current code does not handle the stop signal Docker sends, which causes it to always get a SIGKILL from the Docker daemon after a 10 second timeout. 

The reason for this is that Docker sends a SIGTERM signal first, but the code only listens for SIGINT (CTRL+C) and because of that it does not initiate any shutdown procedure.

To fix this, I added a handler for SIGTERM so that the container stops gracefully and much quicker.

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -55,7 +55,9 @@ server.get("/*", CatchAllHandler);
 
 const listener = server.listen(Number(port), hostname, () => {
   setupDateFns();
-  console.log(`Lemmy-ui v${VERSION} started listening on http://${hostname}:${port}`);
+  console.log(
+    `Lemmy-ui v${VERSION} started listening on http://${hostname}:${port}`,
+  );
 });
 
 const signals = {


src/server/index.tsx
```diff
@@ -52,12 +52,34 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
 server.get("/css/themelist", ThemesListHandler);
 server.get("/*", CatchAllHandler);
 
-server.listen(Number(port), hostname, () => {
+var listener = server.listen(Number(port), hostname, () => {
```
SleeplessOne1917, line 55: Use `const`.

src/server/index.tsx
```diff
@@ -52,12 +52,34 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
   setupDateFns();
-  console.log(`http://${hostname}:${port}`);
+  console.log(`Started listening on http://${hostname}:${port}`);
 });
 
-process.on("SIGINT", () => {
-  console.info("Interrupted");
-  process.exit(0);
+var signals = {
```
SleeplessOne1917, line 60: Use `const`

src/server/index.tsx
```diff
@@ -52,12 +52,34 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
+  'SIGTERM': 15
+};
+
+const shutdown = (signal, value) => {
+  // TODO: Should set a flag here for the listener to reject any further
+  // incoming connections with a HTTP 503 error while shutting down.
+  // Otherwise the connection count may not drop to 0 before timeout.
+  listener.close(() => {
+    console.log(`Lemmy stopped by ${signal} with value ${value}`);
+    process.exit(128 + value);
```
SleeplessOne1917, line 72: What does `128 + value` mean?

src/server/index.tsx
```diff
@@ -52,12 +52,34 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
+   console.error(`Could not close all connections in time, forcing shutdown because of ${signal}...`);
+   process.exit(128 + value);
+  }, 9*1000);
+};
+
+Object.keys(signals).forEach((signal) => {
+  process.on(signal, () => {
+    console.log(`Process received a ${signal} signal`);
+    shutdown(signal, signals[signal]);
+  });
```
SleeplessOne1917, line 84: I think it would be slightly cleaner as the following:
```suggestion
for (const [signal, value] of Object.entries(signals)) {
  process.on(signal, () => {
    console.log(`Process received a ${signal} signal`);
    shutdown(signal, value);
  });
}
```

src/server/index.tsx
dessalines: Would be good to make that a const, and add a comment for it.

src/server/index.tsx
```diff
@@ -52,12 +52,34 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
+  // incoming connections with a HTTP 503 error while shutting down.
+  // Otherwise the connection count may not drop to 0 before timeout.
+  listener.close(() => {
+    console.log(`Lemmy stopped by ${signal} with value ${value}`);
+    process.exit(128 + value);
+  });
+  setTimeout( function () {
+   console.error(`Could not close all connections in time, forcing shutdown because of ${signal}...`);
+   process.exit(128 + value);
+  }, 9*1000);
```
dessalines, line 77: Might as well make this timeout a constant also.

src/server/index.tsx
```diff
@@ -52,12 +52,34 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
+
+const shutdown = (signal, value) => {
+  // TODO: Should set a flag here for the listener to reject any further
+  // incoming connections with a HTTP 503 error while shutting down.
+  // Otherwise the connection count may not drop to 0 before timeout.
+  listener.close(() => {
+    console.log(`Lemmy stopped by ${signal} with value ${value}`);
+    process.exit(128 + value);
+  });
+  setTimeout( function () {
```
dessalines, line 74: You can just do `() =>` instead of function.

dessalines:
Looks like you might need to manually run prettier, since husky didn't set up the git hooks correctly for you. `yarn prettier -w src/...`

src/server/index.tsx
```diff
@@ -52,12 +52,39 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
+
+const shutdown = (signal, value) => {
+  // TODO: Should set a flag here for the listener to reject any further
+  // incoming connections with a HTTP 503 error while shutting down.
+  // Otherwise the connection count may not reach zero before timeout.
+  listener.close(() => {
+    console.log(`Lemmy stopped by ${signal} with value ${value}`);
+    process.exit(exit_signal + value);
+  });
+  setTimeout(function () {
```
dessalines, line 77: This one still needs fixing.

src/server/index.tsx
```diff
@@ -52,12 +52,39 @@ server.get("/css/code-themes/:name", CodeThemeHandler);
...
-  process.exit(0);
-});
+const signals = {
+  SIGHUP: 1,
+  SIGINT: 2,
+  SIGTERM: 15,
+};
+
+const exit_signal = 128; // Fatal error signal code on Linux systems
+const exit_timeout = 8; // Because Docker SIGTERMs after 10 secs
```
dessalines, line 67: Might as well just make this 8000

SleeplessOne1917 (approved):

Issue #2243, opened 2023-11-28
SSR Broken for most pages


dessalines:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

Pages such as inbox, reports, are all broken on first load.

Votes appear to be missing when logged in, and reloading the post page, indicating some issue with SSR.

### Steps to Reproduce

1. Vote on a post and some comments in that post
2. Reload post page
3. Votes appear missing

EDIT: oof, this is gonna be a bad one. I can't reproduce it either locally with the lemmy.ml DB backup, or on voyager.

### Technical Details

NA

### Lemmy Instance Version

main

Nutomic:
https://github.com/LemmyNet/lemmy-ui/issues/2244 also seems to be caused by the same issue. In my case its actually showing posts from blocked communities under subscribed after first load.

dessalines:
I put `rc.9` on lemmy.ml , and it didn't work. Half the initial page loads ignore / don't use auth. @SleeplessOne1917 I'll need your help with this one, its a critical bug we've got to fix before releasing.

Nutomic:
It does send `Cookie:jwt=xyz` for the initial load, but `authorization` header is completely missing.

dessalines:
[This line](https://github.com/LemmyNet/lemmy-ui/blob/main/src/server/handlers/catch-all-handler.tsx#L40) does set the authorization, and I've tested it locally as working (but it could have some issues).

The other issue is the mis-named cookie, which could be renamed to `auth`... I'd just thought we didn't need to use that because we prefer the `authorization` header.

dessalines:
Nope, its only sending the mis-named cookie, not the authorization header.

![Screenshot_20231129_080312_Kiwi Browser_1](https://github.com/LemmyNet/lemmy-ui/assets/930215/0c279dc1-e133-4059-8c1b-f1be24ef48c0)

SleeplessOne1917:
[...]

PR #2247, opened 2023-11-28
Attempting to fix SSR. #2243

dessalines:
## Description

Trying two different things to fix the SSR issue. But since I can't actually replicate the bug locally, these are just guesses.

- Removing `cross-fetch`. This seems to be abandoned, and everything appears to be working without it. I've also had some issues with it elsewhere.
- Fixing a minor issue with setting the headers.

### DIFF
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -2,7 +2,6 @@ import { initializeSite, isAuthPath } from "@utils/app";
 import { getHttpBaseInternal } from "@utils/env";
 import { ErrorPageData } from "@utils/types";
 import * as cookie from "cookie";
-import fetch from "cross-fetch";
 import type { Request, Response } from "express";
 import { StaticRouter, matchPath } from "inferno-router";
 import { renderToString } from "inferno-server";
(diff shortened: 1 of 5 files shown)


src/server/handlers/catch-all-handler.tsx
```diff
@@ -29,15 +28,16 @@ export default async (req: Request, res: Response) => {
...
-      new LemmyHttp(getHttpBaseInternal(), { fetchFunction: fetch, headers }),
+      new LemmyHttp(getHttpBaseInternal(), { headers }),
     );
 
     const auth = req.headers.cookie
       ? cookie.parse(req.headers.cookie).jwt
       : undefined;
 
     if (auth) {
-      client.setHeaders({ Authorization: `Bearer ${auth}` });
```
dessalines, line 40: This probably overwrote the setForwardedHeaders above.

SleeplessOne1917 (approved):

SleeplessOne1917:
Regarding cross-fetch: newer versions of node (I think 18 and above) include a fetch API that works the same as the browser one.

dessalines:
Sweet. I'll start another lemmy-ui deploy for testing.

PR #2258, opened 2023-11-30
Moderation/content action overhaul

SleeplessOne1917:
## Description
This started out as a simple bug fix and turned into a full scale refactor and UI overhaul because the existing UI for this stuff was pretty janky. The screenshots should show what I'm going for.

**Note**: Some existing icons were deleted to use pairs of icons that have the same art style.

### Mod action dropdown - comment
image

### Mod action dropdown - post
image

### Ban dialog
image

### Report dialog
image
The remove dropdown looks the same.

### Purge dialog
image

### Transfer community dialog
image
The appoint as mod and appoint as admin are basically the same.

TODO:
- Get consensus on the verbiage. Waiting until that's settled to update the translations repo.
- Have at least one other dev test locally since this was a large refactor.


src/shared/components/common/mod-action-form-modal.tsx
```diff
@@ -0,0 +1,412 @@
...
+              </button>
+            </footer>
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  handleShow() {
+    console.log("In handle show");
```
dessalines, line 402: Can remove.

src/shared/components/common/mod-action-form-modal.tsx
```diff
@@ -0,0 +1,412 @@
...
+          this.props.isBanned ? "unban" : "ban",
+        );
+        isBanned = this.props.isBanned;
+        break;
+      }
+
+      case "community-ban": {
+        headerText = `${I18NextService.i18n.t(
+          this.props.isBanned ? "unban" : "ban",
+        )} ${getApubName(this.props.creator)} from ${getApubName(
```
dessalines, line 204: from ... this a `<T` style i18n

dessalines (commented):
I'm testing the functionality rn, but you might as well start on i18n, because there's a lot of it. We can always tweak those strings later.

src/shared/components/comment/comment-node.tsx
```diff
@@ -189,71 +155,30 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     if (!deepEqual(this.props, nextProps)) {
       this.setState({
         showEdit: false,
-        showRemoveDialog: false,
-        showBanDialog: false,
-        removeData: false,
-        banType: BanType.Community,
-        showPurgeDialog: false,
-        purgeType: PurgeType.Person,
+        collapsed: false,
```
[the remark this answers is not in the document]
SleeplessOne1917: It's still being used in 12 places and seems to be used for conditionally rendering parts of the comment. Did I make a mistake with the merge from main?

SleeplessOne1917:
@dessalines  Regarding speak as moderator, I got [this comment](https://github.com/LemmyNet/lemmy/issues/4216#issuecomment-1834526001) on a PR I opened for the backend. It seems people think it's a feature and not a bug. It may be worth keeping the current functionality and just changing the verbiage in the dropdown.

cc @Nutomic

src/shared/components/comment/comment-node.tsx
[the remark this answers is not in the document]
dessalines: I just mean it shouldn't set collapsed to false on `componentWillReceiveProps`. The variable stays.

dessalines:
@nutomic got that fixed in https://github.com/LemmyNet/lemmy/pull/4222 (IE, only letting you distinguish your own comment).

asimons04:
> Speak as moderator
> Bug: Its shown for every comment. It should only be shown for your own comment, if you are an admin / mod

Only Lemmy-UI treats it as "speak as moderator".  Last I checked, that was redundant since mods already have badges in LUI.

In Tesseract, and maybe other front ends, the distinguish flag is used to feature/sticky comments from anyone.  Why is that a problem?

src/shared/components/comment/comment-node.tsx
```diff
@@ -585,14 +591,14 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
-  handleSaveComment() {
+  async handleSaveComment() {
     this.props.onSaveComment({
       comment_id: this.commentView.comment.id,
       save: !this.commentView.saved,
     });
   }
 
-  handleBlockPerson() {
+  async handleBlockPerson() {
```
dessalines, line 601: Whats the point of making these async, they should be just pushing forms to the top level pages.

SleeplessOne1917: I made them to make sure control of the form has the correct state when the browser hasn't received from the request yet. Before I did this, you had async functions being called from synchronous functions, which meant that the loading would be set to false before the request was done.

SleeplessOne1917:
I addressed everything that needed to be addressed. The only thing that I can't do much about at the moment are the purge issues you pointed out. I ran into issues with purges, but they didn't break the page like you described. From what I've done:

- Purge post: works as expected
- Purges the comment and not its children. Unsure if this behavior is intentional.
- Purge user: definitely buggy. Posts and comments from the user don't get removed, although you will get an error if you navigate to the user's page

src/shared/components/common/mod-action-form-modal.tsx
```diff
@@ -0,0 +1,462 @@
...
+                    {showExpiresField && (
+                      <div className="col-12 col-lg-6 col-xl-5">
+                        <label className="visually-hidden" htmlFor={expiresId}>
+                          {I18NextService.i18n.t("expires")}
+                        </label>
+                        <input
+                          type="number"
+                          id={expiresId}
+                          className="form-control my-2 my-lg-0"
+                          placeholder="Days until expiration"
```
dessalines, line 257: i18n

src/shared/components/common/mod-action-form-modal.tsx
```diff
@@ -0,0 +1,462 @@
...
+                            className="form-check-label"
+                            title={I18NextService.i18n.t("remove_content_more")}
+                          >
+                            <input
+                              className="form-check-input"
+                              type="checkbox"
+                              onChange={linkEvent(this, handleTogglePermaBan)}
+                              checked={shouldPermaBan}
+                            />
+                            Permanently ban
```
dessalines, line 294: i18n

src/shared/components/post/post.tsx
```diff
@@ -867,6 +896,9 @@ export class Post extends Component<any, PostState> {
   async handleFeaturePost(form: FeaturePost) {
     const featureRes = await HttpService.client.featurePost(form);
     this.updatePost(featureRes);
+    if (featureRes.state === "success") {
+      toast(form.featured ? "Featured post" : "Unfeatured post");
```
dessalines, line 900: i18n

dessalines (commented):
Besides the missing i18n below:

- After any modal is shown then dismissed (such as remove / restore a comment) it breaks scrolling, and I have to do a page refresh.

Everything else seems to be working well tho.

PR #2259, opened 2023-11-30
Revert "Ignore autogenerated css files from being tracked in the repo"

dessalines:
Another premature / untested merge. These css files need to be in the repo, as they aren't auto-generated at all.

Reverts LemmyNet/lemmy-ui#2256

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ package-lock.json
 *.orig
 
 src/shared/translations
-src/assets/css/themes/*.css
 
 stats.json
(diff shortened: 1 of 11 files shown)


SleeplessOne1917:
I thought they were generated by running the `themes:build` script that is listed in the package.json?

dessalines:
Unfortunately not, the auto-theme generation still needs to be worked out. tbh this is one I'd rather not do, and just leave that to the new tailwind app.

PR #2268, opened 2023-12-04
Added aspect ratio to the videos

ismailkarsli:
## Description

Vertical videos are sometimes 2 times bigger than window height. This is very disturbing in terms of UX. I added 16x9 ratio, if there is no other overlooked reason, can you peep accept it?

### Before
Screenshot 2023-12-05 at 00 23 10

### After
Screenshot 2023-12-05 at 00 23 19

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -271,7 +271,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     // if direct video link
     if (url && isVideo(url)) {
       return (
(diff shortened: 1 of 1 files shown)


dessalines (approved):
https://getbootstrap.com/docs/5.0/helpers/ratio/

Seems fine as a kludge for now.

dessalines:
> When watching videos in sequence, they all start muted, even we set a volume in previous video.
Do you think it makes sense to store the previous volume value in globalThis and use that value while in the same session?

I spose that'd be fine.

PR #2272, opened 2023-12-05
Only generate manifest if it isn't already in memory

SleeplessOne1917:
This should hopefully address #2234

### DIFF
--- a/src/server/handlers/manifest-handler.ts
+++ b/src/server/handlers/manifest-handler.ts
@@ -1,4 +1,4 @@
-import { getHttpBaseExternal, getHttpBaseInternal } from "@utils/env";
+import { getHttpBaseInternal } from "@utils/env";
 import type { Request, Response } from "express";
 import { LemmyHttp } from "lemmy-js-client";
 import { wrapClient } from "../../shared/services/HttpService";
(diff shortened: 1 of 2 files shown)


src/server/handlers/manifest-handler.ts
```diff
@@ -9,15 +9,15 @@ let manifest: Awaited<ReturnType<typeof generateManifestJson>> | undefined =
   undefined;
 
 export default async (req: Request, res: Response) => {
-  if (!manifest || manifest.start_url !== getHttpBaseExternal()) {
```
SleeplessOne1917, line 12: `start_url` is hard-coded to `/`, so this if statement was always true with the right hand side of the or expression. Removing that condition should hopefully mean the manifest gets generated if it's not already stored in memory.

src/server/handlers/manifest-handler.ts
```diff
@@ -9,15 +9,15 @@ let manifest: Awaited<ReturnType<typeof generateManifestJson>> | undefined =
...
+  if (!manifest) {
     const headers = setForwardedHeaders(req.headers);
```
dessalines, line 13: Maybe we should do away with this also? The forwardedHeaders includes auth, and might make this uncacheable.

dessalines (commented):
I'm not sure what makes this get cached in memory, and when it doesn't... But feel free to merge when you think its good.

SleeplessOne1917:
> I'm not sure what makes this get cached in memory, and when it doesn't... But feel free to merge when you think its good.

I'm not sure if "cache" is the right word. It's more that the lifetime of top-level module variables lasts until the nodejs process terminates. [This stack overflow post provides a more in-depth explanation.](https://stackoverflow.com/a/62843303)

src/server/handlers/manifest-handler.ts
SleeplessOne1917: "This" being the `if` statement? The entire function?

PR #2273, opened 2023-12-06
About 1/3rd done with adding redux.

dessalines:
TBD


SleeplessOne1917:
Redux will be helpful for handling global state and eliminating prop drilling, but this seems like a needlessly large undertaking depending on how soon we are planning on transitioning to a different framework.

dessalines:
This looks like a big change, but its not too much. Its only moving the initially fetched data into redux. All other state is untouched.

Also I'm removing the UserService, to avoid those race conditions. That one is much more difficult because of how badly javascript deals with optional function and component params.

PR #2278, opened 2023-12-08
Implement logging in via external identity providers

thepaperpilot:
## Description

Implements https://github.com/LemmyNet/lemmy/issues/2930. I believe https://github.com/LemmyNet/lemmy/issues/489 is a duplicate of that issue, and would also be considered implemented by this. Note that this does NOT make lemmy itself an identity provider, and thus does NOT implement https://github.com/LemmyNet/lemmy/issues/1368.

External auth methods can be added via the admin settings, and then buttons are shown on the login page to use those auth methods instead of "basic" auth (username + password). The implementation supports both OAuth or OIDC auth methods, and can register non-existent users as well (if a new setting is explicitly turned on).

Other frontends that wish to support these external auth methods can use the changes in lemmy-ui as a reference. They'll need to show the buttons to go to the authorization URL with the appropriate redirect URI, and then implement the endpoint at that URI that takes the `auth` cookie and navigates to the redirect URI param it was passed. Optionally, frontends can also implement the new admin settings.

## Future Work

Most of these are not implemented because my understanding is lemmy-ui is getting replaced soon-ish anyways and these are tasks that would take awhile to implement which is probably not worth it imo.

- Make frontends have convenient presets for common identity providers (like Google, Github, Discord, etc.) that hides the well-known fields (i.e. just show client ID and secret).
- PKCE support (more secure version of OAuth)
- If auto-registration is disabled, bring non-existent users to modified version of the signup page where the email is pre-filled and readonly, and the password field is hidden
- Improve error handling/messaging (for example if an external auth method fails to save due to a non-unique client ID)

## Related PRs

- [lemmy](https://github.com/LemmyNet/lemmy/pull/4238)
- [lemmy-ui](https://github.com/LemmyNet/lemmy-ui/pull/2278)
- [lemmy-js-client](https://github.com/LemmyNet/lemmy-js-client/pull/219)

## Screenshots

image

image


Die4Ever:
so if you linked your Google account then Lemmy could skip email verification for new accounts? I've seen lots of people have issues with email verifications for their new accounts, new users get confused, emails get lost or sent to spam, instances have their email improperly setup, etc. This seems like it would be less error prone.

I hope this can get finished and merged in eventually.

dessalines:
Superceded by #2731

PR #2281, opened 2023-12-13
Make sure banned toast has correct user.

dessalines:
## Description

The banned toast was previously using the post creator, which is not necessarily the case.


SleeplessOne1917 (approved):
Good catch!

Issue #111, opened 2020-12-16
Add mod tools to user page


Tafiti:
If I moderate a community or I am a site admin, the /u/member page should allow for moderation functions like `ban` and `appoint as mod`. Currently the tools are tied to posts or comments a member makes.

This will allow for a better overview of that member profile, like duration on the instance and which other communities or posts the member is making. 

`https://instance.tld/u/member` should have mod tools like `ban` and `appoint as mod`.

dessalines:
I don't know if I like it too much for user pages, because it would require more context about which communities you mod, and a drop down for that, since you aren't in that community. 

I think I'm preferring a community settings page that can do adds and bans from there. Doing mod actions on posts and comments is still easier and better than either of these two.

Tafiti:
From a UI experience, clicking on a /u/member profile felt much intuitive than the three dots. I suppose it is a matter of preference than it is about functionality. Community settings page may resolve this over time. I had not considered that. 

I will close this issue.

PR #2295, opened 2023-12-19
Use highlight color for new comments on dark themes that won't burn your eyes out

### LINKED ISSUE #2264: Background color for new comments is too bright on darkly
### Requirements

### Summary

With 0.19 new comments receive a yellow background color. Its not a bad idea to highlight them, but the color is way too bright at least on darkly theme. It should be changed to something which is more consistent with the normal background.

Example:
Screenshot_20231201_155843

### Steps to Reproduce

1. Go to lemmy.ml and sort by [new comments](https://lemmy.ml/?dataType=Post&listingType=Subscribed&sort=NewComments)
2. Click on a post and view comments

### Technical Details

-

### Lemmy Instance Version

0.19.0-rc.11

### Lemmy Instance URL

lemmy.ml

SleeplessOne1917:
## Description

Closes #2264. The color I picked isn't perfect, but it's definitely easier on the eyes than what was there before.

image

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -451,3 +451,7 @@ br.big {
 .totp-link {
   width: fit-content;
 }
+
+[data-bs-theme="dark"] .mark {
+  background: rgba(243, 208, 101, 0.3);
+}


dessalines (approved):

Pseudochron:
I don't think that this color looks much better. Yellow background isn't good for white text and a dark theme.

Instead of changing main.css, it would be better to fix it in each of the themes. That way it won't override a site's custom theme. 

The dark themes do actually each specify a color for `--bs-highlight-bg`  and set `mark, .mark { background-color: var(--bs-highlight-bg);}`. E.g. in darkly it's `#333` and in darkly-pureblack it's `#111`. But in the dark themes the color is defined in the `[data-bs-theme=light] { }` section instead of the `[data-bs-theme=dark] { }` section, so it gets ignored.

SleeplessOne1917:
> The dark themes do actually each specify a color for --bs-highlight-bg and set mark, .mark { background-color: var(--bs-highlight-bg);}.

I originally tried setting --bs-highlight-bg, but bootstrap kept overriding it. I chose the solution in this PR after over an hour of wrestling with bootstrap to no avail.

PR #2307, opened 2024-01-02
Store volume level on window object

ismailkarsli:
## Description

Currently every video starts muted. With this PR, the volume value is kept in the window and used while playing other videos.

#### componentDidMount is formatted

Husky auto formatting line 131 while commiting.

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -128,10 +128,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
 
   componentDidMount(): void {
     if (UserService.Instance.myUserInfo) {
-      const { auto_expand, blur_nsfw } = UserService.Instance.myUserInfo.local_user_view.local_user;
+      const { auto_expand, blur_nsfw } =
+        UserService.Instance.myUserInfo.local_user_view.local_user;
       this.setState({
-        imageExpanded:
-          auto_expand && !(blur_nsfw && this.postView.post.nsfw),
+        imageExpanded: auto_expand && !(blur_nsfw && this.postView.post.nsfw),
       });
     }
   }
@@ -214,7 +214,15 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     if (url && isVideo(url)) {
       return (
         <div className="embed-responsive ratio ratio-16x9 mt-3">
-          <video muted controls className="embed-responsive-item col-12">
+          <video
+            onLoadStart={this.handleVideoLoadStart}
+            onPlay={this.handleVideoLoadStart}
+            onMouseOver={this.handleVideoLoadStart}
+            onFocus={this.handleVideoLoadStart}
+            onVolumeChange={this.handleVideoVolumeChange}
+            controls
+            className="embed-responsive-item col-12"
+          >
             <source src={url} type="video/mp4" />
           </video>
         </div>
@@ -764,6 +772,18 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
 
   handleEditCancel() {
     this.setState({ showEdit: false });
+  }
+
+  handleVideoLoadStart(e) {
+    const video = e.target as HTMLVideoElement;
+    const volume = window.volumeLevel;
+    if (volume) video.volume = volume;
+    else window.volumeLevel = video.volume = 0;
+  }
+
+  handleVideoVolumeChange(e) {
+    const video = e.target as HTMLVideoElement;
+    window.volumeLevel = video.volume;
   }
 
   // The actual editing is done in the receive for post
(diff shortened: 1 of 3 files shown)


alectrocute:
I'd suggest using `window.localStorage.[getItem/setItem]` instead to persist. Let's avoid adding things to the `window` object.

ismailkarsli:
> I'd suggest using `window.localStorage.[getItem/setItem]` instead to persist. Let's avoid adding things to the `window` object.

I specifically did it this way so that the volume value starts muted in each session. However, if you suggest otherwise, I can do it with localStorage (or another global store if available).

ismailkarsli:
Alright I've changed it to work with localStorage.

src/shared/components/post/post-listing.tsx
```diff
@@ -214,7 +214,15 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     if (url && isVideo(url)) {
       return (
         <div className="embed-responsive ratio ratio-16x9 mt-3">
-          <video muted controls className="embed-responsive-item col-12">
+          <video
+            onLoadStart={this.handleVideoLoadStart}
```
SleeplessOne1917, line 218: Please use `linkEvent` for these as described in [inferno's docs](https://www.infernojs.org/docs/guides/event-handling).

src/shared/components/post/post-listing.tsx
```diff
@@ -766,6 +774,21 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
+    if (volume) video.volume = Number(volume);
+    else {
+      video.volume = 0;
+      localStorage.setItem("volume_level", "0");
+    }
+  }
+
+  handleVideoVolumeChange(e) {
+    const video = e.target as HTMLVideoElement;
+    localStorage.setItem("volume_level", String(video.volume));
```
SleeplessOne1917, line 789: Nitpick: can use `toString` inatead.

SleeplessOne1917 (changes requested):
Besides the changes requested from my other comments, please run prettier. It looks like some formatting changed.

ismailkarsli:
> Besides the changes requested from my other comments, please run prettier. It looks like some formatting changed.

I've run prettier (in fact it runs with pre-commit hook anyways) but that part of the code still stays formatted. I guess it was not formatted according to the config before.

image

src/shared/components/post/post-listing.tsx
```diff
@@ -128,10 +128,10 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
 
   componentDidMount(): void {
     if (UserService.Instance.myUserInfo) {
-      const { auto_expand, blur_nsfw } = UserService.Instance.myUserInfo.local_user_view.local_user;
+      const { auto_expand, blur_nsfw } =
```
dessalines, line 131: Hrm, weird how this is getting reformatted. As long as its passing the CI prettier check it should be fine tho.

dessalines (approved):

ismailkarsli:
I've checked it and event.target is undefined when used with `linkEvent`. Lemmy fix that

src/shared/components/post/post-listing.tsx
```diff
@@ -766,6 +774,21 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
+      video.volume = 0;
+      localStorage.setItem("volume_level", "0");
+    }
+  }
+
+  handleVideoVolumeChange(e) {
+    const video = e.target as HTMLVideoElement;
+    localStorage.setItem("volume_level", video.volume.toString());
+  }
+
```
SleeplessOne1917, line 791: These could be defined at the module level instead of as properties of the component.

ismailkarsli: You mean somewhere like `src/shared/utils/media`?

SleeplessOne1917: I mean in the same file they're already in, but in the outermost scope.

src/shared/components/post/post-listing.tsx
```diff
@@ -214,7 +214,15 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     if (url && isVideo(url)) {
       return (
         <div className="embed-responsive ratio ratio-16x9 mt-3">
-          <video muted controls className="embed-responsive-item col-12">
+          <video
+            onLoadStart={linkEvent(this, this.handleVideoLoadStart)}
+            onPlay={linkEvent(this, this.handleVideoLoadStart)}
+            onMouseOver={linkEvent(this, this.handleVideoLoadStart)}
+            onFocus={linkEvent(this, this.handleVideoLoadStart)}
+            onVolumeChange={linkEvent(this, this.handleVideoVolumeChange)}
```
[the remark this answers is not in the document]
dessalines: Either one works... its not a bad practice to just always use linkEvent, even if you don't need the component.

src/shared/components/post/post-listing.tsx
ismailkarsli: With the recent changes, I brought back `linkEvent` as @dessalines suggested, should I still move?

SleeplessOne1917 (approved):

PR #2326, opened 2024-01-15
Fix captcha cache control bug

### LINKED ISSUE #2279: Captcha refresh doesn't refresh anymore when pressing the refresh button
### Requirements

### Summary

Pressing the refresh button on the captcha doesn't refresh it anymore. Only sometimes a hard page refresh makes it refresh.

https://github.com/LemmyNet/lemmy-ui/assets/67873169/56cab500-dce0-4743-8c79-0d2b4d1a3f7d

### Steps to Reproduce

1. visit  https://voyager.lemmy.ml/signup
2. press refresh under enter code

### Technical Details

FireFox/edge nothing in console

### Lemmy Instance Version

0.19.0-rc13

### Lemmy Instance URL

voyager.lemmy.ml

SleeplessOne1917:
Fixes #2279

### DIFF
--- a/src/shared/components/home/signup.tsx
+++ b/src/shared/components/home/signup.tsx
@@ -71,9 +71,13 @@ export class Signup extends Component<any, State> {
 
   async fetchCaptcha() {
     this.setState({ captchaRes: LOADING_REQUEST });
+    HttpService.client.setHeaders({
+      "Cache-Control": "private, no-cache",
+    });
     this.setState({
       captchaRes: await HttpService.client.getCaptcha(),
     });
+    HttpService.client.setHeaders({});
 
     this.setState(s => {
       if (s.captchaRes.state === "success") {


src/shared/components/home/signup.tsx
```diff
@@ -71,9 +71,13 @@ export class Signup extends Component<any, State> {
 
   async fetchCaptcha() {
     this.setState({ captchaRes: LOADING_REQUEST });
+    HttpService.client.setHeaders({
```
SleeplessOne1917, line 74: It would be helpful to add functionality to the JS client to set headers for individual requests. This time it doesn't matter too much: the signup page should only be visible to unauthenticated users, so we don't have to worry about overwriting the authentication header. However, I can see this making things needlessly difficult down the line.

dessalines: If its just a few times too, it might be useful to just create either a separate client without headers on demand, or in HTTPService. Adding optional function params to every lemmy-js-client call might be overkill.

dessalines (approved):

MV-GH:
Shouldn't this be rather done in the backend, that it should never cache this response.

MV-GH:
If the response is cached meaning it didn't hit the server at all and thus never set the correct answer. This is probably what caused me to fail the captcha so much. In other words the response should never be cached.

Nutomic:
Fix in https://github.com/LemmyNet/lemmy/pull/4381, this commit can be reverted.

dessalines:
#2335

PR #2334, opened 2024-01-19
Changing security.txt to use github security advisories page.

### LINKED ISSUE #2332: security.txt expires date
### Requirements

### Summary

the date would have to be adjusted:
https://github.com/LemmyNet/lemmy-ui/blob/main/src/server/handlers/security-handler.ts#L14

### Lemmy Instance Version

0.19.2

### Lemmy Instance URL

https://rollenspiel.forum/

dessalines:
- Fixes #2332

### DIFF
--- a/src/server/handlers/security-handler.ts
+++ b/src/server/handlers/security-handler.ts
@@ -4,14 +4,8 @@ export default async ({ res }: { res: Response }) => {
   res.setHeader("content-type", "text/plain; charset=utf-8");
 
   res.send(
-    `Contact: mailto:security@lemmy.ml
-  Contact: mailto:admin@` +
-      process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST +
-      `
-  Contact: mailto:security@` +
-      process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST +
-      `
-  Expires: 2024-01-01T04:59:00.000Z
+    `Contact: https://github.com/LemmyNet/lemmy-ui/security/advisories/new
+  Expires: 2025-01-01T04:59:00.000Z
   `,
   );
 };


Nutomic:
Should we just auto-generate a date which is maybe one year in the future from the build time?

Nothing4You:
I don't think this should just link to the github page.
Having the github page in there for the project itself is probably a good idea, but there should also be a (probably optional) instance security contact as well.
Especially for cases when instances are running an older version with known security issues fixed in the latest version, there isn't any point in raising this on github, but instead this should provide a contact to the instance owner directly.
Additionally, some things can be configuration related, such as bad caching configuration or other services on the same domain, like pict-rs, which wouldn't be for the lemmy project itself.

dessalines:
> Having the github page in there for the project itself is probably a good idea, but there should also be a (probably optional) instance security contact as well.

@Nothing4You That's another good argument for removing this entirely, and just letting instance runners serve their own security.txt . 

@Nutomic Okay I've added a build date to the dockerfile.

Nothing4You:
Maybe it could be integrated in lemmy-ansible instead?

dessalines:
I'd rather not add complication and extra configuring there.

urda:
A unit test should be included that **FAILS** when we pass the expire date again.

Dockerfile
```diff
@@ -32,6 +32,7 @@ COPY .git .git
 
 # Set UI version 
 RUN echo "export const VERSION = '$(git describe --tag)';" > "src/shared/version.ts"
+RUN echo "export const BUILD_DATE_ISO8601 = '$(date -d "+1 year" -u +"%Y-%m-%dT%H:%M:%SZ")';" > "src/shared/build-date.ts"
```
Nutomic, line 35: This name is misleading. Better store the current date and then add one year inside the security handler. That or rename the variable.

dessalines: My bad, I'll fix that now.

dessalines:
This is ready to go IMO. @SleeplessOne1917 take a look.

SleeplessOne1917 (approved):

PR #2336, opened 2024-01-19
Upgrading deps.

dessalines:
Upgrades several key deps: Markdown-it, prettier, sharp, and date-fns.


dessalines:
Gonna merge rn so I can start an RC build.

Issue #2160, opened 2023-10-01
Add the federated, and non-federated URL buttons to every post in the feed


K4LCIFER:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Currently, only the federated URL button
![image](https://github.com/LemmyNet/lemmy-ui/assets/7829231/611d0c31-8402-4602-8f02-d1b5a0dcee95)
 is shown for posts in the feed, and it's only shown for posts that are non-local. For example
![image](https://github.com/LemmyNet/lemmy-ui/assets/7829231/9c0e4b7e-02ce-4b89-81e0-442de2b28424)
 I think, for the sake of UX consistency, it would be best to have both the federated, *and* non-federated buttons
![image](https://github.com/LemmyNet/lemmy-ui/assets/7829231/e29e2eeb-7d2f-4dc9-9c09-ed8d4c522660)
shown on every single post.

Die4Ever:
I agree this is should be made consistent since the user will get used to using the permalink button and then get confused why it's not there. Or worse they get used to using the local link button and ignore the permalink.

PR #2344, opened 2024-01-28
always show fedilinks (issue #2160)

Die4Ever:
## Description

Always show fedilinks for UX consistency and so users don't get in the habit of not using them. Fixes issue #2160

### Before

image

### After

image


dessalines (approved):
Seems a bit redundant, since local links will be the same as the fedilink yet they're still shown. But I'm okay with it.

dessalines:
@SleeplessOne1917 take a look at this one too.

PR #2345, opened 2024-01-28
Move to pnpm

dessalines:
## Description

This moves from `yarn` to `pnpm` . Wasn't too difficult.


deploy.sh
```diff
@@ -9,7 +9,7 @@ new_tag="$1"
 # sudo docker push dessalines/lemmy-ui:$new_tag
 
 # Upgrade version
-yarn version --new-version $new_tag
+npm version $new_tag
```
SleeplessOne1917, line 12: If pnpm has a version command we can use that here.

SleeplessOne1917 (approved):

georgyo:
This change has made the nix packaging of lemmy much harder. Namely because Nix does not currently have a pnpm helper to fetch and build dependencies from the `pnpm-lock.yaml`. I am currently frozen on `0.19.3` because of this change. Looking at my federated instances, it seems like a great number of instances are still on `0.19.3` as well. 

Something that was missing from this relatively major change in building was any reasoning for it. Was there something that pnpm provided that yarn did not? It's fine if there was some reasoning, but it would be nice if that was shared.

dessalines:
[Much faster than yarn or npm](https://refine.dev/blog/pnpm-vs-npm-and-yarn/#improved-speed). Did you ask the nix devs why they don't support pnpm considering how much it is used in a lot of different node projects?

georgyo:
They're working on it,[ I even linked to the ticket](https://github.com/NixOS/nixpkgs/issues/317927). There are existing solutions today for Nix pnpm, they are just complicated. Each of these package managers (npm, yarn, pnpm) have very different lock files. And then their lock file formats are changed pretty frequently as well. 

BTW, for benchmarks listed in that 2nd link you listed,  `Yarn PNP` wins nearly every race. Yarn even has the ability to use [pnpm for the backend](https://yarnpkg.com/configuration/yarnrc#nodeLinker).

In any case, thanks for responding. I appreciate the work done on Lemmy. Just when I see major build system changes, it makes me curious as to why. Especially one that affects all downstream packaging and doesn't even make it into the changelog.

PR #2357, opened 2024-02-05
Update deps, namely sharp

SleeplessOne1917:
Switching to pnpm lets us use the latest version of sharp without issue. I updated the rest of the deps as well.


dessalines (approved):
Nice, thx.

PR #2363, opened 2024-02-16
Fix scroll bug when modal closes from browser back button

### LINKED ISSUE #2350: Scrolling disabled by popup did not get re-enable
### Requirements

### Summary

Popup interface like "view vote" and "purge user" will permanently disable scrolling if you did not click on the "X" or "cancel" button. This is especially bad if you purge a user and when the page reload, you will stuck and unable to scroll. Reload the page will help solve the issue but it's a chore.

### Steps to Reproduce

1. As an admin, click on the "three dot" on the post then select "View Votes" 
2. When the popup interface show, press the back button on your browser
3. Once the page reloaded, that browser tab will have scrolling permanently disabled.

Alternatively:

1. As an admin, find your victim for purging(preferably a CSAM troll)
2. Purge the account
3. Same as point number 3 above. 

### Technical Details

Issue happened in both my phone and my pc.

Tested on Vivaldi(phone, v6.5.3217.103) and Firefox(phone, v122.0.1/pc) 

### Lemmy Instance Version

0.19.3

### Lemmy Instance URL

_No response_

SleeplessOne1917:
Fixes #2350.

Not the ideal solution, as the user can still scroll while the modal is open. However, that is much better than having scrolling be unexpectedly broken.

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -1,3 +1,7 @@
+body {
+  overflow-y: scroll !important;
+}
+
 .navbar-expand-lg .navbar-nav .nav-link {
   padding-right: 0.75rem !important;
   padding-left: 0.75rem !important;


dessalines (approved):
Thx, this bug was super-annoying.

PR #2364, opened 2024-02-16
Prevent password length error silent truncation

### LINKED ISSUE #2172: Password length limit notification
### Requirements

- [X] Is this a UI / front end issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo.

User complained about passwords being "only" 60 characters, but the actual problem is that there is no notification explaining maximum length. I'm guessing it's the backbend that should provide this information as an error.

Notification like; "password length over limit of 60"

Notification like; "get a life"

### Additional context

Length established in this commit; https://github.com/LemmyNet/lemmy/pull/1474

SleeplessOne1917:
Fixes #2172 

password-invalid

### DIFF
--- a/src/shared/components/common/password-input.tsx
+++ b/src/shared/components/common/password-input.tsx
@@ -97,8 +97,8 @@ class PasswordInput extends Component<PasswordInputProps, PasswordInputState> {
                 onInput={onInput}
                 value={value}
                 required
-                maxLength={60}
-                minLength={10}
+                pattern=".{10,60}"
+                title="Password must be between 10 and 60 characters!"
               />
               <button
                 className="btn btn-outline-dark"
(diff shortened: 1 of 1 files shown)


src/shared/components/common/password-input.tsx
```diff
@@ -97,8 +97,8 @@ class PasswordInput extends Component<PasswordInputProps, PasswordInputState> {
                 onInput={onInput}
                 value={value}
                 required
-                maxLength={60}
-                minLength={10}
+                pattern=".{10,60}"
+                title={I18NextService.i18n.t("invalid_password")}
```
SleeplessOne1917, line 101: See [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern#usability_and_accessibility_considerations) for why I'm using the `title` prop for this.

SleeplessOne1917 (commented):
Invalid password verbiage is a bit redundant and could use a revision.

dessalines (approved):

PR #2380, opened 2024-02-28
Lazy load translations and date-fns, server side support for "Browser Default" language

matc-pub:
## Description

Lazy loads i18n translations and date-fns using dynamic imports.

### Before
Both scripts are always loaded.

		client.js: 915.51kB (3.29MB)
		48.client.js: 103.53kB (601.55kB) << date-fns

### After
For "en-US" only `client.js` is loaded. "en-GB" and "en-CA" additionally load one `date-fns-*client.js`, Most languages load one `translation-*.client.js` and one `date-fns-*.client.js`. Languages like "pt-BR", for which a specific (pt-BR) and general (pt) translation exists, load two translations and one date-fns.

		client.js: 724.04kB (2.57MB)
		translation-el.client.js 11.64kB (43.67kB) << largest translation
		date-fns-be-tarask-mjs.client.js 3.66kB (13.23kB) << largest date-fns

`date-fns` are picked to match the language that i18n uses.

Uses the "Accept-Language" header to determine which language to use for server side rendering. This also allows to load translations and date-fns in parallel with `client.js`.

The language code for "pt-BR" is no longer "pt_BR".

"DOMContentLoaded" (used for Notification permission) started working when only `client.js` is needed.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @dessalines @SleeplessOne1917 @alectrocute @jsit
+* @dessalines @SleeplessOne1917


dessalines (approved):
Tested this also, and everything works well. This late-loading stuff is magic to me, but it appears to work. I'll let @SleeplessOne1917 also take a look.

SleeplessOne1917 (approved):

PR #2388, opened 2024-03-11
Lazy loading less common languages for syntax highlighting

matc-pub:
## Description

Adds a plugin to highlight.js to detect when code should be highlighted using a language that is not yet loaded. A wrapped call to `md.render` provides a callback that will be called after a language had to be loaded, this allows components to `forceUpdate`.

### Before

All available languages (listed here: https://highlightjs.org/download) are enabled and can be auto-detected.

### After

Only the "common" set of 35 languages can be auto-detected, the remaining languages will be loaded individually on demand when explicitly specified by the author.

Uncommon languages, when auto-detected, are highlighted using the best fitting language of the common set.

`client.js` size shrinks by about 240kB (compressed)

### DIFF
--- a/src/shared/lazy-highlightjs.ts
+++ b/src/shared/lazy-highlightjs.ts
@@ -1,0 +1,111 @@
+import { HLJSApi, HLJSPlugin, LanguageFn } from "highlight.js";
+import hljs from "highlight.js/lib/core";
+import { bundledSyntaxHighlighters } from "./build-config";
+import { isBrowser } from "@utils/browser";
+import { default as MarkdownIt } from "markdown-it";
+
+class LazyHighlightjs implements HLJSPlugin {
+  public hljs: HLJSApi;
+
+  private loadedLanguages: Set<string> = new Set();
+  private failedLanguages: Set<string> = new Set();
+
+  constructor(hljs: HLJSApi) {
+    this.hljs = hljs;
+    this.hljs.addPlugin(this);
+
+    // For consistent autodetection behaviour.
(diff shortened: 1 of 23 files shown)


dessalines (approved):
Works also, thx!

src/shared/lazy-highlightjs.ts
```diff
@@ -0,0 +1,111 @@
...
+    return promise;
+  }
+
+  private enabled: boolean = false;
+  private current?: {
+    readonly callback: () => void;
+    readonly pending: Set<string>;
+  };
+
+  "before:highlight"(context: { language: string }) {
```
SleeplessOne1917, line 59: You can name class methods with string literals?

webpack.config.js
```diff
@@ -4,6 +4,8 @@ const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const nodeExternals = require("webpack-node-externals");
 const CopyPlugin = require("copy-webpack-plugin");
 const { ServiceWorkerPlugin } = require("service-worker-webpack");
+const { bundledSyntaxHighlighters } = require("./src/shared/build-config");
+const { lazySyntaxHighlighters } = require("./src/shared/build-config");
```
SleeplessOne1917, line 8: If I'm not mistaken, this can be shortened to:

```suggestion
const { bundledSyntaxHighlighters, lazySyntaxHighlighters } = require("./src/shared/build-config");
```

SleeplessOne1917 (approved):

PR #2390, opened 2024-03-12
Do most of the theme handling from the Theme component

matc-pub:
## Description

I noticed some issues with the current way of handling themes:
 -  Code blocks briefly use light theme after page refresh for "browser" theme in dark mode
 - "browser-compact" is not compact and always in light mode
 -  Live theme preview:
    - sometimes mixes light and dark elements
    - "browser" theme can be in wrong dark/light variant
    - Navigating away from dialog without saving keeps theme active
  - Unrelated: AdminSettings doesn't have a theme list after page refresh
 
This change moves all decisions about the current theme into the Theme component. The only thing outside of the Theme component is a script embedded in the SSR which sets the `data-bs-theme` attribute for "browser" and "browser-compact", as the server doesn't know the correct value and waiting for all scripts to load would be to slow. The Theme and Settings components communicate using `"refresh-theme"` and `"set-theme-override"` `CustomEvent`s.

#### Code syntax highlight in wrong color scheme while loading
code-block-wrong-color-scheme

#### Live Preview
Screenshot made without saving the theme. Only "browser-compact" has problems after saving and reloading.

##### "browser-compact" live preview not compact
browser-compact-preview-not-compact

##### Live preview with mixed dark/light elements
browser-preview-mixed

##### Live preview "vaporwave-dark" in light
vaporwave-dark-preview-mixed

### After
No screenshots. Things look as they should.


matc-pub:
One sequence to reproduce the light/dark mixed element. With "darkly-red" saved, change to "darkly", then "browser-compat", then "browser".

Looks like "browser-compact" has to be involved.

dessalines (approved):
Just tested this out, and it seemed to work well, thx!

SleeplessOne1917 (approved):

PR #2396, opened 2024-03-25
Rework query parsing

matc-pub:
## Description

Passes the result of the `getQueryParams` implemention for a component, to the component as `props` and to its `fetchInitialData` as `query`. This makes server and client use the same methods to parse query parameters. Also passes the result of `matchPath` as `match` to `fetchInitialData` , this is the same as `props.match`.

As a result the server side rendering matches a lot closer the client side rendering. Previously the server only used the query parameters to fetch data, but ignored them during `renderToString`.

Fixes an issue with double decoding query parameters, where the result of the first decoding can create malformed tokens for the second decoding. The fix relies on always using `URLSearchParams` to parse and create query strings.

Fixes `<PictrsImage />` with an image like  `"../pictrs/image/...webp?some=thing"` trying to show thumbnails ending in `"?some=thing?thumbnail.."`.


src/shared/components/community/community.tsx
```diff
@@ -234,25 +255,22 @@ export class Community extends Component<
...
       new LemmyHttp(getHttpBaseInternal(), { headers }),
     );
-    const pathSplit = path.split("/");
 
-    const communityName = pathSplit[2];
     const communityForm: GetCommunity = {
       name: communityName,
     };
 
-    const dataType = getDataTypeFromQuery(urlDataType);
```
dessalines, line 252: Getting rid of these is much cleaner, thx.

dessalines (changes requested):
Did you test this? It seems to completely break everything but the initial page loads.

I agree with this overall change tho, its a much cleaner way of doing this.

matc-pub:
> Did you test this? It seems to completely break everything but the initial page loads.

I tested it with a downgraded version of `lemmy-js-client`.

```diff
-    "lemmy-js-client": "0.19.4-alpha.4",
+    "lemmy-js-client": "0.19.4-alpha.2",
``` 
This downgrades cross-fetch. Without this I get `'fetch' called on an object that does not implement interface Window.` when trying to login.

dessalines:
We removed the dependency on cross-fetch in [this one.](https://github.com/LemmyNet/lemmy-js-client/pull/251)

I created a deploy with that included, could you try this version out?

`0.19.4-alpha.14`

src/shared/components/app/app.tsx
```diff
@@ -59,20 +64,35 @@ export class App extends Component<any, any> {
...
+                              ...queryProps,
+                              ...getQueryParams(
+                                routeProps.location.search,
+                                this.isoData.site_res,
+                              ),
+                            };
+                          } else {
+                            // ErrorGuard will catch this
+                          }
+                        }
```
SleeplessOne1917, line 80: If the `else` block is going to be empty, why nest the `if`s instead of just checking `getQueryParams && this.isoData.site_res`?

src/shared/utils/helpers/get-query-params.ts
```diff
@@ -1,21 +1,28 @@
...
-  [K in keyof T]: (param: string) => T[K];
-}): T {
-  if (isBrowser()) {
-    const searchParams = new URLSearchParams(window.location.search);
+  PropsT,
+  FallbacksT extends Empty = Empty,
+>(
+  processors: QueryMapping<PropsT, FallbacksT>,
+  source: string | undefined,
+  fallbacks: FallbacksT,
```
SleeplessOne1917, line 16: A default value would be good here. I noticed that you pass an empty object as the third arg to this function practically everywhere you call it.

src/shared/utils/helpers/get-query-params.ts
```diff
@@ -1,21 +1,28 @@
...
->(processors: {
-  [K in keyof T]: (param: string) => T[K];
-}): T {
-  if (isBrowser()) {
-    const searchParams = new URLSearchParams(window.location.search);
+  PropsT,
+  FallbacksT extends Empty = Empty,
+>(
+  processors: QueryMapping<PropsT, FallbacksT>,
+  source: string | undefined,
```
SleeplessOne1917, line 15: Will `source?; string,` work here?

src/shared/utils/helpers/get-query-params.ts
SleeplessOne1917, line 16: A default value would be good here. I noticed that you pass an empty object as the third arg to this function practically everywhere you call it.
```suggestion
  fallbacks: FallbacksT = {},
```

src/shared/utils/helpers/get-query-params.ts
```diff
@@ -1,21 +1,28 @@
-import { isBrowser } from "@utils/browser";
+type Empty = NonNullable<unknown>;
+
+type QueryMapping<PropsT, FallbacksT extends Empty> = {
+  [K in keyof PropsT]-?: (
```
SleeplessOne1917, line 4: Why the `-` before the `?`? I've never seen that syntax used in typescript before.

matc-pub:
`0.19.4-alpha.14` works for me. Also tested with the Dockerfile. `0.19.4-alpha.14` and `npm update` also works fine.

src/shared/components/app/app.tsx
matc-pub: This was intended to point out that a `RouteComponent` with `getQueryParams` will never see props without that don't contain the query parameters.

src/shared/utils/helpers/get-query-params.ts
matc-pub: Makes specifying a mapping for each property manadatory. Works like `Required<T>`:

```typescript
/**
 * Make all properties in T required
 */
type Required<T> = {
    [P in keyof T]-?: T[P];
};
```

dessalines:
Yep, that new lemmy-js-client dep does work correctly, could you also include that too.

matc-pub:
#2398 is this pr merged with #2397, for lemmy-js-client 0.19.4-alpha.14

dessalines (approved):
Tests good for me, thx!

SleeplessOne1917 (commented):
Once the default arg change is made I'll merge.

matc-pub:
> Once the default arg change is made I'll merge.

Is already part of the [Small getQueryParams cleanup](https://github.com/LemmyNet/lemmy-ui/pull/2396/commits/d14feb39c28438d469830b8883c2dbaf61ad5664) commit

PR #2408, opened 2024-04-06
Add hide post UI

SleeplessOne1917:
Makes UI for https://github.com/LemmyNet/lemmy/pull/4480

Screenshot_20240406_185358

Screenshot_20240406_185552

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,35 +1,35 @@
-lockfileVersion: "6.0"
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
 dependencies:
-  "@babel/plugin-proposal-decorators":
+  '@babel/plugin-proposal-decorators':
     specifier: ^7.24.1
     version: 7.24.1(@babel/core@7.24.3)
-  "@babel/plugin-transform-class-properties":
+  '@babel/plugin-transform-class-properties':
     specifier: ^7.24.1
     version: 7.24.1(@babel/core@7.24.3)
-  "@babel/plugin-transform-runtime":
+  '@babel/plugin-transform-runtime':
     specifier: ^7.24.3
     version: 7.24.3(@babel/core@7.24.3)
-  "@babel/plugin-transform-typescript":
+  '@babel/plugin-transform-typescript':
     specifier: ^7.24.1
     version: 7.24.1(@babel/core@7.24.3)
-  "@babel/preset-env":
+  '@babel/preset-env':
     specifier: ^7.24.3
     version: 7.24.3(@babel/core@7.24.3)
-  "@babel/preset-typescript":
+  '@babel/preset-typescript':
     specifier: ^7.24.1
     version: 7.24.1(@babel/core@7.24.3)
-  "@babel/runtime":
+  '@babel/runtime':
     specifier: ^7.24.1
     version: 7.24.1
-  "@emoji-mart/data":
+  '@emoji-mart/data':
     specifier: ^1.1.2
     version: 1.1.2
-  "@shortcm/qr-image":
+  '@shortcm/qr-image':
     specifier: ^9.0.4
     version: 9.0.4
   autosize:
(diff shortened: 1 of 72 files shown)


src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -212,6 +213,16 @@ export default class ContentActionDropdown extends Component<
...
+                ) : (
+                  <li>
+                    <ActionButton
+                      icon={this.props.postView.hidden ? "eye" : "eye-slash"}
+                      label={
+                        this.props.postView.hidden ? "Unhide Post" : "Hide Post"
+                      }
+                      onClick={this.props.onHidePost}
+                    />
+                  </li>
```
SleeplessOne1917, line 225: Should this also show up for a user's own posts?

src/shared/components/post/post-listing.tsx
```diff
@@ -904,6 +908,13 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     });
   }
 
+  handleHidePost() {
+    return this.props.onHidePost({
+      hide: !this.postView.hidden,
+      post_ids: [this.postView.post.id],
```
SleeplessOne1917, line 914: Why does this take an array of ids instead of just one?

src/shared/components/home/home.tsx
```diff
@@ -1050,6 +1088,26 @@ export class Home extends Component<HomeRouteProps, HomeState> {
...
+          for (const post of prev.postsRes.data.posts.filter(p =>
+            form.post_ids.some(id => id === p.post.id),
+          )) {
+            post.hidden = form.hide;
+          }
+        }
+
+        return prev;
+      });
+
```
dessalines, line 1106: i18n

src/shared/components/home/home.tsx
```diff
@@ -791,6 +801,25 @@ export class Home extends Component<HomeRouteProps, HomeState> {
             onChange={this.handleDataTypeChange}
           />
         </div>
+        {dataType === DataType.Post &&
+          (this.state.siteRes.my_user || UserService.Instance.myUserInfo) && (
+            <div className="col-auto form-check form-switch">
```
dessalines, line 806: Since this is used in two places, externalize it into its own component.

Also, this is causing a lot of wrapping even on desktop, so we need to figure out how to make it skinnier, as well as more uniform with the rest of these toggles/filters.

How about doing it like all the other buttons, with the two options being `eye-on` and `eye-off` icons. Check out src/shared/components/common/data-type-select.tsx for an example. 

And to explain what it does on a click/touch and hold, use `data-tippy-content`.

src/shared/components/post/post.tsx
```diff
@@ -1041,6 +1044,22 @@ export class Post extends Component<PostRouteProps, PostState> {
...
+    if (hideRes.state === "success") {
+      this.setState(s => {
+        if (s.postRes.state === "success") {
+          s.postRes.data.post_view.hidden = form.hide;
+        }
+
+        return s;
+      });
+
+      toast(form.hide ? "Post hidden" : "Post unhidden");
```
dessalines, line 1059: i18n

src/shared/components/community/community.tsx
```diff
@@ -814,6 +859,26 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+            form.post_ids.some(id => id === p.post.id),
+          )) {
+            post.hidden = form.hide;
+          }
+        }
+
+        return prev;
+      });
+
+      toast(form.hide ? "Post hidden" : "Post unhidden");
```
dessalines, line 878: i18n

src/shared/components/community/community.tsx
```diff
@@ -561,6 +566,27 @@ export class Community extends Component<CommunityRouteProps, State> {
             onChange={this.handleDataTypeChange}
           />
         </span>
+        {dataType === DataType.Post &&
+          (this.state.siteRes.my_user || UserService.Instance.myUserInfo) && (
```
dessalines, line 570: See comment above about externalizing this

dessalines (commented):
I tested and this works well, just needs i18n and some minor cleanup.

pnpm-lock.yaml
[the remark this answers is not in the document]
SleeplessOne1917: I added `pnpm-lock.yaml` to `.prettierignore`. This should hopefully avoid the issue in the future.

SleeplessOne1917:
I'll work on extracting the component in a bit.

src/shared/components/home/home.tsx
SleeplessOne1917: > Since this is used in two places, externalize it into its own component.

I started doing that and am now questioning the usefulness of that. When I tried, I ended up with a component with just the input and label taking `showHidden` and `onShowHiddenChangeProps`. I do think a case can be made for extracting the entire filter/sort bar into it's own component. That would allow us to have a single source of truth for this instead of repeating logic on the homepage and community page.

> How about doing it like all the other buttons, with the two options being eye-on and eye-off icons. Check out src/shared/components/common/data-type-select.tsx for an example.

Do you mean the pillbox looking button groups? Like this but with eye and eye slash instead of posts and comments?
image

dessalines: > I started doing that and am now questioning the usefulness of that. When I tried, I ended up with a component with just the input and label taking showHidden and onShowHiddenChangeProps

Sure, but then you only need to change markup / visual related things in one place. That's why we have the listingTypeSelect and SortTypeSelect as their own components too, even though they are pretty simple.

> I do think a case can be made for extracting the entire filter/sort bar into it's own component.

Maybe, although the indivdual Select components are used elsewhere besides just the Home and Community pages iirc. Either way that could be a separate PR as it would be a bigger change. Plus even the Home and Community have different filters (no subscribed vs local vs all filter).

> Do you mean the pillbox looking button groups? Like this but with eye and eye slash instead of posts and comments?

Yep. Just to look uniform with the others.

src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -176,6 +177,17 @@ export default class ContentActionDropdown extends Component<
           </button>
 
           <ul className="dropdown-menu" id={dropdownId}>
+            {type === "post" && (
+              <li>
+                <ActionButton
+                  icon={this.props.postView.hidden ? "eye" : "eye-slash"}
+                  label={I18NextService.i18n.t(
+                    this.props.postView.hidden ? "unhide_postt" : "hide_post",
```
dessalines, line 185: `unhide_postt`

dessalines (approved):
Tried it and it looks good.

PR #2409, opened 2024-04-08
Block urls

SleeplessOne1917:
Frontend change to make the change from https://github.com/LemmyNet/lemmy/issues/4514 usable.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -4,3 +4,4 @@ src/assets/css/themes/*.css
 src/assets/css/code-themes/*.css
 stats.json
 dist
+pnpm-lock.yaml
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,35 +1,35 @@
-lockfileVersion: "6.0"
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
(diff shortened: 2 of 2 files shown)


src/shared/components/common/url-list-textarea.tsx
```diff
@@ -0,0 +1,76 @@
...
+          htmlFor="create-site-block-urls"
+        >
+          Block URLs
+        </label>
+
+        <div className="col-12">
+          <textarea
+            id="create-site-block-urls"
+            className="form-control"
+            placeholder="Put your blocked URLs here, one URL per line."
```
dessalines, line 66: i18n.

src/shared/components/common/url-list-textarea.tsx
```diff
@@ -0,0 +1,76 @@
...
+  };
+
+  render() {
+    return (
+      <div className="mb-3 row">
+        <label
+          className="col-12 col-form-label"
+          htmlFor="create-site-block-urls"
+        >
+          Block URLs
```
dessalines, line 59: i18n.

dessalines (commented):
I tested and it works, just needs i18n. I like the simple line delimited box, as it makes them easier to share around.

dessalines (approved):

PR #2415, opened 2024-04-11
Use mixins and decorators for scroll restoration and tippy cleanup

matc-pub:
## Description

Mixins prevent repetitions across multiple components.

Without decorators the use of a mixin would roughly look like this:

```typescript
class Home_ { /* ... */ }
export const Home = mixin(Home_);
```

With decorators:

```typescript
@mixin
export class Home { /* ... */ }
```

The class decorator functions (the mixins) in this PR are compatible with
"legacy" and "2023-11" proposal-decorators versions. "legacy" transpiles to
slightly smaller javascript. (Switching between versions requires changes in
tsconfig and babelrc.)

Typescript requires two arguments for non-legacy class decorator functions. I
had to add `{ "argsIgnorePattern": "^_" }` to `no-unused-vars`.

### Use tippy.js delegate addon, cleanup tippy instances from a mixin.

The delegate addon creates tippy instances from mouse and touch events with a
matching `event.target`. This is initially significantly cheaper than creating
all instances at once. The addon keeps all created tippy instances alive until
it is destroyed itself.

`tippyMixin` destroys the addon instance after every render, as long as all
instances are hidden. This drops some tippy instances that may have to be
recreated later (e.g when the mouse moves over the trigger again), but is
otherwise fairly cheap (creates one tippy instance).

### Restore scroll positions when resource loading settles.

The history module generates a random string (`location.key`) for every
browser history entry. The names for saved positions include this key.
The position is saved before a route component unmounts or before the
`location.key` changes.

The `scrollMixin` tires to restore the scroll position after every
change of `location.key`. It only does so after the first render for
which the route components `loadingSettled()` returns true.

Things like `scrollToComments` should only scroll when `history.action`
is not "POP".

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -13,7 +13,12 @@
     ["@babel/typescript", { "isTSX": true, "allExtensions": true }]
   ],
   "plugins": [
-    "@babel/plugin-transform-runtime",
+    ["@babel/plugin-proposal-decorators", { "version": "legacy" }],
+    [
+      "@babel/plugin-transform-runtime",
+      // version defaults to 7.0.0 for which non-legacy decorators produce duplicate code
+      { "version": "^7.24.3" }
+    ],
     ["babel-plugin-inferno", { "imports": true }],
     ["@babel/plugin-transform-class-properties", { "loose": true }]
   ]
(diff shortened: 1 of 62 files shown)


dessalines (approved):
I tested this and it seems to work well. Not really a huge fan of these OO hacks, but it does simplify both a lot.

SleeplessOne1917 (approved):
I didn't even know mixins were a thing. I'll have to look into using them for modals as well.

Issue #2401, opened 2024-03-27
[Bug]: Link of comment is the link to the parent


rakoo:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

In Lemmy the link to a comment (given by the chain icon) points to the parent.

Here's an example: comment https://lemmy.ml/comment/9675721 is a top-level comment. Its reply has the same link. The reply to the reply has link https://lemmy.ml/comment/9676388 points to the parent.

(you can check the content by curl-ing the comment: `curl -L -H 'Accept: application/ld+json' https://lemmy.ml/comment/9676388`

### Steps to Reproduce

1. Get the link to a comment, retrieve it with the previous curl command
2. Check the content in the json ActivityPub object
3. Check the content in the ui
4. Both contents should be the same: they are not

### Technical Details

I'm not using lemmy locally, I'm seeing it on multiple instances (sh.itjust.works, jlai.lu, lemmy.ml)

### Version

0.19.3

### Lemmy Instance URL

lemmy.ml, sh.itjustworks, jlai.lu

snan:
[More info here](https://github.com/LemmyNet/lemmy/issues/4582).

dessalines:
In jerboa, the link of the comment is the actual comment, and I think we should replicate that here.

dessalines:
Nevermind, this one is done by the above.

PR #2417, opened 2024-04-11
Fix parent comment linking when not showing context

snan:
The same code is reused for the "show context" button as for the "link
to self" button. I'm not sure that's such a good idea in the long run.

There was a const `parentCommentId` that was always set even when not
having the `showContext` prop, causing the bug #2401.


snan:
(Trying to figure out what the linter wants by trial and error.)

dessalines (approved):

PR #2419, opened 2024-04-12
Make confirm popup for adult consent

SleeplessOne1917:
Makes UI for a [backend change](https://github.com/LemmyNet/lemmy/pull/4393). This isn't fully complete because there are more things that I need to clarify.

How should the `content_warning` site setting interact with the `enable_nsfw` site setting? I assume that `content_warning` should not be able to have a value when `enable_nsfw` is off. However, if I understand correctly, having `enable_nsfw` set to true while `content_warning` is blank still blurs nsfw images by default, while the blurring is disabled if a `content_warning` is set. I think it's pretty unintuitive, since the relation between the content warning and post blurring is not clear from just looking at the settings.

One idea I have around this is including a blurb explaining this, though I imagine there are better alternatives and I want to know if anyone thinks of any.

Another thing I noticed is that the user settings for showing nsfw and blurring nsfw are visible even when the site has `enable_nsfw` set to false. The option for blurring nsfw is also able to be changed independently of showing nsfw, which seems wrong.

Besides the stuff above, what I have left with this is:
- Add UI to site form
- Add translations

I'll get those 2 parts done once the other stuff is hashed out.


src/shared/components/app/app.tsx
```diff
@@ -25,6 +26,19 @@ export class App extends Component<any, any> {
...
+    const siteRes = this.isoData.site_res;
+
+    if (
+      siteRes?.site_view.site.content_warning &&
+      !(siteRes?.my_user || localStorage.getItem(adultConsentLocalStorageKey))
+    ) {
+      if (confirm(siteRes.site_view.site.content_warning)) {
+        localStorage.setItem(adultConsentLocalStorageKey, "true");
+      } else {
+        history.back();
```
SleeplessOne1917, line 38: Should I include an alert like lemmy nsfw does?

dessalines:
> I assume that content_warning should not be able to have a value when enable_nsfw is off.

Seems correct to me, maybe hide the content_warning field if enable_nsfw = false.

> However, if I understand correctly, having enable_nsfw set to true while content_warning is blank still blurs nsfw images by default, while the blurring is disabled if a content_warning is set. I think it's pretty unintuitive, since the relation between the content warning and post blurring is not clear from just looking at the settings.

I suppose content_warning is a rare *override*, that should disable all blurring and nsfw settings. As long as the nsfw check is extracted to a function that takes these into account, it shouldn't be too difficult.

> One idea I have around this is including a blurb explaining this, though I imagine there are better alternatives and I want to know if anyone thinks of any.

I wouldn't worry about that, its up to the server-runners to put that in their content_warning if they wish. 

> Another thing I noticed is that the user settings for showing nsfw and blurring nsfw are visible even when the site has enable_nsfw set to false. The option for blurring nsfw is also able to be changed independently of showing nsfw, which seems wrong.

That could also be fixed here, if you'd like. We could hide those user settings fields because they're pointless if the site doesn't allow NSFW.

src/shared/components/app/app.tsx
dessalines: Yeah I think so. It doesn't have to be a javascript::alert like theirs, but that would probably have the best compatibility. I'd probably prefer a proper modal tho, so that we can markdown render the content warning.

The most important thing is that the site behind the alert is hidden until they click a **let me in** button.

SleeplessOne1917:
> I suppose content_warning is a rare override, that should disable all blurring and nsfw settings. As long as the nsfw check is extracted to a function that takes these into account, it shouldn't be too difficult.

The logic sounds good. The other half of the problem is communicating this to the user in an effective way. While I assume an instance admin should be at least somewhat familiar with the docs, I'd rather try to avoid that with UX design.


>> One idea I have around this is including a blurb explaining this, though I imagine there are better alternatives and I want to know if anyone thinks of any.

> I wouldn't worry about that, its up to the server-runners to put that in their content_warning if they wish.

In this case, I'm not talking about notifying the users about that. I'm talking about making it clear on the site settings UI that admins use.


Everything else you said sounds good and I'll get to work on it.

SleeplessOne1917:
I'm almost done with this issue. The only thing left is making translations. If reviewers are good with the verbiage I picked, I'll make the translations PR.

Here is what the section of the admin settings looks like:
Screenshot_20240414_101655

I recorded some videos of the popup, but github is being dumb and won't let me upload them.

SleeplessOne1917:
Since `enable_nsfw` is going to be deprecated, do you want me to remove the option from the admin settings page? Or should I wait until it's phased out for good?

Nutomic:
So far the setting is still in use, so it doesnt make sense to remove it yet.

dessalines (commented):
Tested, but IMO the modal should hide everything in the background, and might need some re-thinking to make that work.

Even if you set every picture to blur, there could still be some things like large offensive text you wouldn't want anyone going to this site for the first time to see.

The strings you made seem fine to me, so you can i18n them whenever.

Nutomic:
Im not sure how lemmynsfw.com is doing it, somehow they are hiding the page entirely while the popup is showing. Maybe you can figure it out from their code.

https://github.com/LemmyNet/lemmy-ui/compare/main...lemmynsfw:lemmy-ui:main

SleeplessOne1917:
I figured out the blur:
image

@dessalines @Nutomic  With the blur and translations done, this should be good to merge.

dessalines:
Its still showing the site for a second or so:

https://github.com/LemmyNet/lemmy-ui/assets/930215/deb5c43c-123e-45f7-9200-a35e6dd0ef54

src/shared/components/app/app.tsx
```diff
@@ -29,96 +125,17 @@ export class App extends Component<any, any> {
...
-            </div>
-            <Footer site={siteRes} />
-          </div>
-        </Provider>
-      </>
+      <Provider i18next={I18NextService.i18n}>
+        {contentWarning && (
+          <AdultConsentModal contentWarning={contentWarning} />
+        )}
+        <AppForwardRef ref={this.rootRef} />
```
dessalines, line 137: Why not just set an `invisible` class depending on the contentWarning?

dessalines (commented):
Its faster, but still like 100-200ms for me. Would probably be good to try it another way: let the site render, with a default invisible class, that gets removed after the content warning passes.

src/shared/components/app/app.tsx
SleeplessOne1917: Currently I handle the blur by setting the `background-filter` for the top level of the modal element. I've tried doing it by putting a class on app, but that always ended up blurring out the modal as well. Another approach I tried was just having all of the lifecycle methods and event handlers be handled in the original top-level `App` component. This seemed to work, but would cause the loading skeleton to flash in the blurred background each second the redirect time counts down.

I'm going to see if I can handle this all at the very top level `componentDidRoute`. I strongly suspect the issue is due to the order rendering is done in. Before my most recent change, the logic for showing the modal didn't run  until after the top level `App` component mounted and rendered. In the current iteration, `componentDidMount` for the consent modal is being called simultaneously with the `componentDidMount` for the `App` component. I think moving this logic to the very root of the component tree as a parent of `App` should solve the issue.

dessalines: Cool. If all else fails we might have to just resort to a simple confirm dialog on the SSR like lemmynsfw does it.

matc-pub: > In the current iteration, `componentDidMount` for the consent modal is being called simultaneously with the `componentDidMount` for the `App` component.

`setState` in `componentWillMount` will affect the first render of the component. It gets called on the server too.

To avoid ever showing the wrongly un/blurred version after a page refresh, you'd probably need a `create-ssr-html`-hack similar to the auto detected light/dark themes. It's the same problem, the server doesn't know the real style to use, and whatever the server decides to show will otherwise be visible at least until `startClient` is called.

https://github.com/LemmyNet/lemmy-ui/blob/accf1b2d72c48f9b4eb0ee265bf6098a84079813/src/server/utils/create-ssr-html.tsx#L82-L87

SleeplessOne1917: I've tried the cookie approach with and without a data attribute. The HTML and CSS that gets returned from the server before client side JS runs still isn't blurred. I think the problem is that bootstrap modals depend on being run in the browser to render. We'll either have to live with a slight delay for when the blur starts or go back to using the regular confirm dialog.

matc-pub:
https://github.com/LemmyNet/lemmy-ui/pull/2430 is a PR to this PR. (Not sure how this will show up)

The PR reverts the commit with the changes in App.tsx and add does apply the blur analog to `data-bs-theme` attribute.

src/server/utils/create-ssr-html.tsx
```diff
@@ -75,37 +75,19 @@ export async function createSsrHtml(
     .map(x => `<link rel="preload" as="script" href="${x}" />`)
     .join("");
 
-  // Blurring has to happen even if all style sheets fail to load.
```
matc-pub: You should keep this approach to include your styles.

src/shared/components/app/app.tsx
```diff
@@ -13,42 +13,39 @@ import { Navbar } from "./navbar";
...
+          {this.isoData.showAdultConsentModal && (
+            <AdultConsentModal
+              contentWarning={siteView!.site.content_warning!}
+            />
+          )}
+          <div
+            id="app"
+            className="lemmy-site"
+            ref={this.rootRef}
+            data-adult-consent={this.isoData.showAdultConsentModal}
```
matc-pub, line 43: Looks like on the server settting an attribute to false, will remove the attribute, but on the client the attribute becomes "false".  I had to change it to `data-adult-consent={this.isoData.showAdultConsentModal || undefined}` to get it working. Otherwise everything works well for me.

src/shared/components/app/app.tsx
```diff
@@ -13,42 +13,39 @@ import { Navbar } from "./navbar";
...
+          {this.isoData.showAdultConsentModal && (
+            <AdultConsentModal
+              contentWarning={siteView!.site.content_warning!}
+            />
+          )}
+          <div
+            id="app"
+            className="lemmy-site"
+            ref={this.rootRef}
+            data-adult-consent={this.isoData.showAdultConsentModal || null}
```
dessalines, line 43: Much cleaner.

src/shared/components/common/pictrs-image.tsx
```diff
@@ -19,48 +21,46 @@ interface PictrsImageProps {
...
-            "img-blur-icon": icon && blur_image,
-            "img-blur-thumb": thumbnail && blur_image,
-            "ms-2 mb-0 rounded-circle object-fit-cover avatar-overlay":
-              iconOverlay,
-            "avatar-pushup": pushup,
-            "card-img-top": cardTop,
-          })}
-        />
-      </picture>
+      !this.isoData.showAdultConsentModal && (
```
dessalines, line 36: I tested, and it seems to handle the re-rendering fine.

dessalines (approved):
LGTM, since @matc-pub helped with this also, let's have them take a look and approve it also.

BTW @matc-pub , would you like to be a maintainer / collaborator for lemmy-ui? LMK and I'll add you.

dessalines:
Oops I forgot to turn off auto-merge, my bad. Still do a look over and maybe another PR if anything's crucial.

PR #2422, opened 2024-04-12
Only conditionally render most of content action dropdown and workaround for tippy warning

matc-pub:
This only starts rendering the dropdown and each modal dialog when they are shown for first time. For simplicity they keep getting rendered until the component unmounts.

A small fix when the Modal promise resolves after the component umounts. I don't remember how to trigger this, but saw this occasionally happen.

Also a work around for the tippy warning. Turns out tippy doesn't remove all event listeners when it destroys an instance.

### DIFF
--- a/src/shared/components/common/content-actions/content-action-dropdown.tsx
+++ b/src/shared/components/common/content-actions/content-action-dropdown.tsx
@@ -59,24 +59,30 @@ export type ContentPostProps = {
 
 type ContentActionDropdownProps = ContentCommentProps | ContentPostProps;
 
-const dialogTypes = [
-  "showBanDialog",
-  "showRemoveDialog",
-  "showPurgeDialog",
-  "showReportDialog",
-  "showTransferCommunityDialog",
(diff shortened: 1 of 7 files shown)


src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -174,292 +188,305 @@ export default class ContentActionDropdown extends Component<
             aria-expanded="false"
             aria-controls={dropdownId}
             aria-label={I18NextService.i18n.t("more")}
+            onClick={() => this.setState({ dropdownOpenedOnce: true })}
```
SleeplessOne1917, line 191: Should use `linkEvent` with this.

SleeplessOne1917:
I noticed that the dynamic import and setup/cleanup gets repeated everywhere we use the modal. Do you think mixins similar to what you did in the previous PR could alleviate this?

SleeplessOne1917 (approved):

PR #2424, opened 2024-04-16
Adding image upload views for admins and profiles.

dessalines:
## Description

Adds the ability to view and delete media uploads for admins and users

## Screenshots

https://github.com/LemmyNet/lemmy-ui/assets/930215/d2c942ea-e582-4aa4-87e8-fb00cd1e6d5a

context: https://github.com/LemmyNet/lemmy/pull/4631

### DIFF
--- a/src/shared/components/common/media-uploads.tsx
+++ b/src/shared/components/common/media-uploads.tsx
@@ -69,7 +69,7 @@ export class MediaUploads extends Component<Props, any> {
   deleteImageBtn(image: LocalImage) {
     return (
       <button
-        onClick={linkEvent({ i: this, image }, this.handleDeleteImage)}
+        onClick={linkEvent(image, this.handleDeleteImage)}
         className="btn btn-danger"
       >
         {I18NextService.i18n.t("delete")}
@@ -77,7 +77,7 @@ export class MediaUploads extends Component<Props, any> {
     );
   }
 
-  async handleDeleteImage({ image }: { i: MediaUploads; image: LocalImage }) {
+  async handleDeleteImage(image: LocalImage) {
     const form = {
       token: image.pictrs_delete_token,
       filename: image.pictrs_alias,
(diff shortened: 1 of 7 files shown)


src/shared/components/common/media-uploads.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <button
+        onClick={linkEvent({ i: this, image }, this.handleDeleteImage)}
+        className="btn btn-danger"
+      >
+        {I18NextService.i18n.t("delete")}
+      </button>
+    );
+  }
+
+  async handleDeleteImage({ image }: { i: MediaUploads; image: LocalImage }) {
```
SleeplessOne1917, line 80: You're not currently using `i`. You don't need it.

src/shared/components/common/media-uploads.tsx
```diff
@@ -0,0 +1,106 @@
...
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+
+  deleteImageBtn(image: LocalImage) {
+    return (
+      <button
+        onClick={linkEvent({ i: this, image }, this.handleDeleteImage)}
```
SleeplessOne1917, line 72: The value for `i` never gets used in `handleDeleteImage`. IDK if `linkEvent` still has optimization benefits if you're not trying to use `this`.

dessalines: I'll try keeping linkEvent, but removing this.

dessalines:
The blank pages are gonna be tough to figure out, bc I can't replicate. I just tested again, and the linked video above shows a refresh. It might be because this requires an unmerged back-end change that's running with my new `LocalImageView` . I'll leave this as draft until the back-end gets merged.

SleeplessOne1917 (approved):
Tested it again and it works.

PR #2425, opened 2024-04-16
Use muted display names, and always show instance for non-local. #1975 #2064

dessalines:
## Description

Everyone seems to like this one from jerboa anyway, so I'll add it here.

When its a non-local instance, always show the instance, but make it smaller and muted.

### Before

Didn't show instance name if a display name was set

### After

Always shows it (for non-local)

(ignore the localhost string, that's just for testing)

Screenshot_20240415_224838_Fulguris_1

### DIFF
--- a/src/shared/components/person/person-listing.tsx
+++ b/src/shared/components/person/person-listing.tsx
@@ -24,64 +24,47 @@ export class PersonListing extends Component<PersonListingProps, any> {
   }
 
   render() {
-    const person = this.props.person;
+    const { person, useApubName } = this.props;
     const local = person.local;
-    let apubName: string, link: string;
+    let link: string;
+    let serverStr: string | undefined = undefined;
+
+    const name = useApubName ? person.name : person.display_name ?? person.name;
 
     if (local) {
-      apubName = `@${person.name}`;
       link = `/u/${person.name}`;
     } else {
-      const domain = hostname(person.actor_id);
-      apubName = `@${person.name}@${domain}`;
+      serverStr = `@${hostname(person.actor_id)}`;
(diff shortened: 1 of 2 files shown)


src/shared/components/person/person-listing.tsx
```diff
@@ -93,7 +76,10 @@ export class PersonListing extends Component<PersonListingProps, any> {
               icon
             />
           )}
-        <span>{displayName}</span>
+        <span>{name}</span>
+        {serverStr !== undefined && (
```
SleeplessOne1917, line 80: Are there any falsey values you're trying to avoid matching?

src/shared/components/person/person-listing.tsx
```diff
@@ -24,64 +24,47 @@ export class PersonListing extends Component<PersonListingProps, any> {
...
-      displayName = `${displayName} (${apubName})`;
-    }
-
+    const classes = classNames(
+      "person-listing d-inline-flex align-items-baseline",
+      {
+        "text-muted": this.props.muted,
+        "text-info": !this.props.muted,
+      },
+    );
```
SleeplessOne1917, line 49: A ternary using `muted` would be more concise.

SleeplessOne1917 (approved):
I really like this. I think it's a good way of giving users important information while looking nice and not taking up more space. The other comments I made are nitpicks that aren't essential to address.

src/shared/components/person/person-listing.tsx
dessalines: This is using that special `classNames` function, it can't work in that way.

PR #2430, opened 2024-04-18
Blur SSR by default when content warning is set

matc-pub:


SleeplessOne1917 (changes requested):
With these changes, if a user has already consented or is logged in, they always get a blurred document sent back to the browser. This is the opposite problem than we were running into originally.

Also, didn't you say you wanted to prevent the browser from downloading NSFW material entirely? Because even if we get the blur working correctly, that NSFW content is still getting downloaded by img tags.

SleeplessOne1917:
The original PR now works if we're trying to go with the blur. It always makes the document returned from the server blurred when appropriate while also not sending a blurred page back to users who are logged in or already consented.

matc-pub:
> With these changes, if a user has already consented or is logged in, they always get a blurred document sent back to the browser. This is the opposite problem than we were running into originally.

There blur would have been removed on the client before the html was fully parsed.
 
> Also, didn't you say you wanted to prevent the browser from downloading NSFW material entirely? Because even if we get the blur working correctly, that NSFW content is still getting downloaded by img tags.

I believe your thinking of this https://github.com/LemmyNet/lemmy/issues/4625 .

matc-pub:
I'm not in any chat.

SleeplessOne1917:
I am so sorry about that. I got you confused with @MV-GH.

MV-GH:
Yeah that was my concern, whether or not the user had consented. The files are already stored.

matc-pub:
With the cookie approach this should be doable by just adding another guard like auth-guard. But it probably needs a page refresh to update the isoData. The error-guard is currently also stuck, once it sees an error, it will see the same error for the next route too.

PR #2432, opened 2024-04-19
Content warning fixups

matc-pub:
- Require consent of logged in existing users.
- Prevent link clicks in blurred areas.
- Enable tippy again.
- Prevent stacking of countdowns for Back button.
- Fix auto expand blocking for images.

### DIFF
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -145,7 +145,7 @@ export default async (req: Request, res: Response) => {
       errorPageData,
       showAdultConsentModal:
         !!site?.site_view.site.content_warning &&
-        !req.cookies[adultConsentCookieKey],
+        !(site.my_user || req.cookies[adultConsentCookieKey]),
     };
 
     const wrapper = (
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -145,7 +145,7 @@ export default async (req: Request, res: Response) => {
       errorPageData,
       showAdultConsentModal:
         !!site?.site_view.site.content_warning &&
-        !(site.my_user || req.cookies[adultConsentCookieKey]),
+        !req.cookies[adultConsentCookieKey],
     };
 
     const wrapper = (
--- a/src/shared/components/app/app.tsx
+++ b/src/shared/components/app/app.tsx
@@ -14,6 +14,7 @@ import "./styles.scss";
 import { Theme } from "./theme";
 import AnonymousGuard from "../common/anonymous-guard";
 import AdultConsentModal from "../common/adult-consent-modal";
+import { destroyTippy, setupTippy } from "../../tippy";
 
 function handleJumpToContent(event) {
   event.preventDefault();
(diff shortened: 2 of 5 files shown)


src/server/handlers/catch-all-handler.tsx
```diff
@@ -145,7 +145,7 @@ export default async (req: Request, res: Response) => {
       errorPageData,
       showAdultConsentModal:
         !!site?.site_view.site.content_warning &&
-        !(site.my_user || req.cookies[adultConsentCookieKey]),
```
SleeplessOne1917, line 148: Logged in users should not be shown the consent modal.

src/shared/components/app/app.tsx
```diff
@@ -23,6 +24,14 @@ export default class App extends Component<any, any> {
...
   private readonly rootRef = createRef<HTMLDivElement>();
 
+  componentDidMount() {
+    setupTippy(this.rootRef);
+  }
+
+  componentWillUnmount() {
+    destroyTippy();
+  }
+
```
SleeplessOne1917, line 34: Doh! I can't believe I removed those. Good catch.

src/shared/components/post/post-listing.tsx
```diff
@@ -103,7 +103,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   private readonly isoData: IsoDataOptionalSite = setIsoData(this.context);
   state: PostListingState = {
     showEdit: false,
-    imageExpanded: !!this.isoData.site_res?.site_view.site.content_warning,
```
SleeplessOne1917, line 106: Images are supposed to auto expand when content warning is present.

SleeplessOne1917 (changes requested):
The code is fine, but some of your changes go against requirements.

src/shared/components/post/post-listing.tsx
matc-pub: Not the opposite? Collapsed while the modal is showing. Expanded when auto expand is enabled and the image is not nsfw or the user settings disable the blur for nsfw.

src/server/handlers/catch-all-handler.tsx
SleeplessOne1917: I was going by [lemmynsfw's changes](https://github.com/LemmyNet/lemmy-ui/compare/main...lemmynsfw:lemmy-ui:main) for this one. I doubt instances that will need the consent form will shift to using it after the instance has been established long enough to have a userbase.

src/shared/components/post/post-listing.tsx
SleeplessOne1917: I'm using lemmynsfw's behavior as a reference, but TBH this does seem like something that should be a separate site setting.

src/server/handlers/catch-all-handler.tsx
dessalines: I pry agree that logged-in users don't need to be shown the popup. If they're already part of the instance they know what its about.

src/shared/components/post/post-listing.tsx
dessalines: related to this: https://github.com/LemmyNet/lemmy/issues/4643

I'd prefer we just let the new `local_site.default_post_listing_mode` handle this, and add a card view.

SleeplessOne1917 (approved):

PR #2434, opened 2024-04-20
Add fallback image for broken images

### LINKED ISSUE #2433: Fallback image for when images are broken
### Requirements

When images are broken, a blank space is all that is shown. A fallback image for broken images would make this more apparent.

SleeplessOne1917:
Closes #2433.

This is what is looks like with the fallback image:
Screenshot_20240420_140939


src/shared/components/common/pictrs-image.tsx
```diff
@@ -20,13 +21,35 @@ interface PictrsImageProps {
...
 }
 
-export class PictrsImage extends Component<PictrsImageProps, any> {
+interface PictrsImageState {
+  src: string;
+}
+
+function handleImgLoadError(i: PictrsImage) {
+  i.setState({
+    src: `${getStaticDir()}/assets/images/broken-image-fallback.png`,
```
dessalines, line 30: Try [this solution](https://stackoverflow.com/a/13118956/1655478) to see if it works.

``

Then we could avoid having to have to create a src state variable, and hooks, just to handle the error case.

matc-pub: > ``

I tried something similar at some point. Worked fine in dev. But in prod the CSP complained about inline-scripts.

dessalines: Ah, in that case this is good then.

dessalines (approved):

PR #2449, opened 2024-05-05
Prevent onBlur from appending URL scheme to URLs

SleeplessOne1917:
Since the blocked URLs no longer include the scheme, I no longer stick the scheme on the front.


lemmy-translations
SleeplessOne1917, line 1: We should remember to update translations to their latest versions before the 0.19.4 release.

dessalines (approved):

PR #2455, opened 2024-05-11
Show same cake day date independent of timezone

matc-pub:
## Description

Profile displays the UTC+0 date to all users. This alters the effective cake day for some users, but prevents users in different timezones from seeing different dates. It also keeps showing the same date after switching timezones.

The tooltip for "Joined X months ago" still shows the localized exact time and date, which can be a different day than displayed below as "Cake day:".

### DIFF
--- a/src/shared/utils/helpers/is-cake-day.ts
+++ b/src/shared/utils/helpers/is-cake-day.ts
@@ -6,9 +6,9 @@ export function cakeDate(published: string): Date {
   return parse(published.substring(0, 10), "yyyy-MM-dd", new Date(0));
 }
 
-export default function isCakeDay(published: string, current?: Date): boolean {
+export default function isCakeDay(published: string): boolean {
   const createDate = cakeDate(published);
-  const currentDate = current ?? new Date();
+  const currentDate = new Date();
 
   // The day-overflow of Date makes leap days become 03-01 in non leap years.
   return (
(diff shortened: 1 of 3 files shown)


src/shared/utils/helpers/is-cake-day.ts
```diff
@@ -1,34 +1,78 @@
...
 
+// Returns a date in local time with the same year, month and day. Ignores the
+// source timezone. The goal is to show the same date in all timezones.
+export function cakeDate(published: string): Date {
+  return parse(published.substring(0, 10), "yyyy-MM-dd", new Date(0));
+}
+
+export default function isCakeDay(published: string, current?: Date): boolean {
+  const createDate = cakeDate(published);
+  const currentDate = current ?? new Date();
```
SleeplessOne1917, line 11: You could just make this a default param.

src/shared/utils/helpers/is-cake-day.ts
```diff
@@ -1,34 +1,78 @@
...
   return (
-    createDateDayOfYear === currentDateDayOfYear &&
-    getYear(createDate) !== getYear(currentDate)
+    isSameDay(currentDate, setYear(createDate, getYear(currentDate))) &&
+    !isSameYear(currentDate, createDate)
   );
 }
+
+/*
+import { format } from "date-fns";
```
SleeplessOne1917, line 21: Should probably delete these commented out lines.

SleeplessOne1917 (approved):

Issue #2454, opened 2024-05-11
Site logo disappearing


kroese:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

Around Lemmy 0.19.4 beta 3 and higher I run into the issue that the site logo (and profile picture for the admin) are frequently disappearing.

My guess is to what happens is that when you edit the `Admin settings` or the `Profile settings` it does not load the current picture correctly (the box stays empty) and as soon as you save your modifications, this will clear the existing logo/picture.

So currently I workaround it by everytime uploading the logo again when I edit any settings on those pages, to prevent the existing logo from disappearing.

Would be nice to see this fixed before the official 0.19.4 is released, because Im sure this will affect a lot of people.

### Steps to Reproduce

1. Edit Admin settings
2. Hit save
3. Check site logo
4. Logo disappeared

### Technical Details

Samsung Internet browser on Android

### Lemmy Instance Version

0.19.4-beta.6

SleeplessOne1917:
I tested this locally. I was able to reproduce the issue with the site logo, but not a user profile picture. I also noticed that the user settings reload the page on save, where the site settings do not. I'll address the issue I found with the site settings.

SleeplessOne1917:
What routes are you on when editing the site logo vs the admin user profile picture?

SleeplessOne1917:
So you are using the right routes. When you say the site logo is identical to the admin logo, do you mean that you are separately setting them to the same image, or that setting one sets the other?

SleeplessOne1917:
Currently, lemmy-ui doesn't set any cache related headers on requests. It does set cache control headers on responses though: https://github.com/LemmyNet/lemmy-ui/blob/main/src/server/middleware.ts#L37-L65

SleeplessOne1917:
The PR for the admin settings fix is merged. @kroese let us know if you still run into this problem once the latest release is out.

SleeplessOne1917:
Is there a reproducible set of steps that consistently causes this bug? I'm currently at a loss as to what could be causing this problem.

[...]

PR #2456, opened 2024-05-14
Refresh admin settings on submit

SleeplessOne1917:
Addresses #2454.

Was only able to reproduce that issue with the admin settings. Adding in a refresh after saving does the trick.


dessalines (approved):
Sweet, thx.

PR #2474, opened 2024-05-18
Fix video thumbnail override issue

### LINKED ISSUE #2418: MP4 videos in post URLs no longer embed when clicking the preview box next to the post title
### Requirements

### Summary

On 0.19.3, clicking on the play symbol to the left of the post title, an embedded video player will be shown.
On 0.19.4-beta.3, the post will show external link indicator and will not embed the video.

### Steps to Reproduce

1. Create a post linking to an mp4 video, such as https://files.catbox.moe/oytdpp.mp4
2. Open said post on a 0.19.4-beta.3 instance
3. Click the (broken, because it seems to try loading the video as image) preview left of the post
4. Get redirected to the video URL

### Technical Details

Firefox 124.0.2

working: https://sh.itjust.works/post/17246485
not working: https://voyager.lemmy.ml/post/78810

### Lemmy Instance Version

0.19.4-beta.3

### Lemmy Instance URL

https://voyager.lemmy.ml

SleeplessOne1917:
Fixes #2418. Here is what it looks like now:

image

I made as few changes to the logic as I could. I'm afraid to touch it, especially since we don't have any tests set up. This makes me want to work on the leptos UI rewrite even more than I already do.


src/assets/symbols.svg
SleeplessOne1917, line 1: I changed the symbol we use for videos so it looks good both alone and as a watermark.

SleeplessOne1917:
I made a few other thumbnail related changes while I was at it.

Thumbnails now always show up, even on mobile:
image
image

I did this because I noticed that videos were unviewable on mobile unless I clicked the post title once to navigate to the post page, then again to actually see the video. Here is a video post on mobile on the main branch:
image

As you can see, you can't tell this is a video by looking at it.

Finally, always including the thumbnail, even for text-only posts, keeps posts on the feed a consistent height, which I personally think looks better.

.npmrc
```diff
@@ -0,0 +1 @@
+ignore-scripts=true
```
SleeplessOne1917, line 1: I started getting an error about an opencollective postinstall script when I tried to install. Without ignoring scripts now, you can't install dependencies. This only started happening at the time I post this and I'm not sure why.

dessalines (approved):

Issue #2481, opened 2024-05-28
Regression in lemmy-ui allows the flooding of same post from new post ui


KindnessInfinity:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

There is a regression that allows user to spam posts, by clicking create multiple times.

### Steps to Reproduce

1. Open community you have post privileges on
2. Start process to create new post
3. Click "create" multiple times

### Technical Details

vanadium (chromium based browser) version 125.0.6422.113.0

### Lemmy Instance Version

BE: 0.19.4-rc.2

### Lemmy Instance URL

https://lemmy.ml

PR #2487, opened 2024-06-01
Fix some submit button issues

matc-pub:
- Disable submit buttons during submission (#2481)
  - Re-enable submit buttons after failure
  - Don't throw away user input on failure
  - Keep navigation warnings enabled during submission

I used this snippet for testing, which selectively replaces the form with an empty object. It also adds a one second delay before the request is made. (Comment and post examples: [submit_buttons.webm](https://github.com/LemmyNet/lemmy-ui/assets/161147791/4a767c68-9234-4f61-8ab2-2a2078bcc511))

```diff
diff --git a/src/shared/services/HttpService.ts b/src/shared/services/HttpService.ts
index 7ff9121b..01a0cef7 100644
--- a/src/shared/services/HttpService.ts
+++ b/src/shared/services/HttpService.ts
@@ -1,3 +1,4 @@
+import { isBrowser } from "@utils/browser";
 import { getHttpBase } from "@utils/env";
 import { LemmyHttp } from "lemmy-js-client";
 
@@ -56,6 +57,24 @@ class WrappedLemmyHttpClient {
       if (key !== "constructor") {
         this[key] = async (...args) => {
           try {
+            if (isBrowser()) {
+              if (
+                [
+                  "createPost",
+                  "editPost",
+                  "createComment",
+                  "editComment",
+                  "createPrivateMessage",
+                  "editPrivateMessage",
+                ].includes(key)
+              ) {
+                if (!confirm("Cancel to fail")) {
+                  args[0] = {};
+                }
+                await new Promise(r => setTimeout(r, 1000));
+              }
+            }
+
             const res = await this.rawClient[key](...args);
 
             return {
```
Comment and post examples with the above snippet: [submit_buttons.webm](https://github.com/LemmyNet/lemmy-ui/assets/161147791/4a767c68-9234-4f61-8ab2-2a2078bcc511)

### DIFF
--- a/src/shared/components/comment/comment-form.tsx
+++ b/src/shared/components/comment/comment-form.tsx
@@ -2,11 +2,17 @@ import { capitalizeFirstLetter } from "@utils/helpers";
 import { Component } from "inferno";
 import { T } from "inferno-i18next-dess";
 import { Link } from "inferno-router";
-import { CreateComment, EditComment, Language } from "lemmy-js-client";
+import {
+  CommentResponse,
+  CreateComment,
+  EditComment,
+  Language,
+} from "lemmy-js-client";
 import { CommentNodeI } from "../../interfaces";
 import { I18NextService, UserService } from "../../services";
 import { Icon } from "../common/icon";
 import { MarkdownTextArea } from "../common/markdown-textarea";
+import { RequestState } from "shared/services/HttpService";
 
 interface CommentFormProps {
   /**
@@ -20,7 +26,9 @@ interface CommentFormProps {
   allLanguages: Language[];
   siteLanguages: number[];
   containerClass?: string;
-  onUpsertComment(form: EditComment | CreateComment): void;
+  onUpsertComment(
+    form: EditComment | CreateComment,
+  ): Promise<RequestState<CommentResponse>>;
 }
 
 export class CommentForm extends Component<CommentFormProps, any> {
@@ -81,32 +89,35 @@ export class CommentForm extends Component<CommentFormProps, any> {
         : capitalizeFirstLetter(I18NextService.i18n.t("reply"));
   }
 
-  handleCommentSubmit(content: string, language_id?: number) {
+  async handleCommentSubmit(
+    content: string,
+    language_id?: number,
+  ): Promise<boolean> {
     const { node, onUpsertComment, edit } = this.props;
     if (typeof node === "number") {
       const post_id = node;
-      onUpsertComment({
+      return onUpsertComment({
         content,
         post_id,
         language_id,
-      });
+      }).then(r => r.state !== "failed");
     } else {
       if (edit) {
         const comment_id = node.comment_view.comment.id;
-        onUpsertComment({
+        return onUpsertComment({
           content,
           comment_id,
           language_id,
-        });
+        }).then(r => r.state !== "failed");
       } else {
         const post_id = node.comment_view.post.id;
         const parent_id = node.comment_view.comment.id;
-        this.props.onUpsertComment({
+        return onUpsertComment({
           content,
           parent_id,
           post_id,
           language_id,
-        });
+        }).then(r => r.state !== "failed");
       }
     }
   }
--- a/src/shared/components/comment/comment-form.tsx
+++ b/src/shared/components/comment/comment-form.tsx
@@ -94,31 +94,33 @@ export class CommentForm extends Component<CommentFormProps, any> {
     language_id?: number,
   ): Promise<boolean> {
     const { node, onUpsertComment, edit } = this.props;
+    let response: RequestState<CommentResponse>;
+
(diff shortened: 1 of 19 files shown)


src/shared/components/comment/comment-form.tsx
```diff
@@ -2,26 +2,33 @@ import { capitalizeFirstLetter } from "@utils/helpers";
...
+  CommentResponse,
+  CreateComment,
+  EditComment,
+  Language,
+} from "lemmy-js-client";
 import { CommentNodeI } from "../../interfaces";
 import { I18NextService, UserService } from "../../services";
 import { Icon } from "../common/icon";
 import { MarkdownTextArea } from "../common/markdown-textarea";
+import { RequestState } from "shared/services/HttpService";
```
dessalines, line 15: These need to be relative imports, or it'll fail.

src/shared/components/comment/comment-node.tsx
```diff
@@ -87,7 +86,6 @@ interface CommentNodeProps {
   allLanguages: Language[];
   siteLanguages: number[];
   hideImages?: boolean;
-  finished: Map<CommentId, boolean | undefined>;
```
dessalines, line 90: Good, its nice to get rid of these.

src/shared/components/home/home.tsx
```diff
@@ -973,13 +968,19 @@ export class Home extends Component<HomeRouteProps, HomeState> {
     const createCommentRes = await HttpService.client.createComment(form);
     this.createAndUpdateComments(createCommentRes);
 
+    if (createCommentRes.state === "failed") {
```
dessalines, line 971: Thx, these are going to be handy.

dessalines (approved):
I tested this out and it works great, thx a ton. So nice to get rid of that finished map, and all those lifecycle events.

I'm not sure why the non-relative import doesn't throw an error, because I've almost always seen that. I'd change it to relative just in case tho. 

Lets let @SleeplessOne1917 take a look, then we can merge.

src/shared/components/comment/comment-form.tsx
```diff
@@ -83,32 +89,35 @@ export class CommentForm extends Component<CommentFormProps, any> {
...
-        this.props.onUpsertComment({
+        return onUpsertComment({
           content,
           parent_id,
           post_id,
           language_id,
-        });
+        }).then(r => r.state !== "failed");
       }
     }
```
SleeplessOne1917, line 122: I think this would be cleaner as

```typescript
  async handleCommentSubmit(
    content: string,
    language_id?: number,
  ): Promise<boolean> {
    const { node, onUpsertComment, edit } = this.props;
    let response: RequestState<CommentResponse>;

    if (typeof node === "number") {
      const post_id = node;
      response = await onUpsertComment({
        content,
        post_id,
        language_id,
      });
    } else if (edit) {
        const comment_id = node.comment_view.comment.id;
        response = await onUpsertComment({
          content,
          comment_id,
          language_id,
        });
      } else {
        const post_id = node.comment_view.post.id;
        const parent_id = node.comment_view.comment.id;
        response = await onUpsertComment({
          content,
          parent_id,
          post_id,
          language_id,
        });
      }

   return response.state !== "failed";
  }
```

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -575,11 +553,15 @@ export class MarkdownTextArea extends Component<
...
+  async handleSubmit(i: MarkdownTextArea, event: any) {
     event.preventDefault();
     if (i.state.content) {
       i.setState({ loading: true, submitted: true });
-      i.props.onSubmit?.(i.state.content, i.state.languageId);
+      const success = await i.props.onSubmit?.(
+        i.state.content,
+        i.state.languageId,
+      );
+      i.setState({ loading: false, submitted: success ?? true });
```
SleeplessOne1917, line 564: Shouldn't `onSubmit` always return a boolean, making the coalesce for `success` unnecessary?

SleeplessOne1917 (changes requested):
Thanks a lot for helping with this. Most of my efforts on making a nice looking UI have been focused on the UI rewrite, so I really appreciate you improving the UX on the current UI while the new one is under development.

src/shared/components/comment/comment-form.tsx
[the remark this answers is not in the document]
matc-pub: The `Dockerfile` builds and runs fine with absolute import. I added a linter rule to catch these in the future.

src/shared/components/common/markdown-textarea.tsx
matc-pub: `onSubmit` is optional, which makes `success` of type `boolean | undefined`.

SleeplessOne1917 (approved):

PR #2515, opened 2024-06-07
Create post query params

### LINKED ISSUE #1913: Regression: Prefilling `create_post` form from URL parameters no longer works
### Requirements

### Summary

Prior to Lemmy 1.18.0, it was possible to use parameters in the URL to load a "New Post" form with some fields prefilled, e.g.:

https://lemmy.instance/create_post?lang=en&url=https%3A%2F%foo.bar&title=Foo+Bar&community_id=1&body=Description+of+foo+bar

This was super useful for bookmarklets and "share" style URLs.

Unfortunately this does not seem to work anymore. I'm not sure if the names of the GET parameters have changed, or if this feature was dropped along the way.

### Steps to Reproduce

1. Navigate to /create_post?title=Foo+bar
2. Observe that the Title form field is not prepopulated with `Foo bar`

### Technical Details

N/A

### Lemmy Instance Version

0.18.1

### Lemmy Instance URL

https://links.roobre.es

SleeplessOne1917:
Fixes #1913

### DIFF
--- a/src/shared/components/post/create-post.tsx
+++ b/src/shared/components/post/create-post.tsx
@@ -5,8 +5,17 @@ import {
   setIsoData,
   voteDisplayMode,
 } from "@utils/app";
-import { getIdFromString, getQueryParams } from "@utils/helpers";
-import { Choice, RouteDataResponse } from "@utils/types";
+import {
+  getIdFromString,
+  getQueryParams,
+  getQueryString,
+} from "@utils/helpers";
+import {
+  Choice,
+  QueryParams,
+  RouteDataResponse,
+  StringBoolean,
+} from "@utils/types";
 import { Component } from "inferno";
 import { RouteComponentProps } from "inferno-router/dist/Route";
 import {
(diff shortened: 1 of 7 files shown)


src/shared/components/post/create-post.tsx
```diff
@@ -166,10 +196,25 @@ export class CreatePost extends Component<
 
   render() {
     const { selectedCommunityChoice, siteRes, loading } = this.state;
-
-    const locationState = this.props.history.location.state as
```
dessalines, line 170: Removing this broke the cross-post button in `src/shared/components/common/content-actions/cross-post-button.tsx`

I suppose it needs to merge these two objects.

dessalines (commented):
This worked, but it also broke the cross-post button.

src/shared/components/post/create-post.tsx
dessalines: I got it now.

dessalines (approved):

PR #2517, opened 2024-06-07
Fixing copy suggested title when creating posts.

dessalines:
I have no idea why, but these `onBlur` events broke the onClick for copy suggested title.

What are they doing exactly?

### DIFF
--- a/src/shared/components/post/post-form.tsx
+++ b/src/shared/components/post/post-form.tsx
@@ -150,7 +150,7 @@ function copySuggestedTitle(d: { i: PostForm; suggestedTitle?: string }) {
     d.i.setState(
       s => ((s.form.name = sTitle?.substring(0, MAX_POST_TITLE_LENGTH)), s),
     );
-    d.i.setState({ suggestedPostsRes: EMPTY_REQUEST });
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (changes requested):
The `onBlur`s set the url query params when the user clicks outside of the inputs or if the inputs use focus. The alternative is updating them every time they are changed (basically every time the user adds or removes a character while typing in the field), which seems overkill.

Removing all of these will break the create post query param regression fix I did last night. I'll take a look at this now.

dessalines:
Sweet, thx.

matc-pub:
@SleeplessOne1917  I think the problem is that the state is lost every time the url updates. `CreatePost` doesn't have a `mountedSameRouteNavKey` in `src/shared/routes.ts`.

dessalines:
I've removed the lemmy-ui `0.19.4` tag until we get that one fixed, last hour bug. 

If you can't figure this out, is removing these onBlurs breaking anything? Or is it just wasteful.

SleeplessOne1917:
I figured out the problem. I'm gonna clean things up a bit and open a PR.

dessalines:
Cool, I'll close this one then.

PR #2518, opened 2024-06-07
Fix copy title from metadata

### LINKED ISSUE #2516: "Copying Title" function stopped working for create post
### Requirements

### Summary

The title copying button no longer works.

Likely this PR broke title copying: https://github.com/LemmyNet/lemmy-ui/pull/2515

### Steps to Reproduce

1. Go to the create post page
2. Type in a url
3. See the metadata fetched title
4. Click copy
5. Nothing happens

### Technical Details

NA

### Lemmy Instance Version

0.19.4

### Lemmy Instance URL

https://voyager.lemmy.ml

SleeplessOne1917:
Fixes #2516 

Alternative to #2517

### DIFF
--- a/src/shared/components/post/post-form.tsx
+++ b/src/shared/components/post/post-form.tsx
@@ -63,13 +63,14 @@ interface PostFormProps {
   onSelectCommunity?: (choice: Choice) => void;
   initialCommunities?: CommunityView[];
   loading: boolean;
-  onTitleUrlChange?: (title: string) => void;
+  onTitleBlur?: (title: string) => void;
   onUrlBlur?: (url: string) => void;
   onBodyBlur?: (body: string) => void;
   onLanguageChange?: (languageId?: number) => void;
   onNsfwChange?: (nsfw: StringBoolean) => void;
   onThumbnailUrlBlur?: (thumbnailUrl: string) => void;
   onAltTextBlur?: (altText: string) => void;
+  onCopySuggestedTitle?: (url: string, title: string) => void;
 }
 
 interface PostFormState {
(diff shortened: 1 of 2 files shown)


dessalines:
Hrm... didn't work. 

Click it the first time, it doesn't copy. Alter the URL again, then click copy, and it works the 2nd time only.

SleeplessOne1917:
I see the issue. It works the first time if the url field loses focus, but it doesn't work the first time when you go from the url field directly to the copy button.

src/shared/components/post/post-form.tsx
```diff
@@ -175,17 +191,19 @@ function handlePostUrlChange(i: PostForm, event: any) {
...
 
 function handlePostUrlBlur(i: PostForm, event: any) {
-  i.setState({ bypassNavWarning: true });
-  i.props.onUrlBlur?.(event.target.value);
-  i.setState({ bypassNavWarning: false });
+  i.setState({
+    urlBlurTimeout: setTimeout(() => {
+      i.updateUrl(() => i.props.onUrlBlur?.(event.target.value));
+    }, 500),
+  });
```
SleeplessOne1917, line 198: A kludge. It fixes the bug at least.

dessalines (approved):
It still does some reloading, but the filling works correctly now. Good enough for me, and we can fix the other things in the next release.

src/shared/components/post/post-form.tsx
```diff
@@ -354,6 +367,10 @@ export class PostForm extends Component<PostFormProps, PostFormState> {
         },
       };
     }
+
+    if (this.state.form.url) {
+      this.fetchPageTitle();
```
matc-pub, line 372: This calls setState in the constructor. Probably only a problem for dev builds.

matc-pub: And should also be client side only.

dessalines:
There are still a few issues with both cross-posting, and copy title. We can fix them for the next bugfix release tho.

PR #2520, opened 2024-06-07
Keep CreatePost mounted when updating URL, fixes focus resets.

matc-pub:
Clicking on another input element now moves the focus. Without this change the clicked input element is replaced with a new one during URL update.

### DIFF
--- a/src/shared/components/post/create-post.tsx
+++ b/src/shared/components/post/create-post.tsx
@@ -6,6 +6,7 @@ import {
   voteDisplayMode,
 } from "@utils/app";
 import {
+  bareRoutePush,
   getIdFromString,
   getQueryParams,
   getQueryString,
@@ -90,6 +91,7 @@ interface CreatePostState {
   selectedCommunityChoice?: Choice;
   initialCommunitiesRes: RequestState<ListCommunitiesResponse>;
   isIsomorphic: boolean;
+  resetCounter: number; // resets PostForm when changed
 }
 
 type CreatePostPathProps = Record<string, never>;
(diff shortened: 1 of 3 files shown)


src/shared/routes.ts
```diff
@@ -121,6 +121,7 @@ export const routes: IRoutePropsWithFetch<RouteData, any, any>[] = [
     component: CreatePost,
     fetchInitialData: CreatePost.fetchInitialData,
     getQueryParams: getCreatePostQueryParams,
+    mountedSameRouteNavKey: "create_post",
```
SleeplessOne1917, line 124: What does this do?

src/shared/components/post/create-post.tsx
```diff
@@ -237,6 +249,7 @@ export class CreatePost extends Component<
           <div id="createPostForm" className="col-12 col-lg-6 offset-lg-3 mb-4">
             <h1 className="h4 mb-4">{I18NextService.i18n.t("create_post")}</h1>
             <PostForm
+              key={this.state.resetCounter}
```
SleeplessOne1917, line 252: Using a key (outside of a collection) to prevent uneccessary rerenders. Clever.

SleeplessOne1917 (approved):

src/shared/routes.ts
matc-pub: In `App` routes without that property get the `location.key` assigned as key, so that they unmount whenever the location changes. Before that property all routes behaved like that.

src/shared/components/post/create-post.tsx
matc-pub: It still has to render even when the key stays the same. But the when the key changes it has to be recreated and resets its state. I added it so that the "Create Post" link in the top navigation completely resets the form.

PR #2552, opened 2024-06-17
Fix auto download bug

### LINKED ISSUE #2547: [Bug]: "Expand Media" with `auto_expand_media` off, and an embedded_url gives no way to embed in the page.
### Requirements

### Summary

When enabling the "Media Auto Expand" feature and opening the lemmy.ml front page, a video download starts.
Does not happen in lemmy.world.
The cause is this specific post: https://lemmy.ml/post/16978055
It also starts the download, when opening the post directly

### Steps to Reproduce

1. Enable "Media Auto Expand" in settings
2. Open "https://lemmy.ml/post/16978055"

### Technical Details

OS: Windows 11
Browser: Firefox
Instance: lemmy.ml

### Lemmy Instance Version

0.19.4

### Lemmy Instance URL

https://lemmy.ml

SleeplessOne1917:
Fixes #2547

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
       "sortpack"
     ]
   },
-  "packageManager": "pnpm@9.4.0",
+  "packageManager": "pnpm@9.4.0+sha512.f549b8a52c9d2b8536762f99c0722205efc5af913e77835dbccc3b0b0b2ca9e7dc8022b78062c17291c48e88749c70ce88eb5a74f1fa8c4bf5e18bb46c8bd83a",
   "engineStrict": true,
   "importSort": {
     ".js, .jsx, .ts, .tsx": {
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -225,8 +225,8 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
       return <></>;
     }
 
-    // if direct video link
-    if (url && isVideo(url)) {
+    // if direct video link or embedded video link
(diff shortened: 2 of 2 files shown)


SleeplessOne1917:
I noticed that there's also no way to manually expand embedded videos. I'm wondering how that should be handled since embedded videos only show up when a URL is present, so the thumbnail is already being used for a link to the page.

dessalines:
The html `video` I think has a fullscreen button either by default or long-press on most browsers. 

We could potentially use something like [video.js](https://videojs.com/) for either this, or lemmy-ui-leptos, but IMO its just better to use the browser's own player.

SleeplessOne1917:
I'm not referring to a custom video player.

This is a post that just has a video as its URL:
[Screencast_20240617_205159.webm](https://github.com/LemmyNet/lemmy-ui/assets/28871516/3baea001-348e-45c3-a6c6-6e6cfa1a3848)
Notice that you can click on the icon thumbnail to show/hide the video.

Compare that to an embedded video:
[Screencast_20240617_205222.webm](https://github.com/LemmyNet/lemmy-ui/assets/28871516/4a90350d-9212-432a-b5c5-e23c0b9d0aa8)
There's nowhere to click to show the video. Users who don't have auto-expand media turned on will have to follow the link to the page hosting the video if they want to see it.

dessalines:
That's a tough one... if we wanted to be consistent, I spose the video as url example, could not expand when `auto_expand_media` is false, but also open the url directly instead of embedding.

PR #2619, opened 2024-07-21
Prevent broken logo from crashing site

### LINKED ISSUE #2211: When pictrs cant fetch site logo, entire site breaks
### Requirements

### Summary

Sometimes pictrs fails for many reasons, mostly because its not 100% perfect, to give a valid response to the logo call from lemmy-ui then the entire lemmy site no longer loads. This can be resolved by removing the site logo and uploading a new one again, without having to restart anything at all. Then lemmy-ui can get the logo again and the site functions normally once more. 

A better solution to the entire site breaking would be to display a broken image icon or something as the logo so that the site continues to operate when this happens. 

you can also remove the site logo directly from the postgres DB by using update site set icon = null where id = 1; and this also fixes the site again and then you can upload a new one.

Would make lemmy more reliable if this could be fixed, as pictrs cant be relied upon to be 100% reliable in every case every time until the end of the universe, as its just written by a normal human and so on. Bugs happen.

### Steps to Reproduce

1. Have site logo break.
2. Lemmy site no longer loads.

### Technical Details

Docker build with pictrs, same as 99% of lemmy servers.

### Lemmy Instance Version

0.18.5

### Lemmy Instance URL

https://lemmy.myserv.one/

SleeplessOne1917:
Fixes #2211

### DIFF
--- a/src/server/utils/create-ssr-html.tsx
+++ b/src/server/utils/create-ssr-html.tsx
@@ -32,8 +32,8 @@ export async function createSsrHtml(
     `<script nonce="${cspNonce}"`,
   );
 
-  try {
-    if (!appleTouchIcon) {
+  if (!appleTouchIcon) {
+    try {
       appleTouchIcon = site?.site_view.site.icon
         ? `data:image/png;base64,${await sharp(
             await fetchIconPng(site.site_view.site.icon),
@@ -50,12 +50,12 @@ export async function createSsrHtml(
             .toBuffer()
             .then(buf => buf.toString("base64"))}`
         : favIconPngUrl;
+    } catch (e) {
+      console.log(
+        "Could not fetch site logo for apple touch icon. Using default icon.",
+      );
+      appleTouchIcon = favIconPngUrl;
     }
-  } catch (e) {
-    console.log(
-      "Could not fetch site logo for apple touch icon. Using default icon.",
-    );
-    appleTouchIcon = favIconPngUrl;
   }
 
   const erudaStr =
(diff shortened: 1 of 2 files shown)


src/server/utils/create-ssr-html.tsx
```diff
@@ -32,23 +32,30 @@ export async function createSsrHtml(
...
+              right: 20,
+              background: "#222222",
+            })
+            .png()
+            .toBuffer()
+            .then(buf => buf.toString("base64"))}`
+        : favIconPngUrl;
+    }
+  } catch (e) {
+    console.log(
```
SleeplessOne1917, line 55: Should we make translations for these logs?

phiresky:
if the conversion request fails it's going to be a pretty bad ddos opportunity to have all that fetching and attempted sharp conversion happen on every single http request

SleeplessOne1917:
I could try caching the manifest icons in memory like I do with the apple touch icon. Thoughts?

phiresky:
yeah makes sense, assign it to something in the catch so it only happens once

src/server/utils/create-ssr-html.tsx
dessalines: Nah, no need.

dessalines (approved):
Just tested with a locally broken icon and this does currently work.

src/server/utils/generate-manifest-json.ts
```diff
@@ -1,19 +1,43 @@
...
-);
+let icons: Icon[] | null = null;
 
 export default async function (site: Site) {
-  const icon = site.icon ? await fetchIconPng(site.icon) : null;
+  if (!icons) {
+    try {
+      const icon = site.icon ? await fetchIconPng(site.icon) : null;
+
+      icons = await Promise.all(
```
phiresky, line 14: it still looks like the icons array is unassigned if this call fails? so it runs again every time?

PR #2629, opened 2024-07-23
Fixing modlog filtering to allow admins and mods to filter by mod.

### LINKED ISSUE #2590: Make sure modlog `Filter by user` filters not just by the target person, but also the `mod_person_id`
### Requirements

- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

There was a case today where I (as an admin) needed to unban a bunch of misbans from a rogue moderator. 

Unfortunately the modlog filter only filters by the *target* / recipient of the mod action, not the mod themselves.

The modlog person filter should do an `OR` filter on the `mod_person_id` also.

NA

### Additional context

NA

dessalines:
- Fixes #2590

## Description

Fixes modlog filtering to allow filtering by mods or admins.

## Screenshots

Screenshot_20240723_141114_Via

### DIFF
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -907,8 +907,9 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
             options={userSearchOptions}
             loading={loadingUserSearch}
           />
-          {!this.isoData.site_res.site_view.local_site
-            .hide_modlog_mod_names && (
+          {(this.amAdminOrMod ||
+            !this.isoData.site_res.site_view.local_site
+              .hide_modlog_mod_names) && (
             <Filter
               filterType="mod"
               onChange={this.handleModChange}


src/shared/components/modlog.tsx
```diff
@@ -907,8 +907,9 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
             options={userSearchOptions}
             loading={loadingUserSearch}
           />
-          {!this.isoData.site_res.site_view.local_site
-            .hide_modlog_mod_names && (
+          {(this.amAdminOrMod ||
+            !this.isoData.site_res.site_view.local_site
+              .hide_modlog_mod_names) && (
```
SleeplessOne1917, line 912: If I'm not mistaken, this will allow anyone who is a mod of a community to filter by mod even when looking at actions for communities they aren't a mod of.

dessalines: We should be okay, check the `this.amModOrAdmin` function.

SleeplessOne1917 (approved):
I see. The only potential inconvenience I can see is that a mod will have to be looking at the modlog for a specific community (as opposed to the whole site's modlog) if they want to filter by moderator. It's not a big enough deal to hold up this PR though.

dessalines:
Eventually when I add `canMod` https://github.com/LemmyNet/lemmy/issues/4365 , that should help with this.

PR #2643, opened 2024-07-30
Remove min and max length from password input when using login form

### LINKED ISSUE #2381: [Bug]:  New password length requirements breaks login for users
### Requirements

- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

```
lemmy-ui:0.19.1
lemmy:0.19.1
```

In an older version users could register with passwords <10 chars, in the new version frontend code you can only login with 10+ chars. This means users with old shorter passwords cannot login to the board. 

Please add a fix, or a config option to make the password min length shorter so these users may login. Thanks

### Steps to Reproduce

As above.

### Technical Details

As above.

### Version

0.19.1

### Lemmy Instance URL

_No response_

SleeplessOne1917:
Fixes #2381 using @richardARPANET 's suggestion.

Something I noticed working on this is that whitespace is allowed in passwords. This doesn't seem a problem in and of itself, but it's possible to make a password entirely from whitespace. Should this be allowed?

### DIFF
--- a/src/shared/components/common/password-input.tsx
+++ b/src/shared/components/common/password-input.tsx
@@ -99,8 +99,10 @@ class PasswordInput extends Component<PasswordInputProps, PasswordInputState> {
                 onInput={onInput}
                 value={value}
                 required
-                pattern=".{10,60}"
+                pattern=".+"
                 title={I18NextService.i18n.t("invalid_password")}
+                minLength={isNew ? 10 : undefined}
+                maxLength={isNew ? 60 : undefined}
               />
               <button
                 className="btn btn-outline-dark"


src/shared/components/common/password-input.tsx
```diff
@@ -99,8 +99,10 @@ class PasswordInput extends Component<PasswordInputProps, PasswordInputState> {
                 onInput={onInput}
                 value={value}
                 required
-                pattern=".{10,60}"
+                pattern=".+"
```
SleeplessOne1917, line 102: I changed the pattern since length is all we're checking for and min and max length gives more information to users as to what is wrong than using the pattern.

dessalines (approved):

PR #2647, opened 2024-08-01
Fix tribute related bug

SleeplessOne1917:
## Description
A recent change caused the frontend build to crash with this error:

```
lemmy-ui-1  | Node.js v20.16.0
lemmy-ui-1  | /app/node_modules/.pnpm/tributejs@5.1.3/node_modules/tributejs/dist/tribute.js:98
lemmy-ui-1  |   if (window && typeof window.CustomEvent !== "function") {
lemmy-ui-1  |   ^
lemmy-ui-1  | 
lemmy-ui-1  | ReferenceError: window is not defined
lemmy-ui-1  |     at /app/node_modules/.pnpm/tributejs@5.1.3/node_modules/tributejs/dist/tribute.js:98:3
lemmy-ui-1  |     at /app/node_modules/.pnpm/tributejs@5.1.3/node_modules/tributejs/dist/tribute.js:2:83
lemmy-ui-1  |     at Object.<anonymous> (/app/node_modules/.pnpm/tributejs@5.1.3/node_modules/tributejs/dist/tribute.js:5:2)
lemmy-ui-1  |     at Module._compile (node:internal/modules/cjs/loader:1358:14)
lemmy-ui-1  |     at Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
lemmy-ui-1  |     at Module.load (node:internal/modules/cjs/loader:1208:32)
lemmy-ui-1  |     at Module._load (node:internal/modules/cjs/loader:1024:12)
lemmy-ui-1  |     at Module.require (node:internal/modules/cjs/loader:1233:19)
lemmy-ui-1  |     at require (node:internal/modules/helpers:179:18)
lemmy-ui-1  |     at Object.tributejs (/app/dist/js/server.js:4103:18)
```

This change fixes the issue by only importing tributejs when the app is being run in the browser.


package.json
```diff
@@ -60,7 +60,7 @@
     "inferno-router": "^8.2.3",
     "inferno-server": "^8.2.3",
     "jwt-decode": "^4.0.0",
-    "lemmy-js-client": "0.19.4",
+    "lemmy-js-client": "0.20.0-alpha.1",
```
dessalines, line 63: This one is causing the error.

src/shared/markdown.ts
```diff
@@ -38,6 +40,10 @@ export let customEmojisLookup: Map<string, CustomEmojiView> = new Map<
   CustomEmojiView
 >();
 
+if (isBrowser()) {
+  Tribute = await import("tributejs");
+}
+
```
dessalines, line 46: We might just need to add an exception for the require rule, because I don't see any other way around it. Tribute should be smart enough to null check the window object for SSR apps, but it doesn't.

package.json
SleeplessOne1917: I was able to run the app just fine after making the change to `markdown.ts`.

dessalines (approved):

PR #2649, opened 2024-08-03
More moderation history

dessalines:
## Description

Adds ability to view a users moderation history for their item.

Also fixed some tribute issues.

### After

Screenshot_20240803_103300_Via_1_1

Profile page:

Screenshot_20240803_105437_Via

### DIFF
--- a/src/shared/components/common/content-actions/content-action-dropdown.tsx
+++ b/src/shared/components/common/content-actions/content-action-dropdown.tsx
@@ -159,6 +159,26 @@ export default class ContentActionDropdown extends Component<
       (amMod(community.id) || (amAdmin() && community.local)) &&
       !creator_banned_from_community;
 
+    const modHistoryUserTranslation = I18NextService.i18n.t(
+      "user_moderation_history",
+      { user: creator.name },
+    );
+
+    // The link and translation string for the item
+    const { modHistoryItemLink, modHistoryItemTranslation } =
+      type === "post"
+        ? {
+            modHistoryItemLink: `/modlog?postId=${id}`,
+            modHistoryItemTranslation: I18NextService.i18n.t(
+              "post_moderation_history",
+            ),
+          }
+        : {
+            modHistoryItemLink: `/modlog?commentId"=${id}`,
+            modHistoryItemTranslation: I18NextService.i18n.t(
+              "comment_moderation_history",
+            ),
+          };
     return (
       <>
         {type === "comment" && (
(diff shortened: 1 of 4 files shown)


SleeplessOne1917:
I thought we were making this viewable from the profile page?

dessalines:
I suppose I could add it there also.

dessalines:
K added it to the profile page also for admins, pic above.

src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -159,6 +159,26 @@ export default class ContentActionDropdown extends Component<
...
+            modHistoryItemTranslation: I18NextService.i18n.t(
+              "post_moderation_history",
+            ),
+          }
+        : {
+            modHistoryItemLink: `/modlog?commentId=${id}`,
+            modHistoryItemTranslation: I18NextService.i18n.t(
+              "comment_moderation_history",
+            ),
+          };
```
SleeplessOne1917, line 181: Are moderation histories for specific posts and comments something admins use? I can't imagine that any given post or comment will be removed and restored so many times that this would be useful.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -91,26 +89,25 @@ export class MarkdownTextArea extends Component<
...
+        textarea.addEventListener("tribute-replaced", () => {
+          this.setState({ content: textarea.value });
+          autosize.update(textarea);
+        });
+
+        this.quoteInsert();
+
+        if (this.props.focus) {
+          textarea.focus();
+        }
```
SleeplessOne1917, line 110: I like this a lot more than how tributes are currently handled.

SleeplessOne1917 (commented):
I'll test this out locally in about an hour or so when I have time.

src/shared/components/common/content-actions/content-action-dropdown.tsx
dessalines: hrm... Ya I think it could be useful to see things that have been removed / restored multiple times. But more importantly, there might be more recent mod actions done on a person, and you might only want to see the mod history for that item, rather than going to the person and having to scroll through. Like a year old comment or something.

src/shared/components/common/markdown-textarea.tsx
dessalines: It works but I'm also a bit scared that because this isn't a global import anymore, it might be doing a lot of work to re-import tribute for every single comment.

SleeplessOne1917: You might be able to cache it in memory by making it a module scoped variable in `markdown.ts`. You can make it `Translation | undefined` (you can import only types with `import { default as TranslationType } from "tributejs"`, which shouldn't cause issues at runtime), check if it already exists in `setupTribute`, and only do the dynamic import if it doesn't exist.

dessalines: I updated it now, I wasn't able to type it because it uses a lot of generics unfortunately. But I added a console debug message, and it seems to only run once, and correctly.

SleeplessOne1917:
Just tested this. It works. Gonna merge.

SleeplessOne1917 (approved):

PR #2654, opened 2024-08-07
Add ability to fill magnet link title on post creation.

dessalines:
## Description

This adds the ability to copy the download name from a torrent magnet link on post creation.

## Screenshots

Screenshot_20240807_131550_Via

### DIFF
--- a/src/shared/markdown.ts
+++ b/src/shared/markdown.ts
@@ -336,7 +336,8 @@ export function getEmojiMart(
 }
 
 export async function setupTribute() {
-  if (Tribute === null) {
+  // eslint-disable-next-line eqeqeq
+  if (Tribute == null) {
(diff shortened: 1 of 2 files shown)


src/shared/markdown.ts
```diff
@@ -336,7 +336,8 @@ export function getEmojiMart(
 }
 
 export async function setupTribute() {
-  if (Tribute === null) {
+  // eslint-disable-next-line eqeqeq
+  if (Tribute == null) {
```
SleeplessOne1917, line 340: Was strict equals causing a bug?

dessalines: Yep, weirdly enough. It couldn't handle it as an import type, and I got a console error.

dessalines:
Gonna merge this as I'm getting another beta out.

PR #2669, opened 2024-08-25
Enable spellcheck for markdown text area

SleeplessOne1917:
I occasionally misspell words, so it can be annoying when the post/comment text areas for Lemmy don't catch that. This PR adds spellchecking functionality.

I also bumped the dependencies since 0.19.6 is going to release relatively soon.


src/shared/components/common/markdown-textarea.tsx
```diff
@@ -223,6 +223,7 @@ export class MarkdownTextArea extends Component<
                     this.props.maxLength ?? markdownFieldCharacterLimit
                   }
                   placeholder={this.props.placeholder}
+                  spellCheck
```
SleeplessOne1917, line 226: Note that if the user has spellcheck disabled in their browser settings, this will not override it. In other words, this will only work if the user has spellcheck enabled for their browser.

dessalines: Nice, that's simple enough.

dessalines (approved):

PR #2674, opened 2024-08-26
Fix memory leak around emojis on server render

makotech222:
## Description

On serverside render, this customEmojis array gets appended to on every request. This eventually leads to Node.js killing the process for using too much memory.

### DIFF
--- a/src/shared/markdown.ts
+++ b/src/shared/markdown.ts
@@ -252,6 +252,7 @@ export function setupEmojiDataModel(custom_emoji_views: CustomEmojiView[]) {
     custom_emoji_views,
     x => x.custom_emoji.category,
   );
+  customEmojis.length = 0;
   for (const [category, emojis] of Object.entries(groupedEmojis)) {
     customEmojis.push({
       id: category,


src/shared/markdown.ts
```diff
@@ -252,6 +252,7 @@ export function setupEmojiDataModel(custom_emoji_views: CustomEmojiView[]) {
     custom_emoji_views,
     x => x.custom_emoji.category,
   );
+  customEmojis.length = 0;
```
SleeplessOne1917, line 255: Instead of setting the length to 0, it's probably better to check if `customEmojis.length` is equal to zero and only run the logic from lines 256-270 when the length is greater than 0. This should remove the need to constantly deallocate and reallocate all of the stuff in the existing array.

I do fear that my suggested approach would make it so emojis that are added while the process is running won't be picked up during SSR (should be fine for client side rendering), however. If you discover this to be the case during your testing, I'll have to hit the drawing board again.

makotech222: feel free to iterate on it; i'm just upstreaming this fix from our hexbear branch. I don't really have the time to refactor the whole thing.

dessalines (approved):

src/shared/markdown.ts
dessalines: This whole section does need a big refactor, but it would probably be a waste of effort considering we're going to be moving to lemmy-ui-leptos eventually anyway.

PR #2693, opened 2024-09-18
Post scheduling

Nutomic:
## Description

Adds simple input for scheduling a post x minutes in the future. Do you want to pull in a date picker library for this? If yes, is there a specific library that you prefer?

Despite the tag name, this feature can easily be backported to 0.19.6

See https://github.com/LemmyNet/lemmy/pull/5025
### Before
Screenshot_20240918_125644

### After

Screenshot_20240918_125529

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "inferno-router": "^8.2.3",
     "inferno-server": "^8.2.3",
     "jwt-decode": "^4.0.0",
-    "lemmy-js-client": "0.19.6-beta.1",
+    "lemmy-js-client": "0.20.0-alpha.6",
     "lodash.isequal": "^4.5.0",
     "markdown-it": "^14.1.0",
     "markdown-it-bidi": "^0.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -37,7 +37,7 @@ importers:
         version: 6.0.1
       babel-loader:
         specifier: ^9.1.3
-        version: 9.1.3(@babel/core@7.25.2)(webpack@5.94.0(webpack-cli@5.1.4))
+        version: 9.2.1(@babel/core@7.25.2)(webpack@5.94.0(webpack-cli@5.1.4))
       babel-plugin-inferno:
         specifier: ^6.7.1
         version: 6.7.2(@babel/core@7.25.2)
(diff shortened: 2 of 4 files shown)


SleeplessOne1917:
I'll take a shot at making something for this later this week. A datepicker would definitely be nice.

Nutomic:
I can handle the implementation, just not sure which library would be best.

dessalines:
This browser native one is probably fine for now: https://www.w3schools.com/tags/att_input_type_datetime-local.asp

Nutomic:
Added the date picker.

dessalines:
Cool, I'll test this now.

dessalines (approved):
Would be good for @SleeplessOne1917 to look and possibly test also.

Nutomic:
After resolving conflicts in my lemmy-js-client PR and making a new version, there are lots of errors about SortType and tagline/custom emoji changes. Seems like those havent been applied to lemmy-ui yet.

dessalines:
No probs, I'll handle those in another PR.

SleeplessOne1917:
Sorry for not testing this earlier. I was busy with the SSO UI changes. I'll give this a test later tonight once I'm free.

SleeplessOne1917:
I tried testing this locally but am unable to build. Looks like there are some changes to the js client breaking some things.

Nutomic:
@SleeplessOne1917 Its because of other breaking changes which were added to lemmy-js-client. You can test with an earlier commit on this branch (80f124f76242a83e559203b4f5ee0672fdc5e1e7).

src/shared/utils/helpers/get-unix-time.ts
SleeplessOne1917, line 1: Since we already depend on `date-fns`, it could be worth using those to implement the unix time functions. [It has functions for unix time.](https://date-fns.org/v4.1.0/docs/fromUnixTime#)

SleeplessOne1917 (approved):
The change looks fine to me. If you don't feel like making the date-fns related changes, I could always tackle that at a later date.

src/shared/utils/helpers/get-unix-time.ts
Nutomic: Seems like this would be practically identical to the current code, either `fromUnixTime(timestamp)` or `new Date(timestamp)`. Or maybe Im missing something.

dessalines:
Once #2715 is done, a merge from main should fix CI for this one.

Nutomic:
Ready now, had to make some fixes to get edits working correctly.

dessalines (approved):

PR #2695, opened 2024-09-18
Adding a post title only filter to the search page.

dessalines:
- Also fixing restore data on unban.
- Accounting for newly removed block views.

Some other things:

- I didn't use a checkbox, because it didn't inline correctly or look great, so this uses a checkbox-button.
- I *could* make it so this filter only shows up when searching for posts only, but then you wouldn't be able to use it when searching for `All` . 

## Screenshots

This example has "grapes" in the body of a sample post, so when using the post title only filter, it returns nothing.

https://github.com/user-attachments/assets/59169151-fa40-4318-8ca8-d7cdddaa9dd2

cc @Nutomic

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -637,7 +637,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
   async handleBanFromCommunity({
     daysUntilExpires,
     reason,
-    shouldRemove,
+    shouldRemoveOrRestoreData,
   }: BanUpdateForm) {
     const {
       creator: { id: person_id },
@@ -646,10 +646,9 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     } = this.commentView;
 
     const ban = !creator_banned_from_community;
-
     // If its an unban, restore all their data
     if (ban === false) {
-      shouldRemove = false;
+      shouldRemoveOrRestoreData = true;
     }
     const expires = futureDaysToUnixTime(daysUntilExpires);
 
@@ -657,7 +656,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
       community_id,
       person_id,
       ban,
-      remove_data: shouldRemove,
+      remove_or_restore_data: shouldRemoveOrRestoreData,
       reason,
       expires,
     });
@@ -666,7 +665,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
   async handleBanFromSite({
     daysUntilExpires,
     reason,
-    shouldRemove,
+    shouldRemoveOrRestoreData,
   }: BanUpdateForm) {
     const {
       creator: { id: person_id, banned },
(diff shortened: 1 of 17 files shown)


Carlos-Cabello:
I think it would be more user friendly to have this be a part of the dropdown options. As a user, I see title and post as different entities, regardless of what is going on in the backend.

dessalines:
> As a user, I see title and post as different entities, regardless of what is going on in the backend.

These are different concepts (one is type, the other is a filter), and we can't move every filter to a single dropdown. 

The search page also has a community filter that doesn't apply to the person or community search, how would you include those in a single dropdown?

src/shared/utils/app/is-post-blocked.ts
```diff
@@ -6,12 +6,8 @@ export default function isPostBlocked(
...
 ): boolean {
   return (
-    (myUserInfo?.community_blocks
-      .map(c => c.community.id)
-      .includes(pv.community.id) ||
-      myUserInfo?.person_blocks
-        .map(p => p.target.id)
-        .includes(pv.creator.id)) ??
+    (myUserInfo?.community_blocks.map(c => c.id).includes(pv.community.id) ||
+      myUserInfo?.person_blocks.map(p => p.id).includes(pv.creator.id)) ??
```
SleeplessOne1917, line 10: Nitpick: this can be rewritten as:

```suggestion
    (myUserInfo?.community_blocks.some(c => c.id === pv.community.id) ||
      myUserInfo?.person_blocks.some(p => p.id === pv.creator.id)) ??
```

Sadly, array methods in JavaScript aren't lazy like iterator methods in Rust are; this code iterates over all of the blocks to create a new array of just the IDs, then iterates through the new ID array to see if it includes the array you're looking for. Using `some` makes it so a new array isn't created and prevents you from needing to iterate over the blocks more than once. I'm not sure if the performance impact of this is noticeable barring extreme cases, but it's worth keeping in mind.

src/shared/utils/helpers/get-bool-from-string.ts
```diff
@@ -0,0 +1,9 @@
+export default function getBoolFromString(
+  boolStr?: string,
+): boolean | undefined {
+  return boolStr
+    ? boolStr.toLowerCase() === "true"
+      ? true
+      : false
+    : undefined;
+}
```
SleeplessOne1917, line 9: I think this would be cleaner as:

```suggestion
export default (boolStr?: string) => boolStr?.toLowerCase() === "true";
```

This assumes `undefined` behaves the same as `false` for our purposes.  If being able to return `undefined` is important, you could also do the following that's not too far off from what you have:

```suggestion
export default (boolStr?: string) =>
  boolStr ? boolStr.toLowerCase() === "true" : undefined;
```

I know you're not a big fan of using arrow functions outside of passing functions as arguments, but I think they look pretty clean when the entire function is just evaluating one statement, especially a simple one like this.

SleeplessOne1917 (changes requested):
My only hard block before approving is changing the name of the state property in `mod-action-form-modal.tsx`. Other than that, is this in a stable enough place to test locally? There's been a lot of churn in the codebase with all of the breaking changes being merged into main.

src/shared/components/comment/comment-node.tsx
```diff
@@ -646,18 +646,17 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     } = this.commentView;
 
     const ban = !creator_banned_from_community;
-
     // If its an unban, restore all their data
     if (ban === false) {
-      shouldRemove = false;
+      shouldRemoveOrRestoreData = true;
```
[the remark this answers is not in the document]
dessalines: That's correct, true means remove data if not banned, restore if banned. I did test this one locally and it works.

src/shared/utils/app/is-post-blocked.ts
dessalines: Nice, thx.

src/shared/utils/helpers/get-bool-from-string.ts
dessalines: When I did this, typescript lost all return type information for the function. I'll update it a bit tho.

PR #2715, opened 2024-09-24
Fixing titleOnly, PostSort, and CommentSort.

dessalines:
This is still a WIP, since I also have to fix the new patterns for taglines, and custom emojis, which have their own endpoints now.


dessalines:
I did none of this, was all @matc-pub , thx! 

Doesn't really need a review cause it was handled in the PR to this one.

PR #2718, opened 2024-09-25
SortType, Tagline, Emojis

matc-pub:
Changes:
  - Settings: Only comment sort types as options for default comment sort type
  - Community/Home: Only comment sort types in comment view (uses default post sort type, ignores default comment sort type)
  - Post: Fallback to default comment sort type
  - TaglineForm/EmojisForm: Use new API
  - EmojisForm: Reloads custom emojis for emoji-mart when changing pages (after saving)
  - Tagline: Tagline only updates after page refresh
  - Custom emojis: Server doesn't recognize custom emojis, renders all emojis the same (client fixes that)
  - Custom emojis: Client fetches emojis before hydration
  - style.css: em-emoji-picker gets 100% width to avoid emoji-mart flickering with dynamicWidth

This needs these changes to lemmy-js-client 0.20.0-alpha.15
```diff
diff --git a/src/http.ts b/src/http.ts
index 51cb755..a9cfe71 100644
--- a/src/http.ts
+++ b/src/http.ts
@@ -151,6 +151,8 @@ import { UpdateTagline } from "./types/UpdateTagline";
 import { DeleteTagline } from "./types/DeleteTagline";
 import { ListTaglines } from "./types/ListTaglines";
 import { ListTaglinesResponse } from "./types/ListTaglinesResponse";
+import { ListCustomEmojis } from "./types/ListCustomEmojis";
+import { ListCustomEmojisResponse } from "./types/ListCustomEmojisResponse";
 
 enum HttpType {
   Get = "GET",
@@ -1445,15 +1447,29 @@ export class LemmyHttp {
     );
   }
 
+  /**
+   * List custom emojis
+   *
+   * `HTTP.GET /custom_emoji/list`
+   */
+  listCustomEmojis(form: ListCustomEmojis) {
+    return this.#wrapper<ListCustomEmojis, ListCustomEmojisResponse>(
+      HttpType.Get,
+      "/custom_emoji/list",
+      form,
+    );
+  }
+
+
   /**
    * Create a new tagline
    *
-   * `HTTP.POST /tagline`
+   * `HTTP.POST /admin/tagline`
    */
   createTagline(form: CreateTagline) {
     return this.#wrapper<CreateTagline, TaglineResponse>(
       HttpType.Post,
-      "/tagline",
+      "/admin/tagline",
       form,
     );
   }
@@ -1461,12 +1477,12 @@ export class LemmyHttp {
   /**
    * Edit an existing tagline
    *
-   * `HTTP.PUT /tagline`
+   * `HTTP.PUT /admin/tagline`
    */
   editTagline(form: UpdateTagline) {
     return this.#wrapper<UpdateTagline, TaglineResponse>(
       HttpType.Put,
-      "/tagline",
+      "/admin/tagline",
       form,
     );
   }
@@ -1474,12 +1490,12 @@ export class LemmyHttp {
   /**
    * Delete a tagline
    *
-   * `HTTP.Post /tagline/delete`
+   * `HTTP.Post /admin/tagline/delete`
    */
   deleteTagline(form: DeleteTagline) {
     return this.#wrapper<DeleteTagline, SuccessResponse>(
       HttpType.Post,
-      "/tagline/delete",
+      "/admin/tagline/delete",
       form,
     );
   }
@@ -1487,12 +1503,12 @@ export class LemmyHttp {
   /**
    * List taglines
    *
-   * `HTTP.GET /tagline/list`
+   * `HTTP.GET /admin/tagline/list`
    */
   listTaglines(form: ListTaglines) {
     return this.#wrapper<ListTaglines, ListTaglinesResponse>(
       HttpType.Get,
-      "/tagline/list",
+      "/admin/tagline/list",
       form,
     );
   }
```


dessalines:
@matc-pub Could you do a PR to lemmy-js-client with those fixes? lmk when that's done and I'll do another alpha release of it.

dessalines:
@matc-pub Okay `0.20.0-alpha.16` is published.

src/shared/components/common/sort-select.tsx
```diff
@@ -109,6 +47,95 @@ export class SortSelect extends Component<SortSelectProps, SortSelectState> {
...
+        <option disabled aria-hidden="true">
+          {I18NextService.i18n.t("sort_type")}
+        </option>
+        <option value={"Hot"}>{I18NextService.i18n.t("hot")}</option>
+        <option value={"Top"}>{I18NextService.i18n.t("top")}</option>
+        <option value={"Controversial"}>
+          {I18NextService.i18n.t("controversial")}
+        </option>
+        <option value={"New"}>{I18NextService.i18n.t("new")}</option>
+        <option value={"Old"}>{I18NextService.i18n.t("old")}</option>
```
SleeplessOne1917, line 69: The values for these don't need the curly braces.

src/shared/components/common/sort-select.tsx
```diff
@@ -109,6 +47,95 @@ export class SortSelect extends Component<SortSelectProps, SortSelectState> {
...
+        {!this.props.hideHot && [
+          <option key={"Hot"} value={"Hot"}>
+            {I18NextService.i18n.t("hot")}
+          </option>,
+          <option key={"Active"} value={"Active"}>
+            {I18NextService.i18n.t("active")}
+          </option>,
+          <option key={"Scaled"} value={"Scaled"}>
+            {I18NextService.i18n.t("scaled")}
+          </option>,
```
SleeplessOne1917, line 96: [Same as here.](https://github.com/LemmyNet/lemmy-ui/pull/2718/files#r1779737803)

src/shared/components/common/sort-select.tsx
```diff
@@ -109,6 +47,95 @@ export class SortSelect extends Component<SortSelectProps, SortSelectState> {
...
+          {I18NextService.i18n.t("top_three_months")}
+        </option>
+        <option value={"TopSixMonths"}>
+          {I18NextService.i18n.t("top_six_months")}
+        </option>
+        <option value={"TopNineMonths"}>
+          {I18NextService.i18n.t("top_nine_months")}
+        </option>
+        <option value={"TopYear"}>{I18NextService.i18n.t("top_year")}</option>
+        <option value={"TopAll"}>{I18NextService.i18n.t("top_all")}</option>
```
SleeplessOne1917, line 134: [Same as here.](https://github.com/LemmyNet/lemmy-ui/pull/2718/files#r1779737803)

src/shared/components/community/community.tsx
```diff
@@ -171,12 +174,11 @@ function getDataTypeFromQuery(type?: string): DataType {
...
 }
 
-function getSortTypeFromQuery(type?: string): SortType {
-  const mySortType =
-    UserService.Instance.myUserInfo?.local_user_view.local_user
-      .default_sort_type;
-
-  return type ? (type as SortType) : (mySortType ?? "Active");
+function getSortTypeFromQuery(
+  type: string | undefined,
```
SleeplessOne1917, line 178: Can be `type?: string,`

src/shared/components/home/emojis-form.tsx
```diff
@@ -279,37 +298,80 @@ export class EmojiForm extends Component<EmojiFormProps, EmojiFormState> {
...
+          }
+          if (a.custom_emoji.category > b.custom_emoji.category) {
+            return 1;
+          }
+          if (a.custom_emoji.shortcode < b.custom_emoji.shortcode) {
+            return -1;
+          }
+          if (a.custom_emoji.shortcode > b.custom_emoji.shortcode) {
+            return 1;
+          }
```
SleeplessOne1917, line 337: Is it viable to lump these together? e.g.
```suggestion
          if (a.custom_emoji.category < b.custom_emoji.category || a.custom_emoji.shortcode < b.custom_emoji.shortcode) {
            return -1;
          }
          if (a.custom_emoji.category > b.custom_emoji.category || a.custom_emoji.shortcode > b.custom_emoji.shortcode) {
            return 1;
          }
```

src/shared/components/home/emojis-form.tsx
```diff
@@ -279,37 +298,80 @@ export class EmojiForm extends Component<EmojiFormProps, EmojiFormState> {
...
-        this.scrollRef[view.custom_emoji.shortcode].scrollIntoView();
+  async handleEmojiClick(e: any) {
+    const emojiIndex = this.state.allEmojis.findIndex(
+      x => x.custom_emoji.shortcode === e.id,
+    );
+    if (emojiIndex >= 0) {
+      const { shortcode } = this.state.allEmojis[emojiIndex].custom_emoji;
+      const page = Math.floor(emojiIndex / this.itemsPerPage) + 1;
+      await this.handlePageChange(page);
+      await new Promise(r => setTimeout(r));
```
SleeplessOne1917, line 370: What is this line doing?

matc-pub: The timeout gives inferno enough time to add the new elements to the DOM.

src/shared/components/home/emojis-form.tsx
matc-pub: It should compare categories first, and when equal compare shortcodes. Changed it to use String localeCompare.

SleeplessOne1917 (approved):

src/shared/components/common/sort-select.tsx
dessalines, line 1: I'm thinking we should just make separate components for a post sort select, and a comment sort select.

src/shared/components/home/tagline-form.tsx
```diff
@@ -1,58 +1,117 @@
...
         <h1 className="h4 mb-4">{I18NextService.i18n.t("taglines")}</h1>
         <div className="table-responsive col-12">
-          <table id="taglines_table" className="table table-sm table-hover">
+          <table
+            id="taglines_table"
+            className="table table-sm table-hover align-middle"
+          >
             <thead className="pointer">
               <th></th>
+              <th style="width:60px"></th>
```
dessalines, line 71: It'd be nice at some point to do away with every table in this codebase. They're not as responsive friendly (even with side-scrolling), and editing both emojis and taglines has been a pain on mobile.

I'll make an issue for that.

src/shared/components/home/tagline-form.tsx
```diff
@@ -99,65 +158,153 @@ export class TaglineForm extends Component<TaglineFormProps, TaglineFormState> {
...
 
   async handleSaveClick(i: TaglineForm) {
-    i.props.onSaveSite({
-      taglines: i.state.taglines,
-    });
+    const promises: Promise<any>[] = [];
+    for (const editable of i.state.taglines) {
+      if (editable.change === "update") {
+        promises.push(
+          HttpService.client.editTagline(editable.tagline).then(res => {
```
dessalines, line 238: This one to me doesn't seem like the right way to do it. here's

Admin settings screen -> Taglines component -> click to edit Tagline form.

The tagline form then pushes up (IE `onEdit(form: TaglineForm)`, or `onCreate(form: ..)` up the chain, up to the admin settings screen, which is the only one that should be making HTTP calls.

src/shared/markdown.ts
```diff
@@ -265,73 +270,34 @@ export function setupEmojiDataModel(custom_emoji_views: CustomEmojiView[]) {
...
-}
-
-export function removeFromEmojiDataModel(id: number) {
-  let view: CustomEmojiView | undefined;
-  for (const item of customEmojisLookup.values()) {
-    if (item.custom_emoji.id === id) {
-      view = item;
-      break;
-    }
+export async function setupEmojiDataModel(
```
dessalines, line 276: Nice, so much simpler now.

src/shared/utils/app/comment-to-post-sort-type.ts
```diff
@@ -0,0 +1,42 @@
+import { CommentSortType, PostSortType } from "lemmy-js-client";
+
+function assertType<T>(_: T) {}
+
+export default function commentToPostSortType(
+  sort: PostSortType | CommentSortType,
```
dessalines, line 6: Why does this need to handle either type? You don't need to convert from a PostSortType to a PostSortType.

src/shared/utils/app/post-to-comment-sort-type.ts
```diff
@@ -1,18 +1,45 @@
 import { CommentSortType, PostSortType } from "lemmy-js-client";
 
+function assertType<T>(_: T) {}
+
 export default function postToCommentSortType(
-  sort: PostSortType,
+  sort: PostSortType | CommentSortType,
```
dessalines, line 6: Same for this one

dessalines (commented):
This was huge, thx a ton. I tested and everything worked well.

We don't have to necessarily do the screen-level-only API action change I suggested below in this PR, but it would be good to do it at some point. lmk about my other comments below, then we can merge.

src/shared/components/common/sort-select.tsx
matc-pub: CommentSortSelect already exists. I changed SortSelect mostly back to how it was.

src/shared/components/home/tagline-form.tsx
matc-pub: The only downside to the way it is now, is that SSR doesn't render any actual data. But the emoji and tagline tabs are not visible until opening the tab. Moving the HTTP calls would only create properties with same signatures as the API methods. This way also the state for AdminSettings doesn't have to change, which means only one child has to rerender instead of all of them.

src/shared/utils/app/comment-to-post-sort-type.ts
matc-pub: I initially had `sort` for Home and Community as the mixed sort type.

dessalines (approved):
LGTM, thx.

src/shared/components/home/tagline-form.tsx
dessalines: That's true. Its not a blocker for this anyway, and we can always try to move these http calls to the top level screens later.

PR #2730, opened 2024-10-01
UX - Swap "Select Language" and "Cancel/Preview/Reply" button locations around in commentsReverse order of buttons in Reply TextArea

### LINKED ISSUE #1924: UX - Swap "Select Language" and "Cancel/Preview/Reply" button locations around in comments
### Requirements

Right now, when you click the Reply icon under a comment, the Reply button is super far away:

image

This leads to the user needing to move their mouse a far distance. It would be better if the Reply button could be where the Select Language dropdown appears instead

Much love!

N4taaa:
Fixes #1924

## Description

Title. This solution simply changes the order of the elements in the code. A shorter way would be to make `flex-direction: row-reverse;` but seems a bit overkill...

New to the contribution world, please let me know of any errors ^^

### Before
Before

### After
After

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -255,7 +255,6 @@ export class MarkdownTextArea extends Component<
               </label>
             </div>
           </div>
-        </div>
 
           <div className="col-12 d-flex align-items-center flex-wrap mt-2">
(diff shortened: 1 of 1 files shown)


N4taaa:
Works on my machine™, not sure why CI is acting up...

SleeplessOne1917 (approved):

PR #2578, opened 2024-07-02
SSO support

privacyguard:
Implements https://github.com/LemmyNet/lemmy/issues/2930.

This PR is based on https://github.com/LemmyNet/lemmy/pull/4238 and https://github.com/LemmyNet/lemmy-ui/pull/2278 by @thepaperpilot.

We noticed that the original PR is outdated and has a lot of conflicts with the recent changes. We tried to keep the previous commits whenever possible (in lemmy-js-client and lemmy-ui).

### How is works?

Admins can configure external OIDC providers from within the admin settings.
Once an OIDC provider is configured, users will be able to Sign In / Sign Up using external OIDC providers.
Available Configuration

### The usual OIDC endpoints
auto_verify_email: When enabled, users signing up using OIDC won't need to go through email verification.
auto_approve_application: When enabled, users signing up using OIDC won't need manual approval even if applications are required.
account_linking_enabled: When enabled, users attempting with sign up with OIDC using an existing user email would link the OIDC account to the existing user.
Disclaimer
This is our first ever rust contribution.

### Who we are? Why are we contributing to Lemmy?
[Privacy Portal](https://privacyportal.org/) is an OIDC provider and an email aliasing service focused on privacy. We have decided to contribute to select open source projects that empower Free Speech online.
Our OIDC provider services are currently offered free of charge. In the future, we will have a generous free plan that will cover most deployments.
Using Privacy Portal as your OIDC provider offers your users great privacy benefits. User emails will automatically get replaced by single-purpose Privacy Aliases during sign up. Users will be able to enter any name (to be used as username). Users can benefit from email encryption and much more.

PR #2731, opened 2024-10-02
SSO Support - alternate UI

SleeplessOne1917:
The main difference between this and #2578 - UI wise at least - is the admin tab for managing SSO providers. Before, is was a table of forms. Since the forms had many fields, even 1 or 2 providers caused a lot of cognitive load trying to read through. In this PR, existing providers are in a list with collapsable entries, and their values are listed in a more space-efficient list of key-value pairs. Adding and editing providers is handled with a modal now.

There are stilll several changes that need to be done. Chief among them is the need for translations; best to hold off on merging until those are decided on and provided. There is also the matter of helpful information about adding OAuth providers that I raised in the other PR. I'm unsure what approach would be best here: Little help circles like we use with some fields in other parts of the UI? A link to a learning resource so admins know what the values they're setting do (perhaps privacy portal has a good resource for this, or maybe we can add something about it to the Lemmy docs)? Some combination of both?

cc. @privacyguard


dessalines (commented):
The UI seems fine, but I don't know how to test this properly.

And ya we'll have to wait on @privacyguard for the correct i18n translations of these ones: which should get added as a PR to [this file](https://github.com/LemmyNet/lemmy-translations/blob/main/translations/en.json)

dessalines:
> There is also the matter of helpful information about adding OAuth providers that I raised in the other PR. I'm unsure what approach would be best here: Little help circles like we use with some fields in other parts of the UI?

I think linking to an external page describing these would be best.

SleeplessOne1917:
> There are minor UI improvements that could be made such as text overflow.

Where are you seeing the text overflow issues?

> Concerning the use of the Modal to login, it's not very common except by apps that usually have a large number of providers configured. Do we expect Lemmy admins to have 5+ providers configured? Otherwise it would be much cleaner to simply have the providers directly in the page (as a single column under the user/pass login). That said, this is a style preference so you guys should decide on that. We're just giving our opinion.

I am not sure how many providers the average instance would have configured tbh. When there are fewer I agree that that the modal is less clean. The non-modal starts to messier once the number of providers gets big but, as stated, I'm not sure what a realistic number would be. I'm fine not using the modal, though I think they should be in a section with their own heading.

> I think linking to an external page describing these would be best.

Makes sense. We should probably add a page for it to the Lemmy docs before releasing 0.20.0.

> Concerning the translations, is this PR blocked by the translations? Could it wait 2-3 days?

We usually handle translations for a feature as part of the same PR that adds the feature, although strictly speaking it is not. I find that having them be added as part of the same PR makes it easier to track which translations are new in case we forget to add one.

dessalines (approved):
I'll try to get this deployed to voyager.lemmy.ml soon so we can test it.

PR #2755, opened 2024-10-15
Adding mermaid.js support via markdown-it.

### LINKED ISSUE #2754: Add support for Mermaid.js diagrams
### Discussed in https://github.com/LemmyNet/lemmy-ui/discussions/2735

There is a markdown-it plugin for mermaid: https://github.com/wekan/markdown-it-mermaid , that I'll add. Any bugs related to it tho, you'll have to open up over there.

cc @YousufSSyed

dessalines:
- Fixes #2754

EDIT: nm, this doesn't work currently because the library doesn't support ESM : 

`Error [ERR_REQUIRE_ESM]: require() of ES Module /home/xxx/git/lemmy-ui/node_modules/.pnpm/mermaid@10.9.2/node_modules/mermaid/dist/mermaid.core.mjs not supported.`

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@babel/preset-typescript": "^7.24.7",
     "@babel/runtime": "^7.25.4",
     "@emoji-mart/data": "^1.2.1",
+    "@wekanteam/markdown-it-mermaid": "^0.6.4",
     "autosize": "^6.0.1",
     "babel-loader": "^9.1.3",
     "babel-plugin-inferno": "^6.7.1",
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


Nutomic:
This is for some sort of diagram rendering? I would rather not add that because its pretty much out of scope, and then all other frontends and apps would also have to add support for this feature.

dessalines:
Yep. [Here's some more cool ones that hedgedoc for example supports in their markdown.](https://demo.hedgedoc.org/features#MathJax)

I think its fine if we have these markdown add-ons, as long as its understood that we won't maintain the libraries ourselves, and that any apps will have to do their own work supporting them. It could also be kind of a cool incentive / project for people to start adding these renderers to other platforms, like android / iOS.

dessalines:
Closing as even if this were added, it adds to much bloat.

PR #2780, opened 2024-10-29
Allow Arabic and Cyrillic characters when signing up or creating community

### LINKED ISSUE #2206: Cannot create arabic usernames
### Requirements

### Summary

Cannot create arabic username because of the pattern validation in html
Screenshot 2023-10-28 at 23 48 25

### Steps to Reproduce

1. got to Signup page
2. type an arabic username
3. you should see the error

### Technical Details

applies to all devices

### Lemmy Instance Version

dev

### Lemmy Instance URL

_No response_

SleeplessOne1917:
Fixes #2206

Going off of [`is_valid_actor_name`](https://github.com/LemmyNet/lemmy/blob/cdc1cf3bf7afb8406d5eb55ee3a4f8adfebae71c/crates/utils/src/utils/validation.rs#L87) in the backend's codebase, Cyrillic characters should also be allowed.

### DIFF
--- a/src/shared/components/community/community-form.tsx
+++ b/src/shared/components/community/community-form.tsx
@@ -14,6 +14,7 @@ import { ImageUploadForm } from "../common/image-upload-form";
 import { LanguageSelect } from "../common/language-select";
 import { MarkdownTextArea } from "../common/markdown-textarea";
 import { tippyMixin } from "../mixins/tippy-mixin";
+import { validActorRegexPattern } from "../../config";
 
 interface CommunityFormProps {
   community_view?: CommunityView; // If a community is given, that means this is an edit
@@ -129,7 +130,7 @@ export class CommunityForm extends Component<
                 onInput={linkEvent(this, this.handleCommunityNameChange)}
                 required
                 minLength={3}
-                pattern="[a-z0-9_]+"
+                pattern={validActorRegexPattern}
                 title={I18NextService.i18n.t("community_reqs")}
               />
             </div>
--- a/src/shared/config.ts
+++ b/src/shared/config.ts
@@ -52,4 +52,4 @@ export const instanceLinkRegex = new RegExp(
 export const testHost = "0.0.0.0:8536";
 
 export const validActorRegexPattern =
-  "^\\w+|[\\p{Script=Arabic}d_]+|[\\p{Script=Cyrillic}d_]+$";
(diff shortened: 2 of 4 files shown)


src/shared/config.ts
```diff
@@ -50,3 +50,6 @@ export const instanceLinkRegex = new RegExp(
 );
 
 export const testHost = "0.0.0.0:8536";
+
+export const validActorRegexPattern =
+  "^\\w+|[\\p{Script=Arabic}\\d_]+|[\\p{Script=Cyrillic}\\d_]+$";
```
SleeplessOne1917, line 55: `\w` is equivalent to `[a-zA-Z0-9_]`. `\d` is equivalent to `[0-9]`.

src/shared/components/community/community-form.tsx
```diff
@@ -129,7 +130,7 @@ export class CommunityForm extends Component<
                 onInput={linkEvent(this, this.handleCommunityNameChange)}
                 required
                 minLength={3}
-                pattern="[a-z0-9_]+"
+                pattern={validActorRegexPattern}
```
SleeplessOne1917, line 133: While the frontend is preventing the use of uppercase letters for community names, `is_valid_actor_name` on the backend, which is also used for checking community names on creation, still allows them. If preventing uppercase characters is a requirement, a backend change will be needed.

src/shared/config.ts
Nutomic: In the backend we currently have three separate regexes for this, so you can also change it to this single regex.

src/shared/components/community/community-form.tsx
dessalines: Yep that should be done also, although I do see this unique index on the community table.

`"idx_community_lower_actor_id" UNIQUE, btree (lower(actor_id::text))`

dessalines (approved):

src/shared/components/community/community-form.tsx
SleeplessOne1917: Out of curiosity, why is that a btree? Are actor IDs organized in a tree structure?

dessalines: It's the default postgres index type, you can read more about the others here: https://www.postgresql.org/docs/current/indexes-types.html

PR #2781, opened 2024-10-30
Renovate should only apply security updates

Nutomic:
This can avoid accidental breakage from updates like in https://github.com/LemmyNet/joinlemmy-site/issues/431.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,5 +3,15 @@
   "extends": ["config:recommended"],
   "schedule": ["every weekend"],
   "automerge": true,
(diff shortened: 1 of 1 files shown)


dessalines:
I understand doing that with join-lemmy.org, because we automatically git pull and update that site every day in cron. But with lemmy-ui we do lots of test builds before ever doing a release, and renovate helps us keep everything up to date, and find and fix breaking changes sooner rather than letting them build up.

Nutomic:
Alright

PR #2797, opened 2024-11-07
Fix fullsize post image

dessalines:
## Description

This fixes the url used for image expands, to the fullsize one, not the one in the thumbnail column.

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -273,18 +273,22 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
   }
 
   get img() {
-    const { post } = this.postView;
-
     if (this.isoData.showAdultConsentModal) {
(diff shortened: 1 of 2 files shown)


src/shared/components/post/post-listing.tsx
```diff
@@ -315,26 +319,12 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     );
   }
 
-  get imageSrc(): string | undefined {
-    const post = this.postView.post;
-    const url = post.url;
-    const thumbnail = post.thumbnail_url;
-
-    if (thumbnail) {
-      return thumbnail;
```
dessalines, line 324: The culprit was here. It was early returning thumbnail, even for full size images. I got rid of this whole function to be more explicit, and moved the logic into the specific functions.

SleeplessOne1917 (approved):

Issue #2891, opened 2025-01-08
Use <main> on more pages or use role="main" on more pages (Accessibility, ARIA Landmarks)


travis-jeans:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

This is half way between a bug and an enhancement. I say it's a bug because it will cause "buggy" behaviour for screenreader users who don't realize the cause of the Landmark inconsistencies.

### Issue:
Other than the homepage which has `<main class="col-12 col-md-8 col-lg-9" role="main">` other pages like Admin Settings for example are missing the `<main>` element and a `role="main"` instead having a custom class on a `<div>` like `.admin-settings`. Others include `.person-profile`, `.inbox`, `.person-reports`, `.registration-applications`, and there might be more I missed. An example code snippet for the Admin Settings page is `<div class="admin-settings container-lg">`.

#### Theme development
From the perspective of a theme developer this means targeting the main area of the document and excluding the header navigation and footer means targeting `<main>` plus a series of custom classes for different pages as defined above. It is inefficient.

#### Accessibility
Although the skip link works properly, the missing `main` and `role="main"` means screenreader users will have an inconsistent experience - they can quickly navigate to main on the homepage but for other pages such a landmark is missing. It would be better if the main element was on every page with the custom classes like `.person-profile` for profile pages added to the class of main with the `role="main"`. This would be the best option. More about this in Using ARIA landmarks to identify regions of a page [(WCAG 2.1 Techniques)](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11).

An intermediate step would be to add `role="main"` to the `div`s with custom classes like `.admin-settings` so the `div`s are identified as landmarks by screenreaders.

My suggestion is to add `<main>` to more all pages for a consistent screenreader experience plus to make theme development easier by allowing developers to target CSS changes to `main` rather than `main` plus a lot of other classes.

### Steps to Reproduce

1. Go to any page other than the home feed
2. Open view-source
3. Ctrl+F search for "<main" or "role="main"
4. Notice the missing <main> element

### Technical Details

[...]

PR #2944, opened 2025-02-06
Improve screenreader accessibility by adding a main element to more pages (issue #2891)

travis-jeans:
## Description

Changes a `div` element that marks the main content area to a `main` element. Related to issue #2891.

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -203,7 +203,7 @@ export class Community extends Component<CommunityRouteProps, State> {
     showSidebarMobile: false,
     isIsomorphic: false,
   };
-  private readonly mainContentRef: RefObject<HTMLElement>;
+  private readonly mainContentRef: RefObject<HTMLDivElement>;
 
   loadingSettled() {
     return resourcesSettled([
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -422,7 +422,7 @@ export class Community extends Component<CommunityRouteProps, State> {
     return (
       <div className="community container-lg">
         <div className="row">
-          <main className="col-12 col-md-8 col-lg-9" ref={this.mainContentRef}>
+          <div className="col-12 col-md-8 col-lg-9" ref={this.mainContentRef}>
             {this.renderCommunity()}
             {this.selects()}
             {this.listings()}
@@ -430,7 +430,7 @@ export class Community extends Component<CommunityRouteProps, State> {
               nextPage={this.getNextPage}
               onNext={this.handlePageNext}
             />
-          </main>
+          </div>
           <aside className="d-none d-md-block col-md-4 col-lg-3">
             {this.sidebar()}
           </aside>
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -1,5 +1,5 @@
 variables:
-  - &install_pnpm "corepack enable pnpm"
+  - &install_pnpm "npm i -g corepack && corepack enable pnpm"
 
 steps:
   fetch_git_submodules:
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,11 @@ FROM node:alpine as builder
 # - It can break depending on the CPU (https://github.com/LemmyNet/lemmy-ui/issues/1566)
 RUN apk update && apk upgrade && apk add --no-cache curl python3 build-base gcc wget git vips-dev pkgconfig
 
+# Install node-gyp and corepack
+RUN npm install -g node-gyp corepack
+
 # Enable corepack to use pnpm
 RUN corepack enable
-
-# Install node-gyp
-RUN npm install -g node-gyp
 
 WORKDIR /usr/src/app
 
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ The official web app for [Lemmy](https://github.com/LemmyNet/lemmy), written in 
 
 Based off of MrFoxPro's [inferno-isomorphic-template](https://github.com/MrFoxPro/inferno-isomorphic-template).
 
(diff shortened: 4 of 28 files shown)


SleeplessOne1917 (changes requested):
I think the `main` element should be used a bit further down the component tree (the components used by the routes) like is done in [`community.tsx`](https://github.com/LemmyNet/lemmy-ui/blob/d4d66ecf5683278fa5cd581744b966dc8b51f24d/src/shared/components/community/community.tsx#L425) or [`home.tsx`](https://github.com/LemmyNet/lemmy-ui/blob/d4d66ecf5683278fa5cd581744b966dc8b51f24d/src/shared/components/home/home.tsx#L409). That, or existing uses of the `main` element should be pulled out and the only one be put here in `app.tsx`.

While I have difficulty finding best practices for correct ways to arrange semantic elements, a point I've come across very consistently is that there should only ever be one `main` element on the page. As this PR is now, some of the pages will end up with 2 `main` elements, likely replacing one difficulty for another for screen reader users on the affected pages.

travis-jeans:
> While I have difficulty finding best practices for correct ways to arrange semantic elements, a point I've come across very consistently is that there should only ever be one main element on the page. As this PR is now, some of the pages will end up with 2 main elements, likely replacing one difficulty for another for screen reader users on the affected pages.

You're right that each page should only have one `<main>` element. Is there an easy way to tell which pages these are?

dessalines:
Just using grep or ripgrep on each page that you're editing.

SleeplessOne1917:
Besides grep, many code editors have a function for searching text in all the files of a project. VS Code, for example, has this functionality.

travis-jeans:
## Regular expressions
While in `/lemmy-ui/` I ran `grep -o -i -E -r "<main|</main"` to find the instances of opening and closing main tags. 

## Commits
### Main
The only one I was absolutely sure had an extra `main` tag was home so I made that the first commit. 

### Post and Community
The others that showed up was post.tsx and community.tsx.
For post.tsx I could not find references to the `main class="col-12` on the in browser in any of the test posts I had made. Same issue with the communities page. Overall, I'm not sure what these files were actually referring to, and if they should actually have main tags or div tags.

### Local Files
One issue with running grep is that I can't search the rendered HTML files this way. My understanding being new to Inferno, Typescript, and source maps is that the Javascript files render the HTML in the browser so I can't use grep on the server in the browser (as far as I know). 

### Debugging
Is there a way to save the rendered html files to a local directory as a part of setting up the server to help with debugging so searching for duplicate tags is easier?
Alternatively, would it be straightforward to had a comment in the top of rendered files that says "this html file was built with file-name.tsx" if it's not possible to automatically download the html files each time there is an update?
Other suggestions for strategies for tracking what the rendered html file is on the directory would be helpful.
I can submit these as a separate issue as a suggested enhancement if needs to be.

SleeplessOne1917:
You finding a `main` tag in `home.tsx`, `community.tsx`, and `post.tsx` matches what I find searching in my IDE. To You point on saving rendered HTML files, I don't know of a good way of going about it, nor do I think it would be particularly helpful to have (at least outside of some kind of automated integration test suite, anyway). I was able to find a `main` tag on a post on lemmy.ml:

image

Granted, I know you said that you didn't see it in any of the test posts you made. How are you running the code from your code changes? Last time I tried running the UI main branch locally it was blowing up (for reasons unrelated to you; entirely my fault) because it hasn't been updated to use any of the 0.20.0 alpha features.

travis-jeans:
I don't have my own Lemmy server so I am just running a Local instance. To be honest since I wrote about not finding the main `class="col-12` I can actually see it now. Maybe I was on a different branch or I just needed to run `pnpm dev` again, so never mind about that.

So this is what I see after the branch update I made where I changed the main elements on post to div:

Lemmy post with the Developer Tools open. There is a <main class="mt-4 p-0 fl-1"> and a <div class="col-12 col-md-8 col-lg-9 mb-3">.
On Lemmy main branch it looks like this for me:
Lemmy post with the Developer Tools open. There is a <div class="mt-4 p-0 fl-1"> and a <main class="col-12 col-md-8 col-lg-9 mb-3">.
It looks like it worked.

dessalines (changes requested):
Seems fine, except its another stacked PR, so you'll probably have to deal with merge conflicts.

Also run prettier.

travis-jeans:
> Seems fine, except its another stacked PR, so you'll probably have to deal with merge conflicts.
> 
> Also run prettier.

I ran prettier on home.tsx, post.tsx, and community.tsx which didn't lead to any changes since I hadn't changed any white space and just changed the text.

I'm guessing the merge conflict is because I had included the colorList change from #2946 in my main branch at some point by accident but I tried updating it with the latest version. Hopefully it should be possible to merge once the colorList PR is accepted.

dessalines (approved):

dessalines (approved):

src/shared/components/community/community.tsx
```diff
@@ -422,15 +422,15 @@ export class Community extends Component<CommunityRouteProps, State> {
     return (
       <div className="community container-lg">
         <div className="row">
-          <main className="col-12 col-md-8 col-lg-9" ref={this.mainContentRef}>
+          <div className="col-12 col-md-8 col-lg-9" ref={this.mainContentRef}>
```
dessalines, line 425: L206 now needs to be changed to HTMLDivElement. Check the CI error

dessalines (changes requested):

src/shared/components/community/community.tsx
travis-jeans: Fixed it and the Woodpecker CI check says it was successful.

dessalines: Sweet, thx!

dessalines (approved):

Issue #2890, opened 2025-01-08
Define comment tree post left-border colour in CSS


travis-jeans:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

This is a related issue to #2889.

Currently the `border-left` colour of the comment tree is inserted as inline CSS with Javascript and defined in the following file `/lemmy-ui/src/shared/components/comment/comment-node.tsx` on line 386 `border-left: 2px ${moreRepliesBorderColor} solid !important.`

Border colour is defined on line 182-184: `const moreRepliesBorderColor = this.props.node.depth ? colorList[this.props.node.depth % colorList.length] : colorList[0];` Colour can be changed by altering `colorList` values . Color list is defined in this file: `/lemmy-ui/src/shared/utils/app/color-list.ts` It is selecting around the colour wheel for different hues.

It would be helpful if the border colour attribute could be defined in CSS to allow those developing themes to change the colours, currently as inline CSS it cannot be changed with a CSS theme file.

Image description: Screenshot of a comment tree of dummy posts with levels of nesting where the colours of the left border change from red to purple.

![comment-tree](https://github.com/user-attachments/assets/74358f0f-f64b-46d5-b697-54b5deaa2a55)

dessalines:
See comment in other issue.

PR #2946, opened 2025-02-06
colorList includes CSS variables so theme developers can change the comment left border color (issue 2890)

travis-jeans:
## Description

This relates to Issue #2890. 

lemmy-ui programmatically selects comment nesting colours from a list and inserts it with inline CSS. This limits theme developers from changing the colours of the border nesting colours.

This pull request defines comment nesting colours by node number in `:root` of the variables SCSS file and refers to them inside colorList. Consequently it removes the import of the hsl helper function as it is no longer required but puts the output colours into the CSS variables so the output colours are the same, yet editable to those who want to edit Lemmy themes.

For this update to work CSS themes will need to be re-built with sass so the root variables are applied to the new themes.

If the variables are better defined somewhere further upstream I am open to suggestions. However in the variables file this ensures they would apply to all downstream themes once they are re-built.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -1,5 +1,5 @@
 variables:
-  - &install_pnpm "corepack enable pnpm"
+  - &install_pnpm "npm i -g corepack && corepack enable pnpm"
 
 steps:
   fetch_git_submodules:
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,11 +6,11 @@ FROM node:alpine as builder
 # - It can break depending on the CPU (https://github.com/LemmyNet/lemmy-ui/issues/1566)
 RUN apk update && apk upgrade && apk add --no-cache curl python3 build-base gcc wget git vips-dev pkgconfig
 
+# Install node-gyp and corepack
+RUN npm install -g node-gyp corepack
+
 # Enable corepack to use pnpm
 RUN corepack enable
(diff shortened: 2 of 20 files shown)


SleeplessOne1917 (approved):
LGTM. As with the other theme PR you opened, please run `pnpm themes:build` and commit the results before merging.

travis-jeans:
I'm a bit new to PRs but I did `pnpm themes:build` then merged with the latest version of `main` (I think) and tried pushing the branch upstream again. Did it work?

dessalines:
It worked, but you need to run prettier on `src/assets/css/themes/_variables.scss`

travis-jeans:
I ran `npx prettier src/assets/css/themes/_variables.scss` and nothing needed to be changed so I merged with main and then pushed upstream. Is everything okay now?

dessalines:
You need to run prettier -w 

https://woodpecker.join-lemmy.org/repos/135/pipeline/5252/5#L14

dessalines (changes requested):
See comment above

travis-jeans:
> You need to run prettier -w
> 
> https://woodpecker.join-lemmy.org/repos/135/pipeline/5252/5#L14

Thank you! It worked this time.

dessalines (approved):

Issue #1537, opened 2023-06-23
Replace one of the default fonts Helvetica with Verdana and Arimo for legibility


heylix:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Currently, the UI gives you the following string for default fonts:

```
-apple-system,BlinkMacSystemFont,"Droid Sans","Segoe UI",Helvetica,Arial,sans-serif
```

([source](https://github.com/LemmyNet/lemmy-ui/blob/f1d555368f12a3bd03beef4e0558a390a4d8772e/src/assets/css/themes/_variables.litely.scss#L31-L32))

However, Helvetica is notoriously bad to read:

![image](https://github.com/LemmyNet/lemmy-ui/assets/36211486/262c5abd-589a-4c6e-9cde-c89373290e08)

([source](https://www.youtube.com/watch?v=h8IOqUl1zII))

Please replace with Verdana and Arimo.

Maybe also consider replacing Droid Sans, as it is *very* similar to Roboto, which also didn't fare well.

It also seems like this is [inconsistent throughout the code](https://github.com/search?q=repo%3ALemmyNet%2Flemmy-ui%20BlinkMacSystemFont&type=code), e.g. [in the darkly theme SCSS](https://github.com/LemmyNet/lemmy-ui/blob/f1d555368f12a3bd03beef4e0558a390a4d8772e/src/assets/css/themes/_variables.darkly.scss#L32-L34) and [the darkly compact CSS](https://github.com/LemmyNet/lemmy-ui/blob/f1d555368f12a3bd03beef4e0558a390a4d8772e/src/assets/css/themes/darkly-compact.css#L114).

heylix:
It can also be noted that `-apple-system` *may or may not* be SF Pro, and Segoe UI (installed by default on many Windows systems) is in order *after* Droid Sans, which means the worse font may be chosen if Droid Sans is installed. Maybe you can also ship Atkinson Hyperlegible to not have to fall back to Roboto on Android systems, which is less readable for severely dyslexic people.

[...]

PR #2975, opened 2025-02-17
Add Verdana and Arimo fonts to Litely and Darkly variables file before Helvetica (Issue 1537)

travis-jeans:
## Description

There was a user who pointed out that Helvetica has worse legibility than fonts like Verdana and Arimo and requested these be added before Helvetica. This relates to issue 1537 [Replace one of the default fonts Helvetica with Verdana and Arimo for legibility #1537](https://github.com/LemmyNet/lemmy-ui/issues/1537).

I added Verdana and Arimo to the font family list in the `_variables` files for Darkly and Litely.

One thing I wasn't sure about was where to add the `@import` for Arimo since it is a Google Font ([About Arimo (Google Fonts)](https://fonts.google.com/specimen/Arimo/about), so I haven't added it at all. So at this point it will just be loaded if a user has it installed on their system.

Note: PR stacked with #2946

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -1,5 +1,5 @@
 variables:
-  - &install_pnpm "corepack enable pnpm"
+  - &install_pnpm "npm i -g corepack && corepack enable pnpm"
 
 steps:
   fetch_git_submodules:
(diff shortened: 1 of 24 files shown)


dessalines:
Looks good, thx!

SleeplessOne1917 (approved):

dessalines:
@travis-jeans you need to run prettier again.

travis-jeans:
Ran prettier.

dessalines (approved):

PR #3009, opened 2025-03-14
Add crawl-delay to robots.txt

Nutomic:
This should prevent the various crawlers from overloading Lemmy instances. It is not officially part of robots.txt spec but many crawlers support it.

https://stackoverflow.com/a/9130248
https://imagesift.com/about

### DIFF
--- a/src/server/handlers/robots-handler.ts
+++ b/src/server/handlers/robots-handler.ts
@@ -16,7 +16,5 @@ export default async ({ res }: { res: Response }) => {
   Disallow: /password_change
   Disallow: /search/
   Disallow: /modlog
-  
-  User-Agent: *
   Crawl-delay: 60`);
 };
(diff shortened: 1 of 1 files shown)


src/server/handlers/robots-handler.ts
```diff
@@ -16,5 +16,7 @@ export default async ({ res }: { res: Response }) => {
   Disallow: /password_change
   Disallow: /search/
   Disallow: /modlog
-  `);
+  
+  User-Agent: *
+  Crawl-delay: 60`);
```
SleeplessOne1917, line 21: Couldn't this go right beneath the existing `User-Agent: *` from line 6?

Nutomic: True, fixed

dessalines (approved):

SleeplessOne1917 (approved):

PR #3041, opened 2025-03-30
Add support for using Lemmy UI with an external Lemmy instance

SolninjaA:
## Description

As discussed in the Lemmy docs PR (https://github.com/LemmyNet/lemmy-docs/pull/360), this adds support for using the Lemmy UI with an external Lemmy instance.

This PR, combined with the documentation PR, will fix:
- https://github.com/LemmyNet/lemmy/issues/2608
- https://github.com/LemmyNet/lemmy/issues/3400
- https://github.com/LemmyNet/lemmy/issues/3840

This adds `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN` as a new environment variable that can be configured to either `true` or `false`.

## Testing
Lemmy UI was tested with the following consistent environment variables:
```sh
LEMMY_UI_LEMMY_INTERNAL_HOST=192.168.4.235:8536 LEMMY_UI_LEMMY_EXTERNAL_HOST=sol.mn LEMMY_UI_HTTPS=true pnpm dev
```
Additionally, the environment variable `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN` was appended and the value was changed for testing.
| Environment variable | Can connect to external Lemmy server? |
|--------|--------|
| `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN` was not set | false |
| `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN=false` | false |
| `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN=true` | true |  

Part of the testing process was to look at the browser's console and network activity. Everything works as expected, and there were no apparent issues found when testing manually.

## Screenshots
image

Users can login to a Lemmy server that is not running on the same domain, as shown in this image. Posts have been blurred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,13 +10,14 @@ Based off of MrFoxPro's [inferno-isomorphic-template](https://github.com/MrFoxPr
 
 The following environment variables can be used to configure lemmy-ui:
 
-| `ENV_VAR`                      | type     | default          | description                                                                         |
-| ------------------------------ | -------- | ---------------- | ----------------------------------------------------------------------------------- |
-| `LEMMY_UI_HOST`                | `string` | `0.0.0.0:1234`   | The IP / port that the lemmy-ui isomorphic node server is hosted at.                |
-| `LEMMY_UI_LEMMY_INTERNAL_HOST` | `string` | `0.0.0.0:8536`   | The internal IP / port that lemmy is hosted at. Often `lemmy:8536` if using docker. |
-| `LEMMY_UI_LEMMY_EXTERNAL_HOST` | `string` | `0.0.0.0:8536`   | The external IP / port that lemmy is hosted at. Often `DOMAIN.TLD`.                 |
-| `LEMMY_UI_HTTPS`               | `bool`   | `false`          | Whether to use https.                                                               |
-| `LEMMY_UI_EXTRA_THEMES_FOLDER` | `string` | `./extra_themes` | A location for additional lemmy css themes.                                         |
-| `LEMMY_UI_DEBUG`               | `bool`   | `false`          | Loads the [Eruda](https://github.com/liriliri/eruda) debugging utility.             |
-| `LEMMY_UI_DISABLE_CSP`         | `bool`   | `false`          | Disables CSP security headers                                                       |
-| `LEMMY_UI_CUSTOM_HTML_HEADER`  | `string` |                  | Injects a custom script into `<head>`.                                              |
+| `ENV_VAR`                        | type     | default          | description                                                                         |
+|----------------------------------|----------|------------------|-------------------------------------------------------------------------------------|
+| `LEMMY_UI_HOST`                  | `string` | `0.0.0.0:1234`   | The IP / port that the lemmy-ui isomorphic node server is hosted at.                |
+| `LEMMY_UI_LEMMY_INTERNAL_HOST`   | `string` | `0.0.0.0:8536`   | The internal IP / port that lemmy is hosted at. Often `lemmy:8536` if using docker. |
+| `LEMMY_UI_LEMMY_EXTERNAL_HOST`   | `string` | `0.0.0.0:8536`   | The external IP / port that lemmy is hosted at. Often `DOMAIN.TLD`.                 |
+| `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN` | `bool`   | `false`          | If the Lemmy server/backend is hosted on a different domain, this must be enabled.  |
+| `LEMMY_UI_HTTPS`                 | `bool`   | `false`          | Whether to use https.                                                               |
+| `LEMMY_UI_EXTRA_THEMES_FOLDER`   | `string` | `./extra_themes` | A location for additional lemmy css themes.                                         |
+| `LEMMY_UI_DEBUG`                 | `bool`   | `false`          | Loads the [Eruda](https://github.com/liriliri/eruda) debugging utility.             |
+| `LEMMY_UI_DISABLE_CSP`           | `bool`   | `false`          | Disables CSP security headers                                                       |
+| `LEMMY_UI_CUSTOM_HTML_HEADER`    | `string` |                  | Injects a custom script into `<head>`.                                              |
--- a/src/shared/utils/env/get-external-host.ts
+++ b/src/shared/utils/env/get-external-host.ts
@@ -2,19 +2,15 @@ import { isBrowser } from "@utils/browser";
 import { testHost } from "../../config";
 
 export default function getExternalHost() {
-  const externalDomains = process.env.LEMMY_UI_LEMMY_EXTERNAL_DOMAIN === "true";
-
-  if (externalDomains) {
-    return process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST;
-  } else {
-    return isBrowser()
-      ? `${window.location.hostname}${
+  return isBrowser()
+    ? process.env.LEMMY_UI_LEMMY_EXTERNAL_DOMAIN === "true"
+      ? process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST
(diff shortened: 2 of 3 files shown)


src/shared/utils/env/get-external-host.ts
```diff
@@ -3,12 +3,14 @@ import { testHost } from "../../config";
...
+    ? process.env.LEMMY_UI_LEMMY_EXTERNAL_DOMAIN === "true"
+      ? process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST || testHost
+      : `${window.location.hostname}${
+          ["1234", "1235"].includes(window.location.port)
+            ? ":8536"
+            : window.location.port === ""
+              ? ""
+              : `:${window.location.port}`
+        }`
     : process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST || testHost;
```
Nutomic, line 15: Maybe this function could be changed to always return `LEMMY_UI_LEMMY_EXTERNAL_HOST`? That var is defined in all the examples I found, and then no new variable would be needed.

README.md
```diff
@@ -10,13 +10,14 @@ Based off of MrFoxPro's [inferno-isomorphic-template](https://github.com/MrFoxPr
...
-| `LEMMY_UI_CUSTOM_HTML_HEADER`  | `string` |                  | Injects a custom script into `<head>`.                                              |
+| `ENV_VAR`                        | type     | default          | description                                                                         |
+|----------------------------------|----------|------------------|-------------------------------------------------------------------------------------|
+| `LEMMY_UI_HOST`                  | `string` | `0.0.0.0:1234`   | The IP / port that the lemmy-ui isomorphic node server is hosted at.                |
+| `LEMMY_UI_LEMMY_INTERNAL_HOST`   | `string` | `0.0.0.0:8536`   | The internal IP / port that lemmy is hosted at. Often `lemmy:8536` if using docker. |
+| `LEMMY_UI_LEMMY_EXTERNAL_HOST`   | `string` | `0.0.0.0:8536`   | The external IP / port that lemmy is hosted at. Often `DOMAIN.TLD`.                 |
+| `LEMMY_UI_LEMMY_EXTERNAL_DOMAIN` | `bool`   | `false`          | If the Lemmy server/backend is hosted on a different domain, this must be enabled.  |
```
Nutomic, line 18: Should be `LEMMY_UI_LEMMY_HAS_EXTERNAL_DOMAIN` or similar, currently it doesnt sound like a bool.

src/shared/utils/env/get-external-host.ts
SolninjaA: Personally, I think this is the best solution. I just wasn't sure, however, since @dessalines mentioned in [this comment](https://github.com/LemmyNet/lemmy-docs/pull/360#discussion_r2012159313) that:
> You'd need to add another optional env var, and test it, to override the window location one.

If acceptable, I would love to make your proposed changes.

SolninjaA:
I just implemented [Nutomic's suggestion to rename the environment variable](https://github.com/LemmyNet/lemmy-ui/pull/3041#discussion_r2020601128). I'll make the necessary changes based on [this comment](https://github.com/LemmyNet/lemmy-ui/pull/3041#discussion_r2020600452) when I have more information.

src/shared/utils/env/get-external-host.ts
dessalines: I'm fine with that also, it'd simplify it a lot.

SolninjaA: Thanks very much for that information! I'll make the changes when I get the chance.

SolninjaA:
I have made the changes discussed [here](https://github.com/LemmyNet/lemmy-ui/pull/3041/files/d9819428b1c15900d9a4ec2b6334d5133224531d).

Please let me know if there are any problems!
Thanks.

dessalines (commented):
LGTM, thx!

dessalines (approved):

src/shared/utils/env/get-external-host.ts
```diff
@@ -1,14 +1,5 @@
...
-  return isBrowser()
-    ? `${window.location.hostname}${
-        ["1234", "1235"].includes(window.location.port)
-          ? ":8536"
-          : window.location.port === ""
-            ? ""
-            : `:${window.location.port}`
-      }`
-    : process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST || testHost;
+  return process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST || testHost;
```
SleeplessOne1917, line 4: Can use `??` instead of `||` since this isn't a bool.

webpack.config.js
```diff
@@ -86,6 +86,9 @@ module.exports = (env, argv) => {
       new webpack.DefinePlugin({
         "process.env.COMMIT_HASH": `"${env.COMMIT_HASH}"`,
         "process.env.NODE_ENV": `"${mode}"`,
+        "process.env.LEMMY_UI_LEMMY_EXTERNAL_HOST": JSON.stringify(
```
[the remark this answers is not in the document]
SolninjaA: Thanks for your comment!
It seems like it, otherwise I get this error:
```
ReferenceError: sol is not defined
```

Where `sol` is the beginning of my domain before the TLD.

SolninjaA:
I just made a new commit which implements @SleeplessOne1917's [suggested changes](https://github.com/LemmyNet/lemmy-ui/pull/3041#discussion_r2027827033).

src/shared/utils/env/get-external-host.ts
SolninjaA: Thanks very much! I've made that change in [this PR](https://github.com/LemmyNet/lemmy-ui/pull/3041/commits/f071aa84953b2121a985b205ea606b09d4195395).

webpack.config.js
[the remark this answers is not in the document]
matc-pub: The `DefinePlugin` doesn't populate an `env` object with actual values. It replaces its uses with the actual values. E.g. the resulting code would be `return undefined ?? testHost;` or `return "example.org" ?? testHost;` (the quotes are part of the defined value e.g. `"\"example.org\""`). `JSON.strignify` is useful here because it also escapes newline and quotes.

A problem here is that `LEMMY_UI_LEMMY_EXTERNAL_HOST` will be the value at compile time.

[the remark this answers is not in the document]
matc-pub: The catch-all-handler would have to add it to isoData to get the runtime value.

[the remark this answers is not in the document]
SolninjaA: Hm, perhaps that explains why the Docker container didn’t seem to work. I figured it was a problem with me running it, rather than a code issue. But it could only be working with pnpm dev as I pass environment variables to it on compile.

It’s 12am right now so I’ll look at this later, or when I get the chance (probably within a couple of days).

Any particular suggestions on the best way to go about this? 


Also, considering the seemingly apparent discrepancy between Docker and pnpm testing, would you like me to test in Docket hereafter?

Thanks very much!

SolninjaA:
I've had a look at @matc-pub's issue that he pointed out, and how it could be fixed.

I'm grateful for the information that he gave on how to fix the issue:
> The catch-all-handler would have to add it to isoData to get the runtime value.

However, due to my unfamiliarity with the project's structure and Webpack as a whole, I couldn't find a solution based solely from his advice.

I've researched other popular solutions online, and almost all of them involve modifying the Dockerfile, which I feel should be unnecessary.

If possible, it would be great if anyone could elaborate on @matc-pub's [proposed solution](https://github.com/LemmyNet/lemmy-ui/pull/3041#discussion_r2028808390) in further detail.

Thanks!

SleeplessOne1917 (approved):

dessalines (approved):

PR #3043, opened 2025-03-31
Dont render images in private message

Nutomic:


SleeplessOne1917:
Should we make an exception for emojis?

dessalines:
I don't think its worth it. Unless its done directly to that markdown no_images parsing mode, but that can be a different PR.

SleeplessOne1917 (approved):

Issue #2892, opened 2025-01-08
Add missing focus indicators to buttons and navigation menu items


travis-jeans:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

While developing a custom theme [RBlind-Lemmy-Themes](https://codeberg.org/travis-jeans/RBlind-Lemmy-Themes) I added focus indicators to elements that had them missing.

Elements missing focus indicators:
- Posts/Comments tab (homepage)
- Show Hidden Posts / Hide Hidden Posts (homepage)
- User Profile dropdown button (homepage)

Added for the first three with a custom focus indicator color:
`.btn-check:checked + .btn:focus-visible, :not(.btn-check) + .btn:active:focus-visible, .btn:first-child:active:focus-visible, .btn.active:focus-visible, .btn.show:focus-visible { box-shadow: 0 0 0 0.25rem var(--rblind-focus-indicator) !important; }`

Added a focus indicator to the User Profile toggle dropdown: 
`.dropdown-toggle:focus-visible { box-shadow: 0 0 0 0.25rem var(--rblind-focus-indicator) !important; }`

I also added or altered focus indicators to the following elements so their colouring was consistent:

- Skip link button: `.skip-link:focus, .skip-link:focus-visible { top: 0.25rem; left: 0.25rem !important; text-decoration: underline; box-shadow: 0 0 0 0.25rem var(--rblind-focus-indicator) !important; }`
- Navbar toggler and icon buttons: `a.sort-select-icon:focus svg, a[title=RSS]:focus svg, a[title="sorting help"]:focus svg, .navbar-toggler:focus { box-shadow: 0 0 0 0.25rem var(--rblind-focus-indicator) !important; }`
- Give unselected tabs (like Comment) the same focus outline as other buttons. Added .btn-check:focus-visible + .btn { box-shadow: 0 0 0 0.25rem var(--rblind-focus-indicator) !important; }
[...]

PR #3048, opened 2025-04-01
Fix missing hover and focus indicators on the Show Hidden Posts and Hide Hidden Posts buttons, plus added aria-labels

travis-jeans:
## Description

This and my other pull request #3047 aim to fix the remaining issues I can see in #2892. 

Changes I made to the icon buttons for Show/Hide Hidden Posts:

- move the input element outside the label element so the label only wraps around the icon and not the input element 
- make the input element appear first in the DOM to match the other buttons like Subscribed/Local/All
- add aria-labels that match the buttons purpose (I copied the tippy labels for them)

I remember seeing an aria-describedby=tippy element on one of these but it seems to have disappeared after the fix (or maybe I lost track of where it went). Anyway I think this label wasn't functioning as intended because the text was not visible on the screen unless toggled with tippy (mouse hover tooltip, I think). As a general rule, aria-describedby works best with labels that have visible text (that is visible at all times). In the case of icon buttons like these aria-labels are better.

### Navigating to Buttons

In Firefox, once the page has loaded, press Ctrl+K to jump to the URL then press Tab to navigation from button to button until you get to the Show/Hide Hidden Posts buttons. You can navigate backwards with Shift+Tab.

### Aria-labels

To see what happens with the aria-labels, download a copy of Orca if it isn't already installed (Orca is a screenreader for Linux but if on another system there are other options). Tab navigate until you get to the buttons and the aria-labels will be read out. Without these labels the buttons have no audible label and it will say something uninformative like "list item 1 of 2, unselected/selected" which doesn't tell you what the buttons do.

### DIFF
--- a/src/shared/components/common/post-hidden-select.tsx
+++ b/src/shared/components/common/post-hidden-select.tsx
@@ -27,40 +27,41 @@ export default class PostHiddenSelect extends Component<
         className="show-hidden-select btn-group btn-group-toggle flex-wrap"
         role="group"
       >
+        <input
+          id="show-hidden"
+          type="radio"
+          className="btn-check"
+          value="true"
+          checked={showHidden === "true"}
+          onChange={linkEvent(this, handleShowHiddenChange)}
+        />
         <label
           htmlFor="show-hidden"
           className={classNames("pointer btn btn-outline-secondary", {
             active: showHidden === "true",
           })}
           data-tippy-content={I18NextService.i18n.t("show_hidden_posts")}
+          aria-label="Show hidden posts"
         >
           <Icon icon="eye" inline />
-          <input
-            id="show-hidden"
-            type="radio"
-            className="btn-check"
-            value="true"
-            checked={showHidden === "true"}
-            onChange={linkEvent(this, handleShowHiddenChange)}
-          />
         </label>
-
+        <input
+          id="hide-hidden"
+          type="radio"
+          className="btn-check"
+          value="false"
+          checked={showHidden !== "true"}
+          onChange={linkEvent(this, handleShowHiddenChange)}
+        />
         <label
           htmlFor="hide-hidden"
           className={classNames("pointer btn btn-outline-secondary", {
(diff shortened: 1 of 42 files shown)


SleeplessOne1917 (approved):

src/shared/components/common/post-hidden-select.tsx
```diff
@@ -27,40 +27,41 @@ export default class PostHiddenSelect extends Component<
...
+          checked={showHidden !== "true"}
+          onChange={linkEvent(this, handleShowHiddenChange)}
+        />
         <label
           htmlFor="hide-hidden"
           className={classNames("pointer btn btn-outline-secondary", {
             active: showHidden !== "true",
           })}
           data-tippy-content={I18NextService.i18n.t("hide_hidden_posts")}
+          aria-label="Hide hidden posts"
```
dessalines, line 62: I18NextService.i18n.t("hide_hidden_posts")

src/shared/components/common/post-hidden-select.tsx
```diff
@@ -27,40 +27,41 @@ export default class PostHiddenSelect extends Component<
...
+          checked={showHidden === "true"}
+          onChange={linkEvent(this, handleShowHiddenChange)}
+        />
         <label
           htmlFor="show-hidden"
           className={classNames("pointer btn btn-outline-secondary", {
             active: showHidden === "true",
           })}
           data-tippy-content={I18NextService.i18n.t("show_hidden_posts")}
+          aria-label="Show hidden posts"
```
dessalines, line 44: Use `I18NextService.i18n.t("show_hidden_posts")` . Otherwise it won't be internationalized.

src/shared/components/common/post-hidden-select.tsx
travis-jeans: Thanks, I'll modify this.

dessalines (approved):

PR #3049, opened 2025-04-02
Added alt text to the Show/Hide Content buttons (plus-square and minus-square SVG buttons)

travis-jeans:
## Description

The plus and minus reveal buttons (e.g. on post listings next to the heading that display or hide the post) were missing alt text (they had a visually hidden title element but the Orca screenreader never reads it for me) so I added them using the technique described on the [Creating Accessible SVGs (Deque)](https://www.deque.com/blog/creating-accessible-svgs/) page. 

This involves adding to the SVG element: 
- aria-described 
- role="img"
- title element

Originally I tried editing assets/symbols.svg to have <text class="visually-hidden">alt text</text> or <title>alt text</title> like on the technique webpage but those did not work because the <symbol> element can't have a <title> attribute and the screenreader does not read the <text> element when it's set to be visually hidden. So instead I used a switch / case statement instead.

### DIFF
--- a/src/shared/components/common/icon.tsx
+++ b/src/shared/components/common/icon.tsx
@@ -22,45 +22,27 @@ export class Icon extends Component<IconProps, any> {
       this.props.icon === "minus-square"
     ) {
       iconAltText = `${I18NextService.i18n.t("show_content")}`;
-    } else {
-      iconAltText = "";
     }
-
-    if (iconAltText !== "") {
-      return (
-        <svg
-          className={classNames("icon", this.props.classes, {
-            "icon-inline": this.props.inline,
-            small: this.props.small,
-          })}
-          role="img"
-          aria-describedby={`${this.props.icon}-alt`}
-        >
+    return (
+      <svg
+        className={classNames("icon", this.props.classes, {
+          "icon-inline": this.props.inline,
+          small: this.props.small,
+        })}
+        {...(iconAltText
+          ? { role: "img", "aria-describedby": `${this.props.icon}-alt` }
+          : {})}
+      >
+        {iconAltText && (
           <title id={`${this.props.icon}-alt`}>{iconAltText}</title>
-          <use
-            xlinkHref={`${getStaticDir()}/assets/symbols.svg#icon-${
-              this.props.icon
-            }`}
-          ></use>
-        </svg>
-      );
-    } // use aria-describedby or not
-    else {
-      return (
-        <svg
-          className={classNames("icon", this.props.classes, {
-            "icon-inline": this.props.inline,
-            small: this.props.small,
-          })}
-        >
-          <use
-            xlinkHref={`${getStaticDir()}/assets/symbols.svg#icon-${
-              this.props.icon
-            }`}
-          ></use>
-        </svg>
-      );
-    }
+        )}
+        <use
+          xlinkHref={`${getStaticDir()}/assets/symbols.svg#icon-${
+            this.props.icon
+          }`}
+        ></use>
+      </svg>
+    );
   }
 }
 
--- a/src/shared/components/common/icon.tsx
+++ b/src/shared/components/common/icon.tsx
@@ -16,15 +16,14 @@ export class Icon extends Component<IconProps, any> {
   }
 
   render() {
-    let iconAltText: string;
+    let iconAltText: string | undefined;
     if (
       this.props.icon === "plus-square" ||
       this.props.icon === "minus-square"
     ) {
       iconAltText = `${I18NextService.i18n.t("show_content")}`;
-    } else {
-      iconAltText = "";
     }
+
     return (
       <svg
         className={classNames("icon", this.props.classes, {
@@ -35,7 +34,7 @@ export class Icon extends Component<IconProps, any> {
           ? { role: "img", "aria-describedby": `${this.props.icon}-alt` }
           : {})}
       >
-        {iconAltText !== "" && (
+        {iconAltText && (
           <title id={`${this.props.icon}-alt`}>{iconAltText}</title>
         )}
         <use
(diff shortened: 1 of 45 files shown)


src/shared/components/common/icon.tsx
```diff
@@ -16,21 +16,32 @@ export class Icon extends Component<IconProps, any> {
   }
 
   render() {
+    switch (`${this.props.icon}`) {
```
dessalines, line 19: Lots of bugs here, clearly you're not using a code editor that works with typescript.

```ts
let iconAltText`: string;
switch(this.props.icon) {
  case "...": 
    iconAltText = ...
  break;
```

src/shared/components/common/icon.tsx
```diff
@@ -16,21 +16,32 @@ export class Icon extends Component<IconProps, any> {
...
+      case "plus-square":
+        var icon_alt_text = "Show Content";
```
dessalines, line 21: These need to be `i18n`. Never insert english strings.

src/shared/components/common/icon.tsx
dessalines: You're probably better off using if statements anyway.

travis-jeans: Ah, no I'm using Emacs. I also am not that familiar with Typescript but it seems to work in my browser. I'll look into this.

src/shared/components/common/icon.tsx
travis-jeans: Oh, I didn't realize `i18n` was used for managing translation. I'll look into this.

travis-jeans:
I haven't yet done the `i18n` translation part, still working on it. Thought I'd push the changes I had made already though that hopefully fixes some bugs.

travis-jeans:
After adding the `i18n` variable `show_content` to the english database file I tried switching to french in my browser on my local instance (in Lemmy interface settings) and there's no automatic translation involved. Are the translations done on [Weblate]](https://weblate.join-lemmy.org/projects/lemmy/)? So it won't be translated until somebody does it manually? Right now it's showing "Show Content" as I put in the English file.

Sorry if I've done something wrong here, I don't really understand how it's working under the hood yet, the most I was able to get was that there are key value pairs but not how the other translations are populated.

Nutomic:
Yes translations are made via weblate after your changes are merged. Its normal that only English is used for now, but we update the other languages before each release.

travis-jeans:
Sorry just noticed the lemmy translations folder wasn't added to tracking... and I can't seem to get it to. This is what I had changed:

`  "hide_post": "Hide Post",
    "unhide_post": "Unhide Post",
    "show_content": "Show Content",`

Looks like the translation folder has its own git repository so I might have to do it separately.

dessalines:
Thx, yep you did it correctly.

src/shared/components/common/icon.tsx
```diff
@@ -16,23 +16,51 @@ export class Icon extends Component<IconProps, any> {
...
+          })}
+        >
+          <use
+            xlinkHref={`${getStaticDir()}/assets/symbols.svg#icon-${
+              this.props.icon
+            }`}
+          ></use>
+        </svg>
+      );
+    }
```
SleeplessOne1917, line 63: This can be shortened somewhat. I think the following should work:

```suggestion
    let iconAltText: string;
    if (
      this.props.icon === "plus-square" ||
      this.props.icon === "minus-square"
    ) {
      iconAltText = `${I18NextService.i18n.t("show_content")}`;
    }
    return (
      <svg
        className={classNames("icon", this.props.classes, {
          "icon-inline": this.props.inline,
          small: this.props.small,
        })}
        {...(iconAltText ? {role: "img", "aria-describedby": `${this.props.icon}-alt`} : {})}
      >
        {iconAltText && (<title id={`${this.props.icon}-alt`}>{iconAltText}</title>)}
        <use
          xlinkHref={`${getStaticDir()}/assets/symbols.svg#icon-${
            this.props.icon
          }`}
        ></use>
      </svg>
    );
```

The suggestion code will need to be formatted with prettier, and there may be ways to clean it up further.

travis-jeans: I accepted your suggestion, thank you. It seems to work on my side.

As someone mostly unfamiliar with Typescript I actually find this code harder to understand. I looked up short hand if statements but it didn't include code blocks, can you tell me what this method is called so I can read more about it? What I'm having trouble following is the meaning of `{...`,  `{}`, and `iconAltText && (code block)}` part. If iconAltText is undefined does the && short circuit and then delete itself?

travis-jeans: Actually the code failed once uploaded to GitHub because the linter does not like that iconAltText isn't defined. That was one of the reasons I had added the blank `iconAltText=""` before.

travis-jeans: I put it back and it seems to work. Should the `{...{iconAltText ?` part need to evaluate if iconAltText isn't blank now too?

SleeplessOne1917 (approved):

src/shared/components/common/icon.tsx
```diff
@@ -16,21 +16,33 @@ export class Icon extends Component<IconProps, any> {
...
       <svg
         className={classNames("icon", this.props.classes, {
           "icon-inline": this.props.inline,
           small: this.props.small,
         })}
+        {...(iconAltText
+          ? { role: "img", "aria-describedby": `${this.props.icon}-alt` }
+          : {})}
       >
+        {iconAltText !== "" && (
```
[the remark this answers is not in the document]
travis-jeans: Thank you for those suggestions! I tested those changes and they work so I've pushed the updates.

dessalines (approved):
You're a fast learner, thx!

PR #3053, opened 2025-04-03
Hide images in private message reports

Nutomic:
Looks like post-report and comment-report already hide images using [hideImages](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/post/post-report.tsx#L87) param.

The disadvantage with this is that the images are not rendered at all, not even as plain link. So there is no way to view the image or even to know that it is there.

### DIFF
--- a/src/shared/components/private_message/private-message-report.tsx
+++ b/src/shared/components/private_message/private-message-report.tsx
@@ -4,7 +4,7 @@ import {
   PrivateMessageReportView,
   ResolvePrivateMessageReport,
 } from "lemmy-js-client";
-import { mdToHtml } from "../../markdown";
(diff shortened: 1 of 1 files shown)


dessalines (approved):

dessalines:
[There's probably a better way for us to do noImages](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/markdown.ts#L201)... really we don't want to hide them, but just make them not rendered, but as links. I'll see if theres a way to do that with markdown-it.

Nutomic:
@dessalines For that it will be necessary to write a plugin.

https://docs.joshuatz.com/cheatsheets/node-and-npm/markdown-it/

Issue #2438, opened 2024-04-24
[Bug]: Saved Posts RSS Using Posts Feed


ToffeeIsForClosers:
### Requirements

- [X] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support
- [X] Did you check to see if this issue already exists?
- [X] Is this only a single bug? Do not put multiple bugs in one issue.
- [X] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [X] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

Under the Saved (posts) section of a user profile, the RSS feed is not a listing of saved posts but rather the user created posts. 

In contrast, the feed under Posts and the feed under Comments both provide their respective RSS article links properly. Based on that setup, it’s reasonable to expect a Saved post RSS feed to be available. 

So, either the feed icon should be removed to avoid confusion or, better, an RSS feed for Saved posts should be added.

I am a user, not someone running an instance of Lemmy. I am filing the bug but I won’t be able to support its development or testing at the level that may be expected. Let me know if I shouldn’t log issues here.

Posted to backend Lemmy repo as this handles the RSS, by stated description. Let me know if it should go to 
Lemmy-UI.

### Steps to Reproduce

1. Log into Lemmy account and go to Profile > Saved.
2. Click the RSS button to subscribe/open the feed XML.
3. Feed is for user-created posts, not the user-saved posts.

### Technical Details

My test account was on Lemmy.World. I don’t have the technical details unfortunately.

### Version

Not sure

### Lemmy Instance URL

lemmy.world

Nutomic:
Saved items are not publicly available through the api, so they also shouldnt be public through other ways as that would be too unexpected.

dessalines:
We already have private RSS feeds, for example your subscribed rss feed uses your `jwt` as part of the RSS url. 

So we *could* add saved posts as its own RSS, but what would be the point? RSS is mainly for news type feeds where you mark new items as read and they get removed, not a bookmarking service.

I think I'm with @dullbananas here, that this is just a lemmy-ui bug (which I'll fix rn), unless there are good arguments made for adding a saved_only RSS feed.

PR #3054, opened 2025-04-03
Hide RSS button for Uploads view on profile page

Tyoda:
## Description

RSS feeds are not implemented for uploads, and it wouldn't make sense to, similarly to saved posts as discussed in #2438

### Before

rss_before

### After

rss_after


SleeplessOne1917 (approved):
I didn't even know we had RSS feeds for uploads.

dessalines (approved):
Good catch, thx for all these!

PR #3034, opened 2025-03-28
Implement donation dialog

Nutomic:
## Description

Adds a donation dialog which gets shown to users once a year:
Screenshot_20250402_105859

Translations (still need to write a better text): https://github.com/LemmyNet/lemmy-translations/pull/161
js-client: https://github.com/LemmyNet/lemmy-js-client/pull/540
Backend 0.19 backport: https://github.com/LemmyNet/lemmy/pull/5559

To test this checkout `release/0.19` in backend repo, do `cargo run` and in psql do `update local_user set last_donation_notification = '1999-01-08 04:05:06';`.

After this is merged I will also add it to main branch.

PR #3056, opened 2025-04-04
Implement donation dialog

Nutomic:
Add donation dialog to main branch (from https://github.com/LemmyNet/lemmy-ui/pull/3034).

Translations: https://github.com/LemmyNet/lemmy-translations/pull/163
JS client: https://github.com/LemmyNet/lemmy-js-client/pull/547

Currently throws a lot of compile errors, probably requires #2985 to be merged first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "inferno-router": "^8.2.3",
     "inferno-server": "^9.0.0",
     "jwt-decode": "^4.0.0",
-    "lemmy-js-client": "0.20.0-alpha.17",
+    "lemmy-js-client": "1.0.0-alpha.0",
     "lodash.isequal": "^4.5.0",
     "markdown-it": "^14.1.0",
     "markdown-it-bidi": "^0.2.0",
--- a/src/shared/components/home/donation_dialog.tsx
+++ b/src/shared/components/home/donation_dialog.tsx
@@ -1,0 +1,87 @@
+import { Component } from "inferno";
+import { GetSiteResponse } from "lemmy-js-client";
+import * as sanitizeHtml from "sanitize-html";
+import { HttpService, I18NextService } from "../../services";
+import { donateLemmyUrl } from "../../config";
+
+interface Props {
+  site?: GetSiteResponse;
+}
+
+interface State {
+  show: boolean;
+}
+
+export class DonationDialog extends Component<Props, State> {
+  state: State = { show: this.initializeShow() };
+
+  constructor(props: any, context: any) {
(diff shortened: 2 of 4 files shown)


dessalines:
Yep that's correct, I'll change the base to merge it into my dev branch.

src/shared/components/home/donation_dialog.tsx
```diff
@@ -0,0 +1,87 @@
...
+              aria-label={I18NextService.i18n.t("donation_dialog_button_hide")}
+            ></button>
+          </div>
+          <div
+            class="card-text"
+            dangerouslySetInnerHTML={{
+              __html: sanitizeHtml(
+                I18NextService.i18n.t("donation_dialog_message"),
+              ),
+            }}
```
Nutomic, line 57: Should get rid of raw html here similar to [joinlemmy pr](https://github.com/LemmyNet/joinlemmy-site/pull/467/files#diff-bc0ae13d62de3da6e730cdd077d2dcea98df71eaf724377c285b3955b1055698R43)

dessalines: I see that `donation_dialog_message` has html in it. It'd be better to remove that entirely and either: 

- Split it into 2 strings, IE `donation_dialog_message_1` and `donation_dialog_message_2`
- Replace the `<br>`s with `<1>` , as the i18next react uses.

The first is probably better and much easier. Really html shouldn't be used for line breaks, but only for within-sentence formatting where you need to use links, italics, etc.

Nutomic: Alright got it.

Nutomic:
CI is failing because of changes in js-client.

package.json
```diff
@@ -60,7 +60,7 @@
     "inferno-router": "^8.2.3",
     "inferno-server": "^9.0.0",
     "jwt-decode": "^4.0.0",
-    "lemmy-js-client": "0.20.0-alpha.17",
+    "lemmy-js-client": "1.0.0-alpha.0",
```
dessalines, line 63: Not sure that we should do this yet, and compeletely break main. Probably safer to do PRs to my 1.0 dev branch until we get everything compiling.

https://github.com/LemmyNet/lemmy-ui/pull/2985

Nutomic: Agree, otherwise it will be impossible to work on lemmy-ui until that PR is merged.

Nutomic:
Updated

SleeplessOne1917 (approved):

PR #3075, opened 2025-04-09
[0.19] Remove dangerous inner html for donation dialog

Nutomic:
https://github.com/LemmyNet/lemmy-ui/pull/3056#discussion_r2032768311


dessalines (approved):

Nutomic:
Btw it looks like we can use lemmy-translations main branch for 0.19, that way new weblate translations can be included in releases.

dessalines:
That's good. I forget the reason we had to make a branch for it in the first place (possibly for the back-end to compile).

PR #3077, opened 2025-04-11
feat: add extra items to robots.txt disallow

ticoombs:
## Description

Adds more items to robots

## Screenshots

N/A

### DIFF
--- a/src/server/handlers/robots-handler.ts
+++ b/src/server/handlers/robots-handler.ts
@@ -6,7 +6,7 @@ export default async ({ res }: { res: Response }) => {
   res.send(`User-Agent: *
 Disallow: /login
 Disallow: /login_reset
-DIsallow: /signup
+Disallow: /signup
 Disallow: /settings
 Disallow: /instances
 Disallow: /create_community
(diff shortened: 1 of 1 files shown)


src/server/handlers/robots-handler.ts
```diff
@@ -6,7 +6,9 @@ export default async ({ res }: { res: Response }) => {
   res.send(`User-Agent: *
 Disallow: /login
 Disallow: /login_reset
+DIsallow: /signup
 Disallow: /settings
+Disallow: /instances
```
Nutomic, line 11: Allowing /instances makes some sense for bots to discover other Fediverse sites.

src/server/handlers/robots-handler.ts
```diff
@@ -6,7 +6,9 @@ export default async ({ res }: { res: Response }) => {
   res.send(`User-Agent: *
 Disallow: /login
 Disallow: /login_reset
+DIsallow: /signup
```
Nutomic, line 9:
```suggestion
Disallow: /signup
```

SleeplessOne1917 (approved):

dessalines (approved):

PR #3101, opened 2025-04-19
Remove LEMMY_UI_LEMMY_EXTERNAL_HOST from webpack.DefinePlugin

MrKaplan-lw:
## Description

[webpack.DefinePlugin](https://webpack.js.org/plugins/define-plugin/) overrides expressions at compile time, but this env var must be evaluated at runtime

This was already raised in the discussion here but not actually resolved before the PR got merged: https://github.com/LemmyNet/lemmy-ui/pull/3041#discussion_r2028679878

### Before

`LEMMY_UI_LEMMY_EXTERNAL_HOST` env var is always evaluated as `undefined` if it's not defined at build time.

### After

`LEMMY_UI_LEMMY_EXTERNAL_HOST` is evaluated at runtime.

### DIFF
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -28,7 +28,7 @@ import {
 } from "../../shared/services/";
 import { parsePath } from "history";
 import { getQueryString } from "@utils/helpers";
-import { adultConsentCookieKey } from "../../shared/config";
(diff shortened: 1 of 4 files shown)


dessalines (approved):
Good catch, thx.

SleeplessOne1917 (approved):

PR #3105, opened 2025-04-23
Remove instance cache

### LINKED ISSUE #2792: Lemmy UI silently fills Firefox local storage (posts) without automatic deletion/cleaning
### Requirements

### Summary

I noticed under Firefox: Settings -> Privacy my local cookie/cache data was taking some 3G of space which felt off. I opened Manage Data, sorted by size and lo and behold lemm.ee (my home instance) was using all that space. I cleared the cache, verified empty in `about:cache` as empty yet the consumed lemm.ee space it was still listed. I had to resort to using the History function "forget this site" to actually clear out all used space.

I'll try and present my findings, but I'm sort of guessing my way through this as an observer and end user, not a dev or instance admin. @sunaurus runs the instance and helps with Lemmy code, I hope they do not mind me tagging them into this issue.

### Steps to Reproduce

1. Browse your Lemmy instance for a long while, opening many images along the way using the click-to-expand feature of the article list views (click on image on left without opening full post)
2. Observe in Settings -> Privacy your used site data number increasing over time
3. Clear all cache, confirm in `about:cache`, restart Firefox
4. Observe in Settings -> Privacy the storage amount does not decrease as expected

### Technical Details

In your local Firefox profile there exists the `storage/...` tree where each subdir is a named site (such as lemm.ee or lemmy.world, etc.) and several subdirs under that, with "default" being the one used in my observation. A quick look at the files sitting there reveals a lot of objects like this one:

```
$ find ./storage/default/https+++lemm.ee/ -type f
./storage/default/https+++lemm.ee/cache/caches.sqlite
...
./storage/default/https+++lemm.ee/cache/morgue/57/{dcc46c14-70ec-4c6a-8537-1bee4688d239}.final
```
->
```
$ file ./storage/default/https+++lemm.ee/cache/morgue/57/{dcc46c14-70ec-4c6a-8537-1bee4688d239}.final
./storage/default/https+++lemm.ee/cache/morgue/57/{dcc46c14-70ec-4c6a-8537-1bee4688d239}.final: snappy framed data

```

"snappy" is a compression format used by Firefox; a quick install of [snappy-tools](https://sr.ht/~nabijaczleweli/snappy-tools/) let me have a look at a copy of one of them, confirming it was a JPEG I had looked at moments ago (note I had to use the `-i` flag to ignore some checksum errors from rudimentary debugging):

```
$ ./snappy-tools/unsnappy -i test.final > new.file
unsnappy: test.final: chunk of length 65312: checksum 0x6C991DCD != 0xD86CB7E2
unsnappy: test.final: chunk of length 65546: checksum 0x82701F3A != 0xC295B675
unsnappy: test.final: chunk of length 25832: checksum 0x641E20EA != 0xE0E7B4C5
```

That said, it pulled right up in an image viewer as the JPEG. I then shut down Firefox, copied that same sqlite3 file to a test file (it's locked with WAL/SHM files while Firefox is running) and verified the chain of custody on that file after a quick look at the schema and following my nose:

```
$ sqlite3 test.sqlite
sqlite> select request_url_no_query,response_body_id,response_principal_info from entries where response_body_id like '%d239%';
https://lemmy.ca/pictrs/image/2d5b64a8-f43d-4371-a15c-59654555b6ff.jpeg|{dcc46c14-70ec-4c6a-8537-1bee4688d239}|https://lemm.ee/service-worker.js
```

It is here I must stop and see what your team thinks, as I do see what's happening logistically but am unsure what the expectation of either Firefox or Lemmy UI is to be; as an end user this "feels like" permanent local storage is being used/triggered by `service-worker.js` instead of it actually going to cache (the `cache2/` subfolder, e.g. `about:cache` view).

I believe it a reasonable expectation as an end user that these images I'm looking at by the dozens if not hundreds are temporary, transient and should be in the cached data (as per `about:cache` etc.) and not being stored in site local storage which appears to be "forever" (requires manual Forget of the site to clear).

Thank you!

### Lemmy Instance Version

0.19.5

### Lemmy Instance URL

https://lemm.ee/

SleeplessOne1917:
Fixes #2792

It turns out that the problem was just that a max age wasn't specified for the `instance-cache`. This PR makes it so everything in that cache expires in 10 minutes.

### DIFF
--- a/src/shared/services/UserService.ts
+++ b/src/shared/services/UserService.ts
@@ -55,6 +55,9 @@ export class UserService {
 
     HttpService.client.logout();
 
+    // TODO: Remove this in a few releases when this cache has been deleted from most users' browsers
+    window.caches.delete("instance-cache");
+
     if (isAuthPath(location.pathname)) {
       location.replace("/");
     } else {
--- a/package.json
+++ b/package.json
@@ -145,7 +145,7 @@
       "sortpack"
     ]
   },
-  "packageManager": "pnpm@10.2.1+sha512.398035c7bd696d0ba0b10a688ed558285329d27ea994804a52bad9167d8e3a72bcb993f9699585d3ca25779ac64949ef422757a6c31102c12ab932e5cbe5cc92",
+  "packageManager": "pnpm@10.9.0+sha512.0486e394640d3c1fb3c9d43d49cf92879ff74f8516959c235308f5a8f62e2e19528a65cdc2a3058f587cde71eba3d5b56327c8c33a97e4c4051ca48a10ca2d5f",
   "engineStrict": true,
   "importSort": {
     ".js, .jsx, .ts, .tsx": {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -148,6 +148,7 @@ module.exports = (env, argv) => {
               handler: "NetworkFirst",
               options: {
                 cacheName: "instance-cache",
+                maxAgeSeconds: 600,
               },
             },
             {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -148,7 +148,9 @@ module.exports = (env, argv) => {
               handler: "NetworkFirst",
               options: {
                 cacheName: "instance-cache",
-                maxAgeSeconds: 600,
+                expiration: {
+                  maxAgeSeconds: 600,
+                },
               },
             },
             {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -132,25 +132,6 @@ module.exports = (env, argv) => {
           inlineWorkboxRuntime: true,
           runtimeCaching: [
             {
-              urlPattern: ({
-                sameOrigin,
-                url: { pathname, host },
-                request: { method },
-              }) =>
-                (sameOrigin || host.includes("localhost")) &&
-                (!(
-                  pathname.includes("pictrs") || pathname.includes("static")
-                ) ||
-                  method === "POST"),
-              handler: "NetworkFirst",
-              options: {
-                cacheName: "instance-cache",
-                expiration: {
-                  maxAgeSeconds: 600,
-                },
-              },
-            },
-            {
               urlPattern: ({ url: { pathname, host }, sameOrigin }) =>
                 (sameOrigin || host.includes("localhost")) &&
                 pathname.includes("static"),


Nothing4You:
what is the point of this cache?

considering it's network first, i assume this is intended to be usable as a fallback when offline or the server is unreachable. having this limited to 10 minutes doesn't sound like it'd be of much use.

https://developer.chrome.com/docs/workbox/caching-strategies-overview/#network_first_falling_back_to_cache

Nothing4You:
do we need cache for this at all?
is it intended to have some level of offline functionality in the web app?

maybe just remove this cache entirely, as it's not used anyway as long as the server is responsive?

it might explain some oddities i saw in the past where outdated info was shown.

SleeplessOne1917:
I second @tepoza that it's better to push out a change to bust the cache sooner rather than later. Regarding whether we should be caching at all: I think it's worth keeping for things like media and static files. For things from the API, there's no need unless we were to focus on making Lemmy uasable as a PWA offline. While not a bad goal, I don't think offline PWA usage is a priority with us currently trying to make version 1.0.

Nothing4You:
what are you trying to achieve with the cache though? as long as it's set to network first it doesn't provide any performance benefits, as it'll always try to fetch from the origin.
looking at it again now, i'm thinking the original intention of this block (`instance-cache`) was to just have this basically be a "no-cache" rule for the later defined caches? there's probably a better way to achieve that.
i'm not arguing for removal of the other two caches, just the one named `instance-cache`.

Nothing4You:
If you want to get this in for 0.19.12, I don't mind this getting merged in the current state if this isn't going to get resolved soon, but I don't think it adequately addresses #2792 and it should be kept open.

SleeplessOne1917:
> what are you trying to achieve with the cache though? as long as it's set to network first it doesn't provide any performance benefits, as it'll always try to fetch from the origin.

At the time I made these changes, I believe I was trying to make Lemmy downloadable as a PWA. I probably saw that caching was one of the things PWAs let you do and added it in for everything so that there'd be a fallback in case something couldn't be fetched. Looking at it now, I don't think caching the API requests is achieving anything helpful and it's best to nix the caching for things that aren't media or static files.

Like I said before, caching the API requests would only really be helpful if we wanted to make Lemmy usable as an offline PWA since it would fall back to using cached content when a connection to the server can't be established, but there's more that would need to be done to make that a good experience and it's far from being a priority feature at the moment.

I'll push up a change removing the `instance-cache` in a moment.

Nutomic:
Also adjust the title please.

Nothing4You:
I think we would also need to delete the old cache, not just no longer define it, to be sure that the old entries are gone.
If we don't do that the old entries will probably persist until manually cleared by a user.

Something like `await window.caches.delete("instance-cache")` seems like it would do the trick, not what's the ideal way to include this.

SleeplessOne1917:
Mering this now. @dessalines Do I need to make a separate PR to backport this to 0.19.x?

dessalines:
Yep it does need a backport.

src/shared/services/UserService.ts
```diff
@@ -55,6 +55,9 @@ export class UserService {
 
     HttpService.client.logout();
 
+    // TODO: Remove this in a few releases when this cache has been deleted from most users' browsers
+    window.caches.delete("instance-cache");
```
Nutomic, line 59: You can already remove this line from the main branch, it will take a few more months and a couple more 0.19.x releases until 1.0 gets published

PR #3118, opened 2025-05-01
Add `Vary: Cookie` Header (fixes #3117)

### LINKED ISSUE #3117: showAdultConsentModal is being cached by Nginx
### Requirements

### Summary

I wanted to try the content warning feature in lemmynsfw in the last update, but it showed the modal again and again despite pressing the "continue" button.

I had a chance to check it now and found that the problem was that the HTML page containing the `showAdultConsentModal: true` was cached by Nginx for a minute.

This has two solutions:
1. Using Express methods to make html content with `showAdultConsentModal: true` to not cache and making Nginx comply with it.
```js
// for example
if (showAdultConsentModal) {
  res.set('Cache-Control', 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0');
}
```
2. Using localStorage for showAdultConsentModal instead of cookies.

Frankly, I believe it makes more sense to continue with cookies. Depending on your feedback on this matter, I can work on it.

### Steps to Reproduce

1. Activate content_warning in an instance behind Nginx proxy with cache.
2. Open any page. It will prompt content_warning modal.
3. Click continue. It will re-ask again until cache invalidates.
4. You will be able to open any other page because you did accepted the modal and they're not cached.

### Technical Details

Not matters in this case.

### Lemmy Instance Version

0.19.11

### Lemmy Instance URL

https://lemmynsfw.com

xaegit:


src/server/middleware.ts
```diff
@@ -60,6 +60,7 @@ export function setCacheControl(
   }
 
   res.setHeader("Cache-Control", caching);
+  res.setHeader("Vary", "Cookie, Accept-Language");
```
matc-pub, line 63: You should move it up into the outer else clause. The login cookie would otherwise cause the static files to be cached per user. They also don't depend on the browser language.

xaegit: How about now? It doesn't modify cache behaviour of static files now. Though I'm not sure how `Vary` and `Cache-Control: private` works with each other.

xaegit: I just checked `Cache-Control: private` and I think we should still use `Vary: Cookie` for private pages too because it can still cache private pages locally.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#directives

matc-pub (approved):

xaegit:
@MrKaplan-lw You mean ‘application/activity+json’ header stuff?

I thought those requests were passed to backend in the nginx_internal file. But having ‘Accept’ header in ‘Vary’ header of UI wouldn’t hurt anyone either.

It’s up to maintainers call. I am fine.

Nutomic:
Yes `Accept` should also be included.

xaegit:
Added 👍

dessalines (approved):

PR #3120, opened 2025-05-01
Disable blur for NSFW images by default if content_warning exists

xaegit:
I'm trying to get rid of our own fork :) I can create same PR for 0.19 if it makes sense.


dessalines (approved):
Thx, its much better to open PRs to lemmy-ui sooner than later, otherwise you could end up in a merge hell if there are too many changes. A few instances had this problem with forks, so they stay much closer to main now.

Yes please open a PR for `release/v0.19` also if you would. I'll merge both after so we don't forget.

xaegit:
I agree. I forked it for quick development. But as you said, it has its difficulties. Here's 0.19 version: https://github.com/LemmyNet/lemmy-ui/pull/3128

Issue #1728, opened 2023-07-01
High contrast themes


Artoria2e5:
### Requirements

- [X] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [X] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

The combination of my angle-sensitive LCD display and my glasses is not great for the current low-contrast appearance used by Lemmy. I don't actually care about the buttons being green (they are big enough), but it would be great to have the text not be grey.

travis-jeans:
I made a Lemmy theme for the RBlind community which has higher contrast than the default Darkly or Litely themes. You can see a preview of the themes on the announcement page: [Announcing RBlind custom dark and light themes for Lemmy](https://rblind.com/post/3476242)

dessalines:
I would much appreciate you adding these themes to lemmy-ui , so that all servers and people could benefit from using them.

travis-jeans:
> I would much appreciate you adding these themes to lemmy-ui , so that all servers and people could benefit from using them.

I wasn't sure if that was allowed! We want to make some changes to the comment border colours and widths which was the reason for the PRs so once the themes have been updated to have those added I can submit them via PR. Alternatively, I can make a new PR to re-update them each time there's a major update to the themes. What would be better?

PR #3159, opened 2025-05-20
Added the RBlind theme files as discussed in issue #1728

travis-jeans:
### Why upload these themes to Lemmy?

@dessalines requested I upload the RBlind themes awhile back (in issue #1728). 

These CSS themes were made for the [RBlind](rblind.com) instance to improve the accessibility experience on Lemmy. They're based on litely-compact and darkly-compact and include a fair bit of extra CSS, root variables, and changes to bootstrap variables. It would be beneficial for other users to have access to these themes, as mentioned in issue #1728. Because of conflicts experienced using the themes as external style sheets, the best option would be to let Admins and Users access the theme in the Settings panel.

### About the RBlind themes

The history and documentation for these themes is available at [RBlind-Lemmy-Themes (Codeberg repository)](https://codeberg.org/travis-jeans/RBlind-Lemmy-Themes), and there is a GitHub repo as well. The short version is that the themes aim to be functional and visually appealing to the vision impaired community as well as be useful for everyone.

The README also includes screenshots of the themes and other information. The latest Changelog (today) included changes to the comment thread colours and renaming some files for simplicity.

Some areas I'd like some input on:
- Since the RBlind-Light and RBlind-Dark theme have some shared code which is imported by their respective SCSS files from RBlind-Theme.scss, which isn't the case for many of the existing Lemmy themes, would you prefer if I modified the themes to remove the RBlind-Theme.scss file then put a copy of the contents in RBlind-Dark.scss and RBlind-Light.scss to make the build process easier?
- Are there any suggested changes to the GNU license statement (particularly suggestions to avoid duplicates appearing during theme generation)? If it should appear once, should it be at the top of the RBlind-Theme.scss file so it appears at the top of the script? I read some information on the GNU site that license statements should be on every piece of code but in this case maybe it's too much since it ends up duplicating itself.

I don't have a Lemmy instance of my own so it would be beneficial for one or more people to test the themes don't break unexpectedly.

Happy to hear any other feedback about how to best organise the files. Myself and the RBlind admin team would be very happy if the RBlind-Themes could be a part of the official Lemmy themes.


src/assets/css/themes/RBlind-Dark.scss
```diff
@@ -0,0 +1,23 @@
...
+
+This file is part of RBlind Lemmy Themes.
+
+RBlind Lemmy Themes is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+
+RBlind Lemmy Themes is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with RBlind Lemmy Themes. If not, see <https://www.gnu.org/licenses/>. 
+
+*/
```
Nutomic, line 12: Lemmy code is all under [AGPL](https://github.com/LemmyNet/lemmy-ui/blob/main/LICENSE) so this should use the same one. AGPL has stronger  protections, because it also requires code changes running on a server to be open sourced. Also there is no need to include the license in every single file.

travis-jeans: So by uploading the themes to Lemmy will they automatically have the AGPL license applied to them, if I remove the license statements from the files? I think the RBlind Themes are already using the AGPL license actually - here's the link to the [LICENSE](https://codeberg.org/travis-jeans/RBlind-Lemmy-Themes/src/branch/master/LICENSE.md) file.

Do you think there needs to be an author line somewhere / a link to the original repository in the files themselves or will that be evident by the existing repository and myself as the author of the PR (that people can look up)?

Nutomic: Its clear from the git log who made this commit, and the log also includes a reference to this pull request with your description.

dessalines (commented):
Thx. Now many others will benefit from this.

After this is merged, I'll also add this to the `release/v0.19` branch.

SleeplessOne1917 (approved):

travis-jeans:
The RBlind-Theme.scss functions like main.css in the /css/ directory in that it should be imported into both RBlind-Dark and RBlind-Light. To avoid RBlind-Theme.scss being generated into a RBlind-Theme.css during the `pnpm themes:build` command, I've moved it to be next to main.css and renamed it to be a .css file instead of a .scss file.

Only caveat is that having it imported as a css file means [ruby sass can't run it](https://stackoverflow.com/questions/7111610/import-regular-css-file-in-scss-file)  and throws an import error. If the node sass used by lemmy-ui uses ruby maybe I can change it to be a scss file instead.
Actually I just realized the RBlind-Theme.scss file does contain one bootstrap variable so I might need to fix this. Sorry. I'll move the bootstrap part to be in the variables files instead.

travis-jeans:
I moved the bootstrap variable from the css to the variables files. Hopefully the files are in a better position now.

src/assets/css/RBlind-Theme.css
dessalines, line 1: This is a theme, so I'd really prefer this be inside the `css/themes` folder. If you need to make another embedded folder like `css/themes/rblind` , in order for it to work with the script, that'd be better.

travis-jeans: The RBlind-Theme.css file is not intended to be uploaded by the user/admin as a theme, since it produces some css changes but uses the Darkly variables since it does not have any RBlind variables imported. In development it was an intermediate step in theme development. It helps keep the bootstrap variables and the css separate, since most of the css in the light and dark themes are shared.

I could change the name to just RBlind.css to prevent any misunderstanding and also make a directory for it /rblind/ as you've suggested so it doesn't get uploaded to the theme list.

travis-jeans: Just made these changes.

dessalines (approved):
Looks good, thx!

PR #3163, opened 2025-05-23
Rblind theme [release/v0.19]

dessalines:
Cherry-picked from main


travis-jeans:
Just FYI, the "RBlind" should be in Pascal case for releases. This is to match the [instance name](https://rblind.com/) the themes were made for. It also helps distinguish it from the /r/blind subreddit which is also run by [OurBlind](https://ourblind.com/).

SleeplessOne1917 (approved):

PR #3175, opened 2025-06-06
Add a denial reason using the error.message field.

### LINKED ISSUE #1096: Denied applications do not receive any information about their denial

### Issue Summary

When a user applies to our instance and is denied, the user does not receive the denial reason. They are unable to login and unable to re-apply. Their user is created at the time of application, but no one can log into it. 

### Steps to Reproduce

1. Create a test account on an instance with applications. Apply
2. Deny the application
3. Attempt to log in with the test account

### Technical details

n/a

dessalines:
- All the ordinary i18n errors now use error.name .
- Fixes #1096

What it looks like now:

Screenshot_20250606_110311_Via_1

### DIFF
--- a/src/shared/components/home/login.tsx
+++ b/src/shared/components/home/login.tsx
@@ -81,18 +81,19 @@ async function handleLoginSubmit(i: Login, event: any) {
     });
     switch (loginRes.state) {
       case "failed": {
-        if (loginRes.err.message === "missing_totp_token") {
+        if (loginRes.err.name === "missing_totp_token") {
           i.setState({ show2faModal: true });
         } else {
-          // TODO: We shouldn't be passing error messages as args into i18next
-          toast(
-            I18NextService.i18n.t(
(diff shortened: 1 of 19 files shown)


src/shared/components/home/login.tsx
```diff
@@ -81,18 +81,19 @@ async function handleLoginSubmit(i: Login, event: any) {
...
+          let errStr = I18NextService.i18n.t(
+            loginRes.err.name === "registration_application_is_pending"
+              ? "registration_application_pending"
+              : loginRes.err.name,
           );
+          // If there's an error message, append it
+          if (loginRes.err.message) {
+            errStr = `${errStr}: ${loginRes.err.message}`;
+          }
+          toast(errStr, "danger");
```
dessalines, line 96: Main change is here. If theres an error message, it appends it, not using i18n.

SleeplessOne1917 (approved):

PR #3182, opened 2025-06-09
Hide Next button for Overview, Saved, Upvoted

Tyoda:
## Description

Hide the next button for Overview, Saved, Upvoted views on users' profiles when there are no more posts or comments left. Like it's done for the Comments and Posts views already.

Issue persists on the Uploads view but that would require a bit more work.

### Before
before
before-2

### After
after
after-2


dessalines (approved):
Nice, thx!

Issue #3121, opened 2025-05-01
Allow an grace period after publishing that allows editing the comment without "modified tag" showing up


NoSpiner:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

It is annoying when posting comment and you notice a typo and you have to edit it out, but now you have that "edited mark" next to your comment, as if you had changed or added something significant. It would be nice if there was like a 2-3min grace period after posting a comment that allows you to edit and not get that mark.

dessalines:
Wouldn't be too difficult to add, this would be a good first issue for someone.

jfaustino:
> It would be nice if there was like a 2-3min grace period after posting a comment that allows you to edit and not get that mark

Should it be an environment variable or hard coded? Is it ok that other front-ends see the edited tag?

dessalines:
Hard coded is fine. Basically if the modified is within the last 5 minutes or so, then don't show it as edited.

This should be a UI issue only, not on the lemmy back end.

Nutomic:
https://github.com/LemmyNet/lemmy-ui/pull/3197

PR #3197, opened 2025-06-18
Don't show edit mark if comment was edited in less than 5 minutes

jfaustino:
## Description

Implements #3121: Don't show the edit mark in a comment if it was edited in less than 5 minutes from the moment it was published.

### Before

before

### After

after

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -170,7 +170,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     const {
       creator_is_moderator,
       creator_is_admin,
-      comment: { id, language_id, published, distinguished, updated },
+      comment: { id, language_id, published, distinguished },
       creator,
       community,
       post,
@@ -260,7 +260,10 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 counts={counts}
               />
               <span>
-                <MomentTime published={published} updated={updated} />
+                <MomentTime
+                  published={published}
+                  updated={this.commentUpdatedTime}
+                />
               </span>
             </div>
             {/* end of user row */}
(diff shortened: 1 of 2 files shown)


SleeplessOne1917 (approved):

src/shared/components/comment/comment-node.tsx
```diff
@@ -260,7 +260,10 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 counts={counts}
               />
               <span>
-                <MomentTime published={published} updated={updated} />
+                <MomentTime
+                  published={published}
+                  updated={this.commentUpdatedTime}
+                />
```
dessalines: Move this logic to the MomentTime component. That way it will work for not just comment edits, but edits of all fields, like post edits too.

dessalines (changes requested):

src/shared/components/comment/comment-node.tsx
jfaustino: Done.

src/shared/components/common/moment-time.tsx
```diff
@@ -36,15 +36,32 @@ export class MomentTime extends Component<MomentTimeProps, any> {
...
   }
 
+  isRecentlyUpdated(): boolean {
+    if (this.props.updated) {
+      const published = new Date(this.props.published);
+      const updated = new Date(this.props.updated);
+      const updateLimit = addMinutes(published, 5);
+      return isBefore(updated, updateLimit);
+    }
+    return false;
```
dessalines: Use an else block.

src/shared/components/common/moment-time.tsx
```diff
@@ -36,15 +36,32 @@ export class MomentTime extends Component<MomentTimeProps, any> {
...
+  }
+
+  get updatedTime(): string {
```
dessalines: Why is this a get, but not the other one?

src/shared/components/common/moment-time.tsx
```diff
@@ -36,15 +36,32 @@ export class MomentTime extends Component<MomentTimeProps, any> {
...
+    if (this.isRecentlyUpdated()) {
+      return "";
+    }
+    return this.props.updated || "";
+  }
```
dessalines: use an else block, and don't return empty strings. Use undefined or null for missing values.

dessalines (changes requested):

src/shared/components/common/moment-time.tsx
jfaustino: It feels to me that `updatedTime` is a data attribute of the object while `isRecentlyUpdated` sounds more like a method.

src/shared/components/common/moment-time.tsx
jfaustino: Will do. Is it for code style or something else?

jfaustino:
Cool. Remember to close the related issue: https://github.com/LemmyNet/lemmy-ui/issues/3121

PR #3198, opened 2025-06-19
Click anywhere on byline to collapse a comment

philip-peterson:
## Description

This PR intends to increase target clickable area in furtherance of accessibility. Similarly to Reddit, allows the user to click anywhere alongside the byline (username, etc.) to collapse a comment. Affects URLs at `/post/*`.

Does not fix a separate issue.

## Disclaimer

This PR departs a bit from convention by exempting a code change from a couple of lint rules, and here is why I believe these are an exception. (For reference, the lint rules relate to clickable element requirements.): 

The button element for collapsing comments still exists as before, with the key and click handlers, and continues to serve as an accessibility-friendly way to toggle the `collapsed?` state of a CommentNode. This PR simply expands the clickable region (which is itself for accessibility reasons), but having screen readers treat the entire region as a button or clickable element is not a desired behavior, since it may contain other clickable elements within. This is intended as a convenience for graphical users (desktop, mobile).

## Screenshots

No visual difference, but new behavior:

click-anywhere

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -2,7 +2,7 @@ import { colorList, getCommentParentId } from "@utils/app";
 import { futureDaysToUnixTime, numToSI } from "@utils/helpers";
 import classNames from "classnames";
 import { isBefore, parseISO, subMinutes } from "date-fns";
-import { Component, linkEvent } from "inferno";
+import { Component, InfernoMouseEvent, linkEvent } from "inferno";
 import { Link } from "inferno-router";
 import {
   AddAdmin,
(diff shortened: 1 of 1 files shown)


dessalines (commented):
Not a fan of this, and it really isn't necessary since there's already a collapse button. Also its probably confusing for i18n.

philip-peterson:
The issue with a collapse button is that its horizontal position changes at each indentation level. It's also a smaller click target, which can be an issue for those with visual impairments. On mobile, most people are using their right hand which poses difficulty since the collapse button is on the left. When the click region is the entire bar, it makes the core workflow of collapsing to skip past irrelevant discussion much easier.

PR #3217, opened 2025-07-01
[0.19] Add search field to community sidebar

Nutomic:
Add search bar to community sidebar as [discussed on Lemmy](https://lemmy.ml/post/32355919/19527965).

Screenshot_20250701_162412-1

### DIFF
--- a/src/shared/components/community/sidebar.tsx
+++ b/src/shared/components/community/sidebar.tsx
@@ -59,6 +59,7 @@ interface SidebarState {
   leaveModTeamLoading: boolean;
   followCommunityLoading: boolean;
   purgeCommunityLoading: boolean;
+  searchText: string;
 }
 
 @tippyMixin
@@ -73,6 +74,7 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
     leaveModTeamLoading: false,
     followCommunityLoading: false,
     purgeCommunityLoading: false,
+    searchText: "",
   };
 
   constructor(props: any, context: any) {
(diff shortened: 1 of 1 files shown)


dessalines (commented):
I don't think using form actions like that is going to work correctly for the SPA (IE not doing any page reloading).

There is an example of searching in `communities.tsx`. Here are the main places to look at:

- https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/community/communities.tsx#L244
- https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/community/communities.tsx#L262
- https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/community/communities.tsx#L320

You might want to also rebase this off of `release/v0.19` first, to be able to test it. Then do another PR to main after that ones merged.

Nutomic:
I did test with 0.19 but opened the PR for main as there were no conflicts anyway. Changed the target branch now so its easier to test. Also changed it to use linkEvent like the example you linked.

src/shared/components/community/sidebar.tsx
```diff
@@ -162,6 +164,27 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
                   />
                   {this.canPost && this.createPost()}
                   {myUserInfo && this.blockCommunity()}
+                  <form
+                    class="d-flex"
+                    onSubmit={linkEvent(this, this.handleSearchSubmit)}
+                  >
+                    <input
+                      name="q"
+                      type="text"
```
SleeplessOne1917: Could make this `type="search"` instead.

src/shared/components/community/sidebar.tsx
```diff
@@ -162,6 +164,27 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
                   />
                   {this.canPost && this.createPost()}
                   {myUserInfo && this.blockCommunity()}
+                  <form
+                    class="d-flex"
+                    onSubmit={linkEvent(this, this.handleSearchSubmit)}
```
SleeplessOne1917, line 169: Since, like @dessalines said, this is a SPA where requests are driven by JS instead of HTML forms, I think it would be better to replace this `form` element with a [`search` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/search), move the handler for the button's `onClick` instead of a form `onSubmit`, and make the button `type="button"`.

Nutomic: It seems `<search>` element is not supported: 
```
rc/shared/components/community/sidebar.tsx:167:19 - error TS2339: Property 'search' does not exist on type 'JSX.IntrinsicElements'.

167                   <search>
```

src/shared/components/community/sidebar.tsx

SleeplessOne1917 (approved):

PR #3219, opened 2025-07-02
[0.19] Fixed ordering for search results

Nutomic:
Instead of trying to order different types of sort results by date and mixing up the types, use a fixed order so communities are always first, then users, then posts and finally comments. This should make it easier to navigate the results.

Screenshot_20250708_101248

### DIFF
--- a/src/shared/components/search.tsx
+++ b/src/shared/components/search.tsx
@@ -80,6 +80,7 @@ interface SearchProps {
   communityId?: number;
   creatorId?: number;
   page: number;
+  titleOnly: boolean;
 }
 
 type SearchData = RouteDataResponse<{
@@ -119,6 +120,7 @@ export function getSearchQueryParams(source?: string): SearchProps {
       communityId: getIdFromString,
       creatorId: getIdFromString,
       page: getPageFromString,
+      titleOnly: getTitleOnlyFromQuery,
     },
     source,
   );
@@ -137,6 +139,9 @@ function getSortTypeFromQuery(sort?: string): SortType {
 function getListingTypeFromQuery(listingType?: string): ListingType {
   return listingType ? (listingType as ListingType) : defaultListingType;
 }
+
+const getTitleOnlyFromQuery = (titleOnly?: string): boolean =>
+  titleOnly?.toLowerCase() === "true";
 
 const Filter = ({
   filterType,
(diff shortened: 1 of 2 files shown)


src/shared/components/search.tsx
```diff
@@ -213,22 +175,142 @@ const Filter = ({
...
+      <>
+        <h3>{I18NextService.i18n.t("communities")}</h3>
+        {communities.map(c => (
+          <div className="row flex-row">
+            {getListing(
+              <CommunityLink community={c.community} />,
+              c.counts.subscribers,
+              "number_of_subscribers",
+            )}
+          </div>
```
Nutomic, line 195: Not sure why the count is slipping into a new line now. I also want to add the short description and maybe some other info here, so community items dont appear much smaller than post or comment items.

SleeplessOne1917: Looking at what's used in the 0.19.12 UI:

```HTML
<div class="row">
    <div class="col-12">
        <span>
            <a class="community-link " title="Patient Gamers" href="/c/patientgamers">
                <picture>
                    <source srcset="https://lemmy.ml/pictrs/image/131fd58b-e418-4d0f-81c1-2521e920be61.png?format=webp&amp;thumbnail=96" type="image/webp">
                    <source srcset="https://lemmy.ml/pictrs/image/131fd58b-e418-4d0f-81c1-2521e920be61.png">
                    <source srcset="https://lemmy.ml/pictrs/image/131fd58b-e418-4d0f-81c1-2521e920be61.png?format=jpg&amp;thumbnail=96" type="image/jpeg">
                    <img class="overflow-hidden pictrs-image object-fit-cover img-icon me-1" src="https://lemmy.ml/pictrs/image/131fd58b-e418-4d0f-81c1-2521e920be61.png" alt="" title="" loading="lazy">
                </picture>
                <span class="overflow-wrap-anywhere">Patient Gamers</span>
            </a>
        </span>
        <span> - 10.2K subscribers</span>
    </div>
</div>
```

With the change in the PR, [the `col-12` `div` is now wrapping the row](https://github.com/LemmyNet/lemmy-ui/pull/3219/files#diff-de316adc11f4e30079afa914457d3dad9c6b760e132ac4dc11fe66c13719690fR959) instead of the other way around as it was before.

Nutomic: Thanks, was enough to remove the `row` class.

SleeplessOne1917 (approved):

PR #3220, opened 2025-07-02
[0.19] Add checkbox for title only search

Nutomic:
Screenshot_20250702_163813

The style looks a bit odd even though I used the same classes as other checkboxes in the code. Still need to update value from the form input.

### DIFF
--- a/src/shared/components/search.tsx
+++ b/src/shared/components/search.tsx
@@ -80,6 +80,7 @@ interface SearchProps {
   communityId?: number;
   creatorId?: number;
   page: number;
+  titleOnly: boolean;
 }
 
 type SearchData = RouteDataResponse<{
@@ -125,6 +126,7 @@ export function getSearchQueryParams(source?: string): SearchProps {
       communityId: getIdFromString,
       creatorId: getIdFromString,
       page: getPageFromString,
+      titleOnly: getTitleOnlyFromQuery,
     },
     source,
   );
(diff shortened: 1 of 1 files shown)


src/shared/components/search.tsx
```diff
@@ -630,6 +637,18 @@ export class Search extends Component<SearchRouteProps, SearchState> {
             )}
           </button>
         </div>
+        <div className="col-auto form-check ms-2 mt-0 h-min d-flex align-items-center">
```
SleeplessOne1917, line 640: > The style looks a bit odd even though I used the same classes as other checkboxes in the code. 

It could be the `d-flex` and `align-items-center` that's doing it. [The Bootstrap docs](https://getbootstrap.com/docs/5.2/forms/checks-radios/#without-labels) don't use any classes to change the `display` property like `d-flex` does, and [other places in this codebase that use `form-check`](https://github.com/search?q=repo%3ALemmyNet%2Flemmy-ui%20form-check&type=code) also don't include `d-flex`.

SleeplessOne1917:
Should this checkbox always be shown? If I understand correctly, it only applies when searching for posts or all types (since all includes posts).

Nutomic:
> Should this checkbox always be shown? If I understand correctly, it only applies when searching for posts or all types (since all includes posts).

Thats right for 0.19, but in 1.0 it also checks the [name and title fields](https://github.com/LemmyNet/lemmy/blob/main/crates/db_views/search_combined/src/impls.rs#L262) for all other types. So it doesnt make sense to add extra logic for this and later remove it again.

src/shared/components/search.tsx
Nutomic: Without those the checkbox is aligned at the top of the search button and not centered. And I also dont see any way in bootstrap docs to use secondary color for the checkbox.

PR #3251, opened 2025-07-28
[Main] remove all caches

dessalines:


SleeplessOne1917 (approved):

src/shared/services/UserService.ts
```diff
@@ -57,6 +57,8 @@ export class UserService {
 
     // TODO: Remove this in a few releases when this cache has been deleted from most users' browsers
     window.caches.delete("instance-cache");
+    window.caches.delete("static-cache");
+    window.caches.delete("image-cache");
```
Nutomic: You can remove all these, the removal will be done with 0.19.x and is not necessary with 1.0 anymore.

dessalines: Done

Nutomic (approved):

Issue #2601, opened 2024-07-14
Inbox doesn't work if it's the first page to be viewed


dullbananas:
### Requirements

- [X] This is a bug report, and if not, please post to https://lemmy.ml/c/lemmy_support instead.
- [X] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this issue already exists.
- [X] It's a single bug. Do not report multiple bugs in one issue.
- [X] It's a frontend issue, not a backend issue; Otherwise please create an issue on the [backend repo](https://github.com/LemmyNet/lemmy) instead.

### Summary

If you start at the homepage then use the inbox button, then it works.

If you start with the inbox URL (e.g. when clicking the "inbox" link in an email), then it asks you to log in.

### Steps to Reproduce

1. Go to a new browser tab that doesn't already have Lemmy open
2. Go directly to the /inbox URL

### Technical Details

Safari from iOS 17.5.1

### Lemmy Instance Version

0.19.5

### Lemmy Instance URL

https://lemmy.ca

SleeplessOne1917:
This doesn't sound like a bug. How could a non-logged in user see an inbox when inboxes are associated with accounts?

dullbananas:
This happens when I'm logged in

SleeplessOne1917:
I'm unable to reproduce this issue both locally and in production testing with firefox. Do you run into this issue with other browsers, or only Safari?

dullbananas:
It randomly started working correctly a minute ago

dullbananas:
I just got an email notification and I had this problem again with the inbox link

SleeplessOne1917:
When it happened last time, was that also from following an email link?

dullbananas:
Yes

SleeplessOne1917:
I've finally gotten around to trying to fix this and am running into a lot of trouble trying to test this locally. I am unable to set an email for a user when testing this locally. I tried adding the email config from `./config/defaults.hjson` to `./docker/lemmy.hjson` and even then I'm unable to set it. @dullbananas  Does this happen any time when following a link that isn't sent via email? Otherwise, I'm not sure how to go about reproducing the issue, and I don't think this should hold up the 0.19.6 release if that's the case.

dullbananas:
Based on the "steps to reproduce" list, it seems like I was able to replicate it without email a while ago, but I haven't been able to do that recently.

[...]

PR #3260, opened 2025-07-29
Set SameSite: strict for auth cookie (fixes #1059)

### LINKED ISSUE #1059: Cookies should use SameSite attribute.
Cookie “jwt” does not have a proper “SameSite” attribute value. Soon, cookies without the “SameSite” attribute or with an invalid value will be treated as “Lax”. This means that the cookie will no longer be sent in third-party contexts. If your application depends on this cookie being available in such contexts, please add the “SameSite=None“ attribute to it. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite

Nutomic:
Works fine in local testing.


SleeplessOne1917:
[This created issues for users navigating to the site from email notifications.](https://github.com/LemmyNet/lemmy-ui/issues/2601#issuecomment-2313322667)

dessalines:
👍

PR #3262, opened 2025-07-29
Fix host splitting for IPv6 (fixes #2374)

### LINKED ISSUE #2374: lemmy-ui does not support IPv6, splits hostname incorrectly
### Requirements

### Summary

Setting LEMMY_UI_HOST to "[::]:1234" let's lemmy-ui crash, it should result in listening to IPv6 ::.

### Steps to Reproduce

1. Set LEMMY_UI_HOST to "[::]:1234" 
2. Start lemmy
3. Get an error message (see below)

```
(node:1) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: getaddrinfo ENOTFOUND [
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:26)
Emitted 'error' event on Server instance at:
    at GetAddrInfoReqWrap.doListen [as callback] (node:net:2093:12)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:17) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: '['
}

Node.js v21.6.0
``


### Technical Details

The problem seems to be (at least) in index.tsx:

```
const [hostname, port] = process.env["LEMMY_UI_HOST"]
  ? process.env["LEMMY_UI_HOST"].split(":")
  : ["0.0.0.0", "1234"];
```

Splitting on : does not work for IPv6. The code should:

* check for existence of [ and ]
* if yes, assume IPv6
* Split at ]: instead

The problem is without an IPv6 listener, lemmy-ui cannot run in IPv6 only environments.
Lemmy itself (backend) seems to work fine with IPv6 already.

### Lemmy Instance Version

0.19.3

### Lemmy Instance URL

any

Nutomic:


dessalines (approved):
Tested, seems fine.

Issue #3221, opened 2025-07-02
Add the ability to tag users like on RES #2353 (implement UI)


wiki-me:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

A continuation of [this](https://github.com/LemmyNet/lemmy/issues/2353) issue.

PR #3347, opened 2025-08-28
Add ability to create a note for a person, and view vote totals

dessalines:
## Description

Add ability to create a note for a person, and view vote totals.

- Fixes https://github.com/LemmyNet/lemmy-ui/issues/3221
- Fixes https://github.com/LemmyNet/lemmy-ui/issues/3228


src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -562,7 +577,7 @@ export default class ContentActionDropdown extends Component<
     this.setState({ dropdownOpenedOnce: true });
   }
 
-  toggleModDialogShow(
+  toggleDialogShow(
```
dessalines, line 581: Renamed this because its not a mod dialog, but it can use the same code.

wiki-me:
Red usually indicates "stop" or that something is bad or dangerous . but i often use this (i think like 95% of the times) to indicate something positive. using RES black over a white background seems better.

Also when i change something on RES it adds the notes immdediately. I don't think it is a overkill but a nice way to polish the UI and make it a little more efficient, if you decided for example you want to pay more attention to someone you will do that immdediately after writing the note.

SleeplessOne1917:
Codewise this looks good to me. I think @wiki-me's feedback makes sense: a note isn't necessarily negative. `info` or it's subtle variant are good choices [from Bootstrap's theming system](https://getbootstrap.com/docs/5.3/customize/color/#colors).

One thing that stood out to me from #3221 and the older issue it links is that it describes it as a tag system instead of just a note system. I've never used RES, but what comes to mind is that if a user adds a new tag/note to another user, they will be able to select from previously their previously defined tags when making a tag/note about yet another user. Perhaps they can even browse groups of users based on how they tagged/noted them, or place multiple tags on users.

The only thing holding up approval on my end is the coloring bit from the first paragraph. If users think the functionality discussed in the second paragraph would be good to add, I'm pretty sure it could be added in the future without breaking compatibility with the v1 API.

dessalines:
K I've made it live update now.

I tried a few different styles out, but only `text-info` looked decent.

Screenshot_2025-08-31-17-18-36-156_com duckduckgo mobile android

SleeplessOne1917 (approved):

PR #3348, opened 2025-08-28
Add ability to block all persons from an instance.

### LINKED ISSUE #3286: Add person blocking (in addition to the existing community blocking)
### Requirements

These endpoints are now: 

- `userBlockInstanceCommunities`
- `userBlockInstancePersons`

dessalines:
- Adds *UserBlockInstancePersons* alongside the already-existing *UserBlockInstanceCommunities*.
- Also cleans up some of the i18n code for blocking / unblocking items.
- Fixes #3286

## Screenshots

Screenshot_2025-08-28-13-59-31-298_mark via

### DIFF
--- a/src/shared/components/person/settings.tsx
+++ b/src/shared/components/person/settings.tsx
@@ -10,7 +10,8 @@ import {
   setIsoData,
   showLocal,
   updateCommunityBlock,
-  updateInstanceBlock,
+  updateInstanceCommunitiesBlock,
+  updateInstancePersonsBlock,
   updateMyUserInfo,
   updatePersonBlock,
 } from "@utils/app";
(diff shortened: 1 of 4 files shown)


src/shared/components/person/settings.tsx
```diff
@@ -615,36 +629,85 @@ export class Settings extends Component<SettingsRouteProps, SettingsState> {
...
+      <>
+        <h2 className="h5">
+          {I18NextService.i18n.t(
+            "blocked_all_communities_from_these_instances",
+          )}
+        </h2>
+        <ul className="list-unstyled mb-0">
+          {this.state.instanceCommunitiesBlocks.map(i => (
+            <li key={i.id}>
+              <span>
```
SleeplessOne1917: Is the span needed? If you need the button to be inline, you could use `d-inline` or `d-inline-block` on it.

dessalines: I can try that for all these.

dessalines:
> It may be easier to understand if the instance block cards are in the same column instead of same row, but not enough to withold approval for.

This one I don't think is doable, because you can't determine the direction of bootstraps column reflow. On mobile, they'd be all vertical, and we wouldn't want to have the order be `users, block_instance_comms, communities, block_instance_users`. Better to have the instance blocks in order.

src/shared/components/person/settings.tsx
dessalines: Ended up not being necessary

PR #3371, opened 2025-09-01
Improve development instructions to use remote instance

Nutomic:
With these instructions its not necessary to compile the backend locally which makes everything much easier.


dessalines:
I have a [test.sh script here now](https://github.com/LemmyNet/lemmy-ui/blob/main/scripts/test.sh), but its currently not working either.

I found 2 issues, because https://voyager.lemmy.ml/api/v4/post/list is working, but http://voyager.lemmy.ml:8536/api/v4/post/list isn't

- That nginx needs to route / allow regular http requests (IE not just https).
- It also needs to expose the `:8536` backend in its internal docker and nginx.

Nutomic:
Got it working now with some changes to env var handling. Using `LEMMY_UI_BACKEND_REMOTE=voyager.lemmy.ml pnpm dev` you can now do all actions through the remote instance, including login, commenting etc. 

Remaining problem is that you get logged out after reloading the page. Tried [these instructions](https://stackoverflow.com/a/46412839) but couldnt figure out how to set `withCredentials=true` in our case.

dessalines:
No env var changes should be necessary... that test.lemmy.ml nginx just needs to properly route http and expose its `:8356`. 

If you can access http://voyager.lemmy.ml:8536/api/v4/post/list (not working rn), then any lemmy-ui client should work fine.

Nutomic:
It should work with http so that it can also be used with any production instance as backend. Same as [Voyager](https://github.com/aeharding/voyager) and other alternative frontends.

README.md
```diff
@@ -4,7 +4,18 @@ The official web app for [Lemmy](https://github.com/LemmyNet/lemmy), written in
...
+You need to have [pnpm](https://pnpm.io/installation) installed. Then run the following:
+
+```bash
+git clone https://github.com/LemmyNet/lemmy-ui.git
+cd lemmy-ui
+pnpm install
+LEMMY_UI_BACKEND_REMOTE=voyager.lemmy.ml pnpm dev
+```
+
+Finally open `http://0.0.0.0:1234` in your browser. This uses the public test instance `https://voyager.lemmy.ml/` as backend. For more details such as developing with a locally compiled Lemmy backend, read the [documentation](https://join-lemmy.org/docs/contributors/01-overview.html).
```
SleeplessOne1917, line 18: Should the documentation URL be https://join-lemmy.org/docs/contributors/02-local-development.html instead?

Nutomic: To me it makes sense to link the general overview first, but either one works.

Nutomic:
If this approach doesnt work, another option would be to route all api requests through a proxy in the Nodejs process only for development. That way the browser only connects to localhost and wont complain about CORS.

Nutomic:
Got it working. Took some experimentation but the required changes are very small. Just run `LEMMY_UI_BACKEND_REMOTE=voyager.lemmy.ml pnpm dev` to connect with the test instance as backend and interact with it (login, vote, comment etc). The test instance doesnt have pictrs enabled so I wasnt able to test that, but we can fix it later if needed.

With this change contributions will be much easier, as it only requires git and pnpm. No need to setup postgres or compile rust.

README.md
dessalines: Hrm... Ya I agree lets keep the link and just move these instructions to join-lemmy.

README.md
```diff
@@ -13,8 +24,9 @@ The following environment variables can be used to configure lemmy-ui:
 | `ENV_VAR`                      | type     | default          | description                                                                         |
 | ------------------------------ | -------- | ---------------- | ----------------------------------------------------------------------------------- |
 | `LEMMY_UI_HOST`                | `string` | `0.0.0.0:1234`   | The IP / port that the lemmy-ui isomorphic node server is hosted at.                |
-| `LEMMY_UI_LEMMY_INTERNAL_HOST` | `string` | `0.0.0.0:8536`   | The internal IP / port that lemmy is hosted at. Often `lemmy:8536` if using docker. |
-| `LEMMY_UI_LEMMY_EXTERNAL_HOST` | `string` | `0.0.0.0:8536`   | The external IP / port that lemmy is hosted at. Often `DOMAIN.TLD`.                 |
+| `LEMMY_UI_BACKEND_REMOTE`      | `string` |                  | Domain of a remote Lemmy instance to connect using HTTPS and port 403               |
```
dessalines: Is this one optional? Might be good to have `undefined` in the default column, and maybe move this below the others.

Also would be good to add a note that its only necessary if you're connecting to a remote server.

dessalines (commented):
The names of these are probably fine.

My gut feeling is that we could probably remove `LEMMY_UI_BACKEND_EXTERNAL` in some way, and just rely on `window.location`, but don't worry about it for now.

dessalines:
And after this merges we also gotta remember to update `lemmy-ansible` to reflect these new env changes, maybe just leaving as a draft PR so we don't forget for 1.0.

README.md
Nutomic: I want to keep the instruction directly in the repo so its easy to start contributing directly. I can also copy-paste it to the docs so its available in both places.

README.md
Nutomic: Done, but kept it here so its together with the other backend env vars.

Nutomic:
Opened an issue for lemmy-ansible (https://github.com/LemmyNet/lemmy-ansible/issues/271). It would really be easier to keep the env vars unchanged, but `LEMMY_UI_LEMMY_BACKEND_INTERNAL` is just too long.

I definitely want to merge this quickly so that other people can contribute more easily and help with development.

Nutomic:
@SleeplessOne1917 If dessalines also approved already, you can merge it directly ;)

PR #3375, opened 2025-09-02
Indicate when registration is disabled (fixes #2070)

### LINKED ISSUE #2070: Make it more clear when Registrations are disabled
### Requirements

On Instances where account creation is disabled I feel like they should not show the sign-up form with username, email, and password fields because it can be confusing to people as it looks like the Instance is open when it isn't.
gpe8j1fe8rjb1

As an example the person (credit to u/TheTwelveYearOld on Reddit for this screenshot) is trying to sign up to an Instance with closed registrations but due to how the UI works it's not entirely clear and to others not trying to sign up that registrations there are closed and it looks like Registrations are open.

My proposal is to make it so that when registrations are closed it'll show a message, similar to the application form but without a field to put in data explaining that Account creation has been disabled on this Instance and that users cannot create an account here. 

I made a mockup of what something like that might look like in an image editor:
Lemmy Registration Disabled example

Nutomic:
If registration is disabled, disables the "Create account" link in navbar and redirects users away from /signup.

### DIFF
--- a/src/shared/components/app/navbar.tsx
+++ b/src/shared/components/app/navbar.tsx
@@ -100,6 +100,7 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
     const person = this.props.myUserInfo?.local_user_view.person;
     const registrationClosed =
       siteView?.local_site.registration_mode === "Closed";
+
     return (
       <div className="shadow-sm">
         <nav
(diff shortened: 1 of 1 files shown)


src/shared/components/app/navbar.tsx
```diff
@@ -448,8 +450,13 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
...
                       to="/signup"
-                      className="nav-link"
-                      title={I18NextService.i18n.t("sign_up")}
+                      className={
+                        registrationClosed ? "nav-link pe-none" : "nav-link"
+                      }
+                      aria-disabled={registrationClosed}
+                      title={I18NextService.i18n.t(
+                        registrationClosed ? "registration_closed" : "sign_up",
+                      )}
```
Nutomic: It seems like both `pe-none` and `aria-disabled` prevent title from being shown.

dessalines: Instead I would just hide this entire block. IE above this `NavLink`, do `{!registrationClosed && <NavLink ...`

SleeplessOne1917: I second Dessalines on this. Normally, disabling an input communicates that an input is temporarily unavailable (like when a form is submitting), or it is used to prevent a user from putting the form in an incorrect state (like how the send notifications to email checkbox is disabled when no email is specified). Hiding an input is more appropriate when a user shouldn't have access to a functionality in the first place, like how mod and admin actions aren't shown for non mod/admin users. I think a signup button on an instance with registration disabled falls under the latter.

Nutomic: Okay done

dessalines (approved):

SleeplessOne1917 (approved):

PR #3376, opened 2025-09-02
Show community languages in sidebar (fixes #1009)

### LINKED ISSUE #1009: Show which languages are allowed in instance/community sidebar
Instances and communities can now specify which languages are allowed. This information is not visible though. It should be shown in the sidebar.

Nutomic:
Not sure whats the best place for this, I put it here for now:
Screenshot_20250902_130337

For site languages:
Screenshot_20250902_131321

### DIFF
--- a/src/shared/components/community/sidebar.tsx
+++ b/src/shared/components/community/sidebar.tsx
@@ -28,7 +28,7 @@ import { CommunityLink } from "../community/community-link";
 import { PersonListing } from "../person/person-listing";
 import { tippyMixin } from "../mixins/tippy-mixin";
 import CommunityReportModal from "@components/common/modal/community-report-modal";
-import { renderLanguageList } from "@components/common/language-list";
+import { LanguageList } from "@components/common/language-list";
 
 interface SidebarProps {
   community_view: CommunityView;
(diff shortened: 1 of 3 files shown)


src/shared/components/community/sidebar.tsx
```diff
@@ -157,6 +157,9 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
       community: { name, ap_id, id, posting_restricted_to_mods, visibility },
       community_actions: { received_ban_at: bannedFromCommunity } = {},
     } = this.props.community_view;
+    const langs = this.props.allLanguages.filter(x =>
+      this.props.communityLanguages?.includes(x.id),
+    );
```
dessalines: I'd externalize this into a utils function that both can use, and maybe just return the language names.

`function buildLanguangeNames(allLanguages, languageIds): string[]`

Nutomic:
Made it so the helper function directly returns the view as its the same in both cases. Not sure where to put it now, if the helper is under utils it cant compile the html. And under common everything has its own file and own class which is unnecessary in this case.

dessalines:
Maybe src/shared/components/common/language-select.tsx ? Either that or just make a new file under common. 

It also doesn't need to be a full class, you can create a function component. [Example of one](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/common/badges.tsx#L19) . 

EDIT oh nm I see it already is one.

SleeplessOne1917:
Should "Undetermined" show up in language lists?

dessalines:
As an option yes, but I'm not sure what would be the best way to i18n that. Might need to create a special exception just for that one.

Nutomic:
Moved function to a separate file. Also added localization for "undetermined language" with https://github.com/LemmyNet/lemmy-translations/pull/187. I called it "unknown language" to keep things simple.

Nutomic:
Changed woodpecker config from `git submodule update --recursive --remote` to `git submodule update` so it uses the same translation version as in the commit, and doesnt throw an unnecessary error.

PR #3377, opened 2025-09-02
Hide bio for banned users (fixes #961)

### LINKED ISSUE #961: Moderating federated user profiles
For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

Continuation from here: https://github.com/LemmyNet/lemmy/issues/2277#issuecomment-1438624728

> I realize this is a federation issue to some extend (user from lemmy.ml and image from community.xmpp.net), but I am an admin on slrpnk.net and there is no apparent option on the [originally linked example](https://slrpnk.net/u/nintendolover@lemmy.ml) to remove, hide or purge this stuff from being displayed as part of the slrpnk.net domain.

Not sure which way to approach this best in a federated network, but it would be good to at least be able to hide such user profiles so that the material is not publicly accessible on the domain you are administrating.

Nutomic:


src/shared/components/person/profile.tsx
```diff
@@ -567,14 +567,19 @@ export class Profile extends Component<ProfileRouteProps, ProfileState> {
 
         const isUpload = view === "Uploads";
 
+        var bio: string | undefined = "";
+        if (!personRes.person_view.creator_banned) {
+          bio = personRes.person_view.person.bio;
+        }
```
dessalines: `const bio = !personRes.person_view.creator_banned ? personRes.person_view.person.bio : "";`

Nutomic: Right, after using Rust Im not used to ternary operator anymore.

src/shared/components/person/profile.tsx
```diff
@@ -567,14 +567,18 @@ export class Profile extends Component<ProfileRouteProps, ProfileState> {
 
         const isUpload = view === "Uploads";
 
+        var bio = !personRes.person_view.creator_banned
```
dessalines: Use const here.

Only use `const` or `let` in typescript, not var. let means mutable.

src/shared/components/person/profile.tsx
dessalines: Rust is so much better for allowing `let X = if {..` blocks and abandoning ternary.

typescript unfortunately only supports `let x = switch (...` or the ternary.

src/shared/components/person/profile.tsx
SleeplessOne1917: Adding to your point, `const` is for immutable references, but a const object can still be mutated. For example, this is valid:

```typescript
const arr = [1, 2, 3];
arr.push(4);
console.log(arr[arr.length - 1]);
```

Basically, `let` allows a variable to be reassigned and `const` doesn't. So:

```typescript
// Valid
let arr1 = [1, 2, 3];
arr1 = [4, 5, 6]; // Can reassign this since arr1 is defined with let
console.log(arr1[arr1.length - 1]);

// Invalid
const arr2 = [1, 2, 3];
arr2 = [4, 5, 6]; // Will throw an error since arr2 is defined with const
console.log(arr2[arr2.length - 1]);
```

---

The reason usage of `var` is discouraged is because it hoists variables to the top of a function's scope, even if the variable is defined in a more specific scope. For (a contrived) example, the following is valid JS/TS code:

```javascript
function printObj() {
    if (true) {
        var obj = { key: "value" };
    }

    // This will print the object despite the fact that this statement is outside the scope
    // obj is defined in. obj is hoisted to be in the scope of the printObj function.
    console.log(obj);
}
```

dessalines: If we weren't moving to rust, and were gonna stick with lemmy-ui for years, I'd recommend we use something like https://github.com/immutable-js/immutable-js . It at least can enforce immutability for object / list types. 

Its kind of astonishing that so many front end problems come from mutability, and javascript / typescript, the most popular front end languages, do it so terribly.

Nutomic: Thanks for the explanation. Odd that the linter doesnt even catch such a basic problem.

dessalines (approved):

PR #3385, opened 2025-09-04
Setting for community/post notifications

### LINKED ISSUE #3222: Add a "disable inbox replies" option of posts and comments (implement UI)
### Requirements

A continuation of this [issue](https://github.com/LemmyNet/lemmy/issues/3042).

Nutomic:
## Description

Add community/post settings for https://github.com/LemmyNet/lemmy/pull/5604

Translations: https://github.com/LemmyNet/lemmy-translations/pull/185 https://github.com/LemmyNet/lemmy-translations/pull/188

JS client: https://github.com/LemmyNet/lemmy-js-client/pull/684

## Screenshots

Screenshot_20250909_120854

Screenshot_20250909_122626

edit by dess:

- Fixes #3222

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -556,6 +556,7 @@ export class Community extends Component<CommunityRouteProps, State> {
           return (
             <PostListings
               posts={this.state.postsRes.data.posts}
+              showDupes="ShowSeparately"
               markable
               enableNsfw={enableNsfw(siteRes)}
               showAdultConsentModal={this.isoData.showAdultConsentModal}
--- a/src/shared/components/community/sidebar.tsx
+++ b/src/shared/components/community/sidebar.tsx
@@ -176,6 +176,21 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
             <div className="card-body">
               {this.communityTitle()}
               {this.props.editable && this.adminButtons()}
+              {received_ban_at && (
+                <div
+                  className="alert alert-danger text-sm-start text-xs-center mt-2"
+                  role="alert"
+                >
+                  <Icon
+                    icon="ban"
+                    inline
+                    classes="me-sm-2 mx-auto d-sm-inline d-block"
+                  />
+                  <T i18nKey="banned_from_community_blurb" className="d-inline">
+                    #<strong className="fw-bold">#</strong>#
+                  </T>
+                </div>
+              )}
               {!received_ban_at && (
                 <>
                   <SubscribeButton
@@ -186,36 +201,35 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
                     showRemoteFetch={!this.props.myUserInfo}
                   />
                   {this.canPost && this.createPost()}
-                  {this.props.myUserInfo && this.blockCommunity()}
-                  <div className="mb-2 d-flex">
-                    <CommunityNotificationSelect
-                      current={this.state.notifications}
-                      onChange={this.handleNotificationChange}
-                    />
-                  </div>
-                  <form
-                    class="d-flex"
-                    onSubmit={linkEvent(this, this.handleSearchSubmit)}
-                  >
-                    <input
-                      name="q"
-                      type="search"
-                      className="form-control flex-initial"
-                      placeholder={`${I18NextService.i18n.t("search")}...`}
-                      aria-label={I18NextService.i18n.t("search")}
-                      onInput={linkEvent(this, this.handleSearchChange)}
-                      required
-                      minLength={1}
-                    />
-                    <button
-                      type="submit"
-                      class="btn btn-outline-secondary ms-1"
-                    >
-                      <Icon icon="search" />
-                    </button>
-                  </form>
                 </>
               )}
+              <>
+                {this.props.myUserInfo && this.blockCommunity()}
+                <div className="mb-2 d-flex">
+                  <CommunityNotificationSelect
+                    current={this.state.notifications}
+                    onChange={this.handleNotificationChange}
+                  />
+                </div>
+                <form
+                  class="d-flex"
+                  onSubmit={linkEvent(this, this.handleSearchSubmit)}
+                >
+                  <input
+                    name="q"
+                    type="search"
+                    className="form-control flex-initial"
+                    placeholder={`${I18NextService.i18n.t("search")}...`}
+                    aria-label={I18NextService.i18n.t("search")}
+                    onInput={linkEvent(this, this.handleSearchChange)}
+                    required
+                    minLength={1}
+                  />
+                  <button type="submit" class="btn btn-outline-secondary ms-1">
+                    <Icon icon="search" />
+                  </button>
+                </form>
+              </>
               {!this.props.myUserInfo && (
                 <div className="alert alert-info" role="alert">
                   <T
@@ -244,21 +258,6 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
                     classes="me-sm-2 mx-auto d-sm-inline d-block"
                   />
                   <T i18nKey="community_locked_message" className="d-inline">
-                    #<strong className="fw-bold">#</strong>#
-                  </T>
-                </div>
-              )}
-              {received_ban_at && (
-                <div
-                  className="alert alert-danger text-sm-start text-xs-center"
-                  role="alert"
-                >
-                  <Icon
-                    icon="ban"
-                    inline
-                    classes="me-sm-2 mx-auto d-sm-inline d-block"
-                  />
-                  <T i18nKey="banned_from_community_blurb" className="d-inline">
                     #<strong className="fw-bold">#</strong>#
                   </T>
                 </div>
--- a/src/shared/components/post/post.tsx
+++ b/src/shared/components/post/post.tsx
@@ -599,6 +599,7 @@ export class Post extends Component<PostRouteProps, PostState> {
               <PostListing
                 post_view={res.post_view}
                 crossPosts={res.cross_posts}
+                showDupes="Expanded"
                 showBody
                 showCommunity
                 admins={siteRes.admins}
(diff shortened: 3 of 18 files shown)


src/shared/components/post/post.tsx
```diff
@@ -319,7 +338,11 @@ export class Post extends Component<PostRouteProps, PostState> {
       comment_id: getCommentIdFromProps(props),
     });
     if (token === this.fetchPostToken) {
-      this.setState({ postRes });
+      console.log(postRes);
+      this.setState({
+        postRes,
+        notifications: postRes.data.post_view.post_actions.notifications,
```
Nutomic: This doesnt work, not sure whats the right way to init this.

Nutomic:
Im regularly getting CI lint errors related to translations even though it passes locally:
```
src/shared/components/community/sidebar.tsx:79:5 - error TS2322: Type '"notification_mode_all_posts_and_comments"' is not assignable to type 'NoOptionI18nKeys'.

79     i18nKey: "notification_mode_all_posts_and_comments",
       ~~~~~~~
```

src/shared/components/community/sidebar.tsx
```diff
@@ -660,6 +696,15 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
     i.setState({ purgeReason: event.target.value });
   }
 
+  async handleNotificationChange(i: Sidebar, event: any) {
+    i.setState({ notifications: event.target.value });
+    const form = {
+      community_id: i.props.community_view.community.id,
+      mode: i.state.notifications,
+    };
+    await HttpService.client.updateCommunityNotifications(form);
```
dessalines: API calls should only (wherever possible), be in top-level routes. 

So instead add this as a prop:

`onUpdateCommunityNotifs(form: UpdateCommunityNotifications): void;`

And here, do `i.props.onUpdateCommunityNotifs(form)`

Then in the top level routes that use the sidebar, you'll make the API call. Follow `onEditCommunity` for an example of how to do this.

There's examples of this in this file also. There's a lot of reasons why this is the right way to do it... there's a long post from me here somewhere and a lot of videos regarding state in jetpack compose on why this is the best way to do it, but the general rule is that only your *screens* / top level routes should be altering data or making API calls, and the smaller components should simply be passing changes up to them.

src/shared/components/post/post.tsx
dessalines: It could be null / undefined so do ` ?? "RepliesAndMentions".

Also I believe this should be in the constructor, so move this to right below line 329.

`this.state.notifications = this.state.postRes.... ?? "RepliesAndMentions"

src/shared/components/post/post.tsx
```diff
@@ -647,6 +670,22 @@ export class Post extends Component<PostRouteProps, PostState> {
                 {this.state.showSidebarMobile && this.sidebar()}
               </div>
               {this.sortRadios()}
+              <div className="btn-group flex-wrap mb-2" role="group">
+                <select
+                  value={this.state.notifications}
+                  onChange={linkEvent(
+                    res.post_view.post.id,
```
dessalines: You shouldn't need to pass the post id, its already in `state`

src/shared/components/post/post.tsx
```diff
@@ -1371,6 +1413,12 @@ export class Post extends Component<PostRouteProps, PostState> {
     }
   }
 
+  async handleNotificationChange(post_id: number, event: any) {
```
dessalines: Should only need `event`

src/shared/components/post/post.tsx
```diff
@@ -1371,6 +1413,12 @@ export class Post extends Component<PostRouteProps, PostState> {
...
+    const form = { post_id, mode: event.target.value };
+    this.setState({ notifications: form.mode });
+    await HttpService.client.updatePostNotifications(form);
```
dessalines: I'd also consider making a success toast if it went through: like `Notifications updated`

dessalines:
Visually I feel like we need some kind of `bell` icon or text label on the form, or something that shows that its notifications. 

I think a `Notifications` label on that [form field](https://getbootstrap.com/docs/5.3/forms/form-control/) might be best.

src/shared/components/post/post.tsx
Nutomic: Fixed. Would be nice if there was some way to get the default value from Rust somehow, instead of hardcoding it every time.

src/shared/components/community/sidebar.tsx
Nutomic: Done, although it means I need to define the same `handleUpdateCommunityNotifs()` function both in community.tsx and post.tsx. Would appreciate if you could link that explanation post of yours.

src/shared/components/post/post.tsx
Nutomic: Fixed. Problem is that it now complains about postRes being possibly empty, so I need to check `this.state.postRes.state == "success"` each time?

src/shared/components/post/post.tsx
Nutomic: Done

Nutomic:
Added an icon as text would take too much space.

src/assets/symbols.svg
```diff
@@ -311,5 +311,8 @@
     <symbol id="icon-caret-right" fill="currentColor" viewBox="0 0 256 256">
       <path d="M181.66,133.66l-80,80a8,8,0,0,1-11.32-11.32L164.69,128,90.34,53.66a8,8,0,0,1,11.32-11.32l80,80A8,8,0,0,1,181.66,133.66Z" />
     </symbol>
+    <symbol id="icon-bell" fill="currentColor" viewBox="0 0 24 24">
```
dessalines: We already have an `icon-bell` on L149

src/shared/components/community/community.tsx
```diff
@@ -864,6 +866,11 @@ export class Community extends Component<CommunityRouteProps, State> {
     return res;
   }
 
+  async handleUpdateCommunityNotifs(form: UpdateCommunityNotifications) {
+    await HttpService.client.updateCommunityNotifications(form);
+    toast(I18NextService.i18n.t("notifications_updated"));
```
dessalines: Check the `success` like all the other examples in this file.

src/shared/components/community/sidebar.tsx
```diff
@@ -660,6 +700,16 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
...
   }
 
+  async handleNotificationChange(i: Sidebar, event: any) {
+    i.setState({ notifications: event.target.value });
+    const form = {
+      community_id: i.props.community_view.community.id,
+      mode: i.state.notifications,
+    };
+    await i.props.onUpdateCommunityNotifs(form);
+    toast(I18NextService.i18n.t("notifications_updated"));
```
dessalines: Remove this toast.

src/shared/components/community/sidebar.tsx
```diff
@@ -176,6 +202,20 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
...
+                      onChange={linkEvent(this, this.handleNotificationChange)}
+                      className="form-select"
+                    >
+                      {notificationModes.map(mode => (
+                        <option value={mode.value}>
+                          {I18NextService.i18n.t(mode.i18nKey)}
+                        </option>
+                      ))}
+                    </select>
+                  </div>
```
dessalines: Make sure to post a pic of this form also.

src/shared/components/post/post.tsx
```diff
@@ -647,6 +675,19 @@ export class Post extends Component<PostRouteProps, PostState> {
                 {this.state.showSidebarMobile && this.sidebar()}
               </div>
               {this.sortRadios()}
+              <div className="btn-group flex-wrap mb-2" role="group">
```
dessalines: Missing the bell. It should probably look just like the other one (except for the options).

Might be a good idea to externalize it into a component that can be used by both.

src/shared/components/post/post.tsx
```diff
@@ -1099,6 +1144,10 @@ export class Post extends Component<PostRouteProps, PostState> {
     return res;
   }
 
+  async handleUpdateCommunityNotifs(form: UpdateCommunityNotifications) {
+    await HttpService.client.updateCommunityNotifications(form);
```
dessalines: Check for success, and do a toast.

src/shared/components/post/post.tsx
dessalines: Check for success before doing the toast

src/shared/components/community/sidebar.tsx
dessalines: https://developer.android.com/develop/ui/compose/mental-model

This is jetpack-compose specific, but it describes the declarative approach and a lot of the recomposition problems it solves, and applies to all UI programming. 

In short: only pass data down from screens(or data layers) into UI components, and only pass events up from UI components into screens.

src/shared/components/post/post.tsx
Nutomic: Done

src/assets/symbols.svg
Nutomic: Removed

src/shared/components/community/sidebar.tsx

src/shared/components/community/community.tsx
Nutomic: Done

src/shared/components/community/sidebar.tsx
Nutomic: Its at the top.

src/shared/components/post/post.tsx
Nutomic: Done

src/shared/components/post/post.tsx
Nutomic: This is the only remaining problem.

src/shared/components/post/post.tsx
Nutomic: Added component.

Nutomic:
Resolved conflicts with notifications PR so now it looks like this (not banned and banned).

Screenshot_20250911_113440

Screenshot_20250911_113445

PR #3387, opened 2025-09-04
More details about crossposts (fixes #3386)

### LINKED ISSUE #3386: More details for crossposts
### Requirements

For crossposts it should at least show comment count and votes. Here is how mbin does it for example:

Image

[link](https://kbin.earth/m/technology@lemmy.world/t/850733/Decentralized-Social-Media-Is-the-Only-Alternative-to-the-Tech)

Also see https://github.com/LemmyNet/lemmy/issues/3827#issuecomment-3201878721

Nutomic:
## Description

Removes crosspost info from listings and only shows it on the post page itself, with more details.

## Screenshots

Screenshot_20250904_143659

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -37,7 +37,7 @@ import {
   TransferCommunity,
 } from "lemmy-js-client";
 import { relTags, torrentHelpUrl } from "@utils/config";
-import { VoteContentType } from "@utils/types";
+import { ShowDupesType, VoteContentType } from "@utils/types";
 import { mdToHtml, mdToHtmlInline } from "@utils/markdown";
 import { I18NextService } from "../../services";
 import { tippyMixin } from "../mixins/tippy-mixin";
@@ -85,6 +85,7 @@ type PostListingProps = {
   showAdultConsentModal: boolean;
   myUserInfo: MyUserInfo | undefined;
   localSite: LocalSite;
+  showDupes: ShowDupesType;
   onPostEdit(form: EditPost): Promise<RequestState<PostResponse>>;
   onPostVote(form: CreatePostLike): Promise<RequestState<PostResponse>>;
   onPostReport(form: CreatePostReport): Promise<void>;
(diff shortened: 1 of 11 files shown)


src/shared/components/post/post.tsx
```diff
@@ -1507,4 +1522,109 @@ export class Post extends Component<PostRouteProps, PostState> {
...
+                        updated={pv.post.updated_at}
+                      />
+                    </div>
+                  </div>
+                </div>
+              );
+            })}
+          </div>
+        );
+      }
```
Nutomic: This is mostly copied from post-listing.tsx. I also tried to use PostListing directly here with a new param and some adjustments, but thats too complicated.

src/shared/components/post/post.tsx
```diff
@@ -124,6 +132,7 @@ interface PostState {
   maxCommentsShown: number;
   isIsomorphic: boolean;
   lastCreatedCommentId?: CommentId;
+  myUserInfoRes: RequestState<MyUserInfo>;
```
dessalines: Not necessary now, you can get `myUserInfo` from `this.isodata.myuserinfo` for top level routes.

src/shared/components/post/post-listing.tsx
```diff
@@ -623,30 +623,6 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     }
   }
 
-  duplicatesLine() {
```
dessalines, line 626: This shouldn't be removed from post-listing, because it shows dupes for the `home` and `community`. 

You'll probably need a *small*, and an expanded version. The expanded would only get shown on the `post.tsx` page. I think the best way here, is to add a prop to `post-listing`, like

```typescript

type ShowDupesType = "Small" | "Expanded" | "None"
...

PostlistingProps {
...
showDupes: ShowDupesType`
```

src/shared/components/post/post.tsx
```diff
@@ -1507,4 +1522,109 @@ export class Post extends Component<PostRouteProps, PostState> {
...
+    if (this.state.postRes.state === "success") {
+      const dupes = this.state.postRes.data.cross_posts;
+
+      if (dupes && dupes.length > 0) {
+        const myUserInfo =
+          this.state.myUserInfoRes.state === "success"
+            ? this.state.myUserInfoRes.data
+            : undefined;
+        return (
+          <div className="d-flex flex-wrap">
```
dessalines: Instead of flex, I would use bootstrap columns, and probably `col-sm-12 col-md-6 col-lg-3`

src/shared/components/post/post.tsx
dessalines: When you move this to post-listing, you might be able to avoid some copy-pasting by making smaller functional components. If it doesn't work, then don't worry about it.

src/shared/components/post/post-listing.tsx
Nutomic: In my opinion its not necessary to show duplicates in the post listing, Reddit also doesnt do that.

src/shared/components/post/post.tsx
Nutomic: Changed although it doesnt make much difference.

src/shared/components/post/post.tsx
Nutomic: Leaving it like this as post-listing seems to be the wrong place to show duplicates.

src/shared/components/post/post-listing.tsx
dessalines: Its really useful, especially when someone makes a lot of crossposts, and you don't want to see them all duped on your front page. Don't worry about it for now tho, I'll do a PR to this one with some of my suggestions in a few hours.

dessalines (approved):

PR #3388, opened 2025-09-04
Multiple language input using checkboxes (fixes #1935)

### LINKED ISSUE #1935: change style of profile language select box to a pillbox
### Requirements

### Existing Problems
right now with the current language select box:
- you can't search for languages (have to scroll through the list)
- selecting multiple is unclear in the UI (you can't see every language you've selected until you scroll to them and see they're highlighted)
- it's unclear if the server allows selecting multiple languages or not

### My Request
change the language select box to a pillbox, similar to this (two examples below, using select2js):
image
image
live example found [here](https://select2.org/getting-started/basic-usage#multi-select-boxes-pillbox)

### Possible Improvements:
- text searching for specific languages
- make what languages are selected much clearer

### Improvements specific to select2js
- ajax for pagination (loading more records when scrolling to the bottom) and search, so large lists wouldn't have to be loaded all at once

Nutomic:
Screenshot_20250904_161653

### DIFF
--- a/src/shared/components/common/language-select.tsx
+++ b/src/shared/components/common/language-select.tsx
@@ -20,15 +20,27 @@ interface LanguageSelectProps {
   myUserInfo: MyUserInfo | undefined;
 }
 
-export class LanguageSelect extends Component<LanguageSelectProps, any> {
+interface LanguageSelectState {
+  filter: string;
+  selected: number[];
+}
+
+export class LanguageSelect extends Component<
+  LanguageSelectProps,
+  LanguageSelectState
+> {
   private id = `language-select-${randomStr()}`;
+  state: LanguageSelectState = {
+    filter: "",
+    selected: this.props.selectedLanguageIds ?? [],
+  };
 
   constructor(props: any, context: any) {
     super(props, context);
   }
 
   componentDidMount() {
-    this.setSelectedValues();
+    !this.props.multiple && this.setSelectedValues();
   }
 
   // Necessary because there is no HTML way to set selected for multiple in value=
@@ -72,21 +84,7 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
             {I18NextService.i18n.t("undetermined_language_warning")}
           </div>
         )}
-        <div
-          className={classNames(`col-sm-${this.props.multiple ? 9 : 10}`, {
-            "input-group": this.props.multiple,
-          })}
-        >
-          {this.selectBtn}
-          {this.props.multiple && (
-            <button
-              className="btn btn-outline-secondary"
-              onClick={linkEvent(this, this.handleDeselectAll)}
-            >
-              <Icon icon="x" />
-            </button>
-          )}
-        </div>
+        <div>{this.props.multiple ? this.multiSelectBtn : this.selectBtn}</div>
       </div>
     );
   }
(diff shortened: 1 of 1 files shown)


src/shared/components/common/language-select.tsx
```diff
@@ -144,8 +142,85 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
...
+      const index = i.state.selected.indexOf(id);
+      if (index > -1) {
+        i.state.selected.splice(index, 1);
+      }
+    }
+    i.props.onChange(i.state.selected);
+  }
+
+  handleSearchChange(i: LanguageSelect, event: any) {
+    i.state.filter = event.target.value;
```
dessalines: Here and above, be sure to only use `i.setState({`, not `i.state.X = `

Nutomic: Done

src/shared/components/common/language-select.tsx
```diff
@@ -144,8 +146,97 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
...
+          tabIndex={-1}
+        >
+          {!this.props.multiple && (
+            <option selected disabled hidden>
+              {I18NextService.i18n.t("language_select_placeholder")}
+            </option>
+          )}
+          {filteredLangs.map(l => (
+            <li
+              className="list-group-item m-1 p-0 focus-ring border border-0 rounded-0"
```
Nutomic: The margin is necessary to make focus-ring work (for keyboard navigation), but it results in an area which is not clickable...

Nutomic:
One disadvantage of this approach is that it doesnt easily allow to select all languages with Ctrl+A. But that is probably not used much anyway?

src/shared/components/common/language-select.tsx
dessalines: Maybe didn't get pushed up yet.

src/shared/components/common/language-select.tsx
dessalines: Really it'd be better to use a multi-select library like [choices](https://github.com/Choices-js/Choices), because these libraries have though out every weird case. But I wouldn't worry about it for now unless we find a bunch of other problems with this.

src/shared/components/common/language-select.tsx
Nutomic: Oops missed this one.

src/shared/components/common/language-select.tsx
```diff
@@ -144,8 +146,95 @@ export class LanguageSelect extends Component<LanguageSelectProps, any> {
...
+              id={l.id.toString()}
+              tabIndex={0}
+              onClick={linkEvent(this, this.handleCheckboxLanguageChange)}
+              onKeyPress={linkEvent(this, this.handleCheckboxLanguageChange)}
+              role="checkbox"
+              aria-checked={selectedLangs?.includes(l.id)}
```
Nutomic: This throws a lint error:
```
/home/felix/workspace/lemmy/lemmy-ui/src/shared/components/common/language-select.tsx
202:15  error  Non-interactive elements should not be assigned interactive roles  jsx-a11y/no-noninteractive-element-to-interactive-role
```

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-element-to-interactive-role.md

Should I add a `<button>` element inside of `<li>` to fix this?

dessalines: Yep try it.

dessalines:
> One disadvantage of this approach is that it doesnt easily allow to select all languages with Ctrl+A. But that is probably not used much anyway?

Wouldn't worry about that, because undetermined already fits that case anyway.

Nutomic:
Got the lint fixed and also found a better way to add highlighting.

Screenshot_20250910_105707

Issue #546, opened 2022-01-17
Change communities page to use bootstrap rows instead of tables


dessalines:
The communities page doesn't currently look great on mobile, and the table is to blame.

PR #3397, opened 2025-09-08
Extend admin user list with more info

Nutomic:
Converted this to a table in order to show more info for each user. Also changed it to show all users by default. It can also show more information, eg `bot_account`, `post_count`, `admin` or anything else from `person` and local_user` tables.

Screenshot_20250908_124413

Localization: https://github.com/LemmyNet/lemmy-translations/pull/190

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -497,3 +497,7 @@ br.big {
 em-emoji-picker {
   width: 100%;
 }
+
+.language-item:focus {
+  background-color: var(--bs-info-bg-subtle) !important;
+}
--- a/src/shared/components/common/language-select.tsx
+++ b/src/shared/components/common/language-select.tsx
@@ -21,15 +21,31 @@ interface LanguageSelectProps {
   myUserInfo: MyUserInfo | undefined;
 }
 
-export class LanguageSelect extends Component<LanguageSelectProps, any> {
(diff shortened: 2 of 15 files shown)


dessalines:
This is def a good idea. At least email, registered, and admin, and comment / post totals. A few things:

- Use the existing `UserBadges` component to show this extra info. IE add `registerDate` and `email` as optional `UserBadgesProps`
- Always use the `MomentTime` component for dates.
- I've been trying to get away from tables as much as possible, since they're terrible for mobile, and overflow in weird and unpredictable ways. Probably better to refactor these into either cards (and bootstrap rows), or just a simple vertical list.

Nutomic:
Updated. I think a table is fine in this case because admin tasks are presumably done from a desktop most of the time. I made the table [responsive](https://getbootstrap.com/docs/5.3/content/tables/#responsive-tables) so it shouldnt make problems. Theres also a param to [reduce table padding](https://getbootstrap.com/docs/5.3/content/tables/#small-tables) but it doesnt look good on desktop, and would require custom css to use it only on small screens. 

Screenshot_20250909_112925

dessalines:
Lots of ppl do admin tasks from mobile, but I'm already tracking this in #546 , and can do all the tables at once there.

dessalines (approved):

PR #3399, opened 2025-09-08
Use a functional component for LanguageList.

dessalines:
Used a functional component.

Really we should *always* be using these, even internally for a component. But for now we should at least make sure external ones always use components.


Nutomic (approved):

Nutomic:
Thanks, didnt know how this works.

PR #3401, opened 2025-09-09
Adding ability to visit a random community.

### LINKED ISSUE #116: Feature Request: Go to a random community
I'm not sure if there's any interest in implementing something like this, but in my opinion it'd be cool if we could go to a random community by clicking a button like we can on Reddit.

It's fun to sometimes go to a random subreddit and maybe find a new interest.

dessalines:
- Fixes #116

## Description

Adds visit a random community button to the communities page. It uses the current local vs all filter also.

## Screenshots

https://github.com/user-attachments/assets/4ee8dc55-62e3-42d1-a3aa-711000d7f20b

### DIFF
--- a/src/shared/components/community/communities.tsx
+++ b/src/shared/components/community/communities.tsx
@@ -253,14 +253,14 @@ export class Communities extends Component<
             </div>
             <div className="col">
               <button
-                className="btn btn-link"
+                className="btn btn-secondary"
                 onClick={this.handleVisitRandomCommunity}
                 aria-label={I18NextService.i18n.t("visit_random_community")}
                 data-tippy-content={I18NextService.i18n.t(
                   "visit_random_community",
                 )}
               >
-                <Icon icon="shuffle" classes="text-muted small" />
+                <Icon icon="shuffle" />
               </button>
             </div>
             <div className="col-auto">{this.searchForm()}</div>


src/shared/components/community/community-link.tsx
```diff
@@ -70,3 +60,26 @@ export class CommunityLink extends Component<CommunityLinkProps, any> {
     );
   }
 }
+
+export type CommunityLinkAndServerStr = {
+  link: string;
+  serverStr?: string;
+};
+
+export function communityLink(
```
dessalines, line 69: Extracted this to a common function.

Nutomic:
The icon is very small and hard to see. Better a normal button with text "Go to random community" or so. Also having it above the subscribed/local/all buttons makes it seem that it isnt affected by them. So it would make more sense to have it lower, eg between sort dropdown and search field.

dessalines:
I tried that and I really don't like it, as it causes row overflows now, and looks cluttered.

Screenshot_2025-09-09-10-31-36-728_com duckduckgo mobile android


edit: i spose we can go with it anyway tho. It is more clear at least

Nutomic:
Try to shorten the text, eg only "random". On mobile it could even be a button with only icon and no text.

dessalines:
How about I just use the shuffle icon as a link button (where it is now), and keep the tooltip so anyone can see what its about?

Nutomic:
Sounds good.

dessalines:
Screenshot_2025-09-11-09-27-54-564_com duckduckgo mobile android

src/shared/components/community/community-link.tsx
Nutomic: Oh I did the same thing in https://github.com/LemmyNet/lemmy-ui/pull/3409

src/shared/components/community/communities.tsx
```diff
@@ -245,6 +251,18 @@ export class Communities extends Component<
                 onChange={this.handleSortChange}
               />
             </div>
+            <div className="col">
+              <button
+                className="btn btn-link"
```
Nutomic:
```suggestion
                className="btn btn-seconray"
```

Otherwise its difficult to see. Also need to adjust the icon color then.

dessalines:
Looks like this now.

Screenshot_2025-09-11-09-53-11-991_com duckduckgo mobile android

PR #3402, opened 2025-09-09
A few updates to community-post-notifs

dessalines:
## Description

Adds a few improvements, I'll comment on them below.

## Screenshots

I have a PR to remove this overly long string for the replies and mentions:

Screenshot_2025-09-09-12-14-25-357_com duckduckgo mobile android

Now its ok:

Screenshot_2025-09-09-12-17-59-007_com duckduckgo mobile android

### DIFF
--- a/src/shared/components/common/notification-select.tsx
+++ b/src/shared/components/common/notification-select.tsx
@@ -38,7 +38,7 @@ class NotificationSelect<T extends string> extends Component<
         <select
           value={this.props.current}
           onChange={linkEvent(this, handleNotificationChange)}
-          className="form-select w-auto"
+          className="form-select w-100 text-truncate"
         >
           <option disabled aria-hidden="true">
             {I18NextService.i18n.t("notifications")}


src/shared/components/common/notification-select.tsx
```diff
@@ -1,28 +1,125 @@
...
+import {
+  CommunityNotificationsMode,
+  PostNotificationsMode,
+} from "lemmy-js-client";
+import { Component, linkEvent } from "inferno";
 
-interface NotificationSelectProps {
-  value: string;
-  modes: { value: string; i18nKey: NoOptionI18nKeys }[];
-  onChange: (event: any) => void;
```
dessalines, line 8: These should always be well typed. Try not to push up generic html events, but only the actual data types.

src/shared/components/common/notification-select.tsx
```diff
@@ -1,28 +1,125 @@
...
+  any
+> {
+  constructor(props: any, context: any) {
+    super(props, context);
+  }
+
+  render() {
+    return (
+      <>
+        {this.props.showIcon && <Icon icon="bell" classes="me-2" />}
```
dessalines, line 37: Made the icon optional, as it isn't appropriate for the sidebar.

src/shared/components/common/notification-select.tsx
```diff
@@ -1,28 +1,125 @@
...
+        <select
+          value={this.props.current}
+          onChange={linkEvent(this, handleNotificationChange)}
+          className="form-select w-auto"
+        >
+          <option disabled aria-hidden="true">
+            {I18NextService.i18n.t("notifications")}
```
dessalines, line 44: Added a title, similar to the sort select.

src/shared/components/common/notification-select.tsx
```diff
@@ -1,28 +1,125 @@
...
+}
+
+function handleNotificationChange<T extends string>(
+  i: NotificationSelect<T>,
+  event: any,
+) {
+  i.props.onChange(event.target.value);
+}
+
+const postNotifChoices: Choice<PostNotificationsMode>[] = [
```
dessalines, line 64: Moved all these i18n definitions inside here.

src/shared/components/common/notification-select.tsx
```diff
@@ -1,28 +1,125 @@
...
+  choices: Choice<T>[];
+  showIcon?: boolean;
+}
+
+type Choice<T extends string> = {
+  key: NoOptionI18nKeys;
+  value: T;
+};
+
+class NotificationSelect<T extends string> extends Component<
```
dessalines, line 26: Made a generic, private version, and two exported versions: `CommunityNotificationSelect` and `PostNotificationSelect`.

src/shared/components/community/community.tsx
```diff
@@ -867,8 +867,8 @@ export class Community extends Component<CommunityRouteProps, State> {
   }
 
   async handleUpdateCommunityNotifs(form: UpdateCommunityNotifications) {
-    const success = await HttpService.client.updateCommunityNotifications(form);
-    if (success) {
```
dessalines, line 871: This would've always been true. You needed to check the `res.state` . javascript is not a good language.

src/shared/components/post/post.tsx
```diff
@@ -327,8 +313,14 @@ export class Post extends Component<PostRouteProps, PostState> {
...
         isIsomorphic: true,
+      };
+    }
+
+    if (this.state.postRes.state === "success") {
+      this.state = {
+        ...this.state,
         notifications:
-          postRes.data.post_view.post_actions.notifications ??
+          this.state.postRes.data.post_view.post_actions?.notifications ??
```
dessalines, line 323: The remaining problem before: at the end of the constructor, and after every repost fetch, set the notifications inside a `postRes.state === "success"` block.

In the constructor you have to use `this.state = ...`, while everywhere else, you have to use `this.setState({...` Another javascript/react state fail.

src/shared/components/post/post.tsx
```diff
@@ -1418,15 +1415,17 @@ export class Post extends Component<PostRouteProps, PostState> {
...
-  async handleNotificationChange(event: any) {
-    const form = {
-      post_id: this.state.postRes.data.post_view.post.id,
-      mode: event.target.value,
-    };
-    this.setState({ notifications: form.mode });
-    const success = await HttpService.client.updatePostNotifications(form);
-    if (success) {
-      toast(I18NextService.i18n.t("notifications_updated"));
+  async handleNotificationChange(val: PostNotificationsMode) {
```
dessalines, line 1418: See now you don't just have a generic event, but the actual value being passed up.

src/shared/components/common/notification-select.tsx
Nutomic: JS types are confusing for me especially as the compiler doesnt help.

Nutomic:
For me it looks like this in community sidebar because the label for "Replies and mentions" is too long. Added a commit to fix that.

Screenshot_20250910_110749

src/shared/components/common/notification-select.tsx
```diff
@@ -1,28 +1,125 @@
...
+  }
+
+  render() {
+    return (
+      <>
+        {this.props.showIcon && <Icon icon="bell" classes="me-2" />}
+        <select
+          value={this.props.current}
+          onChange={linkEvent(this, handleNotificationChange)}
+          className="form-select w-auto"
```
Nutomic:
```suggestion
          className="form-select w-100 text-truncate"
```

src/shared/components/community/community.tsx
Nutomic: Almost like its actively trying to sabotage you...

Nutomic:
Thanks!

src/shared/components/common/notification-select.tsx
dessalines: Yep, typescript is a noble effort, but it still takes a lot of hacky work to fit js types in there.

PR #3403, opened 2025-09-09
Adding post time filtering with defaults.

### LINKED ISSUE #2187: More specific time filtering
### Requirements

When changing the sort type, instead of having such a large menu with so many 'Top' filtering options, it would be better to have only the most common sorting options and a 'More' option that opens up a window allowing you to input more specific times.

Some quick mockups of what I'm thinking of:
Screenshot 2023-10-18 234043
Screenshot 2023-10-19 000342

Alternatively, you could have a text box next to a drop down selector with 'hours' 'days' 'months' 'years' in the window above:
Screenshot 2023-10-19 000927

dessalines:
- Fixes #2187

## Description

- Adds the `default_post_time_range_seconds` day range select to your user settings. This is extracted into a component that uses days by default.
  - I tried for a long time to find a smart time interval picker library for javascript, IE one that can handle different ranges like months, years, etc, but couldn't. It would be fairly complicated to build one, but someone could always do so in the future. For now, days is fine, because the main purpose of this is to have an rss type *inbox zero*.
- Also adds it to the home and community screens.

## Screenshots

https://github.com/user-attachments/assets/22ab3ba8-4b91-4852-b225-c839292afbba


Nutomic:
A text input is too complicated especially on mobile. Also one day is too long as lots of users use the current options "top six hours" or "top twelve hours". Instead I would use a [range input](https://getbootstrap.com/docs/5.3/forms/range/) with steps for the currently available times (1 hour, 6 hours, 12 hours, 1 day etc). Then it also needs a label to show the current value.

dessalines:
A range input would not be good, as they take up way too much space. 

I agree tho it'd be better to be more fine grained... I just didn't want to have to build my own internal library where instead of static days, it would be a dropdown for hours, days, months, years, etc. That means that converting from seconds to that and using `Math.floor`would be a little more complicated.

I'll make a post on some of the askprogramming communities if anyone knows of a library. If not, I guess I'll have to do the work above.

https://lemmy.ml/post/35949063

Nutomic:
The range input could be inside the dropdown although that may be difficult to use. Another option is to keep using a dropdown (next to the existing dropdown for sort type) with the same time intervals available in 0.19.

dessalines:
The [dropdown is easy enough in bootstrap](https://getbootstrap.com/docs/5.3/forms/input-group/#buttons-with-dropdowns). The annoying part would be the seconds to interval simplifying. I spose I'll have to do it though.

dessalines:
I've added the smart time picker now.

It rounds down to the largest interval on any page load. This shows it in action:

https://github.com/user-attachments/assets/a53951a6-0f21-4542-abd6-d9dde72a3f83


Not sure whether to have the dropdown options be uppercase or not. It unfortunately needs to be wider to handle the longer strings like `seconds` and `minutes`.

Nutomic:
Its too complicated with two separate inputs like this. Just make a normal dropdown with entries:
- Hour
- Six Hours
- Twelve Hours
- Day
- Week
- Month
- Three Months
- Six Months
- Nine Months
- Year
- All Time

So the same intervals that are available now. More specific selection can be added later.

dessalines:
This is already far more powerful than that. Its better give users the option to use any time frame they like, rather than having to keep updating a set of pre-filled defaults.

Nutomic:
Its definitely more powerful. Problem is that it will be very annoying to use, especially on mobile. For most users the current dropdown times are absolutely sufficient so they shouldnt be completely removed.

dessalines:
We're gonna need some others feedback on this: 

cc @SleeplessOne1917  @matc-pub

matc-pub:
For most people it's going to be confusing to enter "All Time" as a  number. A dropdown is a lot more convenient on mobile. We could also do both, either as a dropdown that provides presets or a dropdown that has a "Custom Time" option.

- A dropdown with presets could just set the appropriate time range.
- A dropdown with "Custom Time" option could save some space by hiding the number input when not needed, it would have to default to "Custom Time" for users with a default time. The event handler could probably just set both the selection property and the time property, so that `fetch` can always use the time.

Another thing is that the unit in `TimeIntervalSelect` should be a property that is part of the url, to avoid changing units when looking at a post and then going back. I believe currently it changes 40 days to display as 1 month.

dessalines:
In that case then I'll create a second version that doesn't have a text input field, and just uses a simple select dropdown, with a few popular defaults.

matc-pub:
I think you should keep the number input for now and just add a dropdown with the presets here. The event handler can just have switch statement that calls e.g. `this.updateUrl({ postTimeRange: 60*60 });`. It's going to be simple enough to just not render the number input, or render it only in some condition, if that's what we end up doing.

dessalines:
I'm having trouble thinking about how this will both look and function, trying to keep both. But I'll see if I can mix in some `preset` dropdowns that hide the numeric input.

matc-pub:
You don't have to hide it. I'm just thinking if we remove the numeric input now, we're probably not going to add it back.

dessalines:
K I think I got it. Selecting any of the presets pre-fills the number and unit.

We can add or remove any of these presets as necessary.

Screenshot_2025-09-12-13-27-42-463_com duckduckgo mobile android-edit

dessalines:
> Another thing is that the unit in TimeIntervalSelect should be a property that is part of the url, to avoid changing units when looking at a post and then going back. I believe currently it changes 40 days to display as 1 month. 

Also w/ regard to this, that's intentional behavior. I want the time-interval-select to smartly round down to the smallest time unit for the given seconds. It should be a *friendly representation only* of the `postTimeRange` seconds in the url, not the exact seconds. I don't want to have to include units in the url, it should only have seconds, and let the front end be smart about displaying those seconds.

IE even though it says "1 month", if you look in at the seconds in URL its actually limiting to 40 days.

src/shared/components/common/time-interval-select.tsx
```diff
@@ -0,0 +1,168 @@
...
+            .map(c => c.unit)
+            .map(cUnit => (
+              <li>
+                <button
+                  className="dropdown-item"
+                  onClick={() => handleTimeIntervalUnitChange(this, cUnit)}
+                >
+                  {I18NextService.i18n.t(cUnit)}
+                </button>
+              </li>
```
SleeplessOne1917: This is one place where destructuring would be a bit easier to follow:
```suggestion
            .map(({ unit }) => (
              <li>
                <button
                  className="dropdown-item"
                  onClick={() => handleTimeIntervalUnitChange(this, unit)}
                >
                  {I18NextService.i18n.t(unit)}
                </button>
              </li>
```

matc-pub:
This is way better than what I had in mind. I was thinking of a second dropdown.

> IE even though it says "1 month", if you look in at the seconds in URL its actually limiting to 40 days.

I'm ok with that. But it can be annoying if you change your mind after visiting a post, at the same time that's probably rare enough.

Nutomic:
Will probably need some changes based on user feedback, for now its good enough.

PR #3406, opened 2025-09-10
Use params instead of string concat (fixes #1350)

### LINKED ISSUE #1350: [Bug]: Bad translation units
### Requirements

### Summary

I found the incorrect German translation `Einrichten deine Seite`. When I tried to fix it, I realized it isn't possible because it consists of two translation keys:

```ts
    `${siteSetup
        ? capitalizeFirstLetter(i18n.t("edit"))
        : capitalizeFirstLetter(i18n.t("setup"))
    } ${i18n.t("your_site")}`
```

In German, this would be written as `Richte deine Seite ein` (imperative), or `Deine Seite einrichten` (infinitive). Importantly, **words can appear in different positions depending on the language**. In some languages, words can even be split (`einrichten -> richte ... ein`).

Searching the code, I found a few more issues:

- `${i18n.t("blocked")} ${data.person_view.person.name}`
- `${i18n.t("unblocked")} ${data.person_view.person.name}`
- `${i18n.t("blocked")} ${data.community_view.community.name}`
- `${i18n.t("unblocked")} ${data.community_view.community.name}`
  - The above examples are translated as `Blockiert <Name>` and `Blockierung aufgehoben <Name>`, but should be translated as `<Name> blockiert` and `Blockierung von <Name> aufgehoben`
- `${i18n.t("purge")} ${cv.creator.name}`
- `<span> {i18n.t("ban")} {cv.creator.name} </span>`
- `{i18n.t("joined")}{" "}
                <MomentTime
                  published={pv.person.published}
                  showAgo
                  ignoreUpdated
                />`
- `<button
              type="submit"
              className="btn btn-secondary"
              aria-label={i18n.t("ban")}
            >
              {i18n.t("ban")} {pv.person.name}
            </button>`
- `({this.unreadCount} {i18n.t("new")})`
- `${i18n.t("purge")} ${post.creator.name}`
- `<span>
                    {i18n.t("ban")} {post.creator.name}
                  </span>`
- `${i18n.t("cross_posted_from")} ${post.ap_id}\n\n${body.replace(
          /^/gm,
          "> "
        )}`

I also found this date formatting code: `moment.utc(pv.person.published).local().format("MMM DD, YYYY")`. But in Germany we use `DD. MMM YYY`, e.g. `24. März 2023`.

### Solutions

There are ways to implement this properly: There's a `<T>` component to insert translated text, e.g.

```jsx
<T i18nKey="subscribe_to_communities">
    #<Link to="/communities">#</Link>
</T>
```

And it's possible to pass parameters to translations, e.g.

```js
i18n.t("number_of_points", {
  count: Number(this.postView.counts.score),
  formattedCount: Number(this.postView.counts.score),
})
```

The date can be easily fixed with the `LL` format, which prints year, month and date in a locale-aware way:

```js
>> moment().local().format("LL")
<- "June 17, 2023"
>> moment.locale("de");
>> moment().local().format("LL")
<- "17. Juni 2023"
```

### Steps to Reproduce

It can be observed by creating a new Lemmy instance locally while your preferred browser language is set to German. Here, the subtitle reads "Einrichten deine Seite".

### Version

main branch (commit fef0352)

### Lemmy Instance URL

_No response_

Nutomic:
Most of these were already fixed, but I found a few more. Especially by searching the codebase for `} ${` and `} {` (string concat in html). There may be some others missing, eg string concat done from Typescript.

There are also some cases of string concat in titles, these are probably fine so I left them unchanged.

https://github.com/LemmyNet/lemmy-translations/pull/192


src/shared/components/common/registration-application.tsx
```diff
@@ -69,8 +69,9 @@ export class RegistrationApplication extends Component<
           />
         </div>
         <div>
-          {I18NextService.i18n.t("created")}:{" "}
-          <MomentTime showAgo published={ra.published_at} />
+          {I18NextService.i18n.t("created_at", {
+            time: formatDate(ra.published_at),
+          })}
```
dessalines: This should still use the MomentTime IMO. The hover / click and hold will still show the real date.

src/shared/components/post/post-listing.tsx
```diff
@@ -1010,9 +1010,8 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     const body = post.body;
 
     return body
-      ? `${I18NextService.i18n.t("cross_posted_from")} ${
-          post.ap_id
-        }\n\n${body.replace(/^/gm, "> ")}`
+      ? `${I18NextService.i18n.t("cross_posted_from", post.ap_id)}
```
dessalines: Use the `@${hostname(community.ap_id)}`;` . 

Also I believe these need objects, ie 

`I18NextService.i18n.t("cross_posted_from", {ap_id: post.ap_id})

src/shared/components/common/moment-time.tsx
```diff
@@ -13,7 +13,7 @@ interface MomentTimeProps {
   ignoreUpdated?: boolean;
 }
 
-function formatDate(input: string) {
+export function formatDate(input: string) {
```
dessalines: Then you can keep this internal.

src/shared/components/common/registration-application.tsx
Nutomic: I dont know if there are languages which would display these in a different order. Its probably fine to leave it like before. Reverted.

src/shared/components/post/post-listing.tsx
Nutomic: `ap_id` is right, this gets inserted into the post body form when creating a new crosspost. So it should link the original post. Had the wrong translation key so it was throwing an error for the param before, fixed now.

Btw as this is based on the cross-poster's interface language but gets shown to other users, translating like this can result in the wrong language. But thats another issue.

dessalines: oooh ok gotcha.

PR #3409, opened 2025-09-10
In search results show number of posts/comments for users/communities

Nutomic:
Screenshot_20250910_155957

### DIFF
--- a/src/shared/components/search.tsx
+++ b/src/shared/components/search.tsx
@@ -58,7 +58,7 @@ import { CommentNodes } from "./comment/comment-nodes";
 import { HtmlTags } from "./common/html-tags";
 import { Spinner } from "./common/icon";
 import { ListingTypeSelect } from "./common/listing-type-select";
-import { buildPersonLink, PersonListing } from "./person/person-listing";
+import { PersonListing } from "./person/person-listing";
 import { PostListing } from "./post/post-listing";
 import { getHttpBaseInternal } from "../utils/env";
 import { RouteComponentProps } from "inferno-router/dist/Route";
@@ -69,7 +69,7 @@ import { SearchSortSelect } from "./common/sort-select";
 import { SearchableSelect } from "./common/searchable-select";
 import { UserBadges } from "./common/user-badges";
 import { Badges } from "./common/badges";
-import { CommunityLink, communityLink } from "./community/community-link";
+import { CommunityLink } from "./community/community-link";
 
 interface SearchProps {
   q?: string;
@@ -184,19 +184,15 @@ const communityListing = (
       <>
         <h3>{I18NextService.i18n.t("communities")}</h3>
         {communities.map(c => (
-          <a
-            className="d-flex align-items-center"
-            href={communityLink(c.community, false).link}
-          >
-            <span className="p-1 h-100">
-              <CommunityLink community={c.community} myUserInfo={myUserInfo} />
-            </span>
+          <div>
+            <CommunityLink community={c.community} myUserInfo={myUserInfo} />
             <Badges
+              className="ms-1 d-inline-flex"
               communityId={c.community.id}
               subject={c.community}
               lessBadges
             />
-          </a>
+          </div>
         ))}
         <hr class="border m-2" />
       </>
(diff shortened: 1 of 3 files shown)


src/shared/components/search.tsx
```diff
@@ -207,15 +217,21 @@ const personListing = (
...
                 person={p.person}
                 showApubName
                 myUserInfo={myUserInfo}
-              />,
-              p.person.comment_count,
-              "number_of_comments",
-            )}
+              />
+            </span>
+            <span>{` - ${I18NextService.i18n.t("number_of_posts", {
```
dessalines: Instead of these spans, add these as optional badges to `<UserBadges`

We could also probably show them on the `Profile.tsx` page.

src/shared/components/search.tsx
```diff
@@ -184,11 +184,21 @@ const communityListing = (
...
           <div>
-            {getListing(
-              <CommunityLink community={c.community} myUserInfo={myUserInfo} />,
-              c.community.subscribers,
-              "number_of_subscribers",
-            )}
+            <span>
+              <CommunityLink community={c.community} myUserInfo={myUserInfo} />
+            </span>
+            <span>{` - ${I18NextService.i18n.t("number_of_subscribers", {
```
dessalines: There's a `<Badges` that can take in a community already. 

We should probably make that component private, and export a `CommunityBadges` and `LocalSiteBadges` that extend it. If you need my help with that lmk.

Nutomic:
Updated with badges. The layout is still a bit odd for community with the many badges. I also added some build link helper functions to make a bigger clickable area. Was surprised such methods dont seem to exist yet.

Nutomic:
Screenshot_20250911_125701

---

Screenshot_20250911_125651

Nutomic:
Reduced the number of community badges shown in search results, and made badges consistent between person and community.

Screenshot_20250911_151231

Nutomic:
Also enabled count badges for the votes modal:

Screenshot_20250911_152012

src/shared/components/community/community-link.tsx
```diff
@@ -15,31 +15,34 @@ interface CommunityLinkProps {
   myUserInfo: MyUserInfo | undefined;
 }
 
+export function buildCommunityLink(community: Community): string {
```
dessalines: If you merge from main, there's already a `communityLink` function at the bottom of this file, I needed to use anyway.

It return a `LinkAndServerString` that you could move to `utils/types.tsx`, and have the person exported link work in the same way.

EDIT: Looking down, you don't need these anyway. Those links are already clickable.

src/shared/components/search.tsx
```diff
@@ -183,23 +184,15 @@ const communityListing = (
       <>
         <h3>{I18NextService.i18n.t("communities")}</h3>
         {communities.map(c => (
-          <div>
-            <span>
+          <a
+            className="d-flex align-items-center"
+            href={buildCommunityLink(c.community)}
+          >
```
dessalines: This isn't necessary. The `CommunityLink` already is a clickable link.

src/shared/components/search.tsx
```diff
@@ -216,23 +209,22 @@ const personListing = (
...
-            <span>{` - ${I18NextService.i18n.t("number_of_posts", {
-              count: Number(p.person.post_count),
-              formattedCount: numToSI(p.person.post_count),
-            })}`}</span>
-            <span>{` - ${I18NextService.i18n.t("number_of_comments", {
-              count: Number(p.person.comment_count),
-              formattedCount: numToSI(p.person.comment_count),
-            })}`}</span>
-          </div>
+          <a href={buildPersonLink(p.person)} className="d-flex">
```
dessalines: Same, the `<a` isn't necessary, person listing is already a link.

dessalines:
> I also added some build link helper functions to make a bigger clickable area. Was surprised such methods dont seem to exist yet.

Don't do that, because those badges might be clickable also, or have tooltips. The community name / person name being clickable is enough. 

Everything else looks good.

Nutomic:
The username is very small to click, especially for mobile. Only clickable badge is mod log which is not shown here. Tooltips are still working. For usability its better to have a larger clickable area.

dessalines:
There's no need to make a special case just for search, where everywhere else, clicking on the badges should not be a link. 

If you really want to add this, then it should be inside those components anyway, with an optional bool. I highly recommend against it tho, the username links are already clickable everywhere in the UI and ppl don't seem to have a problem with it. Also who knows what functionality we might want to add to those badges anyway.

Nutomic:
Okay removed it.

PR #3412, opened 2025-09-11
Updating to new 1.0 names.

dessalines:
Just a few updates from `lemmy-js-client`, no logical changes.

### DIFF
--- a/src/shared/components/private_message/private-message.tsx
+++ b/src/shared/components/private_message/private-message.tsx
@@ -4,9 +4,9 @@ import {
   CreatePrivateMessageReport,
   DeletePrivateMessage,
   EditPrivateMessage,
-  MarkPrivateMessageAsRead,
   MyUserInfo,
   Person,
+  PrivateMessageId,
   PrivateMessageView,
 } from "lemmy-js-client";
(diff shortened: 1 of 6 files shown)


src/shared/components/private_message/private-message.tsx
```diff
@@ -339,10 +339,10 @@ export class PrivateMessage extends Component<
 
   handleMarkRead(i: PrivateMessage) {
     i.setState({ readLoading: true });
-    i.props.onMarkRead({
-      private_message_id: i.props.private_message_view.private_message.id,
-      read: !i.props.read,
-    });
+    i.props.onMarkRead(
+      i.props.private_message_view.private_message.id,
```
Nutomic, line 343: The notification_id gets passed in as `key` so it would be easier to use that directly. Then you can directly return `MarkNotificationAsRead` from here, and dont need to `find` in notifications.tsx

dessalines: These `PrivateMessage` and `CommentNode` components don't have notifications as a prop, which is good because they're used in a lot of places that has nothing to do with notifs. So they should only push up their ids, and let the notifications screen handle finding them.

PR #3416, opened 2025-09-12
Collapse removed comments that have no children by default.

### LINKED ISSUE #3398: Collapse removed comments without children by default
### Requirements

Recently we've been spammers resorting to using scripts to automatically post thousands of comments in short periods of time. Even after removing those comments, they will still stick around and consume a large amount of screen space.
It might be useful to not remove them entirely for purposes of transparency, but having them collapsed by default if they don't have any children would significantly improve user experience.

It won't entirely address the issue, as without purging them, these comments will continue to be listed, but they will at least consume a lot less space on the screen:

Image

dessalines:
- Fixes #3398

https://github.com/user-attachments/assets/0087fa93-d6f9-4125-9494-58b81a310580

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -139,8 +139,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     showEdit: false,
     // Collapse comments that have no children and are removed by default
     collapsed:
-      this.commentView.comment.child_count === 0 &&
-      this.commentView.comment.removed,
+      !this.commentView.comment.child_count && this.commentView.comment.removed,
     viewSource: false,
     showAdvanced: false,
     createOrEditCommentLoading: false,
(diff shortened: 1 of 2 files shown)


src/shared/components/comment/comment-node.tsx
```diff
@@ -137,7 +137,10 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
   state: CommentNodeState = {
     showReply: false,
     showEdit: false,
-    collapsed: false,
+    // Collapse comments that have no children and are removed by default
+    collapsed:
+      this.commentView.comment.child_count === 0 &&
```
SleeplessOne1917: No need to change: since 0 is falsy, the left half of the and can be `!this.commentView.comment.child_count`.

SleeplessOne1917 (approved):

dessalines:
Cool. Gonna merge so the renovate PRs can pass, since main is broken rn because of `creator.banned`.

PR #3440, opened 2025-09-13
Remove all tables to better handle mobile devices.

### LINKED ISSUE #546: Change communities page to use bootstrap rows instead of tables

dessalines:
## Description

Apologize for the long PR, but as part of this I needed to do a lot of cleanup.

- Removed all the tables, in favor of bootstraps `list-inline` (which uses flex-wrap), or bootstrap columns. Many of these like the modlog, were a lot of work.
- Needed to re-do the entire `taglines-form.tsx` and `emojis-form.tsx`, so that they properly pushed up events, instead of doing internal http calls. I left `instance-blocks.tsx`, which is also built using internal http calls, to be done another time. #3453 
- There were several tables that were used instead of bootstrap forms... so I converted them to inline forms.
- Re-did the main color schemes (litely / darkly) to be more muted (btn-secondary is supposed to be a muted gray, not a color). Like jerboa, we should avoid colors outside of our main elements (user-listings and community-listings) and use badges, cards, or text adjustments instead. I had to regenerate themes after updating the necessary `_variables.scss` files.
- Removed some filled icons in favor of outlined ones.
- Fixed some logical issues around modding found when testing.
- I also updated to the new lemmy-js-client version.
- Fixes #546 

## Screenshots

I'll try to include both mobile, and tablet views for the relevant ones.

### Front page

Screenshot_2025-09-18-14-59-35-658_mark via_1

### Community

Screenshot_2025-09-18-15-01-29-287_mark via

### Modlog

Screenshot_2025-09-18-15-14-54-604_mark via

### Users

Screenshot_2025-09-18-14-59-53-214_mark via

### Taglines

Screenshot_2025-09-18-15-00-19-902_mark via

### Emojis

Screenshot_2025-09-18-15-36-00-913_mark via


Nutomic:
On mobile it also doesnt look good. Better to keep the tables and use a strict max width for columns with longer values ellipsized and reduced margins. Then everything can easily fit on the screen and remain nicely organized into columns.

dessalines:
I'm using proper `list-inlines` now which uses `flex-wrap` and just moves overflowing content to new rows. 

I'm done trying to make tables responsive, and seeing apps that keep using them. The data should just flow properly into a new row instead of forcing mobile users to view less content, see squished content, or have to side scroll. None of that is necessary with [flex-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap).

I'm doing other things to pretty-up lemmy-ui as part of this, and am def open to other suggestions about how to display these stats, but not by using tables anymore. I'll post pics of everything I changed when its ready.

Nutomic:
Im against merging this, its much harder to read both on mobile and desktop. Tables work well because the same information is aligned vertically, but here the alignment is completely different if there is a longer username at the beginning. Its especially noticable in the modlog and user list. 

Only in a few cases like media uploads this change makes sense because a table isnt really needed.

dessalines:
Vertical alignment of columns isn't important, and its especially useless on mobile, since you can only have a max of like 2 columns anyway. I've done a lot of work here to make sure all data is shown, and no one has to side-scroll anymore.

I don't think there are many people who prefer side scolling, to just having text properly reflow into new lines. Take a look at the modlog for instance:

Current:



New: 



---

I'm def willing to do things to make this cleaner, like add badges or text differentiators to make things stand out. If you're dead set on keeping tables in lemmy-ui, then we're gonna need some feedback from @matc-pub and @SleeplessOne1917 .

Nutomic:
Would it be feasible to keep the tables on large screens, and use this inline display only on mobile?

SleeplessOne1917:
I think the move away from tables makes sense for most of the places you did the change. I do think a different design could help with this change though.

Looking at the screenshots you posted, especially the ones for users, modlog, and communities, still looks like tables despite not being tables. I think the Transformed heading in [this article](https://medium.com/appnroll-publication/5-practical-solutions-to-make-responsive-data-tables-ff031c48b122) (sadly the heading doesn't have an anchor, so you'll need to scroll down) offers a good approach for situations like this that would even go well with Nutomic's suggestion to keep the tables on larger screen sizes. Note that even though the example in the article uses pretty plain looking individual tables for what would be a row in a desktop sized table, it could be anything where each "row" (using scare quotes now since they wouldn't be rows anymore) is self contained and visually separated from each other "row". Not sure how easy this would be to implement using Bootstrap though.

Things like custom emojis, taglines, and media uploads shouldn't have even been tables in the first place imo. I could see something card-like working well for these. Emojis (once already uploaded) could display a list of cards with the image and the name/shortcode and allow users to change them with a form that appears when they click an "edit" button. The taglines could be similar (though maybe not cards): perhaps a list of the rendered taglines separated by horizontal rules that have buttons for editing and deleting, where pressing the former opens up the text area to edit it with?

dessalines:
> Would it be feasible to keep the tables on large screens, and use this inline display only on mobile?

Its possible, but IMO not preferable to have to maintain two separate versions.

> Looking at the screenshots you posted, especially the ones for users, modlog, and communities, still looks like tables despite not being tables.

I'll go back to the drawing board for some of these then, and create *card* implementations for user and community, and use bootstraps standard grid system. Then I can use the standard bootstrap responsive classes like `col-12 col-md-6 col-lg-4` etc.

For taglines and emojis, since they're forms, I've already used `form-inline`. To make them into cards, we'd just change some classes around, but they're 99% done, and would only need some minor tweaks: https://getbootstrap.com/docs/5.3/forms/overview/#overview

dessalines:
Closed in favor of #3479

PR #3442, opened 2025-09-15
[0.19] Development instructions using remote instance

Nutomic:
Backport of https://github.com/LemmyNet/lemmy-ui/pull/3371. Confirmed that it works to connect with lemmy.ml using `LEMMY_UI_BACKEND_REMOTE=lemmy.ml pnpm dev` and also works with a local backend using `pnpm dev`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ You need to have [pnpm](https://pnpm.io/installation) installed. Then run the fo
 git clone https://github.com/LemmyNet/lemmy-ui.git
 cd lemmy-ui
 pnpm install
-LEMMY_UI_BACKEND_REMOTE=enterprise.lemmy.ml pnpm dev
+LEMMY_UI_DEBUG=true LEMMY_UI_BACKEND_REMOTE=enterprise.lemmy.ml pnpm dev
 ```
 
 Finally open `http://0.0.0.0:1234` in your browser. This uses the public test instance `https://enterprise.lemmy.ml/` as backend. You can also connect to any production instance running version 0.19, for example `LEMMY_UI_BACKEND_REMOTE=lemmy.ml`. For more details such as developing with a locally compiled Lemmy backend, read the [documentation](https://join-lemmy.org/docs/contributors/01-overview.html).


README.md
```diff
@@ -4,17 +4,31 @@ The official web app for [Lemmy](https://github.com/LemmyNet/lemmy), written in
...
 
+## Development
+
+You need to have [pnpm](https://pnpm.io/installation) installed. Then run the following:
+
+```bash
+git clone https://github.com/LemmyNet/lemmy-ui.git
+cd lemmy-ui
+pnpm install
+LEMMY_UI_BACKEND_REMOTE=lemmy.ml pnpm dev
```
dessalines:
```suggestion
LEMMY_UI_DEBUG=true LEMMY_UI_BACKEND_REMOTE=enterprise.lemmy.ml pnpm dev
```

dessalines: Better to use enterprise.lemmy.ml, not to use lemmy.ml for development.

dessalines: I'm also going to update the `scripts/test.sh` on `main`

Nutomic: I want to show that it works with any production instance. This can be useful if you notice an issue while browsing, and want to reproduce/fix it locally without manually setting up the right test data first. When you open the site it should be clear that you are dealing with production data and any changes will be public.

dessalines: As long as a comment is there, then that's good enough. We have 3 test instances, lets have ppl use them.

Nutomic: Okay updated.

README.md
```diff
@@ -4,17 +4,31 @@ The official web app for [Lemmy](https://github.com/LemmyNet/lemmy), written in
...
 
+## Development
+
+You need to have [pnpm](https://pnpm.io/installation) installed. Then run the following:
+
+```bash
+git clone https://github.com/LemmyNet/lemmy-ui.git
+cd lemmy-ui
+pnpm install
+LEMMY_UI_BACKEND_REMOTE=enterprise.lemmy.ml pnpm dev
```
dessalines:
```suggestion
LEMMY_UI_DEBUG=true LEMMY_UI_BACKEND_REMOTE=enterprise.lemmy.ml pnpm dev
```

Nutomic: Looks like this is to enable `eruda` which seems identical to the normal browser dev tools. I dont see whats the use of it.

SleeplessOne1917: I think eruda is supposed to help with testing on mobile in addition to laptop/desktop. I've never used it myself though (I just resize windows to test that the UI is responsive). Maybe @dessalines or @matc-pub use it?

dessalines: I def use it a lot. I develop on android and eruda is a full debugging suite that lets you inspect items, view console, cookies, network, etc.

dessalines: There's also a few other things that `LEMMY_UI_DEBUG` adds so it should definitely be used here.

Nutomic: Other places where this var is used also check for `process.env["NODE_ENV"] === "development"`. So wouldnt it make more sense to get rid of LEMMY_UI_DEBUG and only check NODE_ENV?

dessalines: Sure I wouldn't be opposed, but IMO in a different PR, and I wouldn't worry about `release/v0.19`

Nutomic: Okay will do, then lets merge this one as is.

Issue #31, opened 2020-09-22
Add i18next support to modlog page.


dessalines:
Requires some complicated interpolations for all the modlog actions.

alectrocute:
This appears to still be something we need.

PR #3443, opened 2025-09-15
Add new modlog filters

Nutomic:
Defined here: https://github.com/LemmyNet/lemmy/blob/main/crates/db_schema/src/lib.rs#L96

### DIFF
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -924,7 +924,7 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
               </option>
               <option value={"AdminAdd"}>Adding Admin to Site</option>
               <option value={"AdminBlockInstance"}>
-                Block a federated Instance
+                Blocking a federated instance
               </option>
               <option value={"AdminAllowInstance"}>
                 Allow a federated instance
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -931,12 +931,12 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
               </option>
               {this.isoData.myUserInfo?.local_user_view.local_user.admin && (
                 <>
-                  <option value={"AdminPurgePerson"}>Purge a Person</option>
+                  <option value={"AdminPurgePerson"}>Purging a Person</option>
                   <option value={"AdminPurgeCommunity"}>
-                    Purge a Community
+                    Purging a Community
                   </option>
-                  <option value={"AdminPurgePost"}>Purge a Post</option>
-                  <option value={"AdminPurgeComment"}>Purge a Comment</option>
+                  <option value={"AdminPurgePost"}>Purging a Post</option>
+                  <option value={"AdminPurgeComment"}>Purging a Comment</option>
                 </>
               )}
             </select>
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -924,21 +924,15 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
               </option>
               <option value={"AdminAdd"}>Adding Admin to Site</option>
               <option value={"AdminBlockInstance"}>
-                Blocking a federated instance
+                Blocking a federated Instance
               </option>
               <option value={"AdminAllowInstance"}>
-                Allowing a federated instance
+                Allowing a federated Instance
               </option>
-              {this.isoData.myUserInfo?.local_user_view.local_user.admin && (
-                <>
-                  <option value={"AdminPurgePerson"}>Purging a Person</option>
-                  <option value={"AdminPurgeCommunity"}>
-                    Purging a Community
-                  </option>
-                  <option value={"AdminPurgePost"}>Purging a Post</option>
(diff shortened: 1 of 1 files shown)


src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+              {this.isoData.myUserInfo?.local_user_view.local_user.admin && (
+                <>
+                  <option value={"AdminPurgePerson"}>Purge a Person</option>
+                  <option value={"AdminPurgeCommunity"}>
+                    Purge a Community
+                  </option>
+                  <option value={"AdminPurgePost"}>Purge a Post</option>
+                  <option value={"AdminPurgeComment"}>Purge a Comment</option>
+                </>
+              )}
```
Nutomic: Should purges be visible to everyone? I think it makes sense to show them only to admins. Also needs a check in the backend api handler then.

dessalines: Yep its fine to show them to everyone. It can't show the content anyway of course, it just has a line like `admin purged a community`.

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
                 Transferring Communities
               </option>
-              <option value={"ModAdd"}>Adding Mod to Site</option>
-              <option value={"ModBan"}>Banning From Site</option>
+              <option value={"ModChangeCommunityVisibility"}>
+                Changing Community visibility
+              </option>
+              <option value={"AdminAdd"}>Adding Admin to Site</option>
+              <option value={"AdminBlockInstance"}>
+                Block a federated Instance
```
dessalines:
```suggestion
                Blocking a federated instance
```

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+              </option>
+              <option value={"AdminAllowInstance"}>
+                Allow a federated instance
```
dessalines:
```suggestion
                Allowing a federated instance
```

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+              </option>
+              {this.isoData.myUserInfo?.local_user_view.local_user.admin && (
+                <>
+                  <option value={"AdminPurgePerson"}>Purge a Person</option>
```
dessalines:
```suggestion
                  <option value={"AdminPurgePerson"}>Purging a Person</option>
```

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+                  <option value={"AdminPurgeCommunity"}>
+                    Purge a Community
```
dessalines:
```suggestion
                    Purging a Community
```

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+                  </option>
+                  <option value={"AdminPurgePost"}>Purge a Post</option>
```
dessalines:
```suggestion
                  <option value={"AdminPurgePost"}>Purging a Post</option>
```

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+                  <option value={"AdminPurgeComment"}>Purge a Comment</option>
```
dessalines:
```suggestion
                  <option value={"AdminPurgeComment"}>Purging a Comment</option>
```

src/shared/components/modlog.tsx
```diff
@@ -906,16 +906,39 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
...
+              <option value={"ModChangeCommunityVisibility"}>
+                Changing Community visibility
+              </option>
+              <option value={"AdminAdd"}>Adding Admin to Site</option>
+              <option value={"AdminBlockInstance"}>
+                Blocking a federated instance
+              </option>
+              <option value={"AdminAllowInstance"}>
+                Allowing a federated instance
+              </option>
```
SleeplessOne1917: Capitalization is inconsistent between these. Compare "Changing Community visibility" to "Purging a Community" for example. I think capitalizing all words except for articles ("a", "the", "an", etc.) and prepositions ("on", "to", etc.) makes sense and is consistent with what we already have.

Also, not necessarily needed for this PR, but we should make translations for these in i18n.

dessalines: Def true, one of our oldest issues: #31

dessalines (approved):

SleeplessOne1917 (approved):

Nutomic:
Changed "instance" to uppercase and also made purge actions visible for everyone.

PR #3445, opened 2025-09-15
Extend readme

Nutomic:
Added some parts of the backend repo, particularly badges and general project description. We could also copy-paste the remaining sections.


dessalines (approved):
This is good enough IMO, no need to copy everything from lemmy.

PR #3446, opened 2025-09-16
UI changes if current user is banned (fixes #989)

### LINKED ISSUE #989: A site ban leaves the user uninformed, seeing '404: site_ban' on all urls of the instance

When a user receives a site-ban, the user sees a white page with only the text '404: site_ban' for all urls of the instance.
I see two issues;
- If the user knows of the modlog page, the user also gets the site_ban message for that url (because of the cookie i guess)
- If the user doesn't know about the modlog page, the user has no way to find out why (s)he received a (temporary) site ban, because there's no info on the site_ban page.

The '404: site_ban' page should refer to the https://lemmy.ml/modlog page for finding out what happened.
That page shouldn't be restricted by the site_ban.

Clearing cookie's or using another browser would provide access to the modlog page.
But still the user probably wouldn't know of the modlog page

### Additional context

.

Nutomic:
Add dialog and disable various actions if the current user is banned. Also hide most user settings.

Screenshot_20250916_130824

Translation: https://github.com/LemmyNet/lemmy-translations/pull/197


src/shared/components/comment/comment-form.tsx
```diff
@@ -64,7 +64,10 @@ export class CommentForm extends Component<CommentFormProps, any> {
             buttonTitle={this.buttonTitle}
             replyType={typeof this.props.node !== "number"}
             focus={this.props.focus}
-            disabled={this.props.disabled}
+            disabled={
+              this.props.disabled ||
+              this.props.myUserInfo.local_user_view.banned
```
dessalines: It would be better to move this up. I see only comment-node.tsx and post.tsx have this anyway.

Also use the function you created.

src/shared/utils/app.ts
```diff
@@ -769,3 +769,7 @@ export function commentViewToPersonContentCombinedView(
     ...cv,
   };
 }
+
+export function disableInput(user: MyUserInfo | undefined): boolean {
```
dessalines: How about rename this to `userNotLoggedInOrBanned` . Then its clearer up the chain *why* you're disabling things.

Also might be clearer to do `userLoggedInAndNotBanned`, IE an enable, and 

`return user?.local_user_view?.banned ?? false`

src/shared/components/person/settings.tsx
```diff
@@ -223,6 +224,7 @@ export class Settings extends Component<SettingsRouteProps, SettingsState> {
     show2faModal: false,
     importSettingsRes: EMPTY_REQUEST,
     exportSettingsRes: EMPTY_REQUEST,
+    userBanned: this.isoData.myUserInfo?.local_user_view.banned ?? false,
```
dessalines: You created a function for this already, doesn't need to be in the state.

src/shared/components/home/home.tsx
```diff
@@ -420,6 +421,13 @@ export class Home extends Component<HomeRouteProps, HomeState> {
           <div className="row">
             <div className="col-12 col-md-8 col-lg-9">
               <DonationDialog myUserInfo={this.isoData.myUserInfo} />
+              {this.isoData.myUserInfo?.local_user_view.banned && (
```
dessalines, line 427: Use the function you created.

src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -252,7 +252,8 @@ export default class ContentActionDropdown extends Component<
                     />
                   </li>
                 )}
-                {this.amCreator ? (
+                {this.amCreator &&
+                !this.props.myUserInfo?.local_user_view.banned ? (
```
dessalines: Use the function.

src/shared/components/home/banned-dialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export function BannedDialog({ expires }: BannedDialogProps) {
+  const title = expires
+    ? I18NextService.i18n.t("banned_dialog_title_temporary", {
+        expires: formatRelativeDate(expires),
+      })
+    : I18NextService.i18n.t("banned_dialog_title_permanent");
+  return (
+    <div class="alert alert-danger" role="alert">
```
dessalines: replace all these `class` with `className` . Its a tsx component thing to prefer it.

src/shared/components/home/banned-dialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <div class="d-flex">
+        <h4 class="alert-heading flex-grow-1">{title}</h4>
+      </div>
```
dessalines: Where is this from? You don't need d-flex or grow, it should already wrap the title fine.

Use card-title and card-text, with `text-bg-danger` on the card.

https://getbootstrap.com/docs/5.3/components/card/

src/shared/components/home/banned-dialog.tsx
SleeplessOne1917: I think it's not just preferred, but required. `class` is a reserved word in JS/TS. I know that React requires `className` because of this and that Inferno advertises itself of React-like. However, the actual Inferno docs say nothing on the matter.

src/shared/utils/app.ts
SleeplessOne1917: I find this a lot clearer.

src/shared/components/home/banned-dialog.tsx
Nutomic: There are a lot of places in the code using `class` (try `rg " class="`). Anyway changed it.

src/shared/components/home/banned-dialog.tsx
Nutomic: Youre right, I copied it from the donation dialog.

Nutomic:
Made all these changes.

src/shared/utils/app.ts
```diff
@@ -769,3 +769,7 @@ export function commentViewToPersonContentCombinedView(
     ...cv,
   };
 }
+
+export function userNotLoggedInOrBanned(user: MyUserInfo | undefined): boolean {
+  return user?.local_user_view?.banned ?? false;
+}
```
dessalines: This might be incorrect? 

Some of the disabled=`userNotLoggedInOrBanned` will be false if there is no user provided.

The not not language is confusing me. 

It maybe should be `userLoggedInAndNotBanned = `!user?.local_user_view?.banned ?? false`

Make sure you test this.

src/shared/components/home/home.tsx
```diff
@@ -420,6 +422,13 @@ export class Home extends Component<HomeRouteProps, HomeState> {
           <div className="row">
             <div className="col-12 col-md-8 col-lg-9">
               <DonationDialog myUserInfo={this.isoData.myUserInfo} />
+              {userNotLoggedInOrBanned(this.isoData.myUserInfo) && (
```
dessalines: I guess for this you maybe can't use your function after all. It should only show when banned, not when you're just not logged in.

src/shared/components/home/banned-dialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+    ? I18NextService.i18n.t("banned_dialog_title_temporary", {
+        expires: formatRelativeDate(expires),
+      })
+    : I18NextService.i18n.t("banned_dialog_title_permanent");
+  return (
+    <div className="alert alert-danger text-bg-danger" role="alert">
+      <h4 className="alert-heading">{title}</h4>
+      <div className="card-text">
+        {I18NextService.i18n.t("banned_dialog_body")}
+      </div>
```
dessalines: Okay so you went with using an alert, not a card. Don't use card-text then: https://getbootstrap.com/docs/5.3/components/alerts/#additional-content

Always just copy-paste from the bootstrap examples when you're not sure.

src/shared/utils/app.ts
Nutomic: Tested it again (logged in, banned and not logged in) and everything works as expected. Besides I changed this function just as you suggested.

Nutomic: Turns out the function you suggested was wrong, I should have tested it properly. Now it works.

dessalines:
Looks good, feel free to merge whenever.

PR #3447, opened 2025-09-16
Add report actions (fixes #501)

### LINKED ISSUE #501: Cannot access mod tools on /reports page
When someone reports a post/comment, shows up on the /reposts page,

but when clicking on the _three dots_ (more) to access the mod tools they don't show up, 
i get an error instead:

`Uncaught TypeError: this.props.moderators[0] is undefined`

Nutomic:
- [x] Remove item
- [x] Ban creator from site
- [x] Ban creator from community
- [ ] Anything else?

### DIFF
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -192,7 +192,7 @@ export class CommentReport extends Component<
               ? "unban_from_community"
               : "ban_from_community",
           )}
-          inline
+          inlineWithText
           icon={comment_view.creator_banned ? "unban" : "ban"}
           noLoading
           onClick={this.handleModBanFromCommunity}
@@ -201,9 +201,9 @@ export class CommentReport extends Component<
         {this.props.myUserInfo?.local_user_view.local_user.admin && (
           <ActionButton
             label={I18NextService.i18n.t(
-              comment_view.creator_banned ? "unban" : "ban",
+              comment_view.creator_banned ? "unban_from_site" : "ban_from_site",
             )}
-            inline
+            inlineWithText
             icon={comment_view.creator_banned ? "unban" : "ban"}
             noLoading
             onClick={this.handleAdminBan}
--- a/src/shared/components/common/content-actions/action-button.tsx
+++ b/src/shared/components/common/content-actions/action-button.tsx
@@ -8,6 +8,7 @@ interface ActionButtonPropsBase {
   icon: string;
   iconClass?: string;
   inline?: boolean;
+  inlineWithText?: boolean;
   noLoading?: boolean;
 }
 
@@ -49,14 +50,14 @@ export default class ActionButton extends Component<
   }
 
   render() {
-    const { label, icon, iconClass, inline } = this.props;
+    const { label, icon, iconClass, inline, inlineWithText } = this.props;
 
     return (
       <button
         className={classNames(
           "btn btn-link btn-sm",
-          inline
-            ? "btn-animate text-muted py-0"
+          inline || inlineWithText
+            ? "btn-animate text-body py-0"
             : "d-flex align-items-center rounded-0 dropdown-item",
         )}
         onClick={linkEvent(this, handleClick)}
@@ -69,7 +70,7 @@ export default class ActionButton extends Component<
         ) : (
           <Icon classes={classNames("me-2", iconClass)} icon={icon} inline />
         )}
-        {!inline && label}
+        {(!inline || inlineWithText) && label}
       </button>
     );
   }
--- a/src/shared/components/post/post-report.tsx
+++ b/src/shared/components/post/post-report.tsx
@@ -179,7 +179,6 @@ export class PostReport extends Component<PostReportProps, PostReportState> {
           label={I18NextService.i18n.t(
             pv.post.removed ? "restore_post" : "remove_post",
           )}
-          inline
           icon={pv.post.removed ? "restore" : "x"}
           noLoading
           onClick={() => this.setState({ showRemovePostDialog: true })}
@@ -189,8 +188,7 @@ export class PostReport extends Component<PostReportProps, PostReportState> {
           label={I18NextService.i18n.t(
             pv.creator_banned ? "unban_from_community" : "ban_from_community",
           )}
-          inline
-          icon={pv.creator_banned ? "unban" : "ban"}
+          icon={pv.creator_banned ? "unban_from_site" : "ban_from_site"}
           noLoading
           onClick={this.handleModBanFromCommunity}
           iconClass={`text-${pv.creator_banned ? "success" : "danger"}`}


src/shared/components/person/reports.tsx
```diff
@@ -135,8 +144,38 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
+              await HttpService.client.removeComment(removeCommentForm);
+              this.setState({ removeCommentForm: null });
+              this.update();
+            }}
+            modActionType="remove-comment"
+            isRemoved={!removeCommentForm.removed}
+            onCancel={() => this.setState({ removeCommentForm: null })}
+            show={true}
+          />
+        )}
```
Nutomic, line 198: Is there any way to handle this generically, without a separate block for each type? Because it quickly gets very verbose. I tried to define a state `removeForm: RemovePost | RemoveComment` but there seems to be no good way to know the actual type and call the correct api endpoint.

src/shared/components/comment/comment-report.tsx
```diff
@@ -172,4 +186,12 @@ export class CommentReport extends Component<
       resolved: !i.props.report.comment_report.resolved,
     });
   }
+
+  handleRemoveComment() {
+    this.setState({ loading: true });
+    this.props.onRemoveComment({
+      comment_id: this.props.report.comment_report.comment_id,
+      removed: !this.props.report.comment.removed,
+    });
```
dessalines: This pry isn't going to work. It needs to load the modal to be able to set a reason, which is required now.

src/shared/components/person/reports.tsx
```diff
@@ -65,6 +68,8 @@ interface ReportsState {
   siteRes: GetSiteResponse;
   cursor?: DirectionalCursor;
   isIsomorphic: boolean;
+  removePostForm: RemovePost | null;
+  removeCommentForm: RemoveComment | null;
```
dessalines: Don't use null, use `removePostForm?: RemovePost`

src/shared/components/person/reports.tsx
```diff
@@ -84,6 +89,8 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
     messageType: "All",
     siteRes: this.isoData.siteRes,
     isIsomorphic: false,
+    removePostForm: null,
+    removeCommentForm: null,
```
dessalines: Can remove these lines.

src/shared/components/person/reports.tsx
```diff
@@ -135,8 +144,38 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
+    const removeCommentForm = this.state.removeCommentForm;
     return (
       <div className="person-reports container-lg">
+        {removePostForm && (
+          <ModActionFormModal
+            onSubmit={async (reason: string) => {
+              removePostForm.reason = reason;
+              await HttpService.client.removePost(removePostForm);
+              this.setState({ removePostForm: null });
+              this.update();
```
dessalines: Use handler functions, there are a ton of them in this file.

And you need to use `setState` when setting things, you can't just do `removePostForm.reason=  reason`

src/shared/components/person/reports.tsx
```diff
@@ -135,8 +144,38 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
+            }}
+            modActionType="remove-post"
+            isRemoved={!removePostForm.removed}
+            onCancel={() => this.setState({ removePostForm: null })}
```
dessalines: Same

src/shared/components/person/reports.tsx
```diff
@@ -135,8 +144,38 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
+            show={true}
+          />
+        )}
+        {removeCommentForm && (
+          <ModActionFormModal
+            onSubmit={async (reason: string) => {
```
dessalines: Same

src/shared/components/person/reports.tsx
dessalines: No, they're different forms anyway. Better to just be verbose and define specific handler functions for each.

src/shared/components/person/reports.tsx
```diff
@@ -341,20 +380,24 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
-            key={i.type_ + i.post_report.id}
-            report={i}
-            enableNsfw={enableNsfw(siteRes)}
-            showAdultConsentModal={this.isoData.showAdultConsentModal}
-            myUserInfo={this.isoData.myUserInfo}
-            localSite={siteRes.site_view.local_site}
-            admins={this.isoData.siteRes.admins}
-            onResolveReport={this.handleResolvePostReport}
-          />
+          <>
```
dessalines: Can remove this one.

dessalines:
The only other mod related action I could see that you might want to take on reports, is purging, and locking. Up to you if you want to add those or not, but once you get one working with those mod action modals, the rest should be easy.

`interface ContentActionDropdownPropsBase` has them all.

src/shared/components/comment/comment-report.tsx
Nutomic: Yes that change will be tricky. It also requires UI changes to include a reason for instance blocks. Not gonna take it as part of this pr.

src/shared/components/person/reports.tsx
Nutomic: This is not supposed to change the state, just copy the form into a shorter variable name.

Nutomic:
Purging is used very rarely, and would be a big problem if its clicked accidentally. So better to leave it out. Or it could be hidden behind a context menu later.

For this PR I will keep it limited to remove and ban actions. Also added button "Show community reports" as these are hidden from admins by default now:
Screenshot_20250918_160650

https://github.com/LemmyNet/lemmy-translations/pull/202

I noticed that `/api/v4/report/list` returns the wrong value for `creator_banned`, it is always false. `creator_banned_from_community` works correctly. The same value is correct in case of post listing which uses the same select expression. Maybe it uses the wrong user alias and actually checks if the report creator is banned.

dessalines:
> For this PR I will keep it limited to remove and ban actions. Also added button "Show community reports" as these are hidden from admins by default now:

Not sure I agree with that. We've had discussions about *report sovereignty* on other issues, but IMO people don't care who sees their report; they just want action taken on it ASAP. That and community mods are inactive on all but the largest servers, so admins are more likely to be there and able to take action faster.

> I noticed that /api/v4/report/list returns the wrong value for creator_banned, it is always false. creator_banned_from_community works correctly. The same value is correct in case of post listing which uses the same select expression. Maybe it uses the wrong user alias and actually checks if the report creator is banned.

Haven't looked over the PR yet, but ya that should be fixed in the back end, with a test verifying. Open up an issue there if there isn't one already. A lot of times these front end PRs are the ones that actually expose issues in the back end.

src/shared/components/comment/comment-report.tsx
dessalines: Its kinda the whole point of the PR... if we're adding mod actions to the report screen, it needs to properly load the mod action modals so that a reason can be added (especially now that they're required. Maybe @SleeplessOne1917 can help out here.

src/shared/components/person/reports.tsx
```diff
@@ -135,8 +175,51 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
 
   render() {
+    const removePostForm = this.state.removePostForm;
+    const removeCommentForm = this.state.removeCommentForm;
+    const banFromCommunityForm = this.state.banFromCommunityForm;
+    const adminBanForm = this.state.adminBanForm;
     return (
       <div className="person-reports container-lg">
+        {removePostForm && (
+          <ModActionFormModal
```
dessalines: These modals should be inside the component, not the screen, just like they are in `post-listing` or `comment-node` with the dropdowns there.

src/shared/components/person/reports.tsx
```diff
@@ -606,40 +722,110 @@ export class Reports extends Component<ReportsRouteProps, ReportsState> {
...
     this.findAndUpdatePostReport(res);
-    if (this.state.unreadOrAll === UnreadOrAll.Unread) {
-      this.refetch();
-      UnreadCounterService.Instance.updateReports();
-    }
+    this.update();
+  }
+
+  handleRemovePost(form: RemovePost) {
+    this.setState({ removePostForm: form });
```
dessalines: Then these should not be in the state, but call the `client.removePost`

Check out `post.tsx` for how the top level screens should handle these mod actions.

Nutomic: Makes sense I moved remove post and remove comment modals into the components. Ban actions are the same in both cases so Im leaving it in reports.tsx to avoid code duplication.

src/shared/components/comment/comment-report.tsx
Nutomic: The modals and everything else is working fine. However I dont want to handle the js-client upgrade with mandatory reason in this PR, as there are also a lot of unrelated changes needed.

Nutomic:
> Not sure I agree with that. We've had discussions about report sovereignty on other issues, but IMO people don't care who sees their report; they just want action taken on it ASAP. That and community mods are inactive on all but the largest servers, so admins are more likely to be there and able to take action faster.

This was already discussed in https://github.com/LemmyNet/lemmy/pull/5350

> Haven't looked over the PR yet, but ya that should be fixed in the back end, with a test verifying. Open up an issue there if there isn't one already. A lot of times these front end PRs are the ones that actually expose issues in the back end.

https://github.com/LemmyNet/lemmy/issues/6011

src/shared/components/comment/comment-report.tsx
```diff
@@ -172,4 +231,38 @@ export class CommentReport extends Component<
...
+      },
+    });
+  }
+
+  async handleRemoveComment2(reason: string) {
+    var form = this.state.removeCommentForm!;
+    form.reason = reason;
+    this.props.onRemoveComment(form);
+    this.setState({ removeCommentForm: undefined });
+  }
```
dessalines: - Don't use `!` in all but the most extreme cases, it can cause runtime crashes. Use a default instead if its missing.
- Don't use var in typescript, only const or let.
- Don't set state via anything but `setState`. javascript is a terrible language for letting you do this when these fields should be read-only. 
- This doesn't need to be async, there are no `awaits`.
- Should just be renamed handleRemoveComment

src/shared/components/comment/comment-report.tsx
```diff
@@ -172,4 +231,38 @@ export class CommentReport extends Component<
...
+
+  handleRemoveComment() {
+    this.setState({
+      loading: true,
+      removeCommentForm: {
+        comment_id: this.props.report.comment_report.comment_id,
+        removed: !this.props.report.comment.removed,
+      },
+    });
+  }
```
dessalines: I don't see why this is necessary? Its missing both the reason, and its not pushing up the form anyway.

src/shared/components/comment/comment-report.tsx
```diff
@@ -15,17 +16,27 @@ import { PersonListing } from "../person/person-listing";
...
   admins: PersonView[];
   onResolveReport(form: ResolveCommentReport): void;
+  onRemoveComment(form: RemoveComment): void;
+  onModBanFromCommunity(form: BanFromCommunityData): void;
+  onAdminBan(form: BanFromSiteData): void;
 }
 
 interface CommentReportState {
   loading: boolean;
+  removeCommentForm?: RemoveComment;
```
dessalines: Not sure why this is needed in the state. Ideally you click the modal, the modal pushes up the reason to this component, and you can use the `comment_id` and `removed` either from the `comment_view` prop.

src/shared/components/comment/comment-report.tsx
```diff
@@ -172,4 +231,38 @@ export class CommentReport extends Component<
...
+    this.setState({ removeCommentForm: undefined });
+  }
+
+  handleModBanFromCommunity() {
+    this.setState({ loading: true });
+    this.props.onModBanFromCommunity({
+      person: this.props.report.comment_creator,
+      community: this.props.report.community,
+      ban: !this.props.report.creator_banned,
+    });
```
dessalines, line 249: This is the weird thing about not having these modals inside here. Now its pushing up half a form, without the reason included.

src/shared/components/comment/comment-report.tsx
```diff
@@ -161,6 +177,49 @@ export class CommentReport extends Component<
...
+            onClick={this.handleAdminBan}
+            iconClass={`text-${comment_view.creator_banned ? "success" : "danger"}`}
+          />
+        )}
+        {removeCommentForm && (
+          <ModActionFormModal
+            onSubmit={this.handleRemoveComment2}
+            modActionType="remove-comment"
+            isRemoved={comment_view.comment.removed}
+            onCancel={() => this.setState({ removeCommentForm: undefined })}
```
dessalines: Use a handler function.

dessalines (commented):
> This was already discussed in https://github.com/LemmyNet/lemmy/pull/5350

Alright then, that's fine.

src/shared/components/comment/comment-report.tsx
Nutomic: Changed it. There are two methods needed, one for the action button click and then one for the dialog submit. Also async is required by `ModActionFormModal.onSubmit`.

src/shared/components/comment/comment-report.tsx
Nutomic: Right, removed

src/shared/components/comment/comment-report.tsx
Nutomic: Removed although I had to add a bool to render the dialog.

src/shared/components/comment/comment-report.tsx
Nutomic: Seems much more complicated with a handler than having this single line directly here.

src/shared/components/comment/comment-report.tsx
Nutomic: Thats true, but having each report view implement its own logic for modal doesnt make much sense either.

SleeplessOne1917:
Looking at the screenshot in [this comment](https://github.com/LemmyNet/lemmy-ui/pull/3447#issuecomment-3307672525), I can't distinguish between the ban button for the instance vs the community. Also, not sure if the comments I was mentioned in are still an issue since the code they appear on are both marked as outdated. Let me know if my help is still needed with those and how I can help.

Nutomic:
> Looking at the screenshot in https://github.com/LemmyNet/lemmy-ui/pull/3447#issuecomment-3307672525, I can't distinguish between the ban button for the instance vs the community. 

That is true, I used the same icons that were used in other parts of the code which are both the same. Is there another icon I can use, or rather convert them to text buttons?

> Also, not sure if the comments I was mentioned in are still an issue since the code they appear on are both marked as outdated. Let me know if my help is still needed with those and how I can help.

That was about upgrading to the latest lemmy-js-client which makes `reason` mandatory for all mod actions (https://github.com/LemmyNet/lemmy/pull/5982). This requires some big changes like adding an input for instance blocks, so Im not going to do it in this PR.

src/shared/components/comment/comment-report.tsx
```diff
@@ -15,17 +16,27 @@ import { PersonListing } from "../person/person-listing";
...
   admins: PersonView[];
   onResolveReport(form: ResolveCommentReport): void;
+  onRemoveComment(form: RemoveComment): void;
+  onModBanFromCommunity(form: BanFromCommunityData): void;
+  onAdminBan(form: BanFromSiteData): void;
 }
 
 interface CommentReportState {
   loading: boolean;
+  showRemoveCommentDialog: boolean;
```
dessalines, line 39: 👍

src/shared/components/comment/comment-report.tsx
```diff
@@ -161,6 +176,49 @@ export class CommentReport extends Component<
...
           )}
         </button>
+        <ActionButton
+          label={I18NextService.i18n.t(
+            comment_view.comment.removed ? "restore_comment" : "remove_comment",
+          )}
+          inline
+          icon={comment_view.comment.removed ? "restore" : "x"}
+          noLoading
+          onClick={() => this.setState({ showRemoveCommentDialog: true })}
```
dessalines, line 186: I really don't like not having the handlers be separate functions, like we've done everywhere else in this codebase, but I'm not gonna block because of it.

dessalines (commented):
I know we'll need to fix the modBan pretty soon anyway by moving the modals inside, but I spose we can do that in another PR.

Use whichever icon you like, we can always change it later.

Nutomic:
I didnt find any good alternative icon, instead I changed it to show with text:
Screenshot_20250925_111201

SleeplessOne1917 (approved):

PR #3448, opened 2025-09-16
Add button to expand all images (fixes #1273)

### LINKED ISSUE #1273: Adding a "show all images" button

Instead of clicking on each post individually, I'd like an equivalent of the "show images" button from RES that opens all available images.

This proposed new button would convert all the image previews for the current page's posts into their full, non-preview versions.

N/A

### Additional context

Apologies if this is already a feature and I've missed it!
If someone could confirm this needs adding I'd be happy to have a go at implementing it myself.

Nutomic:
Add button on home/community post listings to expand/collapse all post images.

Screenshot_20250916_155151


src/shared/components/community/community.tsx
```diff
@@ -712,6 +715,16 @@ export class Community extends Component<CommunityRouteProps, State> {
             />
           </div>
         )}
+        <div className="col-auto ps-0">
+          <button
+            class="btn btn-secondary"
+            onClick={() => {
+              this.setState({ expandAllImages: !this.state.expandAllImages });
+            }}
```
dessalines: Extract this into a handler function.

src/shared/components/home/home.tsx
```diff
@@ -853,6 +856,16 @@ export class Home extends Component<HomeRouteProps, HomeState> {
...
         </div>
+        <div className="col-auto ps-0">
+          <button
+            class="btn btn-secondary"
+            onClick={() => {
+              this.setState({ expandAllImages: !this.state.expandAllImages });
+            }}
+          >
+            <Icon icon={this.state.expandAllImages ? "minus" : "plus"} />
+          </button>
```
SleeplessOne1917: Same feedback as I left on the community page.

src/shared/components/community/community.tsx
```diff
@@ -712,6 +715,16 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+          >
+            <Icon icon={this.state.expandAllImages ? "minus" : "plus"} />
+          </button>
```
SleeplessOne1917: Visually showing only the icon is fine, but there should be some visually hidden text for screen reader users. You can add a span as a child of the button and hide it using the [`visually-hidden` Bootstrap class](https://getbootstrap.com/docs/5.2/getting-started/accessibility/#visually-hidden-content).

[This accessibility section](https://www.accessibility-developer-guide.com/knowledge/aria/bad-practices/#adding-missing-labels) explains the benefits of this approach over using `aria-label`.

SleeplessOne1917 (changes requested):
Looking at how it's implemented, if I understand correctly, pressing the (un)expand all buttons will toggle whether or not they're expanded unless they have been manually (un)expanded, in which case it will ignore any subsequent togglings of the (un)expand all button. I think that's good for now, but correct me if I understand it wrong.

Marking as requesting change due to the button text issue I pointed out in other comments.

src/shared/components/community/community.tsx
Nutomic: Makes sense. I did it in a different way, by adding a prop `altText` for `Icon`. That one seems to handle accesibility in a different way, but it can be fixed in another PR.

Nutomic:
> Looking at how it's implemented, if I understand correctly, pressing the (un)expand all buttons will toggle whether or not they're expanded unless they have been manually (un)expanded, in which case it will ignore any subsequent togglings of the (un)expand all button. I think that's good for now, but correct me if I understand it wrong.

Yes correct.

src/shared/components/common/icon.tsx
dessalines, line 1: Don't alter this file, we don't use altText on icons, only aria labels and tippy on the icon buttons. Look at any button with an icon for an example.

src/shared/components/community/community.tsx
dessalines: See comment above, use aria labels on the buttons like we do for every icon button. 

I disagree with that article, [it seems to never want you to use aria-labels](https://www.accessibility-developer-guide.com/knowledge/aria/bad-practices/#approach-using-aria) ever.

Just follow the standard we use for every icon button already in lemmy-ui instead of trying to recreate from scratch.

Use `aria-label` and `data-tippy-content` with i18n strings. Check out `src/shared/components/common/vote-buttons.tsx` for an example.

Nutomic: Updated

dessalines (approved):
Looks good. In the future we should avoid using `componentWillReceiveProps` , and just let it rerender when there's any prop change.

PR #3449, opened 2025-09-16
RSS feed should use local domain (fixes #2012)

### LINKED ISSUE #2012: RSS feeds should be different for kbin communities
### Requirements

### Summary

For example, the RSS feed that is linked to in https://lemmy.world/c/tech@kbin.social is https://kbin.social/feeds/m/tech.xml?sort=Active, but it should be https://kbin.social/rss?magazine=tech

### Steps to Reproduce

.

### Technical Details

.

### Lemmy Instance Version

0.18.3

### Lemmy Instance URL

_No response_

Nutomic:
Using the remote domain doesnt work, because it may be a private instance, or may use a different software (eg Piefed or Mbin) with different RSS paths. Also checked other RSS feeds and they are already doing this correctly.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -4,9 +4,15 @@ set -e
 CWD="$(cd -P -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd -P)"
 cd "$CWD/../"
 
-export LEMMY_UI_BACKEND_INTERNAL=0.0.0.0:8536
-export LEMMY_UI_BACKEND_EXTERNAL=voyager.lemmy.ml:8536
-export LEMMY_UI_HTTPS=true
 export LEMMY_UI_DEBUG=true
+
+# Use this to develop on voyager.lemmy.ml
+export LEMMY_UI_BACKEND_REMOTE=voyager.lemmy.ml
+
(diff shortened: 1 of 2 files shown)


src/shared/utils/app.ts
```diff
@@ -127,9 +128,13 @@ export function commentsToFlatNodes(
   return nodes;
 }
 
-export function communityRSSUrl(actorId: string, sort: string): string {
-  const url = new URL(actorId);
-  return `${url.origin}/feeds${url.pathname}.xml${getQueryString({ sort })}`;
+export function communityRSSUrl(community: Community, sort: string): string {
+  var domain = "";
```
dessalines: Use:

`const domain = hostname(community.ap_id)`

Nutomic: Changed

dessalines: The community local check is pointless, just use `const domain = ...`

Nutomic: Okay that works although it makes for longer urls.

dessalines: Ohh ok, I guess the local check is fine there. I added it again using ternary, feel free to merge whenever.

dessalines (approved):

PR #3451, opened 2025-09-17
Private community

Nutomic:
Requires https://github.com/LemmyNet/lemmy/pull/6003 and updated js client which will rename some things. Otherwise its fully working and ready to review.

Screenshots:

Screenshot_20250917_144732
Screenshot_20250917_144835
Screenshot_20250917_144845
Screenshot_20250924_131422

Translations: https://github.com/LemmyNet/lemmy-translations/pull/201

### DIFF
--- a/src/shared/components/community/pending-follows.tsx
+++ b/src/shared/components/community/pending-follows.tsx
@@ -11,7 +11,7 @@ import {
   QueryParams,
   RouteDataResponse,
 } from "@utils/types";
-import { Component, linkEvent } from "inferno";
+import { Component } from "inferno";
 import {
   ApproveCommunityPendingFollower,
   LemmyHttp,
@@ -37,9 +37,10 @@ import { RouteComponentProps } from "inferno-router/dist/Route";
 import { IRoutePropsWithFetch } from "@utils/routes";
 import { InfernoNode } from "inferno";
 import { PendingFollow } from "@components/common/pending-follow";
-import { registrationStateRadios } from "@components/person/registration-applications";
-
-type ViewState = "unread" | "all" | "denied";
+import {
+  RegistrationState,
+  RegistrationStateRadios,
+} from "@components/common/registration-state-radios";
 
 type PendingFollowsData = RouteDataResponse<{
   listPendingFollowsResponse: ListCommunityPendingFollowsResponse;
@@ -51,11 +52,11 @@ interface PendingFollowsState {
 }
 
 interface PendingFollowsProps {
-  viewState: ViewState;
+  viewState: RegistrationState;
   cursor?: DirectionalCursor;
 }
 
-function stateFromQuery(view?: string): ViewState {
+function stateFromQuery(view?: string): RegistrationState {
   switch (view) {
     case "unread":
     case "all":
@@ -107,6 +108,8 @@ export class PendingFollows extends Component<
 
     this.handlePageChange = this.handlePageChange.bind(this);
     this.handleApproveFollower = this.handleApproveFollower.bind(this);
+    this.handlePendingFollowsStateChange =
+      this.handlePendingFollowsStateChange.bind(this);
 
     // Only fetch the data if coming from another route
     if (FirstLoadService.isFirstLoad) {
@@ -181,10 +184,10 @@ export class PendingFollows extends Component<
   selects() {
     return (
       <div className="mb-2">
-        {registrationStateRadios(
-          linkEvent(this, this.handlePendingFollowsStateChange),
-          this.props.viewState,
-        )}
+        <RegistrationStateRadios
+          state={this.props.viewState}
+          onClickHandler={this.handlePendingFollowsStateChange}
+        />
       </div>
     );
   }
@@ -209,8 +212,8 @@ export class PendingFollows extends Component<
     );
   }
 
-  handlePendingFollowsStateChange(i: PendingFollows, event: any) {
-    i.updateUrl({ viewState: event.target.value, cursor: undefined });
+  handlePendingFollowsStateChange(val: RegistrationState) {
+    this.updateUrl({ viewState: val, cursor: undefined });
   }
 
   handlePageChange(cursor?: DirectionalCursor) {
@@ -230,7 +233,7 @@ export class PendingFollows extends Component<
       new LemmyHttp(getHttpBaseInternal(), { headers }),
     );
     // TODO: this is always undefined after page reload
-    const state = viewState ?? "Unread";
+    const state = viewState ?? "unread";
     return {
       listPendingFollowsResponse: headers["Authorization"]
         ? await client.listCommunityPendingFollows({
--- a/src/shared/components/person/registration-applications.tsx
+++ b/src/shared/components/person/registration-applications.tsx
@@ -3,7 +3,6 @@ import {
   cursorComponents,
   getQueryParams,
   getQueryString,
-  randomStr,
   resourcesSettled,
 } from "@utils/helpers";
 import { scrollMixin } from "../mixins/scroll-mixin";
@@ -12,8 +11,7 @@ import {
   QueryParams,
   RouteDataResponse,
 } from "@utils/types";
-import classNames from "classnames";
-import { Component, linkEvent } from "inferno";
+import { Component } from "inferno";
 import {
   ApproveRegistrationApplication,
   LemmyHttp,
@@ -39,8 +37,10 @@ import { PaginatorCursor } from "@components/common/paginator-cursor";
 import { RouteComponentProps } from "inferno-router/dist/Route";
 import { IRoutePropsWithFetch } from "@utils/routes";
 import { InfernoNode } from "inferno";
-
-type RegistrationState = "unread" | "all" | "denied";
+import {
+  RegistrationState,
+  RegistrationStateRadios,
+} from "@components/common/registration-state-radios";
 
 type RegistrationApplicationsData = RouteDataResponse<{
   listRegistrationApplicationsResponse: ListRegistrationApplicationsResponse;
@@ -110,6 +110,8 @@ export class RegistrationApplications extends Component<
 
     this.handlePageChange = this.handlePageChange.bind(this);
     this.handleApproveApplication = this.handleApproveApplication.bind(this);
+    this.handleRegistrationStateChange =
+      this.handleRegistrationStateChange.bind(this);
 
     // Only fetch the data if coming from another route
     if (FirstLoadService.isFirstLoad) {
@@ -197,10 +199,10 @@ export class RegistrationApplications extends Component<
     return (
       <div className="mb-2">
         <span className="me-3">
-          {registrationStateRadios(
-            linkEvent(this, this.handleRegistrationStateChange),
-            this.props.view,
-          )}
+          <RegistrationStateRadios
+            state={this.props.view}
+            onClickHandler={this.handleRegistrationStateChange}
+          />
         </span>
       </div>
     );
@@ -227,8 +229,8 @@ export class RegistrationApplications extends Component<
     );
   }
 
-  handleRegistrationStateChange(i: RegistrationApplications, event: any) {
-    i.updateUrl({ view: event.target.value, cursor: undefined });
+  handleRegistrationStateChange(val: RegistrationState) {
+    this.updateUrl({ view: val, cursor: undefined });
   }
 
   handlePageChange(cursor?: DirectionalCursor) {
@@ -301,66 +303,4 @@ export class RegistrationApplications extends Component<
       return s;
     });
   }
-}
-
-export function registrationStateRadios(
-  handler,
-  currentState: RegistrationState,
-) {
-  const radioId = randomStr();
-
-  return (
-    <div className="btn-group btn-group-toggle flex-wrap mb-2" role="group">
-      <input
-        id={`${radioId}-unread`}
-        type="radio"
-        className="btn-check"
-        value={"unread"}
-        checked={currentState === "unread"}
-        onChange={handler}
-      />
-      <label
-        htmlFor={`${radioId}-unread`}
-        className={classNames("btn btn-outline-secondary pointer", {
-          active: currentState === "unread",
-        })}
-      >
-        {I18NextService.i18n.t("unread")}
-      </label>
-
-      <input
-        id={`${radioId}-all`}
-        type="radio"
-        className="btn-check"
-        value={"all"}
-        checked={currentState === "all"}
-        onChange={handler}
-      />
-      <label
-        htmlFor={`${radioId}-all`}
-        className={classNames("btn btn-outline-secondary pointer", {
-          active: currentState === "all",
-        })}
-      >
-        {I18NextService.i18n.t("all")}
-      </label>
-
-      <input
-        id={`${radioId}-denied`}
-        type="radio"
-        className="btn-check"
-        value={"denied"}
-        checked={currentState === "denied"}
-        onChange={handler}
-      />
-      <label
-        htmlFor={`${radioId}-denied`}
-        className={classNames("btn btn-outline-secondary pointer", {
-          active: currentState === "denied",
-        })}
-      >
-        {I18NextService.i18n.t("denied")}
-      </label>
-    </div>
-  );
 }
--- a/src/shared/components/common/registration-state-radios.tsx
+++ b/src/shared/components/common/registration-state-radios.tsx
@@ -1,0 +1,83 @@
+import { I18NextService } from "@services/index";
+import { randomStr } from "@utils/helpers";
+import classNames from "classnames";
+import { linkEvent } from "inferno";
+import { Component } from "inferno";
+
+export type RegistrationState = "unread" | "all" | "denied";
+
+interface RegistrationStateRadiosProps {
+  state: RegistrationState;
+  onClickHandler(val: RegistrationState): void;
+}
+
+export class RegistrationStateRadios extends Component<
+  RegistrationStateRadiosProps,
+  object
+> {
+  constructor(props: any, context: any) {
+    super(props, context);
+  }
+
+  render() {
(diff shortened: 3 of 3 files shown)


src/shared/components/app/navbar.tsx
```diff
@@ -159,6 +183,7 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
                   </NavLink>
                 </li>
               )}
+              {/* TODO: what is this section for and why does it duplicate everything? */}
```
dessalines, line 186: I'd have to look over, but it likely has do to with mobile.

src/shared/components/common/pending-follow.tsx
dessalines, line 1: `pending-follow.tsx`, named like all the other files.

src/shared/components/common/pending-follow.tsx
```diff
@@ -0,0 +1,113 @@
...
+> {
+  state: PendingFollowState = {
+    approveLoading: false,
+    denyLoading: false,
+  };
+
+  constructor(props: any, context: any) {
+    super(props, context);
+  }
+  componentWillReceiveProps(
```
dessalines, line 36: We should really get away from using this. Normally you should only need to set this in the state, IE do `state: PendingFollowState = { whateverVar: this.props....`

Then when new data comes in, it'll rerender based on the changed props. Don't worry about it for now tho.

src/shared/components/common/pending-follow.tsx
```diff
@@ -0,0 +1,113 @@
...
+        approveLoading: false,
+        denyLoading: false,
+      });
+    }
+  }
+
+  render() {
+    const p = this.props.pending_follow;
+    return (
+      <div className="d-flex flex-column">
```
dessalines: Use the bootstrap grid classes if you need to. `registration-application.tsx` does this.

src/shared/components/community/communities.tsx
```diff
@@ -43,6 +43,7 @@ import { RouteComponentProps } from "inferno-router/dist/Route";
 import { scrollMixin } from "../mixins/scroll-mixin";
 import { isBrowser } from "@utils/browser";
 import { PaginatorCursor } from "@components/common/paginator-cursor";
+import { community_visibility_label } from "./sidebar";
```
dessalines: Typescript functions should be camelCase https://ts.dev/style/#syntax

src/shared/components/community/community.tsx
```diff
@@ -438,23 +438,45 @@ export class Community extends Component<CommunityRouteProps, State> {
   }
 
   render() {
+    const res =
+      this.state.communityRes.state === "success" &&
+      this.state.communityRes.data;
+    const can_view_community =
```
dessalines: camelcase.

src/shared/components/community/pending-follows.tsx
```diff
@@ -0,0 +1,315 @@
...
+    return `${I18NextService.i18n.t(
+      "community_pending_follows",
+    )} - ${this.isoData.siteRes.site_view.site.name}`;
+  }
+
+  render() {
+    const state = this.state.appsRes.state;
+    const pending = state === "success" && this.state.appsRes.data.items;
+    return (
+      <div className="registration-applications container-lg">
```
dessalines: Do a grep for `registration` to try and find all these cases and replace them with pending-follows.

src/shared/components/community/pending-follows.tsx
dessalines, line 1: Looks good.

src/shared/components/community/sidebar.tsx
```diff
@@ -166,9 +168,14 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
...
     const {
       community: { name, ap_id, id, posting_restricted_to_mods, visibility },
-      community_actions: { received_ban_at } = {},
+      community_actions: { received_ban_at, follow_state } = {},
     } = this.props.community_view;
 
+    const visibility_label = community_visibility_label(visibility);
+    const visibility_description = (visibility_label +
+      "_desc") as NoOptionI18nKeys;
+    const can_view_community =
```
dessalines: camelCase.

src/shared/components/community/sidebar.tsx
```diff
@@ -200,35 +207,42 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
                     loading={this.state.followCommunityLoading}
                     showRemoteFetch={!this.props.myUserInfo}
                   />
-                  {this.canPost && this.createPost()}
+                  {this.canPost && can_view_community && this.createPost()}
```
dessalines: I spose it is necessary to hide this.

src/shared/components/community/sidebar.tsx
```diff
@@ -166,9 +168,14 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
...
       community: { name, ap_id, id, posting_restricted_to_mods, visibility },
-      community_actions: { received_ban_at } = {},
+      community_actions: { received_ban_at, follow_state } = {},
     } = this.props.community_view;
 
+    const visibility_label = community_visibility_label(visibility);
+    const visibility_description = (visibility_label +
+      "_desc") as NoOptionI18nKeys;
+    const can_view_community =
+      visibility !== "Private" || follow_state === "Accepted";
```
dessalines: I think you used this logic elsewhere, so it might be a good idea to add it as a utils helper function.

src/shared/components/community/sidebar.tsx
```diff
@@ -767,3 +768,28 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
     );
   }
 }
+
+export function community_visibility_label(
+  visibility: CommunityVisibility,
+): NoOptionI18nKeys {
+  var visibility_label = "community_visibility_";
```
dessalines: Don't use var. Just use early return and remove all the breaks.

src/shared/services/UnreadCounterService.ts
```diff
@@ -79,10 +83,24 @@ export class UnreadCounterService {
     }
   }
 
+  public async updatePendingFollows() {
+    if (this.shouldUpdate) {
+      // TODO: remove community_id param
```
dessalines: I see this is part of the back-end PR you have. Anyway you can probably remove the comment because a newly generated lemmy-js-client will annoy us about this.

src/shared/utils/roles.ts
```diff
@@ -72,6 +72,15 @@ export function moderatesSomething(
   return amAdmin(myUserInfo) || (myUserInfo?.moderates?.length ?? 0) > 0;
 }
 
+export function moderatesPrivateCommunity(
+  myUserInfo: MyUserInfo | undefined,
+): boolean {
+  return (
+    (myUserInfo?.moderates?.filter(c => c.community.visibility === "Private")
+      ?.length ?? 0) > 0
```
dessalines: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/shared/components/community/communities.tsx
Nutomic: This should really be a lint. I tried to set `"camelcase": [1, {ignoreImports: true, ignoreDestructuring: true}]` for eslint but it still gives 870 warnings, most of them for imports. Also --fix doesnt work...

src/shared/components/community/pending-follows.tsx
Nutomic: Done

src/shared/components/community/sidebar.tsx
Nutomic: That would be much more verbose as it needs to convert to `NoOptionI18nKeys` every time.

src/shared/components/common/pending-follow.tsx
Nutomic: Seems fine like this.

src/shared/components/community/communities.tsx
dessalines: Yep I def agree. Open that up as another issue, because it still might be possible.

src/shared/components/common/pending-follow.tsx
dessalines: We don't and shouldn't use flex like this anywhere. Its better to use bootstrap grid classes, which is the standard for lemmy-ui.

src/shared/components/community/sidebar.tsx
dessalines: Then use `let`.

`var` is deprecated in typescript, only use let or const. This could also be done with a `const Map` also. Look at `sort-select.tsx` if you want to see a good example of how that's done.

src/shared/components/person/registration-applications.tsx
```diff
@@ -356,3 +302,65 @@ export class RegistrationApplications extends Component<
     });
   }
 }
+
+export function registrationStateRadios(
```
dessalines: Doesn't need to be exported.

src/shared/components/community/communities.tsx
Nutomic: https://github.com/LemmyNet/lemmy-ui/issues/3484

src/shared/components/community/sidebar.tsx
Nutomic: Fixed an opened https://github.com/LemmyNet/lemmy-ui/pull/3485

src/shared/components/common/pending-follow.tsx
Nutomic: Okay changed.

src/shared/components/person/registration-applications.tsx
Nutomic: Yes because Im using it from pending-follows.tsx (instead of copy-pasting the code).

src/shared/services/UnreadCounterService.ts
```diff
@@ -79,10 +83,23 @@ export class UnreadCounterService {
...
   }
 
+  public async updatePendingFollows() {
+    if (this.shouldUpdate) {
+      const pendingFollowsRes =
+        await HttpService.client.getCommunityPendingFollowsCount({
+          community_id: 0,
+        });
+      if (pendingFollowsRes.state === "success") {
+        this.pendingFollowCountSubject.next(pendingFollowsRes.data.count);
```
SleeplessOne1917, line 91: What does a community ID of 0 mean?

src/shared/components/community/sidebar.tsx
```diff
@@ -767,3 +767,35 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
     );
   }
 }
+
+export function communityVisibilityLabel(
+  visibility: CommunityVisibility,
+): NoOptionI18nKeys {
+  let visibility_label = "community_visibility_";
+  // TODO: might be easier with a library that converts camel case to snake case
```
SleeplessOne1917: Outside the scope of this story, but `ts_rs` can make the generated TS types have all of their names in camelCase with `#[ts(rename_all = "camelCase")]`. See https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html.

If you wanted to, you could even use the [similarly named `serde` macro](https://serde.rs/container-attrs.html) to make it so even the JSON that gets accepted/returned by the server has camelCase fields with `#[serde(rename_all = "camelCase")]`. On the off chance you wanted to do this, you wouldn't need to have the `rename_all` `ts_rs` attributes since that crate has serde compatability as a default feature.

src/shared/components/community/pending-follows.tsx
```diff
@@ -0,0 +1,283 @@
...
+        {registrationStateRadios(
+          linkEvent(this, this.handlePendingFollowsStateChange),
+          this.props.viewState,
+        )}
+      </div>
+    );
+  }
+
+  applicationList(pending: PendingFollowView[]) {
+    if (this.props.viewState === "Denied") {
```
SleeplessOne1917: Is similar filtering needed for unread?

src/shared/services/UnreadCounterService.ts
Nutomic: I removed this parameter in the backend because it doesnt make sense, the endpoint should return the count for all moderated communities instead. However I didnt update the js client as it has too many unrelated breaking changes.

https://github.com/LemmyNet/lemmy/pull/6003/files#diff-e5ca7df2592314dbc01ee19ba4575b092fc18c025af24b1d9a2236eda4fa7a5fL9

src/shared/components/community/sidebar.tsx
Nutomic: https://github.com/LemmyNet/lemmy/issues/6062

src/shared/components/community/pending-follows.tsx
Nutomic: Unread is handled by the api: https://github.com/LemmyNet/lemmy-ui/pull/3451/files#diff-8cddb1772f8cf466d9ddfe165c6e091706863d8bb138f8349501f9a0cb3e7201R237

SleeplessOne1917: That makes sense. What I don't understand is why filtering for follow state isn't also done on the backend.

Nutomic: I copied the logic from [registration-applications.tsx](https://github.com/LemmyNet/lemmy-ui/blob/main/src/shared/components/person/registration-applications.tsx#L264) which also works in the same way.

src/shared/components/community/sidebar.tsx
```diff
@@ -170,6 +172,10 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
       community_actions: { received_ban_at } = {},
     } = this.props.community_view;
 
+    const visibilityLabel = communityVisibilityLabel(visibility);
+    const visibilityDescription = (visibilityLabel +
+      "_desc") as NoOptionI18nKeys;
```
matc-pub, line 178: We should avoid these `as NoOptionI18nKeys` casts, as they keep "working" even when the keys change. Typescript can detect invalid keys when they are template literals.
```typescript
function example(x: "thing" | "other_thing"): NoOptionI18nKeys {
  return `prefix_${x}`;
  // ^ Type '"prefix_thing" | "prefix_other_thing"' is not assignable to type 'NoOptionI18nKeys'.
}
```

Nutomic: All this should become unnecessary once js-client is updated with these changes: https://github.com/LemmyNet/lemmy/pull/6067

src/shared/components/common/pending-follow.tsx
```diff
@@ -0,0 +1,113 @@
...
+          />
+          <UserBadges
+            classNames="ms-1"
+            myUserInfo={this.props.myUserInfo}
+            creator={p.person}
+            showCounts
+          />
+        </span>
+        <span className="col">
+          {p.follow_state === "ApprovalRequired" && (
```
dessalines: Surprised this isn't throwing a case error now.

src/shared/components/common/pending-follow.tsx
```diff
@@ -0,0 +1,113 @@
...
+    );
+  }
+
+  handleApprove(i: PendingFollow) {
+    i.setState({ approveLoading: true });
+    i.props.onApproveFollower({
+      follower_id: i.props.pending_follow.person.id,
+      community_id: i.props.pending_follow.community.id,
+      approve: true,
+    });
```
dessalines, line 102: You don't have to do this for this PR, but in the future:

`loading` should never be a state in components that don't do API calls. It should only be a state in top-level components / screen ones like `home` or `post`. 

In these it should always be a prop. 

When you do the API call at the screen level, you can add a `approveLoading` state, set it before and after the API call, then pass it as a prop anywhere you need.

src/shared/components/community/community-form.tsx
```diff
@@ -268,9 +268,24 @@ export class CommunityForm extends Component<
...
-              <option value="Public">{I18NextService.i18n.t("public")}</option>
-              <option value="LocalOnly">
-                {I18NextService.i18n.t("local_only")}
+              <option value="Public">
+                {I18NextService.i18n.t("community_visibility_public")}
+              </option>
+              <option value="Unlisted">
+                {I18NextService.i18n.t("community_visibility_unlisted")}
+              </option>
+              <option value="LocalOnlyPublic">
```
dessalines: Check the case on all these, because it'll be unchecked. Or maybe we'll handle that in another issue.

src/shared/components/community/community.tsx
```diff
@@ -115,7 +115,7 @@ import {
   CommentsLoadingSkeleton,
   PostsLoadingSkeleton,
 } from "../common/loading-skeleton";
-import { Sidebar } from "./sidebar";
+import { canViewCommunity, Sidebar } from "./sidebar";
```
dessalines: Might want to move that to `utils/app.tsx` or somewhere else.

Nutomic: Done

Nutomic:
Merged from main and updated the enum variants to snake case. Also tested that its working.

src/shared/components/common/pending-follow.tsx
Nutomic: The logic is copied from registration-application.tsx. It would be possible to move `approveLoading` and `denyLoading` to the top-level view, but then it would have to be a number to affect only the specific view and not the full list. And the prop would have to be passed as `approveLoading={pending_follow.community.id == this.state.approveLoadingId}`. Seems too complicated and not worth the effort.

src/shared/components/person/registration-applications.tsx
dessalines: This needs to be made into a proper component or functional component, with `onChange` props, not passing in handlers. 

It'd probably be a good idea to move it to the `common` folder with the rest of them also.

Check out any `listing-type-select` for a simple example.

Nutomic: Done

PR #3454, opened 2025-09-19
Fix community link in modlog title (fixes #2209)

### LINKED ISSUE #2209: Modlog for remote community incorrectly links to local community with the same name
### Requirements

### Summary

The sidebar of https://lemmy.ml/c/microblogmemes@lemmy.world links to https://lemmy.ml/modlog/124413, which contains the modlog of [`!microblogmemes@lemmy.world`](https://lemmy.world/c/microblogmemes).
The modlog page at https://lemmy.ml/modlog/124413 show the following section at the top:
> [/c/microblogmemes ](https://lemmy.ml/c/microblogmemes)Modlog
(There is also a space as part of the link instead of after the link, but that's not in scope for this report.)

This link incorrectly leads to a (in this case non-existing) local community `!microblogmemes@lemmy.ml`.

### Steps to Reproduce

1. Visit remote community, such as https://lemmy.ml/c/microblogmemes@lemmy.world
2. Click modlog link in sidebar
3. Click community name above filters
4. Be on 404 page for non-existing local community

### Technical Details

image

### Lemmy Instance Version

BE: 0.18.5, frontend version is not exposed

### Lemmy Instance URL

https://lemmy.ml

Nutomic:
Screenshot_20250919_145556

### DIFF
--- a/src/shared/components/community/community-link.tsx
+++ b/src/shared/components/community/community-link.tsx
@@ -76,9 +76,9 @@ export function communityLink(
     return { link: `/c/${community.name}` };
   } else {
     const serverStr = `@${hostname(community.ap_id)}`;
-    const link = !realLink
-      ? `/c/${community.name}${serverStr}`
-      : community.ap_id;
+    const link = realLink
+      ? community.ap_id
+      : `/c/${community.name}${serverStr}`;
 
     return { link, serverStr };
   }
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -45,7 +45,7 @@ import { HtmlTags } from "./common/html-tags";
 import { Icon, Spinner } from "./common/icon";
 import { MomentTime } from "./common/moment-time";
 import { SearchableSelect } from "./common/searchable-select";
-import { CommunityLink } from "./community/community-link";
+import { communityLink, CommunityLink } from "./community/community-link";
 import { PersonListing } from "./person/person-listing";
 import { getHttpBaseInternal } from "../utils/env";
 import { IRoutePropsWithFetch } from "@utils/routes";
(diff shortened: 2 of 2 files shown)


src/shared/components/community/community-link.tsx
```diff
@@ -76,9 +76,9 @@ export function communityLink(
     return { link: `/c/${community.name}` };
   } else {
     const serverStr = `@${hostname(community.ap_id)}`;
-    const link = !realLink
-      ? `/c/${community.name}${serverStr}`
-      : community.ap_id;
+    const link = realLink
+      ? community.ap_id
+      : `/c/${community.name}${serverStr}`;
```
Nutomic, line 81: Just flipped these to remove `!` which was confusing.

src/shared/components/modlog.tsx
```diff
@@ -911,9 +911,11 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
               <>
                 <Link
                   className="text-body"
-                  to={`/c/${communityResp.community_view.community.name}`}
+                  to={
+                    communityLink(communityResp.community_view.community).link
+                  }
                 >
-                  /c/{communityResp.community_view.community.name}
+                  {communityLink(communityResp.community_view.community).link}
```
Nutomic, line 918: Dont see any way to create a variable here, and doesnt seem worth the effort making a separate function that is only used once.

dessalines: That's fine. Its pretty annoying to do a lot of variable manipulation inside react html, and the only way to do it correctly is to extract to a function / component.

But also, you can probably just use `<CommunityLink` here anyway. And if we need to do styling, use props to manipulate that.

Nutomic: Tried to use that but it looks very strange with different colors and text sizes, and then the "modlog" text with a different style right after it.

dessalines: Alright, this is fine then.

PR #3461, opened 2025-09-19
Add more settings

dessalines:
WIP


Nutomic:
This one is also missing translations. For keyword blocking it may be cleaner to use newline as separator instead of comma. Otherwise its difficult to review as its mixed with changes from the other PR. I assume you went through the whole list of backend settings to add new ones and remove those which dont exist anymore?

dessalines:
This isn't ready yet, it has to come after #3479 

> I assume you went through the whole list of backend settings to add new ones and remove those which dont exist anymore?

Many of them.

> For keyword blocking it may be cleaner to use newline as separator instead of comma.

I'll do this now.

PR #3475, opened 2025-09-24
Convert buttons for view all, show context to links (fixes #3229)

### LINKED ISSUE #3229: Unable to middle-click "Show all comments" nor "Show context"
### Requirements

### Summary

In the following video, I demonstrate
trying to middle click, to open in a different page, both of these buttons

https://github.com/user-attachments/assets/5a6abee3-35e9-4c25-842c-a702a03f5784

To no effect.

### Steps to Reproduce

Go to a comment where you cannot see all comments

for instance

https://lemmy.ml/post/32996842/19796691

then try to middle click (or right click and new tab) these buttons

Image

It will not work.

Only left click works, which means you have to lose the current view, instead of openning these in another tab

### Technical Details

firefox  140.0.4
windows 10

### Lemmy Instance Version

BE: 0.19.12

### Lemmy Instance URL

https://lemmy.ml/post/32996842/19796691

Nutomic:


src/shared/components/post/post.tsx
```diff
@@ -937,19 +941,19 @@ export class Post extends Component<PostRouteProps, PostState> {
         <div>
           {!!getCommentIdFromProps(this.props) && (
             <>
-              <button
-                className="ps-0 d-block btn btn-link text-muted"
-                onClick={linkEvent(this, this.handleViewAllComments)}
+              <a
+                className="ps-0 d-block btn btn-link text-muted text-start"
+                href={this.handleViewAllComments()}
```
dessalines: Using `<a` will cause a page reload. Always use `<Link` , unless going to an external page.

This specific case is a little weird though since it needs to *refresh* the same route, so I don't know that using `<Link` will work. If it doesn't we should probably just say this is a no-fix, because not having to do a whole page reload is better.

Nutomic: Changed, works fine.

Nutomic:
While testing this I noticed a strange problem on the main branch: clicking "show context" on [this page](https://voyager.lemmy.ml/post/413/23#comment-23) but keeps showing the same comment instead of the parent. And after clicking "show context" a few more times, it doesnt do anything. Could be a backend issue.

dessalines:
Yep that might be a backend issue, something weird is going on there.

dessalines (approved):

SleeplessOne1917 (approved):

PR #3479, opened 2025-09-25
Remove tables in favor of bootstrap columns

### LINKED ISSUE #546: Change communities page to use bootstrap rows instead of tables

dessalines:
## Description

First I tried using `list-inline`, but no one seemed to like that. Then I tried creating cards, but I really didn't like them either.

So I kept the basic table structure from before, but converted them to use bootstrap columns. The header even stays now. 

- Similar to #3440 , but instead of using `list-inline`, converted the tables to bootstrap columns.

See notes on the above issue for all the other changes.

Fixes #546 

### Modlog

Screenshot_2025-09-25-09-14-11-582_mark via

### Users

Screenshot_2025-09-25-09-15-32-383_mark via

### Communities

Screenshot_2025-09-25-09-14-22-797_mark via

### Instances

Screenshot_2025-09-25-09-17-20-548_mark via

### Uploads

Screenshot_2025-09-25-09-15-40-600_mark via

### DIFF
--- a/src/shared/components/home/tagline-form.tsx
+++ b/src/shared/components/home/tagline-form.tsx
@@ -57,7 +57,7 @@ export class TaglineForm extends Component<TaglineFormProps, TaglineFormState> {
             {!this.state.clearMarkdown && (
               <MarkdownTextArea
                 initialContent={this.state.content}
-                focus={true}
+                focus
                 onContentChange={s => this.handleTaglineChange(this, s)}
                 hideNavigationWarnings
                 allLanguages={[]}


Nutomic:
On desktop it looks good. However on mobile views the text is mixed up and its not clear in what order to read it. This is especially noticable in the modlog. It would probably look better if you remove all columns on mobile, and render the whole modlog entry as a single line of text.

Other pages look fine. Btw it seems you broke something in the theme because everything looks grey now.

dessalines:
> This is especially noticable in the modlog. It would probably look better if you remove all columns on mobile, and render the whole modlog entry as a single line of text.

You mean every item is its own row on mobile? IE `col-12` . Because there's no way to get it all on one row.

The gray is intentional, as I regenerated the themes to use a gray for `btn-secondary`. I'd like to use less colors than before, only using them in very rare cases, or for the core items like user and community. This matches what I've done with jerboa, and google's m3 theme guidelines.

Nutomic:
I mean to treat each modlog entry as simple text with line wraps, instead of having separate columns.

The theme change should be a separate PR.

dessalines:
I don't have time to separate these, there's too much work and fixes in this PR.

> I mean to treat each modlog entry as simple text with line wraps

I'll try using col-auto, that should handle variable widths.

dessalines:
I tried `col-auto` for everything, but it really only looked good for the modlog. The other places it looks better to have consistent columns, rather than variable width.

Nutomic:
The modlog is still unreadable:
Screenshot_20250929_113417



I would show the modlog as normal text like this:
Screenshot_20250929_113502

This is with the following quick and dirty changes in modlog.tsx:838. So there should be a conditional to return the existing code for large screens, and the code below for small screens:
```html
        <>
          <div className="row" key={id}>
            <span className="col d-inline">
              <MomentTime published={publishedAt} /><span className="px-1">
              {this.amAdminOrMod && moderator ? (
                <PersonListing person={moderator} myUserInfo={myUserInfo} />
              ) : 
                this.modOrAdminText(moderator)}
              </span>
              {data}
            </span>
          </div>
          <hr />
        </>
```

dessalines:
Okay I've removed col-auto in favor of just `col` (IE `col-4`) now:

I don't like it as it squishes all the main content into a single skinny column, but at least it won't cause right-scrolling.

Screenshot_2025-10-01-11-36-05-94_a87fd7db6caa850b517aa6fa9d2fcd0e

matc-pub:
For the modlog mobile table maybe col-6 for Time and Mod, and col-12 for Action. That should wrap action to a new row. The column labels could be hidden, the column values make sense on their own.

dessalines:
> For the modlog mobile table maybe col-6 for Time and Mod, and col-12 for Action.

I agree, and that's what I originally had it as (and is shown in the pics at the top), but that got shot down.

matc-pub:
Ah. The `reason` can look like it's just part of the action text. People scanning a modlog are likely mostly interested in the action. Giving it a different background or some other visual difference to the reason could work. Or maybe the difference should be highlighted in reason, as it is user input and not some template text.

matc-pub:
Bootstrap can also reorder things with "order-1". Reordering things so that Action is the first row, Time and Mod the second row, and Reason the third row could also work.

dessalines:
Both the action and the reason are potentially long and variable-length strings, so it makes sense to have them on their own rows.

Time and user are more limited, so `col-6` is probably fine there. We probably don't need to mess with order. 

I'm changing it back to `col-6` for time and mod, and `col-12` for the others, since two of us think it looks better.

matc-pub:
Maybe Reason could also be presented as a quote of the mod/admin.

dessalines:
I wouldn't be opposed to that, although lets handle that in another issue. I have a few PRs that are dependent on this one.

Nutomic:
This is how the modlog looks now. Not ideal but you can open an issue to do further improvements later.

Screenshot_20251002_112738

However this PR also has major theme changes which are not discussed at all. In my opinion it looks very ugly to turn everything grey all of a sudden. So move those changes to a separate PR to discuss them properly, and then we can merge the table changes here.

dessalines:
It would take too much work to extract all the fixes I did here.

The only real theme changes is changing our `btn-secondary` color to gray, which is how it is in the default bootstrap theme, and removing card background colors to align with bootstrap v5.3. 

As stated in the previous PR, I want to have lemmy-ui be a more muted / less loud theme, and try to only use colors for our primary entities: users and communities. This means most buttons, icons, etc will stand out less, with users and communities being more easily recognizable, and standing out more. We should try to use text decorations or font weights to make things stand out, rather than colors, but even that probably isn't necessary in most cases.

Lets just merge this and we can discuss theme color changes in another PR, but I'd like us to use a more muted theme going forward anyway.

dessalines:
I've got 2 PRs stacked behind this one, so we should get this merged soon.

Nutomic (approved):
Had another look, some of the themes in 0.19 like darkly already use gray buttons so its not as big a change as i thought. Still next time make a separate PR for unrelated changes please.

Set it to auto-merge but there are some conflicts to resolve.

PR #3480, opened 2025-09-25
Fetch emoji data separately (fixes #3470)

### LINKED ISSUE #3470: Make lemmy-ui less bloated
### Requirements

### Summary

Visualized by [BundleAnalyzerPlugin](https://www.npmjs.com/package/webpack-bundle-analyzer):

Image

As you can see a large part of the code size comes from `date-fns` which seems to be loaded twice and includes data for all available languages. There is also a lot of data for translation strings, with each language being ~50kb resulting in 1.6MB total. And there is a library mathematica.js which doesnt seem to be used anywhere in the code.

So it would be possible to reduce the total size a lot by moving language specific data into separate files, and only loading data for the language which is currently active. This can be done with [code splitting](https://webpack.js.org/guides/code-splitting/).

Nutomic:
Based on instructions from [emoji-mart readme](https://github.com/missive/emoji-mart?tab=readme-ov-file#-data). 

Not sure if this is working correctly as client.js is the same size as before both in debug and prod builds.


Nutomic:
Emojis are working, but not sure if the size reduction is really working. Gonna put a new alpha on voyager.lemmy.ml and see how it looks there.

Nutomic:
Seems like the size optimization didnt work. lemmy.ml (0.19) shows 1.9 MB size for client.js while voyager.lemmy.ml (1.0) shows 2.0 MB.

PR #3481, opened 2025-09-25
Block community from context menu (fixes #3306)

### LINKED ISSUE #3306: 'Block Community' should be an option in the vertical ellipsis menu
### Requirements

Title says it all. At the moment I can't block a toxic community _without_ first entering that community and seeing it's conent - or - edit a block list manually. A menu shortcut alongside 'Block User' makes complete sense to me, and as a bonus would satisfy [this ticket](https://github.com/LemmyNet/lemmy-ui/issues/3267) as well!

Nutomic:
Screenshot_20250925_160451

### DIFF
--- a/src/shared/components/common/content-actions/content-action-dropdown.tsx
+++ b/src/shared/components/common/content-actions/content-action-dropdown.tsx
@@ -35,7 +35,7 @@ interface ContentActionDropdownPropsBase {
   onEdit: () => void;
   onDelete: () => Promise<void>;
   onReport: (reason: string) => Promise<void>;
-  onBlockUser: () => Promise<void>;
+  onBlockPerson: () => Promise<void>;
   onRemove: (reason: string) => Promise<void>;
   onBanFromCommunity: (form: BanUpdateForm) => Promise<void>;
   onAppointCommunityMod: () => Promise<void>;
@@ -157,7 +157,7 @@ export default class ContentActionDropdown extends Component<
 
   render() {
     // Possible enhancement: Priority+ pattern instead of just hard coding which get hidden behind the show more button.
-    const { onSave, type, onDelete, onBlockUser, onEdit, moderators } =
+    const { onSave, type, onDelete, onBlockPerson, onEdit, moderators } =
       this.props;
     const {
       id,
@@ -313,7 +313,7 @@ export default class ContentActionDropdown extends Component<
                       <ActionButton
                         icon="slash"
                         label={I18NextService.i18n.t("block_user")}
-                        onClick={onBlockUser}
+                        onClick={onBlockPerson}
                       />
                     </li>
                     {type === "post" && (
--- a/src/shared/components/common/content-actions/content-action-dropdown.tsx
+++ b/src/shared/components/common/content-actions/content-action-dropdown.tsx
@@ -36,6 +36,7 @@ interface ContentActionDropdownPropsBase {
   onDelete: () => Promise<void>;
   onReport: (reason: string) => Promise<void>;
   onBlockPerson: () => Promise<void>;
+  onBlockCommunity: () => Promise<void>;
   onRemove: (reason: string) => Promise<void>;
   onBanFromCommunity: (form: BanUpdateForm) => Promise<void>;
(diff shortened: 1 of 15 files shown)


src/shared/components/home/home.tsx
```diff
@@ -983,6 +987,13 @@ export class Home extends Component<HomeRouteProps, HomeState> {
     }
   }
 
+  async handleBlockCommunity(form: BlockCommunity) {
+    const blockCommunityRes = await HttpService.client.blockCommunity(form);
+    if (blockCommunityRes.state === "success") {
+      updateCommunityBlock(blockCommunityRes.data, this.isoData.myUserInfo);
+    }
+  }
```
Nutomic, line 996: I noticed that this logic gets duplicated a lot, just in this PR I had to write it three times. And its the same for other API calls. So it would make sense to move all these to a combined file `api.tsx` which can be imported everywhere else.

dessalines: Yep, a lot of these could get cleaned up.

src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -311,9 +313,18 @@ export default class ContentActionDropdown extends Component<
...
-                        onClick={onBlock}
+                        onClick={onBlockUser}
                       />
                     </li>
+                    {type === "post" && (
+                      <li>
+                        <ActionButton
+                          icon="slash"
+                          label={I18NextService.i18n.t("block_community")}
+                          onClick={this.props.onBlockCommunity}
```
dessalines: Be consistent with the other block, and import it the same way.

src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -35,7 +35,7 @@ interface ContentActionDropdownPropsBase {
   onEdit: () => void;
   onDelete: () => Promise<void>;
   onReport: (reason: string) => Promise<void>;
-  onBlock: () => Promise<void>;
+  onBlockUser: () => Promise<void>;
```
dessalines: Both comments and posts have a community, so you might as well add `onBlockCommunity` right below here also.

Also might be better to call this `onBlockPerson`, because its named that elsewhere already, like in `comment-node.tsx`

Nutomic: > Both comments and posts have a community, so you might as well add onBlockCommunity right below here also.

I didnt add that because `CommentSlimView` doesnt have the community. I see now that there is a method to get the community in `comment-node.tsx` so it would work. Still, most of the time when browing comments you are viewing a post, and then it wouldnt make sense to have an option to block the community on every single comment. So it seems fine like this for now, and block community for comments could be added later only on the homepage comment listing.

Also renamed to onBlockPerson.

src/shared/components/common/content-actions/content-action-dropdown.tsx
Nutomic: Doesnt seem possible with post or comment specific props. Others are also like this.

src/shared/components/home/home.tsx
Nutomic: https://github.com/LemmyNet/lemmy-ui/issues/3482

src/shared/components/common/content-actions/content-action-dropdown.tsx
dessalines: Check the `handleBanFromCommunity` on `comment-node.tsx`, because it's an example of how to get the community even for the commentslim variant. These exceptions are going to start piling up unless we stay consistent and keep all the mod actions that you can actually do to both comments and posts.

Nutomic: Okay added

dessalines (approved):
Looks good. 

I know this seems verbose to push up all these required events to the screens, but its so nice to be able to take advantage of the type checking. Its how jerboa does it also and it saves a lot of headaches.

SleeplessOne1917 (approved):

PR #3485, opened 2025-09-26
Enable various lints

Nutomic:
Not sure why most of these rules were explicitly disabled when they arent even throwing any errors.


eslint.config.mjs
```diff
@@ -44,41 +44,12 @@ export default [
...
-      "no-return-assign": 0,
-      "no-throw-literal": 0,
-      "no-trailing-spaces": 0,
-      "no-unused-expressions": 0,
-      "no-useless-constructor": 0,
-      "no-useless-escape": 0,
-      "no-var": 0,
-      "prefer-const": "error",
-      "prefer-rest-params": 0,
-      "prettier/prettier": "error",
```
Nutomic, line 77: prefer-const and prettier are already enabled by default.

eslint.config.mjs
```diff
@@ -44,41 +44,12 @@ export default [
   {
     files: ["src/**/*.js", "src/**/*.mjs", "src/**/*.ts", "src/**/*.tsx"],
     rules: {
-      "@typescript-eslint/ban-ts-comment": 0,
       "@typescript-eslint/no-explicit-any": 0,
```
Nutomic, line 47: Removing this results in 450 errors.

PR #3486, opened 2025-09-26
Indicate removed/deleted comments (fixes #1144)

### LINKED ISSUE #1144: Add custom CSS to deleted/removed comments
Right now there is no way to differentiate if a comment has been removed or if someone just wrote that italicized. Adding some sort of property, greyed out text or background for example, would make it easy to tell that.

Nutomic:
Screenshot_20250926_162115

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -221,13 +221,16 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
 
   render() {
     const node = this.props.node;
-    const cv = this.commentView;
     const {
       creator_is_moderator,
       creator_banned_from_community,
+      creator_banned,
+
       comment_actions: { like_score: my_vote } = {},
       creator_is_admin,
       comment: {
+        deleted,
+        removed,
         id,
         language_id,
         published_at,
@@ -235,8 +238,10 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
         updated_at,
         child_count,
         locked,
+        post_id,
       },
       creator,
+      person_actions,
     } = this.commentView;
 
     const moreRepliesBorderColor = this.props.node.depth
@@ -248,8 +253,6 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
       !this.state.collapsed &&
       node.children.length === 0 &&
       child_count > 0;
-    const deleted = this.props.node.comment_view.comment.deleted;
-    const removed = this.props.node.comment_view.comment.removed;
 
     return (
       <li className="comment list-unstyled">
@@ -285,7 +288,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 myUserInfo={this.props.myUserInfo}
               />
 
-              {cv.comment.distinguished && (
+              {distinguished && (
                 <Icon icon="shield" inline classes="text-danger ms-1" />
               )}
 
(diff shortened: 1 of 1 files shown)


src/shared/components/comment/comment-node.tsx
```diff
@@ -332,6 +335,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                   <Icon icon="lock" classes="icon-inline text-danger" />
                 </span>
               )}
+              {deleted_or_removed && <Icon icon="trash" classes="ms-1" />}
```
dessalines: Why not just copy the block right above this instead of writing from scratch. There's an example with the correct classes, and a tippy right above this.

src/shared/components/comment/comment-node.tsx
```diff
@@ -248,6 +248,9 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
       !this.state.collapsed &&
       node.children.length === 0 &&
       child_count > 0;
+    const deleted_or_removed =
```
dessalines: Use camelCase in typescript

Nutomic: Damn it every time.

src/shared/components/comment/comment-node.tsx
Nutomic: Made it more consistent and updated the screenshot above. Using red icon seemed to strong so I removed that. Also marked the reply button as disabled if the comment is deleted or locked.

src/shared/components/comment/comment-node.tsx
dessalines: Haha its okay, rust is the only language which does case correctly IMO.

src/shared/components/comment/comment-node.tsx
```diff
@@ -326,10 +328,20 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
+                </span>
+              )}
+              {(deleted || removed) && (
+                <span
+                  className="mx-1"
+                  data-tippy-content={I18NextService.i18n.t(
+                    deleted ? "deleted" : "removed",
+                  )}
+                >
+                  <Icon icon="trash" classes="icon-inline" />
```
dessalines: I believe we use the `x` icon with `text-danger` for removed, so this might be a little cleaner as two separate blocks.

Up to you though

Nutomic: Okay updated but left it with the grey icon. Should be clear enough from the text.

src/shared/components/comment/comment-node.tsx
```diff
@@ -341,7 +341,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                     deleted ? "deleted" : "removed",
                   )}
                 >
-                  <Icon icon="trash" classes="icon-inline" />
+                  <Icon icon={deleted ? "trash" : "x"} classes="icon-inline" />
```
dessalines: You'll also need to conditionally add `text-danger`. This would be much easier as two separate blocks anyway.

SleeplessOne1917 (approved):

src/shared/components/comment/comment-node.tsx
Nutomic: In which case would you use text-danger? To me it looks better with the grey icon because red draws too much attention to something that cannot be interacted with. And anyway theres the text indicating that it was removed/deleted.

dessalines: A red X is what we've been using to show removed items. I'll do a PR to this one shortly.

dessalines:
https://github.com/LemmyNet/lemmy-ui/pull/3501

PR #3499, opened 2025-09-29
Show sidebar on create post page (fixes #1522)

### LINKED ISSUE #1522: Show rules when creating a post
### Requirements

In the post creation form, there should be an easy way to see the communtiy's rules. This is helpful when the rules are easy to forget, and it makes rule violations less likely.

Nutomic:
Desktop:
Screenshot_20250929_125212

Mobile:
Screenshot_20250929_125230

By the way sidebar text is not being rendered currently, probably requires update of js-client.

### DIFF
--- a/src/shared/components/community/sidebar.tsx
+++ b/src/shared/components/community/sidebar.tsx
@@ -43,6 +43,7 @@ interface SidebarProps {
   enableNsfw?: boolean;
   showIcon?: boolean;
   editable?: boolean;
+  hideButtons?: boolean;
   myUserInfo: MyUserInfo | undefined;
   onDeleteCommunity(form: DeleteCommunity): void;
   onRemoveCommunity(form: RemoveCommunity): void;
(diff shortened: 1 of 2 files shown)


src/shared/components/post/create-post.tsx
```diff
@@ -253,15 +254,17 @@ export class CreatePost extends Component<
...
+      : undefined;
     return (
       <div className="create-post container-lg">
         <HtmlTags
           title={this.documentTitle}
           path={this.context.router.route.match.url}
         />
         <div className="row">
-          <div id="createPostForm" className="col-12 col-lg-6 offset-lg-3 mb-4">
+          <div id="createPostForm" className="col-12 col-lg-6 offset-lg-2 mb-4">
```
Nutomic, line 267: Not sure whats the best way to center this horizontally.

src/shared/components/post/create-post.tsx
```diff
@@ -293,6 +296,24 @@ export class CreatePost extends Component<
...
+              {I18NextService.i18n.t("sidebar")}{" "}
+              <Icon
+                icon={
+                  this.state.showSidebarMobile ? `minus-square` : `plus-square`
+                }
+                classes="icon-inline"
+              />
+            </button>
+            {this.state.showSidebarMobile && this.sidebar()}
+          </div>
```
Nutomic, line 313: This shows the sidebar button below all the input fields which is easy to miss. Another option would be to put sidebar button next to the community selector inside PostForm but thats a bit more complicated to implement.

src/shared/components/post/create-post.tsx
```diff
@@ -411,4 +432,34 @@ export class CreatePost extends Component<
...
+        <Sidebar
+          community_view={this.state.selectedCommunity}
+          moderators={[]} // TODO: fetch GetCommunityResponse?
+          admins={this.isoData.siteRes.admins}
+          enableNsfw={enableNsfw(this.isoData.siteRes)}
+          showIcon
+          allLanguages={this.isoData.siteRes.all_languages}
+          siteLanguages={this.isoData.siteRes.discussion_languages}
+          myUserInfo={this.isoData.myUserInfo}
+          hideButtons={true}
```
dessalines, line 452: You can just put `hideButtons`

src/shared/components/post/create-post.tsx
dessalines: We really shouldn't even be bothering with offsets, and just use a simple `container` instead of `container-lg` . 

I wouldn't worry about this for now, what you have is good.

src/shared/components/post/create-post.tsx
dessalines: What you have is good, and I can't think of any cleaner way to do it for mobile. 

On the Home and Community pages, we put all these buttons at the top, but it seems weird to do so before a community is picked.

src/shared/components/community/sidebar.tsx
```diff
@@ -404,6 +413,7 @@ export class Sidebar extends Component<SidebarProps, SidebarState> {
 
   sidebarMarkdown() {
     const { sidebar } = this.props.community_view.community;
+    console.log(sidebar);
```
dessalines: Can remove.

dessalines (approved):
Feel free to merge whenever.

src/shared/components/post/create-post.tsx
Nutomic: https://github.com/LemmyNet/lemmy-ui/pull/3504

src/shared/components/community/sidebar.tsx

src/shared/components/post/create-post.tsx
Nutomic: It looks slightly odd with different margins on both sides. But thats probably better than moving the form around when a community gets added and the sidebar gets rendered.

src/shared/components/post/create-post.tsx
Nutomic: It would be possible to put the sidebar button next to the community selection dropdown. But that means passing `sidebar()` into `PostForm`, or making another duplicate of that HTML.

PR #3507, opened 2025-10-02
Check for untranslated error messages (fixes #3478)

### LINKED ISSUE #3478: Check that translation exists for every backend error
### Requirements

This basically means iterating over `LemmyErrorType` (how?) and then checking that a translation string with this name exists. It would also make sense to prefix all of these translations with `error_`.

Edit: It seems that this is [impossible](https://stackoverflow.com/questions/40863488/how-can-i-iterate-over-a-custom-literal-type-in-typescript) with the way LemmyErrorType is represented in lemmy-js-client. And ts-rs doesnt seem to have any option to generate a different enum format (https://github.com/Aleph-Alpha/ts-rs/issues/23).

Nutomic:
Error array added in https://github.com/LemmyNet/lemmy-js-client/pull/720

New errors strings added in https://github.com/LemmyNet/lemmy-translations/pull/207.

The check is currently failing because of these changes: https://github.com/LemmyNet/lemmy/pull/6036

### DIFF
--- a/generate_translations.js
+++ b/generate_translations.js
@@ -62,7 +62,6 @@ fs.readFile(`${translationDir}${baseLanguage}.json`, "utf8", (_, fileStr) => {
 
   const translationKeys = entries.map(e => e[0]);
   let missingErrorTranslations = false;
-  console.log(lemmyjsclient);
   lemmyjsclient.AllLemmyErrors.forEach(e => {
     if (!translationKeys.includes(e)) {
       missingErrorTranslations = true;
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "inferno-router": "^9.0.4",
     "inferno-server": "^9.0.4",
     "jwt-decode": "^4.0.0",
-    "lemmy-js-client": "1.0.0-api-updates-4.1",
+    "lemmy-js-client": "1.0.0-export-regex.0",
     "markdown-it": "^14.1.0",
     "markdown-it-bidi": "^0.2.0",
     "markdown-it-container": "^4.0.0",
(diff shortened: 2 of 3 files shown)


src/shared/services/I18NextService.ts
```diff
@@ -114,6 +115,21 @@ export async function verifyTranslationImports(): Promise<ImportReport> {
...
+    name: "",
+  });
+  const translation_keys = Object.keys(en["translation"]);
+  AllLemmyErrors.forEach(e => {
+    if (translation_keys.includes(e)) {
+      report.success.push(e);
+    } else {
+      report.error.push({ id: e, error: "Missing error translation" });
+    }
+  });
```
Nutomic: Looks like CI doesnt run this. Putting it into `generate_translations.js` also doesnt work because I cant import from lemmy-js-client there.

dessalines: Do `const lemmyjsclient = require("lemmy-js-client");` in the generate translations. Then you should be able to use it.

Nutomic: Thanks fixed

dessalines (approved):
Nice, this is going to be handy, as we can now just `toast(i18n.t(error))`.

generate_translations.js
```diff
@@ -58,6 +60,18 @@ fs.readFile(`${translationDir}${baseLanguage}.json`, "utf8", (_, fileStr) => {
     }
   }
 
+  const translationKeys = entries.map(e => e[0]);
+  var missingErrorTranslations = false;
```
SleeplessOne1917: `let`, not `var`.
```suggestion
  let missingErrorTranslations = false;
```

SleeplessOne1917 (approved):

Nutomic:
This one also requires updated js client as some errors were moved to untranslated.

Nutomic:
Updated, ready to merge.

https://github.com/LemmyNet/lemmy-translations/pull/218

https://github.com/LemmyNet/lemmy-js-client/pull/757

PR #3511, opened 2025-10-02
Use btn-sm for post and comment actions buttons.

### LINKED ISSUE #3503: Increase size of smaller UI elements on mobile
### Requirements

The buttons below posts (and other similar ones of this size, these are just the ones I notice it the most) are really hard to hit with a finger. 

In this image I have marked roughly how much of the screen my finger obstructs when I try pressing something: 

Image

As you can see, the top button is perfectly fine, the target area of the button is roughly half the size of my finger, plenty of space to hit it accurately. However, the bottom buttons are somewhere between 1/4th and 1/5th the size of my finger. I miss these all the time.

I think part of the issue might be that most of these buttons have the CSS class "btn-sm" (except the "link" button) which gives them no vertical padding, making their bounding box smaller than the line they are on. See:

Image

However, even with that fixed, I think they could/should be around 1.25 to 1.5 times bigger, at least so my finger does not completely obscure multiple buttons at once.

dessalines:
- Fixes #3503

Screenshot_2025-10-02-15-25-19-643_mark via

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -409,7 +409,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                   {this.props.showContext && this.getLinkButton()}
                   {this.props.markable && (
                     <button
-                      className="btn btn-link btn-animate text-muted"
+                      className="btn btn-sm btn-link btn-animate text-muted"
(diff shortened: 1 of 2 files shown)


Nutomic:
This seems like the opposite of what was requested in the issue. Now all buttons have an even smaller clickable area.

Issue #757, opened 2022-08-19
Cardview


Kap2022:
https://lemmy.ml/post/428464

A card view would make the Lemmy experience on a cell phone so much better and a lot more like the experience encountered on other sites.

At the moment on my cell phone, the pictures are too small. think how much better Memes would be if you could see the picture as you scroll through. But also I think it would generate a lot more user engagement overall.

If you do a cardview, make sure you put the headline under the picture. There has been a lot of studies done to look at how people read content. They focus on the picture and read the headline underneath the picture more frequently then they read the headline if it is above the picture.

dessalines:
Will probably be a while before I get to this.

Nutomic:
This would only make the experience better if the content consists mostly of images. I would consider Lemmy more like a forum, which is primarily based on text, so I wouldnt want such cards. I suppose it could be optional though.

Kap2022:
I noticed Hexbear are porting some of their stuff up stream.

When you visit their site they have the images displayed larger then normal Lemmy.

dessalines:
Yep I see this as being a radio switch with `(Card, List)` , maybe to the right of sorting. Although we're running low on real estate there.

Kap2022:
That would work, but it would be good if the admin could set the default for the site - Card or List - in the Admin Settings.

The reason is visitors who are new to Lemmy might not know where the setting is and not hang around long enough to find out if it visually appealing to them.

Cris-Edmundson:
Just wanted to voice I would also like this, since this thread hasn't been active for quite a while.

Thanks for all the work you do on lemmy!

dessalines:
Should come along with #3428

PR #3512, opened 2025-10-03
Adding card and small card views

dessalines:
- Also cleaning up the post-listing images a lot.
- Fixes https://github.com/LemmyNet/lemmy-ui/issues/3428
- Fixes https://github.com/LemmyNet/lemmy-ui/issues/757

https://github.com/user-attachments/assets/42709600-08b6-4290-92d9-8cc95d32dcbe

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -768,8 +768,15 @@ export class Community extends Component<CommunityRouteProps, State> {
     this.updateUrl({ dataType, cursor: undefined });
   }
 
-  handlePostListingModeChange(val: PostListingMode) {
+  async handlePostListingModeChange(val: PostListingMode) {
     this.setState({ postListingMode: val });
+
+    // Also, save your user settings to this mode
+    if (this.isoData.myUserInfo) {
+      await HttpService.client.saveUserSettings({
+        post_listing_mode: val,
+      });
+    }
   }
 
   handleShowHiddenChange(show?: StringBoolean) {
--- a/src/shared/components/home/home.tsx
+++ b/src/shared/components/home/home.tsx
@@ -948,8 +948,15 @@ export class Home extends Component<HomeRouteProps, HomeState> {
     this.updateUrl({ listingType: val, cursor: undefined });
   }
 
-  handlePostListingModeChange(val: PostListingMode) {
+  async handlePostListingModeChange(val: PostListingMode) {
     this.setState({ postListingMode: val });
+
+    // Also, save your user settings to this mode
+    if (this.isoData.myUserInfo) {
+      await HttpService.client.saveUserSettings({
+        post_listing_mode: val,
+      });
+    }
   }
 
   handleDataTypeChange(val: DataType) {


Nutomic:
- There are some errors about missing translation keys
- View mode selection should use icon buttons instead of dropdown
- Should reduce horizontal margin for vote buttons in card/small_card mode to have more space for text and actions
- Card view wastes a lot of space if there are posts with image next to posts without. In that case the item below should be moved up into the empty space:
Screenshot_20251007_120904
- The only difference between card and small_card is that card expands all images? I would make it so that card only uses a single column so that images are larger, and it can be used as a gallery view

dessalines:
> View mode selection should use icon buttons instead of dropdown

Are you referring to how this looks?

Screenshot_2025-10-07-15-15-33-354_mark via-edit

I'd rather not add yet another distinct select representation. Lets stick with things we have, or make them all icon buttons. If the latter, it should be a separate issue.

> Card view wastes a lot of space if there are posts with image next to posts without. In that case the item below should be moved up into the empty space:

The only way around that is to use a [masonry like layout](https://getbootstrap.com/docs/5.3/examples/masonry/) I wouldn't be opposed to it.

> The only difference between card and small_card is that card expands all images? 

Yep.

> I would make it so that card only uses a single column so that images are larger, and it can be used as a gallery view

With wide screens, this ends up being a nightmare, because most images then will be too tall for a single screen. We have this issue in jerboa also, here's what it ends up looking like:

Screenshot_2025-10-07-15-10-07-413_com jerboa-edit

In general, the wider the screen, the more columns you should add... but I'll try the masonry approach.

dessalines:
Okay I've added showing the post body when its a card, and a masonry layout:


https://github.com/user-attachments/assets/686c6f16-c7b1-4a54-908c-948d16e00d6f

Nutomic:
> > View mode selection should use icon buttons instead of dropdown
> 
> Are you referring to how this looks?
> 
> Screenshot_2025-10-07-15-15-33-354_mark via-edit
> 
> I'd rather not add yet another distinct select representation. Lets stick with things we have, or make them all icon buttons. If the latter, it should be a separate issue.

Yes this would be better. 

The masonry layout looks good. I checked how it looks on mobile, and there images are not expanded, all the modes are almost identical. Maybe its a bug, I cant even expand images by clicking the thumbnail with mobile.

Another thing, card view should have some separator between items, or a proper card background. Right now its hard to see where one card ends and the next one begins.

Screenshot_20251008_151753

Alignment for post body is wrong. Either make it narrower to match the image and buttons, or make the image/buttons wider.

Screenshot_20251008_151934


In card/small_card mode, very long titles take up to 9 lines. These should be shortened to take 1 or 2 lines at most.
 
Edit: Also it would make sense that changing the listing mode gets directly stored in user settings so it is remembered after reload. This would also make sense for other sort options so you dont have to go to the profile to make changes.

Nutomic:
Heres a problem I noticed on the main branch, but it even more noticable here.

main:
Screenshot_20251010_160500

this pr:
Screenshot_20251010_162331

- Vote buttons should be centered horizontally (blue)
- Thumbnail height should match height of the text and buttons to the right

dessalines:
I think I'll have to do a heavy rework of the existing card and list views.

I'm going to try to match jerboa pretty closely. I was hoping to hold off this work for leptos, but I think its unavoidable. Might take me a while.

dessalines:
Finally finished this up. Made it to match jerboa / boost a bit.

- Refactored into proper functional components. 
- Split everything up and cleaned up post-listings massively. 
- Got rid of pointless `d-none` mobile-only blocks
- Made two components: `PostListingList` and `PostListingCard`.
- Added a `ShowBody: "Hidden | "Preview" | "Full"` for displaying a faded out preview on the home / community page.

The only work that needs to be done, is to clean up the actions bar, but saving that for #3527 as this is already large enough.

# Desktop view

https://github.com/user-attachments/assets/eb2e0b60-63b5-4a7c-bc22-53715b2add43

# Phone view

https://github.com/user-attachments/assets/8428e094-620d-4c09-8059-5daa9b9e6123

Nutomic:
- Moving thumbnails to the right looks very odd, I would prefer to keep it on the left like before
- Card views should also keep vote buttons on the left. Moving them to the action bar results in overflow:
Screenshot_20251016_145803

- Loading skeleton doesnt match the actual views in card mode:
Screenshot_20251016_150010

- The line break in card view on username and community doesnt look good. Would be better to move date and language next to the link domain to save a line
Screenshot_20251016_150139

- Behaviour for clicking the title of a post with external link has changed. With 0.19 it opens the comments, but with this change it opens the external link instead.

dessalines:
The left-aligned images I'm doing away with. I'm making this match jerboa / boost, which are UIs I really like.

The post action bar I'm leaving for the linked issue, but I'm going to make it match jerboa also (left aligned comments, right-aligned actions), and moving most of the actions into the dropdown. That'll be much cleaner.

The loading skeleton I can't do much about, because its using bootstrap's placeholder-glow. I don't even know how it'd work with a masonry style layout, but that should be left to another issue if someone wants to try it out.

For the line breaks:

- I removed the *ago* text to match jerboa, so now instead of 10 days ago, it just says 10 days. I'd rather it be `10d`, but I didn't see a way to do that easily with date-fns
- I hid the person avatar, which saves some space (and most apps seem to do this anyway). You'll still see the avatar on comments and everywhere else, just not the `PostCreatedLine`. I still show the community avatar, and also show the person avatar for community.tsx.
- I also decreased some margins. I looked at both jerboa and boost, and the line wraps for long names are fairly unavoidable. I used to try to do a 2-row layout for the title in jerboa, but it didn't look as clean, so I went back to a single row.
- I moved the postbadges to next to the url line also.

I fixed the link behavior so that it only goes external if its either media, or its on the post screen (IE showBody === Full).

Nutomic (approved):
Okay looks good. Im merging this directly because we still have lots of other major features to add. The polishing can be done later when we are closer to release, via separate issues/pull requests.

PR #3522, opened 2025-10-07
Change format for inserted person/community links (fixes #1743)

### LINKED ISSUE #1743: Two nested <a> tags created for autocompleted community URLs
### Requirements

### Summary

When typing a comment and using the !community@instance shortcut to let the editor automatically create a link, the result is something like this: 
image

When you post that comment (`[!dota2@lemm.ee](https://lemm.ee/c/dota2)`), the resulting DOM looks like this: 
image

The actual HTML rendered is `<a href="https://lemm.ee/c/dota2"><a href="/c/dota2@lemm.ee" class="community-link">!dota2@lemm.ee</a></a>`, so nested `<a>` tags, which is invalid HTML, `<a>` tags are not allowed to contain other `<a>` tags.

I would assume that the code that automatically creates a link out of the text `!dota2@lemm.ee` (without a `[]()` link around it) is trying to make a new link even though the text is already linked.

See [this test post](https://lemm.ee/post/564081):

* `[!dota2@lemm.ee](https://lemm.ee/c/dota2)` creates two nested `<a>`s
* `!dota2@lemm.ee` creates only one `<a>`

### Steps to Reproduce

1. create a comment/post with the body `[!dota2@lemm.ee](https://lemm.ee/c/dota2)`

### Technical Details

Windows, Firefox 115.0b9

### Lemmy Instance Version

UI: 0.18.1-rc.7, BE: 0.18.1-rc.4

### Lemmy Instance URL

https://lemm.ee

Nutomic:
There is no need for the markdown link as lemmy-ui also renders a link for simply `!name@example.com`

### DIFF
--- a/src/shared/utils/markdown.ts
+++ b/src/shared/utils/markdown.ts
@@ -349,7 +349,7 @@ export async function setupTribute() {
         trigger: "@",
         selectTemplate: (item: any) => {
           const it: PersonTribute = item.original;
-          return `[${it.key}](${it.view.person.ap_id})`;
+          return it.key;
         },
         values: debounce(async (text: string, cb: any) => {
           cb(await personSearch(text));
@@ -366,7 +366,7 @@ export async function setupTribute() {
         trigger: "!",
         selectTemplate: (item: any) => {
           const it: CommunityTribute = item.original;
-          return `[${it.key}](${it.view.community.ap_id})`;
+          return it.key;
         },
         values: debounce(async (text: string, cb: any) => {
           cb(await communitySearch(text));


dessalines (commented):
Wait, lemmy-ui does, but that doesn't guarantee any other markdown platform or renderer will.

These are inserted, so they should probably use the markdown.

Nutomic:
People already use this feature because it is supported on lemmy-ui. This means other frontends and apps already need to support it. So the insert button should also use this format by default, both to make the markdown easier to read and to avoid generating invalid html like described in the issue.

dessalines:
Alright.

PR #3525, opened 2025-10-08
Consistent size for action buttons (ref #3503)

Nutomic:
Dont use `btn-sm` but instead use `py-0` for all action buttons (this was previously inconsistent). Previously post action buttons were 31 or 23px high. Now post buttons are rendered larger and use less vertical space with 23 or 26px (its difficult to make them all the exact same height).

### DIFF
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -804,7 +804,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
         {this.commentsButton}
         {canShare() && (
           <button
-            className="btn btn-sm btn-link btn-animate text-muted py-0"
+            className="btn btn-link btn-animate text-muted py-0"
             onClick={linkEvent(this, this.handleShare)}
             type="button"
           >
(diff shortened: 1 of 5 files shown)


src/shared/components/post/post-listing.tsx
```diff
@@ -977,7 +977,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     return (
       <button
         type="button"
-        className="btn btn-sm btn-link text-body link-opacity-75 link-opacity-100-hover py-0 align-baseline"
+        className="btn btn-link link-dark link-opacity-75 link-opacity-100-hover align-baseline py-0"
```
dessalines: link-dark got re-added here.

Nutomic:
Before:

Screenshot_20251008_145455

After:
Screenshot_20251008_145441

dessalines:
I prefer the small ones, the larger ones are too spaced out.

Honestly though we should a lot of these to the dropdown menu anyway, especially the share links. #3527

Lets let @SleeplessOne1917 be the tie-breaker.

SleeplessOne1917:
I like the after version better. It can be hard to press tiny buttons on mobile.

Also worth noting that the after picture looks more spaced out because the share button isn't in the before picture for some reason.

PR #3528, opened 2025-10-08
Making admins its own panel in the admin settings.

### LINKED ISSUE #3460: Move Admin list to Users tab on admin-setting
### Requirements

See title, it makes sense to have this on the users tab, then more space can be used for the site settings form, cause there's a lot now.

dessalines:
- This couldn't use the same users-table, since one uses a `localUserView`, and the other a `personView`
- Fixes #3460

## Screenshots

Screenshot_2025-10-08-11-40-50-219_mark via

### DIFF
--- a/src/shared/components/home/admin-settings.tsx
+++ b/src/shared/components/home/admin-settings.tsx
@@ -222,7 +222,7 @@ export class AdminSettings extends Component<
                     {I18NextService.i18n.t("site_config")}
                   </h1>
                   <div className="row">
-                    <div className="col-12 col-md-6">
+                    <div className="col-12">
                       <SiteForm
                         showLocal={showLocal(this.isoData)}
                         onSaveSite={this.handleEditSite}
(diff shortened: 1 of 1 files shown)


Nutomic:
I dont think the admin list really needs its own tab with a full table. It would be enough to show a simle list of admin names on the users tab either above or below the table. That way we avoid having too many admin tabs, 8 or 9 is already a lot especially on mobile.

Also the site tab doesnt look good on desktop as it is stretched so wide. Would be neat if it could be shown as two columns instead.

Another option would be to get rid of the navbar at the top, and instead use a vertical nav on the left side. That works better to show many tabs, and also uses the horizontal space on desktop better.

dessalines:
Alright, I'll move it below the users on the same tab, but I'd like it to be a table.

I've center justified the site tab and made it `col-md-6 for md and over screens:

Screenshot_2025-10-10-11-39-26-637_mark via

PR #3529, opened 2025-10-09
Disable comment form if post is removed or deleted

Nutomic:


src/shared/components/post/post.tsx
```diff
@@ -966,7 +974,11 @@ export class Post extends Component<PostRouteProps, PostState> {
             community={postRes.data.community_view.community}
             viewType={this.props.view}
             maxCommentsShown={this.state.maxCommentsShown}
-            postLocked={postRes.data.post_view.post.locked}
+            postLocked={
+              postRes.data.post_view.post.locked ||
+              postRes.data.post_view.post.removed ||
+              postRes.data.post_view.post.deleted
+            }
```
dessalines: Rename this field to something like

`postLockedOrRemovedOrDeleted`

dessalines:
Is there an issue about this anywhere? We should try to open one if it doesn't exist, then close it with the PR.

Nutomic:
Renamed the prop and added helper function. Dont see any reason to open an issue for something so minor.

PR #3530, opened 2025-10-09
Fixup instance_blocks.tsx

### LINKED ISSUE #3453: Fixup `instance_blocks.tsx` tab
### Requirements

### Summary

As part of #3440 , I fixed the custom emojis and taglines to use proper pushing up of events rather than doing API calls inside components, but this still needs to be done for instance blocks

### Steps to Reproduce

See above

### Technical Details

NA

### Lemmy Instance Version

main

### Lemmy Instance URL

main

dessalines:
## Description

- This no longer does internal http calls.
- Uses inline forms which add a reason and optional expiration time for
blocks.
- Extracts the instance table used in instances.tsx, so that it can be used for these lists.
- Fixes #3453

Caveats: This doesn't handle unblock / disallow reasons, which are required in the API now for some reason. That would require a modal or something that can be added later.

## Screenshots

https://github.com/user-attachments/assets/9b448d01-d5be-4a1c-946f-f1205f5cacb2

### DIFF
--- a/src/shared/components/common/blocking-keywords-textarea.tsx
+++ b/src/shared/components/common/blocking-keywords-textarea.tsx
@@ -1,0 +1,61 @@
+import { linkEvent, Component } from "inferno";
+import { I18NextService } from "../../services/I18NextService";
+
+interface Props {
+  keywords: string[];
+  onUpdate(keywords: string[]): void;
+}
+
+interface State {
+  text: string;
+}
+
+function handleTextChange(i: BlockingKeywordsTextArea, event: any) {
+  i.setState({ text: event.target.value });
+}
+
+function handleTextBlur(i: BlockingKeywordsTextArea, _event: any) {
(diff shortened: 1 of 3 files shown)


Nutomic:
Expiration time for blocks should also be shown in the list. With https://github.com/LemmyNet/lemmy/pull/6054 it will be easier to make reason optional for reverts while keeping it mandatory for the blocking itself. Question is which actions should require a reason and which not.

dessalines:
> Expiration time for blocks should also be shown in the list.

I'm going to need to open up a back-end issue for this, because we currently don't have that info, except via the modlog.

Its hard to know which actions should require a reason, and which shouldn't. Considering all the `removeComment` and such have a required reason, we should probably just keep it required.

dessalines:
Lets leave expiration date for another time. Its still visible via the modlog anyway.

PR #3544, opened 2025-10-17
Add blurhash image using unlazy.

### LINKED ISSUE #3523: Add blur-hash from `image_details` to have a blurred image on preload
### Requirements

Take advantage of lemmy's new blurhash feature, added to `image_details`

dessalines:
This uses the library [unlazy](https://unlazy.byjohann.dev/guide/) to load load lazy placeholders from blurhash strings returned by the `postView.image_details` response.

~~I've tried several libraries, including [fast-blurhash](https://github.com/mad-gooze/fast-blurhash?tab=readme-ov-file), and client-side, they all take at least 500ms to convert a blurhash string to the base64 image string.~~

~~@asonix looked into mastodon, and found that they're likely doing it client-side also, which means they suffer from the same problem.~~ 

~~We can either accept that cost, or a better solution might be :~~ 

~~- Since we already store the `blurhash` string in our `image_details` table, then we could just calculate the base64 in the rust back-end, save it to the `image_details` table, then return it directly.~~ 

~~If anyone else thinks that's a good idea, then I'll open up an issue in lemmy's back-end for that. If its not worth it, then this is as good as we can do.~~

Fixes #3523 

~~EDIT: changed my mind, going to add the base64 saving into the lemmy back-end. Now front ends won't have to do this work.~~

https://github.com/LemmyNet/lemmy/pull/6091

# Example:

https://github.com/user-attachments/assets/401d8efc-88ed-4b51-9baf-e1698d649dfa

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -61,7 +61,7 @@ steps:
         from_secret: docker_username
       password:
         from_secret: docker_password
-      platforms: linux/amd64, linux/arm64
+      platforms: linux/amd64,linux/arm64
       tag: ${CI_COMMIT_TAG}
     when:
       - event: tag
@@ -75,7 +75,7 @@ steps:
         from_secret: docker_username
       password:
         from_secret: docker_password
-      platforms: linux/amd64, linux/arm64
+      platforms: linux/amd64,linux/arm64
       tag: dev
     when:
       - event: cron
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,13 +9,13 @@ ARG TARGETARCH
 # - libvips binaries are not available for ARM32
 # - It can break depending on the CPU (https://github.com/LemmyNet/lemmy-ui/issues/1566)
 RUN \
-    --mount=type=cache,target=/var/cache/apk,id=apk-${TARGETARCH},sharing=private \
-    set -x && apk update && apk upgrade && apk add curl python3 build-base gcc wget git vips-dev pkgconfig
+  --mount=type=cache,target=/var/cache/apk,id=apk-${TARGETARCH},sharing=private \
+  set -x && apk update && apk upgrade && apk add curl python3 build-base gcc wget git vips-dev pkgconfig py3-pip make g++
 
 # Install node-gyp and corepack
 RUN \
-    --mount=type=cache,target=/root/.npm \
-    npm install -g node-gyp corepack
+  --mount=type=cache,target=/root/.npm \
+  npm install -g -f node-gyp corepack
 
 # Enable corepack to use pnpm
 RUN corepack enable
@@ -28,8 +28,8 @@ ENV npm_config_target_libc=musl
 # Cache deps
 COPY package.json pnpm-lock.yaml ./
 RUN \
-    --mount=type=cache,target=/root/.local/share/pnpm/store \
-    pnpm i
+  --mount=type=cache,target=/root/.local/share/pnpm/store \
+  pnpm i
 # Build
 COPY generate_translations.js \
   tsconfig.json \
@@ -46,14 +46,14 @@ RUN echo "export const VERSION = '$(git describe --tag)';" > "src/shared/version
 RUN echo "export const BUILD_DATE_ISO8601 = '$(date -u +"%Y-%m-%dT%H:%M:%SZ")';" > "src/shared/build-date.ts"
 
 RUN \
-    --mount=type=cache,target=/root/.local/share/pnpm/store \
-    pnpm i
+  --mount=type=cache,target=/root/.local/share/pnpm/store \
+  pnpm i
 RUN \
-    --mount=type=cache,target=/root/.local/share/pnpm/store \
-    pnpm prebuild:prod
+  --mount=type=cache,target=/root/.local/share/pnpm/store \
+  pnpm prebuild:prod
 RUN \
-    --mount=type=cache,target=/root/.local/share/pnpm/store \
-    pnpm build:prod
+  --mount=type=cache,target=/root/.local/share/pnpm/store \
+  pnpm build:prod
 
 RUN rm -rf ./node_modules/import-sort-parser-typescript
 RUN rm -rf ./node_modules/typescript
@@ -66,8 +66,8 @@ ARG TARGETARCH
 ENV NODE_ENV=production
 
 RUN \
-    --mount=type=cache,target=/var/cache/apk,id=apk-${TARGETARCH},sharing=private \
-    apk update && apk add curl vips-cpp
+  --mount=type=cache,target=/var/cache/apk,id=apk-${TARGETARCH},sharing=private \
+  apk update && apk add curl vips-cpp
 
 COPY --from=builder --chown=node:node /usr/src/app/dist /app/dist
 COPY --from=builder --chown=node:node /usr/src/app/node_modules /app/node_modules
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lemmy-ui",
-  "version": "1.0.0-alpha.9",
+  "version": "1.0.0-alpha.11.1",
   "description": "An isomorphic UI for lemmy",
   "repository": "https://github.com/LemmyNet/lemmy-ui",
   "license": "AGPL-3.0",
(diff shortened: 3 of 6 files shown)


src/shared/components/app/app.tsx
```diff
@@ -1,3 +1,5 @@
+// @ts-expect-error has a weird import error
+import { lazyLoad } from "unlazy";
```
dessalines: See https://github.com/johannschopplich/unlazy/issues/66

dessalines:
Turns out I didn't need to save a `base64` in the back-end. 

I found a workaround for the [slow blurhash computation](https://github.com/johannschopplich/unlazy/issues/67) issue.

matc-pub:
I'm not sure how useful this is in SSR. The blur images are uncompressed bitmaps encoded in base64. When I save one as image, it's about 4kb, which is almost the size of a normal 96px thumbnail. This still saves on round trip times, but at up to 20 blur images per page, it can also almost double the size of the initial page load. Home on lemmy.ml is about 50kb compressed. The base64 data uris only compress at about 50%, and Firefox takes for 1000 blur images at 32px about 400ms.

This approach could solve these issues, as long as unlazy runs early enough (https://unlazy.byjohann.dev/placeholders/hash-based.html#client-side-decoding):
```html
<img
  data-src="image.jpg"
  data-blurhash="LKO2:N%2Tw=w]~RBVZRi};RPxuwH"
>
```

Another solution would be to use 1px (2 to 8px have color banding issues in chromium) in SSR, and let CSR update it with the 32px version, or even manually update the blur images before hydration.

Times measured in Firefox for 1000 createPngDataUri calls and data uri lengths:
```
1000x blurhash: 64 px,        uri length: 22466, 1603.92ms
1000x blurhash: 32 px,        uri length:  5826,  414.86ms
1000x blurhash: 16 px,        uri length:  1602,  117.02ms
1000x blurhash:  8 px,        uri length:   514,    33.8ms
1000x blurhash:  4 px,        uri length:   226,   12.14ms
1000x blurhash:  2 px,        uri length:   146,    5.94ms
1000x blurhash:  1 px,        uri length:   122,    4.82ms
1000x blurhash: undefined px, uri length:  5826,   433.3ms

```

```javascript
import { createPngDataUri } from "unlazy/blurhash";
window.onload = () => {
  const blurhash = "LEHV6nWB2yk8pyo0adR*.7kCMdnj";
  const limit = 1000;
  for (const size of [undefined, 1, 2, 4, 8, 16, 32, 64].reverse()) {
    const uri = createPngDataUri(blurhash, { size: size });
    const label = `${limit}x blurhash: ${size} px, uri length: ${uri.length}`;
    console.time(label);
    for (let x = 0; x < limit; x++) {
      createPngDataUri(blurhash, { size: size });
    }
    console.timeEnd(label);
  }
};
```

File sizes for 32px as png and text document with data uri:
```
# du -b *blur*
4351    32blur.png
5827    32blur.uri
2786    32blur.png.gz
3752    32blur.uri.gz
```

matc-pub:
Looks like webpack can generate a third script, that `createSsrHtml` can read from disk and write beside isoData. The new script with just `loadLazy` is about 3.5kb compressed.

I didn't really test this, only that it ends up running early enough:

```diff
diff --git a/src/embedded/index.ts b/src/embedded/index.ts
new file mode 100644
index 00000000..badf921f
--- /dev/null
+++ b/src/embedded/index.ts
@@ -0,0 +1,4 @@
+// @ts-expect-error has a weird import error
+import { lazyLoad } from "unlazy";
+
+lazyLoad();
diff --git a/src/server/utils/create-ssr-html.tsx b/src/server/utils/create-ssr-html.tsx
index 1d5212db..e4ed2181 100644
--- a/src/server/utils/create-ssr-html.tsx
+++ b/src/server/utils/create-ssr-html.tsx
@@ -8,10 +8,14 @@ import { IsoDataOptionalSite } from "@utils/types";
 import { buildThemeList } from "./build-themes-list";
 import { fetchIconPng } from "./fetch-icon-png";
 import { findLanguageChunkNames } from "@services/I18NextService";
+import path from "path";

 const customHtmlHeader = process.env["LEMMY_UI_CUSTOM_HTML_HEADER"] || "";

 let appleTouchIcon: string | undefined = undefined;
+import { readFileSync } from "node:fs";
+
+const embeddedScript = readFileSync(path.resolve("./dist/js/embedded.js"));

 export async function createSsrHtml(
   root: string,
@@ -86,6 +90,8 @@ export async function createSsrHtml(
     <script nonce="${cspNonce}">
     window.isoData = ${serialize(isoData)};

+    ${embeddedScript}
+
     if (!document.documentElement.hasAttribute("data-bs-theme")) {
       const light = window.matchMedia("(prefers-color-scheme: light)").matches;
       document.documentElement.setAttribute("data-bs-theme", light ? "light" : "dark");
diff --git a/webpack.config.js b/webpack.config.js
index 6f19344d..e995545d 100644
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -138,6 +138,16 @@ module.exports = (env, argv) => {
     ],
   };

+  const embeddedConfig = {
+    ...base,
+    entry: "./src/embedded/index.ts",
+    target: "browserslist", // looks up package.json
+    output: {
+      ...base.output,
+      filename: "js/embedded.js",
+    },
+  };
+
   if (mode === "development") {
     // serverConfig.cache = {
     //   type: "filesystem",
@@ -153,5 +163,5 @@ module.exports = (env, argv) => {
     serverConfig.plugins.push(new BundleAnalyzerPlugin());
   }

-  return [serverConfig, clientConfig];
+  return [serverConfig, clientConfig, embeddedConfig];
 };
```

dessalines:
> Looks like webpack can generate a third script, that createSsrHtml

Could you do a PR with that? Thx.

> This approach could solve these issues, as long as unlazy runs early enough (https://unlazy.byjohann.dev/placeholders/hash-based.html#client-side-decoding):

I tried the `data-blurhash` approach, but in practice it seemed to take a long time before loading the images, whereas the blur-hash base64 blur rendered immediately on page load. I'll take another try at it.

dessalines:
When using the `data-blurhash` / client side rendering approach, I get this:

https://github.com/user-attachments/assets/29c74d02-8cba-4895-8ab5-d82bfb8545e9

I can verify that it is actually loading the blurhash (by temporarily removing data-src), but as you can see we get a weird image not loaded picture when not using the `src` attribute directly.

dessalines:
Using a tiny base64 placeholder for src seems to fix the error image showing:


https://github.com/user-attachments/assets/bc1f8693-3c54-4093-80f2-f60619a323b5

matc-pub:
> we get a weird image not loaded picture when not using the src attribute directly

I believe this is connected to eruda, with eruda in Firefox I get this warning:

```
Layout was forced before the page was fully loaded. If stylesheets are not yet loaded this may cause a flash of unstyled content.
```
The  browser has to figure what to display for those images too early. Eruda probably triggers this, when placing the dev tools button, by reading the current client size.

dessalines:
I tried commenting out eruda (and removing the base64 svg hack I added), but it still shows the same problem. Anyways the hack I added seems to work fine.

src/shared/components/post/post-thumbnail.tsx
```diff
@@ -124,11 +124,16 @@ type ImgThumbProps = {
...
+  const imageDetails = postView.image_details;
+
   return postView.post.thumbnail_url ? (
     <PictrsImage
       src={postView.post.thumbnail_url}
       thumbnail
       alt={postView.post.alt_text}
+      width={imageDetails?.width}
+      height={imageDetails?.height}
+      blurhash={imageDetails?.blurhash}
```
Nutomic: Pass in `imageDetails` as a single parameter, instead of each var separately.

Nutomic (approved):

matc-pub:
I made a PR against this one for the early lazyLoad call.

Another small issue I noticed, is that Firefox briefly flashes the alt-text when the alt-attribute is set.

dessalines:
> Another small issue I noticed, is that Firefox briefly flashes the alt-text when the alt-attribute is set.

We can pry handle that in another issue.

PR #3550, opened 2025-10-21
Clean up post and comment actions bars

### LINKED ISSUE #3527: Reduce number of items in action bars
### Requirements

The number of items in the action bar is too many:

Image

The link ones should be moved to the dropdown menu.

The check mark should also probably use unfilled / filled icons, rather than a green color.

dessalines:
## Description

This PR cleans up both the post and comment action bars, to be more uniform, and match jerboa / boost.

- Decluttered the action bars by moving a lot of actions into the dropdowns.
- You no longer need to be logged in to see the dropdown, and it will correctly show a restricted set of actions (mainly links)
- Consolidated the Vote-Display and Vote-Buttons components, since they're now shown in the same spot.
- Hiding person avatars correctly when they're banned (wherever that data is available).
- Removed the `OP` badge in favor of adding a badge to the OPs name, for comment threads.
- Enforced a better mandatory postCreatorId and community providing for the `CommentViewSlim` variant.

### Showing the new post and comment bars

Screenshot_2025-10-22-16-27-14-262_mark via

### Post bar on the main screen

Screenshot_2025-10-22-16-27-28-536_mark via

### A logged in dropdown (string updates in another PR)

Screenshot_2025-10-22-16-27-42-614_mark via

### Not-logged-in dropdown

Screenshot_2025-10-22-16-28-08-576_com duckduckgo mobile android

Fixes #3527

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -1,4 +1,4 @@
-import { colorList, userNotLoggedInOrBanned } from "@utils/app";
+import { colorList, userNotLoggedInOrBanned, hideImages } from "@utils/app";
 import { numToSI } from "@utils/helpers";
 import { futureDaysToUnixTime } from "@utils/date";
 import classNames from "classnames";
@@ -198,6 +198,11 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
       node.view.children.length === 0 &&
       child_count > 0;
 
+    const hideImages_ = hideImages(
+      this.props.hideImages ?? false,
+      this.props.myUserInfo,
+    );
+
     return (
       <li className="comment list-unstyled">
         <article
@@ -255,7 +260,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
                 <CommentContent
                   comment={comment}
                   viewSource={this.state.viewSource}
-                  hideImages={this.props.hideImages}
+                  hideImages={hideImages_}
                 />
                 <div className="comment-bottom-btns d-flex justify-content-end justify-content-md-start column-gap-1.5 flex-wrap text-muted fw-bold mt-1 align-items-center">
                   <>
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ pnpm install
 LEMMY_UI_BACKEND_REMOTE=voyager.lemmy.ml pnpm dev
 ```
 
-Finally open `http://0.0.0.0:1234` in your browser. This uses the public test instance `https://voyager.lemmy.ml/` as backend. For more details such as developing with a locally compiled Lemmy backend, read the [documentation](https://join-lemmy.org/docs/contributors/01-overview.html).
+Finally open `http://localhost:1234` in your browser. This uses the public test instance `https://voyager.lemmy.ml/` as backend. For more details such as developing with a locally compiled Lemmy backend, read the [documentation](https://join-lemmy.org/docs/contributors/01-overview.html).
 
 ## Configuration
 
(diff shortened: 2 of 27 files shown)


Nutomic:
So essentially, all actions except upvote, downvote and reply are moved to the dropdown menu, and also the entire context menu is moved from the left to the right.

This change makes sense for small screens where not much space is available, and having buttons on the left or right is not much difference. But on large screens the buttons are too far from the actual content they belong to, and should go back to the left.

 Also it doesnt make sense to leave so much space empty, only to end up with a very large dropdown menu. So on large screens there should be more buttons shown directly and not in the context menu, specifically common actions (link, fedilink, save). The edit and delete buttons should always be in the action bar (including mobile), as they are only present for own comments.

dessalines:
> This change makes sense for small screens where not much space is available, and having buttons on the left or right is not much difference. But on large screens the buttons are too far from the actual content they belong to, and should go back to the left.

This adds too much clutter, even on large screens. I'm just going to match jerboa / boost from now on. 

Plus the dropdown actions are much clearer, with both an icon and text displayed. I also don't want to have to support two different versions, moving things inside and outside of the dropdown based on screen sizes. Someone else can do that later if they want I suppose, although it would just be adding clutter and complication back in. I've been able to get rid of a lot of duplicated `d-none` blocks in my recent refactors, while still making all content and actions available to all screen sizes.

There might also be some smart context menu that can do this automatically, but for now I'm just sticking to simple bootstrap dropdown menus.

All this can be tweaked later.

Nutomic:
At least put edit and delete back in the context menu for own comments, and move it to the left side again.

dessalines:
The edit and delete are shown on the comment dropdown:

Screenshot_2025-10-24-13-45-23-160_mark via

And I would like this (and all the actions and context dropdowns) to stay right-aligned, but someone can add a setting later on for lefties.

dessalines:
If someone wants to add that extra complication back in, they can, but I don't think its a good idea. I also just checked all the other UIs, and they do the same thing, no difference between mobile and desktop, nearly all actions and links behind a dropdown:

The only one that's different is jerboa and boost, which has the bookmark also (but in jerboa I'm probably going to remove the bookmark). 

app | pic
--- | ---
voyager |  Screenshot_2025-10-27-10-28-41-49_a87fd7db6caa850b517aa6fa9d2fcd0e
boost | Screenshot_2025-10-27-10-27-53-52_044f1e467fa7f6fc20a65e6d6e356754
jerboa | Screenshot_2025-10-27-10-28-19-81_a10f969cf1c4adf2c503042b459b8476
photon | Screenshot_2025-10-27-10-29-06-29_a87fd7db6caa850b517aa6fa9d2fcd0e

src/shared/components/comment/comment-node.tsx
```diff
@@ -280,238 +210,99 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
-                          : mdToHtml(this.commentUnlessRemoved, () =>
-                              this.forceUpdate(),
-                            )
-                      }
+                <CommentContent
+                  comment={comment}
+                  viewSource={this.state.viewSource}
+                  hideImages={this.props.hideImages}
+                />
+                <div className="comment-bottom-btns d-flex justify-content-end column-gap-1.5 flex-wrap text-muted fw-bold mt-1 align-items-center">
```
Nutomic:
```suggestion
                <div className="comment-bottom-btns d-flex justify-content-end justify-content-md-start column-gap-1.5 flex-wrap text-muted fw-bold mt-1 align-items-center">
```

This way buttons are right aligned on mobile so its easy to reach with the thumb, and left aligend on desktop so you dont have to move the cursor very far.

dessalines: I tested and I don't like it as much, and it messes with the dropdown positioning.

PR #3554, opened 2025-10-24
Add infinite scrolling setting.

### LINKED ISSUE #3424: Add `infinite_scroll_enabled` to user settings.
### Requirements

Add `infinite_scroll_enabled` to user settings.

dessalines:
- This isn't actually used in lemmy-ui (currently), but other apps might take advantage of it, and we should let them be able to set it through lemmy-ui.
- Fixes #3424

Screenshot_2025-10-24-08-42-10-392_mark via

### DIFF
--- a/src/shared/components/person/settings.tsx
+++ b/src/shared/components/person/settings.tsx
@@ -291,6 +291,7 @@ export class Settings extends Component<SettingsRouteProps, SettingsState> {
           show_downvotes,
           show_upvote_percentage,
           show_person_votes,
+          infinite_scroll_enabled,
         },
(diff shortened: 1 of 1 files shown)


Nutomic:
It would be extremely confusing for users if there is a setting which doesnt work. If lemmy-ui doesnt support infinite scroll then it also shouldnt show any setting. Then other app devs also wont know about the setting and wont use it, so we should just remove it from the backend.

dessalines:
Understandable. I'd like to leave the setting in the back end, because other apps might implement it, or we might eventually add it here.

PR #3555, opened 2025-10-24
Adding enable_animated_images, and hide_media setting.

### LINKED ISSUE #3425: Add `enable_animated_images` to user settings.
### Requirements

Add `enable_animated_images` to user settings.

### LINKED ISSUE #3427: Add `hide_media` to user settings.
### Requirements

Add `hide_media` to user settings.

dessalines:
- Fixes #3427
- Fixes #3425

Adds the setting, and the logic to disable media displaying and animated images.

I spent some time testing out the logic and it works.

Screenshot_2025-10-24-10-15-19-257_mark via

### DIFF
--- a/src/shared/components/post/common.tsx
+++ b/src/shared/components/post/common.tsx
@@ -280,7 +280,6 @@ export function PostImg({
   const url = post.url;
   const thumbnail = post.thumbnail_url;
   const imageSrc = url && isImage(url) ? url : thumbnail;
-  const imageDetails = postView.image_details;
 
   return !hideImages(hideImage, myUserInfo) &&
     imageSrc &&
@@ -290,8 +289,7 @@ export function PostImg({
         <PictrsImage
           src={imageSrc}
           alt={post.alt_text}
-          width={imageDetails?.width}
-          height={imageDetails?.height}
+          imageDetails={postView.image_details}
           nsfw={postView.post.nsfw || postView.community.nsfw}
         />
       </a>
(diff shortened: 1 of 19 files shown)


src/shared/components/post/common.tsx
```diff
@@ -275,7 +282,9 @@ export function PostImg({ postView, showAdultConsentModal }: PostImgProps) {
   const imageSrc = url && isImage(url) ? url : thumbnail;
   const imageDetails = postView.image_details;
 
-  return imageSrc ? (
+  return !hideImages(hideImage, myUserInfo) &&
+    imageSrc &&
+    !hideAnimatedImage(imageSrc, myUserInfo) ? (
```
dessalines, line 286: The main confusing thing about all of our media checking, is that its done at different levels, often intermediary ones.

It might make sense to move all the logic into the lowest level image component, `PictrsImage`, at least then it would be in one place. That can be done as a later cleanup tho, for now I just tried to keep things at their existing level, the `PostImg` and `PostThumbnail` components.

src/shared/utils/media.ts
```diff
@@ -4,6 +4,12 @@ export function isImage(url: string) {
   return imageRegex.test(url);
 }
 
+const animatedImageRegex = /(http)?s?:?(\/\/[^"']*\.(?:gif))/;
```
Nutomic, line 7: How about moving .git into `videoRegex` below and then using the existing `isVideo` function?

dessalines: IMO this setting is specifically about gifs, probably for things like user and community avatars, which some ppl don't like to see.

PR #3572, opened 2025-10-27
Use localhost as testhost

Nutomic:
Without this, all POST requests in Firefox throw `Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://0.0.0.0:8536/api/v4/account/auth/login. (Reason: CORS request did not succeed). Status code: (null).` Now its working again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ pnpm install
 LEMMY_UI_BACKEND_REMOTE=voyager.lemmy.ml pnpm dev
 ```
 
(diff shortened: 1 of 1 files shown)


dessalines:
Change this in the readme also.

And I'm guessing the reason this is failing for you, is because you haven't set `LEMMY_UI_BACKEND_EXTERNAL` to the correct host (which in your case of testing, would be `localhost:8536`

dessalines:
BTW I figured out the problem with voyager.lemmy.ml . Internal docker setups that use LetsEncrypt, should be setting `LEMMY_UI_HTTPS=false`, not true.

Nutomic:
Changed readme.

PR #3574, opened 2025-10-28
Make instance page working again (fixes #3571)

### LINKED ISSUE #3571: Support filtering for `instances.tsx`, support new API.
### Requirements

In updating lemmy-js-client for #3556 , I broke a lot of the `GetFederatedInstances` functionality, which now uses views. 

It also now supports domain filtering, which should be added as page params.

Nutomic:


src/shared/components/home/instances.tsx
```diff
@@ -157,30 +186,105 @@ export class Instances extends Component<InstancesRouteProps, InstancesState> {
...
+    this.props.history.push(`/instances${getQueryString(queryParams)}`);
+  }
+
+  renderRadios() {
+    const allStates: State[] = [
+      { value: "all", i18n: "all" },
+      { value: "linked", i18n: "linked_instances" },
+      { value: "allowed", i18n: "allowed_instances" },
+      { value: "blocked", i18n: "blocked_instances" },
+    ];
```
Nutomic, line 243: With this it always shows the allowed and blocked tabs even when they are empty. Would be good to add fields `GetFederatedInstancesResponse.has_allowlist` and `GetFederatedInstancesResponse.has_blocklist`.

src/shared/utils/routes.ts
```diff
@@ -255,6 +259,7 @@ export const routes: IRoutePropsWithFetch<RouteData, any, any>[] = [
     path: `/instances`,
     component: Instances,
     fetchInitialData: Instances.fetchInitialData,
+    getQueryParams: getInstancesQueryParams,
```
Nutomic, line 262: Lint is complaining about something here but not sure how to fix.

src/shared/components/home/instances.tsx
```diff
@@ -78,27 +120,51 @@ export class Instances extends Component<InstancesRouteProps, InstancesState> {
     }
   }
 
+  componentDidUpdate(prevProps: InstancesRouteProps) {
+    if (this.props.location.key !== prevProps.location.key) {
+      if (this.props.history.action !== "POP") {
+        this.searchInput.current?.select();
+      }
+    }
```
SleeplessOne1917: Since nothing else happens in the top level if block besides the second level if block, this can just be a single if:
```suggestion
    if (this.props.location.key !== prevProps.location.key && this.props.history.action !== "POP") {
        this.searchInput.current?.select();
    }
```

src/shared/utils/routes.ts
SleeplessOne1917: I suspect it might be an issue with how `InstancesFetchConfig` is defined. Looking at its definition in `instances.tsx`, I see:

```typescript
export type InstancesFetchConfig = IRoutePropsWithFetch<
  InstancesData,
  Record<string, never>,
  Record<string, never>
>;
```

Comparing this to a similar type (`SearchFetchConfig` from `search.tsx`) that isn't having a lint error we have:

```typescript
export type SearchFetchConfig = IRoutePropsWithFetch<
  SearchData,
  SearchPathProps,
  SearchProps
>;
```

`InstancesData` is defined in a similar way to `SearchData`, and `SearchPathProps` is just an alias for `Record<string, never>`, just like `InstancesFetchConfig`'s second generic param. I believe the issue is the thrid generic param and that `InstancesFetchConfig` should be defined thusly:

```typescript
export type InstancesFetchConfig = IRoutePropsWithFetch<
  InstancesData,
  Record<string, never>,
  InstancesProps
>;

**Comment by matc-pub** [comment]
```diff
diff --git a/src/shared/components/home/instances.tsx b/src/shared/components/home/instances.tsx
index ce6046f4..7c984133 100644
--- a/src/shared/components/home/instances.tsx
+++ b/src/shared/components/home/instances.tsx
@@ -75,9 +75,8 @@ interface InstancesProps {
   cursor?: DirectionalCursor;
   domain_filter?: string;
 }
-
 type InstancesRouteProps = RouteComponentProps<Record<string, never>> &
-  Record<string, never>;
+  InstancesProps;
 export type InstancesFetchConfig = IRoutePropsWithFetch<
   InstancesData,
   Record<string, never>,
```

src/shared/components/home/instances.tsx
Nutomic: Done, also fixed it in search.tsx where I copied it from.

src/shared/components/common/state-radios.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Component } from "inferno";
+
+export interface State {
+  value: string;
+  i18n: NoOptionI18nKeys;
+}
+
+export interface StateRadioProps {
+  className?: string;
+  allStates: State[];
```
dessalines: Don't call this `State`, that has a specific meaning in react, meaning the component state. Really this is a `RadioOption[]`
, and this component is `Radios` or `RadioButtonGroup`

dessalines: I'd call these:

`allOptions`, `currentOption`, and `onClickOption` or `onClick`. 

Don't call it handler because its not actually *handling* the event, its just the `onClick` event.

src/shared/components/common/state-radios.tsx
```diff
@@ -0,0 +1,62 @@
...
+  currentState: string;
+  onClickHandler(val: string): void;
+}
+
+export class StateRadio extends Component<StateRadioProps, object> {
```
dessalines: Its fine for this to be a component, but in the future, if it only has props and no state, then you can just use a functional component.

https://www.infernojs.org/docs/guides/components

Check out `src/shared/components/post/post-action-bar.tsx` for an example.

src/shared/components/common/state-radios.tsx
```diff
@@ -0,0 +1,62 @@
...
+  constructor(props: any, context: any) {
+    super(props, context);
+  }
+
+  render() {
+    const radioId = randomStr();
+    return (
+      <div
+        className={classNames(
+          "btn-group btn-group-toggle flex-wrap",
```
dessalines: In general I'd like to start moving away from button groups, and use more dropdown selects. If you look at lemmy-ui on mobile, a lot of these button groups overflow in weird ways. Plus google's android app guidelines seems to recomment against them for the most part.

src/shared/components/home/instances.tsx
```diff
@@ -22,11 +27,38 @@ import { Icon, Spinner } from "../common/icon";
...
+import { linkEvent } from "inferno";
+import { PaginatorCursor } from "@components/common/paginator-cursor";
+import { createRef } from "inferno";
+
+function getKindFromQuery(kind?: string): GetFederatedInstancesKind {
+  return kind ? (kind as GetFederatedInstancesKind) : "all";
+}
+function getDomainFilterFromQuery(domain_filter?: string): string | undefined {
+  return domain_filter;
+}
```
dessalines: This one seems pointless.

src/shared/components/home/instances.tsx
```diff
@@ -157,30 +202,102 @@ export class Instances extends Component<InstancesRouteProps, InstancesState> {
...
+  handleChange(state: GetFederatedInstancesKind) {
+    this.updateUrl({ kind: state });
+    this.fetchInstances();
+  }
+
+  handlePageChange(cursor?: DirectionalCursor) {
+    this.updateUrl({ cursor });
+  }
+
+  async updateUrl(state: Partial<InstancesProps>) {
```
dessalines: Remove async.

src/shared/components/home/instances.tsx
dessalines: I wouldn't worry too much about it, its fine to show them.

src/shared/components/home/instances.tsx
```diff
@@ -157,30 +202,102 @@ export class Instances extends Component<InstancesRouteProps, InstancesState> {
...
 
 interface InstanceListProps {
-  items: Instance[];
-  /// Only use link for non-blocked
-  blocked?: boolean;
-  onRemove?(instance: string): void;
+  instances: FederatedInstanceView[];
   hideNoneFound?: boolean;
+  onRemove?(instance: string): void;
+  cursor?: DirectionalCursor;
```
dessalines, line 289: This shouldn't need a cursor.

src/shared/components/search.tsx
```diff
@@ -451,10 +451,11 @@ export class Search extends Component<SearchRouteProps, SearchState> {
...
   componentDidUpdate(prevProps: SearchRouteProps) {
-    if (this.props.location.key !== prevProps.location.key) {
-      if (this.props.history.action !== "POP") {
-        this.searchInput.current?.select();
-      }
+    if (
+      this.props.location.key !== prevProps.location.key &&
+      this.props.history.action !== "POP"
+    ) {
+      this.searchInput.current?.select();
```
matc-pub, line 458: We should probably remove this focus call. On mobile this opens the keyboard for every search type change. The existing call from `componentDidMount` should already do the initial focusing.

src/shared/components/common/state-radios.tsx
Nutomic: Button groups are useful because they require one less click for common actions. For actions which are used rarely a dropdown does make sense.

src/shared/components/home/instances.tsx
Nutomic: It has pagination so it needs a cursor.

src/shared/components/search.tsx
Nutomic: Its not related to this change so please open a separate issue or pull request.

PR #3575, opened 2025-10-29
Attempt to optimize initial page load

Nutomic:
[Chrome dev tools](https://developer.chrome.com/docs/devtools/performance/overview) have some optimization suggestions so I tried them:
- Moved all styles and scripts to footer
- Also use deferred loading for main styles

As a result the "Insight" for render blocking requests goes away, as the browser doesnt have to wait for css to finish loading before rendering the page. However this means the page is first rendered without styling and results in significant cumulative layout shift.

Measurements with 4x cpu slowdown and slow 4g network, eruda commented out:
main branch: LCP 5.6s, CLS 0
voyager (prod build): LCP 3.7s, CLS 0
this branch: LCP 5.25s, CLS 0.9

And with fast 4g network:
main branch: LCP 1.6s, CLS 0
voyager (prod build): 0.8s, CLS 0
this branch: LCP 1.3s, CLS 1

So while it solves one problem it creates another, and its not clear that its worth the benefit. We could potentially create a smaller css file which is loaded before html and only blocks rendering for a short time (similar to what I did for icons). Maybe there are tools to create this automatically.

Edit: Tried with [minimalcss](https://github.com/peterbe/minimalcss) but it produces empty output. Also tried [penthouse](https://github.com/pocketjoso/penthouse) via the hosted version but using that doesnt resolve the layout shift at all. Tried [uncss](https://github.com/uncss/uncss) as well but it didnt reduce CLS either.

### DIFF
--- a/src/server/utils/create-ssr-html.tsx
+++ b/src/server/utils/create-ssr-html.tsx
@@ -91,34 +91,10 @@ export async function createSsrHtml(
     <!DOCTYPE html>
     <html ${helmet.htmlAttributes.toString()}>
     <head>
-    <script nonce="${cspNonce}">
-    window.isoData = ${serialize(isoData)};
-
-    ${embeddedScript}
-    </script>
-    ${lazyScripts}
-  
-    <!-- A remote debugging utility for mobile -->
-    ${erudaStr}
-  
-    <!-- Custom injected script -->
-    ${customHtmlHeaderWithNonce}
-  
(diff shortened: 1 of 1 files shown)


dessalines:
Probably not worth it... deferring the entire js app and the styles isn't going to look good.

SleeplessOne1917:
I second Dessalines here. I don't think the minor speedup is worth introducing a flash of unstyled content.

PR #3579, opened 2025-10-30
Merge modlog tables

Nutomic:
Working fine like before:
Screenshot_20251030_121402

### DIFF
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -215,11 +215,11 @@ function processModlogEntry(
       return {
         modlog,
         moderator,
-        data: (
+        data: target_community && (
           <>
             <span>Purged Post From</span>
             <CommunityLink
-              community={target_community!}
+              community={target_community}
               myUserInfo={myUserInfo}
             />
             {reason && (
@@ -236,12 +236,12 @@ function processModlogEntry(
       return {
         modlog,
         moderator,
-        data: (
+        data: target_person && (
           <>
             <span>{is_revert ? "Appointed " : "Removed "}</span>
             <span>
               <PersonListing
-                person={target_person!}
+                person={target_person}
                 myUserInfo={myUserInfo}
                 banned={false}
               />
@@ -256,12 +256,12 @@ function processModlogEntry(
       return {
         modlog,
         moderator,
-        data: (
+        data: target_person && target_community && (
           <>
             <span>{is_revert ? "Appointed " : "Removed "}</span>
             <span>
               <PersonListing
-                person={target_person!}
+                person={target_person}
                 myUserInfo={myUserInfo}
                 banned={false}
               />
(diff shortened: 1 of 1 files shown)


src/shared/components/modlog.tsx
```diff
@@ -95,118 +95,94 @@ interface ModlogProps {
...
+    target_community,
+    target_person,
+    target_post,
+  } = view;
+
+  switch (view.modlog.kind) {
+    /*
     default:
       // FIXME: placeholder for version mismatch between js-client and backend
       // Without a default case typescript is able to report missing types
```
SleeplessOne1917: Is the mismatch mentioned in the FIXME comment still an issue?

Nutomic: Not sure what the issue is exactly, as lemmy and lemmy-ui are always supposed to run the same version. So its good if this throws an error if a variant isnt handled so we can add it. Gonna remove the whole block.

src/shared/components/modlog.tsx
```diff
@@ -317,55 +253,43 @@ function processModlogEntry(
...
         moderator,
-        publishedAt: published_at,
         data: (
           <>
-            <span>{removed ? "Removed " : "Appointed "}</span>
+            <span>{is_revert ? "Appointed " : "Removed "}</span>
             <span>
               <PersonListing
-                person={other_person}
+                person={target_person!}
```
dessalines: I'm torn on whether to leave these `!` in here. 

I know that these fields really are `not null` depending on the type, but its also a bad practice to use them for production apps since they can crash UIs. 

I'll leave it up to you, but it might be cleaner to wrap the whole span in `target_person && <span ...` , and remove all these exclamation points. Then it's guaranteed no crashes.

src/shared/components/modlog.tsx
```diff
@@ -896,7 +778,7 @@ export class Modlog extends Component<ModlogRouteProps, ModlogState> {
     return amAdmin(this.isoData.myUserInfo) || amMod_;
   }
 
-  modOrAdminText(person?: Person): string {
+  modOrAdminText(person: Person | null): string {
```
dessalines, line 782: Should be ` | undefined`

Nutomic: Doesnt compile, the Person comes from `ModlogView` which has it defined as `moderator: Person | null`.

src/shared/components/modlog.tsx
Nutomic: Okay changed them all.

src/shared/components/modlog.tsx
dessalines: Then that needs to be fixed in the backend by adding `ts(optional_fields)` . I thought we added a check to make sure that we aren't exporting ` | null`, but I guess not.

Nutomic: Fixed in https://github.com/LemmyNet/lemmy-js-client/pull/765. But now it also includes the removal of `BanFromCommunityResponse` and I dont see how to get that working without rewrites. Better to handle in a separate PR and merge this one already.

dessalines: I have the `BanFromCommunityResponse` stuff fixed in a separate PR, I'll open soon.

Issue #3577, opened 2025-10-29
New modlog API support


dessalines:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

The modlog API result is changing, to support a more simplified DB structure with less joins:

https://github.com/LemmyNet/lemmy/pull/6054

The modlog page will need to be reworked to support it.

This would also be a good time to do translations correctly, #31

Issue #3578, opened 2025-10-29
Remove all constructor `bind`s and internal class method event handlers


dessalines:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

In a lot of the component constructors, you'll find lines like these:

```ts
this.handleEditCommunity = this.handleEditCommunity.bind(this);
```

Then there will be an internal method `handleEditCommunity()` which appears to take in no params, and uses `this`. These are all unchecked, and require usage of `this`, which is also dangerously unchecked. 

Handlers should be moved out of class methods, into top level private functions, and called like:

```ts
onBanFromCommunity={form => handleModBanFromCommunity(props, form)}
```

Or use `linkEvent`

A clean example of how these should be done is in `src/shared/components/post/post-action-bar.tsx`

This is already fixed for many components, but we should make this consistent and do it for all of them.

dessalines:
As part of this, I'm going to add typescript typed linting: https://typescript-eslint.io/getting-started/typed-linting/

It shows **a lot** of errors outside of just the bind issues, but I think going through these is worth the effort.

#3611

PR #3580, opened 2025-10-30
 Notify about mod actions (fixes #3488) 

### LINKED ISSUE #3488: New notification items
### Requirements

- https://github.com/LemmyNet/lemmy/pull/5604
- https://github.com/LemmyNet/lemmy/pull/5981

Edit: Requires https://github.com/LemmyNet/lemmy/issues/6070

Nutomic:
Still needs a better design. Using the tables from modlog doesnt seem to make sense here, but I also dont want to duplicate all the code to generate views from modlog items. 

Screenshot_20251030_152621

--- 

Requires https://github.com/LemmyNet/lemmy-ui/issues/3577 and https://github.com/LemmyNet/lemmy/pull/6115

### DIFF
--- a/src/shared/components/person/notifications.tsx
+++ b/src/shared/components/person/notifications.tsx
@@ -311,9 +311,9 @@ export class Notifications extends Component<
   messageTypeRadios() {
     const allStates: RadioOption[] = [
       { value: "all" },
-      { value: "replies" },
-      { value: "mentions" },
-      { value: "messages" },
+      { value: "reply", i18n: "replies" },
+      { value: "mention", i18n: "mentions" },
+      { value: "private_message", i18n: "messages" },
       { value: "mod_action", i18n: "modlog" },
     ];
 


dessalines:
If you can, try to replicate the top bar of comment or post. IE

- Time top right
- Moderator top bar, along with any context.
- Reason and expiration where the comment would be.

Nutomic:
Adjusted it like this:
Screenshot_20251106_154805

src/shared/components/common/content-actions/content-action-dropdown.tsx
```diff
@@ -359,11 +360,10 @@ export default class ContentActionDropdown extends Component<
...
                                 this.props.postView.post_actions?.read_at,
                             })}
-                            label={
-                              this.props.postView.post_actions?.read_at
-                                ? I18NextService.i18n.t("mark_as_unread")
-                                : I18NextService.i18n.t("mark_as_read")
-                            }
+                            label={mark_as_read_i18n(
+                              this.props.postView.post_actions?.read_at !==
+                                null,
```
dessalines: Would probably be better to do ` ?? false` rather than `!== null`

src/shared/components/common/state-radios.tsx
dessalines, line 1: I thought this already got added, and it was `option-radios`.

src/shared/components/person/notifications.tsx
```diff
@@ -487,6 +435,16 @@ export class Notifications extends Component<
             />
           )
         );
+      case "mod_action": {
+        return (
+          <NotificationModlogItem
+            notification={item.notification}
+            modlog_view={item.data as ModlogView}
```
dessalines: The other ones use `i` but this one uses `item` for some reason. You might be able to avoid the `as ModlogView` cast then.

src/shared/components/common/content-actions/content-action-dropdown.tsx
Nutomic: That would return `string | false` instead of bool.

src/shared/components/person/notifications.tsx
Nutomic: Right that works but Im renaming it to `data`.

src/shared/components/person/notifications.tsx
```diff
@@ -514,12 +474,13 @@ export class Notifications extends Component<
...
-    i.setState({
-      messageType: event.target.value as NotificationDataType,
+  async handleMessageTypeChange(val: NotificationDataType) {
+    this.setState({
+      messageType: val,
       cursor: undefined,
     });
-    await i.refetch();
+    // TODO: somehow it cannot refetch from here
+    this.refetch();
```
Nutomic: This one is not working, all API requests from here are failing with `CORS: missing origin`. Other places use `linkEvent` but that doesnt work here as the param is not an event.

src/shared/components/common/state-radios.tsx
Nutomic: Right its `radio-button-group`

src/shared/components/common/content-actions/content-action-dropdown.tsx
dessalines: Gotcha... change to `!== undefined then`

src/shared/components/person/notifications.tsx
```diff
@@ -302,78 +310,20 @@ export class Notifications extends Component<
...
-            active: this.state.messageType === "private_message",
-          })}
-        >
-          {I18NextService.i18n.t("messages")}
-        </label>
-      </div>
+      <RadioButtonGroup
+        allOptions={allStates}
+        currentOption={this.state.messageType}
+        onClick={this.handleMessageTypeChange}
```
dessalines: Do `onClick={val => handleMessageTypeChange(this, val)`

And add `i: Notifications` as the first param to that function.

And remove `this` from within that function, and only use `i.refetch()`

src/shared/components/person/notifications.tsx
dessalines: See comment above.

src/shared/components/person/notifications.tsx
dessalines: Then you can also remove the bind. I'm gonna do that for all cases soon anyway: #3578

src/shared/components/modlog.tsx
```diff
@@ -95,118 +95,94 @@ interface ModlogProps {
...
       return {
         id: 0,
-        publishedAt: nowBoolean(true)!,
+        moderator: null,
+        published_at: nowBoolean(true)!,
         data: (
           <>
             <span>Placeholder for:</span>{" "}
-            <span>{(view as { type_: string }).type_}</span>
+            <span>{(view.modlog as { kind: string }).kind}</span>
```
dessalines: Should all this be commented out?

src/shared/components/modlog.tsx
dessalines, line 1: I don't see why there are changes to the modlog in this PR.

src/shared/components/person/notifications.tsx
Nutomic: Done but it still throws the same error as before.

src/shared/components/modlog.tsx
Nutomic: These changes are all gone after merging from main branch.

src/shared/components/person/notifications.tsx
dessalines: I'll test this out now and see what's going on.

dessalines: K, finally got a test bed set up. I'm not getting any console errors, everything is working fine.

Are you sure that you were using the correct lemmy back-end that adds all this?

dessalines:
Mark these as draft until the back-end is merged.

src/shared/components/person/notifications.tsx
Nutomic: Looks like that was the problem, now its working.

Nutomic:
Had to adjust the state values, now everything is working fine. Backend PR is also merged.

PR #3606, opened 2025-11-10
Dont alter crosspost body for own post (fixes #3603)

### LINKED ISSUE #3603: Don't put crosspost text bodies inside quote blocks?
### Requirements

When I press the cross-post button, it includes a link to the original post and wraps the post around quote blocks, sometimes messing up the formatting. This is highly opinionated, if I'm the one who wrote the post then why would I need to say if I posted it on other forums? That information isn't very relevant to other users either. Lemmy should really handle crossposts better.

**Edit**: should’ve been more specific, I mean I still want the post body but not in quotes, and I don’t want the original post linked.

cross-posted from:  https://lemmy.world/post/9999

> Paragraph ...

Nutomic:


src/shared/components/post/post-action-bar.tsx
```diff
@@ -221,9 +221,15 @@ function crossPostParams(post: Post): CrossPostParams {
...
-function crossPostBody(post: Post): string | undefined {
+function crossPostBody(
+  post: Post,
+  myUserInfo?: MyUserInfo,
+): string | undefined {
   const body = post.body;
 
+  if (post.creator_id === myUserInfo?.local_user_view?.person?.id) {
+    return body;
+  }
```
dessalines: This means that bodies for other posts will be completely lost.

The body can have a lot of context for the link, which is a good thing to show even if its not your post. We should probably just change the format, may to a `---` then the body, instead of big quote block.

dessalines: So its really just the `cross-posted-from` line that should be ignored if its your own user, and possibly the quote markup. The post body could just be copied plain.

Nutomic: Thats exactly what this PR does. When crossposting someone elses post it keeps the current behaviour, adding the link and quotes. When crossposting your own post it keeps the body as is (no need to quote yourself, and the link can be found from the user profile).

dessalines: Those early returns are confusing

https://github.com/LemmyNet/lemmy-ui/pull/3608

PR #3608, opened 2025-11-11
Self crosspost cleanup

dessalines:


SleeplessOne1917 (approved):

src/shared/components/post/post-action-bar.tsx
```diff
@@ -227,11 +227,11 @@ function crossPostBody(
...
-    return body;
-  }
+  // If its you're own post, don't include the x-posted-from line, or the quotes
   return body
-    ? `${I18NextService.i18n.t("cross_posted_from_url", { ap_id: post.ap_id })}
+    ? post.creator_id === myUserInfo?.local_user_view?.person?.id
+      ? body
+      : `${I18NextService.i18n.t("cross_posted_from_url", { ap_id: post.ap_id })}
       \n\n${body.replace(/^/gm, "> ")}`
     : undefined;
```
Nutomic: I find nested ternary operator much more confusing than early return. Better to use if else for one of them.

dessalines: I also kind of hate them, and all this could be eliminated if typescript actually had if assignments like modern languages. I'll refactor in a little different way tho that uses muts

Issue #1166, opened 2023-06-11
Remove undetermined from language selection


Tealk:
### Is your proposal related to a problem?

Often "undetermined" is offered and set as the default language, and I often see posts where no language is selected.
This bypasses all the language filtering.

### Describe the solution you'd like

That the selection "undetermined" is hidden when creating contributions and comments

Nutomic:
Undetermined should be the default, we cant expect users to set the language manually all the time. However I suppose it would be fine to remove it as a dropdown option.

Chais:
I agree it should be the overall default, but it should also be configurable. Ideally on every level.  
I imagine something like this:

- A fresh instance's language defaults to "undetermined" but the admin can configure it.
- Each community on that instance defaults to the instance's language, with an option to override (helpful for multilingual instances)
- When posting to a community the language defaults to the community's language with an option to override (multilingual instances again)
- When commenting on a post the language defaults to the post's language with an option to override (bit of an edge-case. for example English natives asking for resources for learning French in a French community on a bilingual instance)
- When replying to a comment the language defaults to the comment's language with an option to override again (I don't expect this to be useful very often, but at this point, why divert from the pattern ¯\\\_(ツ)\_/¯ )

This way we establish sane defaults on every level, but there's always the option to change it. But having sane defaults means users should rarely have to change them.

kevincox:
Another idea:

Something that would help a lot is allowing user to select their "primary posting languages". I basically only use English online but I honestly get tired of selecting it for every comment. So I probably send a lot of "undetermined" posts and comments out of laziness. For my use case being able to set the default language would mean lots of accurately tagged posts and comments.

Obviously it is more complicated for people who speak multiple languages. But maybe if they could select their "primary posting lagnuages" and it could be intersected with the languages allowed in the communities? If there is not a unique overlap at least their primary languages could be at the top of the list to make it easier than hunting through the entire list. Even then I suspect lots of multi-lingual people have a strong bias towards a particular language that it may make sense to have it set as the default for them, however this risks mis-tagging if they forget to switch the default which is probably worse than "undetermined".

This may be useful in compliment to @Chais's idea. Although quite possibly if most instances/communities gain a language tag then personal preference will be mostly irrelevant. Although it could help a bit for multilingual communities. Although these are fairly rare they definitely exist, especially for things like cities. But in this case it may be better just to follow the parent as suggested above. Only in a situation like the top-level post would the user's language be a main default.

Pros:
[...]

PR #3614, opened 2025-11-13
Use interface language as default language for posts/comments  (ref #1166)

Nutomic:


src/shared/components/common/language-select.tsx
```diff
@@ -188,11 +198,6 @@ export class LanguageSelect extends Component<
           style="max-height: 200px;"
           tabIndex={-1}
         >
-          {!this.props.multiple && (
-            <option selected disabled hidden>
-              {I18NextService.i18n.t("language_select_placeholder")}
-            </option>
-          )}
```
Nutomic, line 195: Unnecessary as the whole block is only for multi select.

matc-pub:
This will create a lot of mislabeled posts, which in my opinion isn't an improvement over unlabeled posts.

src/shared/components/common/language-select.tsx
```diff
@@ -104,6 +104,18 @@ export class LanguageSelect extends Component<
       this.props.myUserInfo,
     );
 
+    // If no language is explicitly selected, take interface_language of current user as default.
+    if (!selectedLangs && this.props.allLanguages && this.props.myUserInfo) {
+      const default_lang_code =
+        this.props.myUserInfo.local_user_view.local_user.interface_language;
+      const default_lang = this.props.allLanguages.find(
+        l => l.code === default_lang_code,
+      );
```
matc-pub, line 113: `interface_language` is likley `"browser"` for most users, which also makes `default_lang` undefined for most users.

Nutomic: The interface language is automatically set on registration based on `Accept-Language` header. I tested this and it works as expected.

- https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_crud/src/user/create.rs#L493
- https://github.com/LemmyNet/lemmy/blob/main/crates/api/api_crud/src/user/create.rs#L538

Nutomic:
Will use a different approach for this, by creating a plugin for language detection.

PR #3617, opened 2025-11-13
Add RES-style keyboard navigation

### LINKED ISSUE #984: RES style keyboard shortcuts

CyberShadow:
# Add RES-style keyboard navigation

Implements keyboard shortcuts for navigating and interacting with posts and comments across Lemmy UI.

Closes #984

### Posts (homepage, community pages, user profiles)

**Navigation:**
| Key | Action |
|-----|--------|
| `j` | Next post |
| `k` | Previous post |
| `J` | Last post |
| `K` | First post |

**Pagination:**
| Key | Action |
|-----|--------|
| `n` | Next page |
| `p` | Previous page |
| `P` | First page |

**Actions:**
| Key | Action |
|-----|--------|
| `a` | Upvote (toggle) |
| `z` | Downvote (toggle) |
| `s` | Save/unsave (toggle) |
| `x` | Expand post body (toggle) |
| `e` | Edit own post |
| `.` | Show advanced menu (toggle) |

**Quick Navigation:**
| Key | Action |
|-----|--------|
| `c` | Go to post comments (current tab) |
| `C` | Go to post comments (new tab) |
| `l` | Open post link (current tab) |
| `L` | Open post link (new tab) |
| `u` | Go to user profile (current tab) |
| `U` | Go to user profile (new tab) |
| `r` | Go to community (current tab) |
| `R` | Go to community (new tab) |

### Comments (post detail pages)

**Navigation:**
| Key | Action |
|-----|--------|
| `j` | Next comment |
| `k` | Previous comment |
| `J` | Next sibling comment |
| `K` | Previous sibling comment |
| `p` | Parent comment |
| `t` | Thread root (top-level comment) |

**Actions:**
| Key | Action |
|-----|--------|
| `a` | Upvote (toggle) |
| `z` | Downvote (toggle) |
| `s` | Save/unsave (toggle) |
| `r` | Reply to comment |
| `e` | Edit own comment |
| `Enter` | Collapse/expand comment |
| `.` | Show advanced menu (toggle) |

**Quick Navigation:**
| Key | Action |
|-----|--------|
| `u` | Go to user profile (current tab) |
| `U` | Go to user profile (new tab) |

## Implementation

I tried to take a least-intrusive approach and isolate the behavior to its own component, as well as to reduce code duplication. However, I may have over-engineered it a little, as this adds a lot more code in total than the predecessor PR.

On the other hand, a good chunk of the new code is comments and type declarations, which I hope contribute positively to overall complexity.

## Testing

I've tested this manually on:
- Homepage post navigation (j/k) and pagination (n/p/P)
- Community page post navigation and pagination
- Post detail page with comments (j/k, J/K sibling nav, p/t parent/thread nav)
- User profile pages (mixed posts and comments) with pagination
- Empty pages (pagination shortcuts still work)
- Keyboard shortcuts while typing (properly ignored)
- Comment tree navigation (sibling, parent, thread root)
- First/last navigation (J/K in post lists)

## Credits

Based on the implementation by @HamzahMansour in PR #1892.

Some code is by Claude (Sonnet 4.5).

### DIFF
--- a/src/shared/components/post/post-listing-list.tsx
+++ b/src/shared/components/post/post-listing-list.tsx
@@ -1,10 +1,29 @@
 import { ShowBodyType, ShowCrossPostsType } from "@utils/types";
 import {
+  AddAdmin,
+  AddModToCommunity,
+  BanFromCommunity,
+  BanPerson,
+  BlockCommunity,
+  BlockPerson,
   CreatePostLike,
+  CreatePostReport,
+  DeletePost,
+  FeaturePost,
+  HidePost,
   Language,
   LocalSite,
+  LockPost,
+  MarkPostAsRead,
   MyUserInfo,
+  NotePerson,
+  PersonView,
   PostView,
+  PurgePerson,
+  PurgePost,
+  RemovePost,
+  SavePost,
+  TransferCommunity,
 } from "lemmy-js-client";
 import { VoteButtons } from "@components/common/vote-buttons";
 import { postIsInteractable, userNotLoggedInOrBanned } from "@utils/app";
@@ -12,11 +31,13 @@ import { PostThumbnail } from "./post-thumbnail";
 import { PostCreatedLine, PostName } from "./common";
 import { CrossPosts } from "./cross-posts";
 import { CommentsButton } from "./post-action-bar";
+import { PostActionDropdownWrapper } from "./post-action-dropdown-wrapper";
 import { mdToHtml } from "@utils/markdown";
 
 type Props = {
   postView: PostView;
   crossPosts: PostView[];
+  admins: PersonView[];
   allLanguages: Language[];
   showCommunity: boolean;
   showBody: ShowBodyType;
@@ -26,14 +47,35 @@ type Props = {
   myUserInfo: MyUserInfo | undefined;
   localSite: LocalSite;
   showCrossPosts: ShowCrossPostsType;
-  onPostVote(form: CreatePostLike): void;
-  onScrollIntoCommentsClick(e: MouseEvent): void;
+  markable: boolean;
   imageExpanded?: boolean;
+  onEditClick(): void;
+  onPostVote(form: CreatePostLike): void;
+  onPostReport(form: CreatePostReport): void;
+  onBlockPerson(form: BlockPerson): void;
+  onBlockCommunity(form: BlockCommunity): void;
+  onLockPost(form: LockPost): void;
+  onDeletePost(form: DeletePost): void;
+  onRemovePost(form: RemovePost): void;
+  onSavePost(form: SavePost): void;
+  onFeaturePost(form: FeaturePost): void;
+  onPurgePerson(form: PurgePerson): void;
+  onPurgePost(form: PurgePost): void;
+  onBanPersonFromCommunity(form: BanFromCommunity): void;
+  onBanPerson(form: BanPerson): void;
+  onAddModToCommunity(form: AddModToCommunity): void;
+  onAddAdmin(form: AddAdmin): void;
+  onTransferCommunity(form: TransferCommunity): void;
+  onHidePost(form: HidePost): void;
+  onPersonNote(form: NotePerson): void;
+  onScrollIntoCommentsClick(e: MouseEvent): void;
+  onMarkPostAsRead(form: MarkPostAsRead): void;
 };
 
 export function PostListingList({
   postView,
   crossPosts,
+  admins,
   allLanguages,
   showCommunity,
   showBody,
@@ -43,9 +85,29 @@ export function PostListingList({
   myUserInfo,
   localSite,
   showCrossPosts,
-  onPostVote,
-  onScrollIntoCommentsClick,
+  markable,
   imageExpanded,
+  onEditClick,
+  onPostVote,
+  onPostReport,
+  onBlockPerson,
+  onBlockCommunity,
+  onLockPost,
+  onDeletePost,
+  onRemovePost,
+  onSavePost,
+  onFeaturePost,
+  onPurgePerson,
+  onPurgePost,
+  onBanPersonFromCommunity,
+  onBanPerson,
+  onAddModToCommunity,
+  onAddAdmin,
+  onTransferCommunity,
+  onHidePost,
+  onPersonNote,
+  onScrollIntoCommentsClick,
+  onMarkPostAsRead,
 }: Props) {
   return (
     <div>
@@ -75,11 +137,42 @@ export function PostListingList({
             allLanguages={allLanguages}
             myUserInfo={myUserInfo}
           />
-          <CommentsButton
-            postView={postView}
-            type_="text"
-            onScrollIntoCommentsClick={onScrollIntoCommentsClick}
-          />
+          <div className="d-flex align-items-center gap-2">
+            <CommentsButton
+              postView={postView}
+              type_="text"
+              onScrollIntoCommentsClick={onScrollIntoCommentsClick}
+            />
+            <PostActionDropdownWrapper
+              postView={postView}
+              admins={admins}
+              showBody={showBody}
+              markable={markable}
+              viewOnly={viewOnly}
+              viewSource={false}
+              myUserInfo={myUserInfo}
+              onViewSource={() => {}}
+              onEditClick={onEditClick}
+              onPostReport={onPostReport}
+              onBlockPerson={onBlockPerson}
+              onBlockCommunity={onBlockCommunity}
+              onLockPost={onLockPost}
+              onDeletePost={onDeletePost}
+              onRemovePost={onRemovePost}
+              onSavePost={onSavePost}
+              onFeaturePost={onFeaturePost}
+              onPurgePerson={onPurgePerson}
+              onPurgePost={onPurgePost}
+              onBanPersonFromCommunity={onBanPersonFromCommunity}
+              onBanPerson={onBanPerson}
+              onAddModToCommunity={onAddModToCommunity}
+              onAddAdmin={onAddAdmin}
+              onTransferCommunity={onTransferCommunity}
+              onHidePost={onHidePost}
+              onPersonNote={onPersonNote}
+              onMarkPostAsRead={onMarkPostAsRead}
+            />
+          </div>
           {postView.post.body && showBody === "full" && (
             <PostBody body={postView.post.body} />
           )}
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -303,6 +303,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
           <PostListingList
             postView={p.postView}
             crossPosts={p.crossPosts}
+            admins={p.admins}
             allLanguages={p.allLanguages}
             showCommunity={p.showCommunity}
             showBody={p.showBody}
@@ -312,9 +313,29 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
             myUserInfo={p.myUserInfo}
             localSite={p.localSite}
             showCrossPosts={p.showCrossPosts}
-            onPostVote={p.onPostVote}
-            onScrollIntoCommentsClick={p.onScrollIntoCommentsClick}
+            markable={p.markable}
             imageExpanded={this.state.imageExpanded}
+            onEditClick={() => handleEditClick(this)}
+            onPostVote={p.onPostVote}
+            onPostReport={p.onPostReport}
+            onBlockPerson={p.onBlockPerson}
+            onBlockCommunity={p.onBlockCommunity}
+            onLockPost={p.onLockPost}
+            onDeletePost={p.onDeletePost}
+            onRemovePost={p.onRemovePost}
+            onSavePost={p.onSavePost}
+            onFeaturePost={p.onFeaturePost}
+            onPurgePerson={p.onPurgePerson}
+            onPurgePost={p.onPurgePost}
+            onBanPersonFromCommunity={p.onBanPersonFromCommunity}
+            onBanPerson={p.onBanPerson}
+            onAddModToCommunity={p.onAddModToCommunity}
+            onAddAdmin={p.onAddAdmin}
+            onTransferCommunity={p.onTransferCommunity}
+            onHidePost={p.onHidePost}
+            onPersonNote={p.onPersonNote}
+            onScrollIntoCommentsClick={p.onScrollIntoCommentsClick}
+            onMarkPostAsRead={p.onMarkPostAsRead}
           />
         );
       case "card":


Nutomic:
In general the shortcuts are working well. Various things I noticed:
- The highlighted background color for posts should only be used after a shortcut is first pressed. For users on mobile or using mouse for navigation it should not be visible.
- Save post doesnt give any feedback so its not clear that it worked.
- Context menu for posts on homepage is broken, `.` shortcut is also broken
- It would make sense to close the current comment editor with `Esc` or similar.
- A user guide on how to use these shortcuts would be nice. Either directly in the UI or in the documentation. Anyway this can be added later.

CyberShadow:
Thanks for testing!

> The highlighted background color for posts should only be used after a shortcut is first pressed. For users on mobile or using mouse for navigation it should not be visible.

We can try this. In RES this is not a problem because you can explicitly enable or disable keyboard navigation.

> Save post doesnt give any feedback so its not clear that it worked.

Agreed. I'll add a toast. Looks like we'll need new strings for localization after all.

> Context menu for posts on homepage is broken, `.` shortcut is also broken

For list mode, it looks like there is no context menu button at all. Since the <kbd>.</kbd> key simply clicks the button, we can't show it there in list mode.

It should be working OK in card and small card mode.

Do you think we should add the ⋮ button but hide it? Or we can add the button at all times, since it seems useful to have either way, but that's a little out of scope for this PR and more of an overall design question. Happy to do either.

> A user guide on how to use these shortcuts would be nice. Either directly in the UI or in the documentation. Anyway this can be added later.

I have a draft for the documentation repository, let me upload it as well.

CyberShadow:
> > Save post doesnt give any feedback so its not clear that it worked.
> 
> Agreed. I'll add a toast. Looks like we'll need new strings for localization after all.

Another option would be to add a badge to the post / comment itself to visually indicate that the post is saved, but that's also a small design change.

Nutomic:
Right it looks like the context menu from post listing was removed on main branch. Imo that should be added back. Inside the post view `.` doesnt work to open the context menu, it only highlights the menu button and then you need to click enter to open it.

CyberShadow:
> The highlighted background color for posts should only be used after a shortcut is first pressed. For users on mobile or using mouse for navigation it should not be visible.

Done.

> Save post doesnt give any feedback so its not clear that it worked.

Toast added. It seemed useful even when doing this with the mouse, so I made it show regardless of how the post is saved/unsaved.

> Right it looks like the context menu from post listing was removed on main branch. Imo that should be added back.

OK, added some patches which do this.

image

> A user guide on how to use these shortcuts would be nice. Either directly in the UI or in the documentation. Anyway this can be added later.

Here it is: https://github.com/LemmyNet/lemmy-docs/pull/395

CyberShadow:
> Inside the post view `.` doesnt work to open the context menu, it only highlights the menu button and then you need to click enter to open it.

I'm having quite a lot of trouble reproducing this. I managed to reproduce this once, and then no longer. I've tried multiple browsers and variations of the circumstances.

I suspect this has something to do with the Dropdown asset not being loaded at the time.

Do you have a way to reproduce this problem reliably?

Nutomic:
Now the problem also disappeared for me, maybe it only happens on the first load or something like that?

CyberShadow:
It's possible, but this is stretching the limits of my front-end development knowledge, so I can't answer for sure.

If it's transient and the feature is not a commonly used one (I certainly have not used the RES equivalent personally), perhaps some flakiness is acceptable until we have more information? Up to you.

Nutomic:
Yes thats fine if it cant be reproduced. Lets see what @dessalines says.

src/shared/components/comment/comment-node.tsx
```diff
@@ -127,6 +137,54 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
 
   constructor(props: any, context: any) {
     super(props, context);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
```
dessalines, line 140: Don't use binds (we're in the process of getting rid of all of them). Look at all the other handler functions at the bottom, and do:

`function handleKeyDown(i: CommentNode, event: KeyboardEvent)`

If an existing file is already using a ton of binds, then don't worry about it, we'll handle it in a later cleanup.

src/shared/components/comment/comment-node.tsx
```diff
@@ -107,6 +115,9 @@ type CommentNodeProps = {
   onPurgeComment(form: PurgeComment): void;
   onPersonNote(form: NotePerson): void;
   onLockComment(form: LockComment): void;
+  // Keyboard navigation props
+  highlightedCommentId?: number | null;
```
dessalines, line 119: Just make this `isHighlighted: boolean;` You can do the check a level up.

src/shared/components/comment/comment-node.tsx
```diff
@@ -205,12 +267,22 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
...
             "border-top border-light": !this.props.noBorder,
             mark: this.isCommentNew || distinguished,
+            "keyboard-selected": this.isHighlighted,
           })}
+          onClick={() => {
+            if (this.props.onCommentClick) {
+              this.props.onCommentClick(id);
+            }
+          }}
+          onKeyDown={this.handleKeyDown}
```
dessalines, line 283: `onKeyDown={e => handleKeyDown(this, e)`

src/shared/components/comment/comment-node.tsx
```diff
@@ -107,6 +115,9 @@ type CommentNodeProps = {
   onPurgeComment(form: PurgeComment): void;
   onPersonNote(form: NotePerson): void;
   onLockComment(form: LockComment): void;
+  // Keyboard navigation props
+  highlightedCommentId?: number | null;
+  onCommentClick?(commentId: number): void;
```
dessalines, line 120: What is this for? For navigation to the comment? Or what keyboard command uses it?

src/shared/components/mixins/keyboard-shortcuts-mixin.ts
```diff
@@ -0,0 +1,160 @@
...
+  MyUserInfo,
+} from "lemmy-js-client";
+import { TContextRouter } from "inferno-router/dist/Router";
+import type { ContentItem } from "@utils/keyboard-shortcuts-actions";
+
+/**
+ * Interface for components that can use keyboard shortcuts for navigation
+ * Generic over the item type (PostView, PersonContentCombinedView, etc.)
+ */
+export interface KeyboardShortcutsComponent<T> {
```
dessalines, line 24: Why is commentnode not using this also?

src/shared/components/person/person-details.tsx
```diff
@@ -85,14 +96,117 @@ interface PersonDetailsProps {
...
+{
+  state: KeyboardNavigationState = {
+    highlightedIndex: 0,
+    expandedPostIndices: new Set<number>(),
+  };
+
   constructor(props: any, context: any) {
     super(props, context);
+    this.handleHighlight = this.handleHighlight.bind(this);
+    this.togglePostExpand = this.togglePostExpand.bind(this);
```
dessalines, line 117: Remove binds.

src/shared/components/person/person-details.tsx
```diff
@@ -85,14 +96,117 @@ interface PersonDetailsProps {
...
+    this.setState({ highlightedIndex: index });
+  }
+
+  scrollToIndex(index: number) {
+    this.scrollToItem(index);
   }
 
-  renderItemType(i: PersonContentCombinedView): InfernoNode {
+  // Custom key handler for component-specific keys
+  handleCustomKeys(
```
dessalines, line 138: Make this a top-level function with `handleCustomKeys(i: PersonDetails, ....`

src/shared/components/person/person-details.tsx
```diff
@@ -142,7 +258,11 @@ export class PersonDetails extends Component<PersonDetailsProps, any> {
             postListingMode="small_card"
             showCommunity
             crossPosts={[]}
-            showBody={"preview"}
+            showBody={getShowBodyMode(this.state.expandedPostIndices, idx)}
+            isExpanded={isExpanded}
+            isHighlighted={isHighlighted}
+            postIndex={idx}
```
dessalines, line 264: This shouldn't be necessary, you're already passing the entire postView which has the id.

src/shared/components/person/person-details.tsx
```diff
@@ -185,11 +305,33 @@ export class PersonDetails extends Component<PersonDetailsProps, any> {
...
     return (
-      <div>
-        {combined.map(i => [
-          this.renderItemType(i),
-          <hr key={i.type_} className="my-3" />,
-        ])}
+      <div
+        className="person-details"
+        role="feed"
+        aria-label="User posts and comments"
```
dessalines, line 311: i18n

src/shared/components/person/person-details.tsx
```diff
@@ -185,11 +305,33 @@ export class PersonDetails extends Component<PersonDetailsProps, any> {
...
 
     return (
-      <div>
-        {combined.map(i => [
-          this.renderItemType(i),
-          <hr key={i.type_} className="my-3" />,
-        ])}
+      <div
+        className="person-details"
+        role="feed"
```
dessalines, line 310: just remove

src/shared/components/post/post-action-dropdown-wrapper.tsx
dessalines, line 1: What is the point of this wrapper? I can't see anything else that it adds, and if it does, just move it to the PostActionDropdown.

src/shared/components/post/post-listing-card.tsx
```diff
@@ -152,15 +152,18 @@ export class PostListingCard extends Component<
...
                   myUserInfo={p.myUserInfo}
+                  imageExpanded={p.imageExpanded}
+                  showAdultConsentModal={p.showAdultConsentModal}
                 />
               </div>
             )}
           </div>
           {!p.smallCard && isImagePost && (
-            <PostImg
+            <PostThumbnail
```
dessalines, line 162: ???

Change this back.

src/shared/components/post/post-listing-list.tsx
```diff
@@ -1,44 +1,113 @@
...
   viewOnly: boolean;
+  showAdultConsentModal: boolean;
   myUserInfo: MyUserInfo | undefined;
   localSite: LocalSite;
   showCrossPosts: ShowCrossPostsType;
+  markable: boolean;
+  imageExpanded?: boolean;
+  onEditClick(): void;
   onPostVote(form: CreatePostLike): void;
+  onPostReport(form: CreatePostReport): void;
```
dessalines, line 54: I guess all these are necessary now for the `PostListingList` too.

src/shared/components/post/post-listing-list.tsx
```diff
@@ -91,3 +196,20 @@ export function PostListingList({
...
+    <article id="postContent" className="my-2">
+      <div className="col-12 card card-body">
+        <div
+          className="md-div"
+          dangerouslySetInnerHTML={mdToHtml(body, () => {})}
+        />
+      </div>
+    </article>
+  );
+}
```
dessalines, line 215: Delete this, it already exists in post-listing-card.tsx#L232

You could probably move it to `src/shared/components/post/common.tsx`

src/shared/components/post/post-listing.tsx
```diff
@@ -33,9 +33,20 @@ import { PostForm } from "./post-form";
...
+} from "@utils/keyboard-shortcuts";
+import {
+  handleKeyboardShortcut,
+  type KeyboardShortcutHandlers,
+} from "@utils/keyboard-shortcuts-actions";
+import classNames from "classnames";
 
 type PostListingState = {
   showEdit: boolean;
+  imageExpanded: boolean;
```
dessalines, line 49: We removed all imageExpanded, and now just use postBodyType (list, card, smallcard), so you can get rid of this.

src/shared/components/post/post-listing.tsx
```diff
@@ -78,21 +89,41 @@ type PostListingProps = {
...
 export class PostListing extends Component<PostListingProps, PostListingState> {
   state: PostListingState = {
     showEdit: false,
+    // imageExpanded set in componentWillMount based on postListingMode
+    imageExpanded: false,
   };
 
   constructor(props: any, context: any) {
     super(props, context);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
```
dessalines, line 111: Remove bind.

src/shared/components/post/post-listing.tsx
```diff
@@ -101,14 +132,146 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     });
   }
 
+  componentDidUpdate(
```
dessalines, line 135: What is going on here?

`isFocused` should be a prop. 

I don't see why you need a masonry updates, we're already doing that elsewhere.

isHighlighted should be a prop and I don't see why a masonry update is necessary.

src/shared/components/post/post-listing.tsx
```diff
@@ -101,14 +132,146 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
+
+    // Update layout when entering/leaving edit mode
+    if (prevState.showEdit !== this.state.showEdit) {
+      // Wait for PostForm to render and expand to full height before recalculating layout
+      // Triple RAF to handle timing variations in form rendering
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          requestAnimationFrame(() => {
+            masonryUpdate();
+          });
```
dessalines, line 182: 💀

src/shared/components/post/post-listing.tsx
```diff
@@ -101,14 +132,146 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
+        });
+      });
+    }
+  }
+
   componentWillUnmount(): void {
     this.unlisten();
   }
 
+  focusListing() {
```
dessalines, line 192: I don't get the flow of this at all. The `isFocused` should be a prop, and scrolling should be done in `onComponentMount`

src/shared/components/post/post-listing.tsx
```diff
@@ -101,14 +132,146 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
...
+    const postElement = document.querySelector(
+      `[data-post-index="${this.props.postIndex}"]`,
+    ) as HTMLElement;
+    if (postElement) {
+      postElement.focus();
+      postElement.scrollIntoView({ behavior: "instant", block: "nearest" });
+    }
```
dessalines, line 199: You're scrolling to a post, but from the individual post listing component? Why isn't this done at the collection level?

src/shared/components/post/post-listing.tsx
dessalines, line 1: Doesn't seem to use the mixin again.

src/shared/components/post/post-listings.tsx
```diff
@@ -69,13 +82,108 @@ interface PostListingsProps {
...
+    highlightedIndex: 0,
+    expandedPostIndices: new Set<number>(),
+  };
+
   duplicatesMap = new Map<number, PostView[]>();
 
   constructor(props: any, context: any) {
     super(props, context);
+    this.handleHighlight = this.handleHighlight.bind(this);
+    this.toggleExpand = this.toggleExpand.bind(this);
```
dessalines, line 105: Remove binds

src/shared/components/post/post-listings.tsx
```diff
@@ -69,13 +82,108 @@ interface PostListingsProps {
...
   onScrollIntoCommentsClick(): void;
+  onNextPage?(): void;
+  onPrevPage?(): void;
+  onFirstPage?(): void;
 }
 
-export class PostListings extends Component<PostListingsProps, any> {
+@keyboardShortcutsMixin
+export class PostListings
+  extends Component<PostListingsProps, KeyboardNavigationState>
```
dessalines, line 92: Rename this to PostListingsState or somethin. Keyboardnavigationstate is feature, not this componens.

src/shared/components/post/post-listings.tsx
```diff
@@ -69,13 +82,108 @@ interface PostListingsProps {
...
+  implements KeyboardShortcutsComponent<PostView>
+{
+  state: KeyboardNavigationState = {
+    highlightedIndex: 0,
+    expandedPostIndices: new Set<number>(),
```
dessalines, line 97: This is a really weird place to store this. Don't you already have it in `PostListing`? And is this imageExpanded, bodyExpanded?

src/shared/components/post/post.tsx
```diff
@@ -133,6 +145,9 @@ interface PostState {
   lastCreatedCommentId?: CommentId;
   notifications: PostNotificationsMode;
   editLoading: boolean;
+  // Keyboard navigation state
+  highlightedCommentId: CommentId | null; // null means post is highlighted
```
dessalines, line 149: `highlightedCommentId?: CommentId`

src/shared/components/post/post.tsx
```diff
@@ -258,8 +285,12 @@ export class Post extends Component<PostRouteProps, PostState> {
     isIsomorphic: false,
     notifications: "replies_and_mentions",
     editLoading: false,
+    highlightedCommentId: null, // Start with post highlighted
+    imageExpanded: false, // Start with image not expanded
```
dessalines, line 289: Remove.

If you need to show the expanded image, use `post_listing_mode: card` vs `small_card`

src/shared/components/post/post.tsx
dessalines, line 1: Another one without the mixin

src/shared/components/post/post.tsx
```diff
@@ -249,6 +264,18 @@ export class Post extends Component<PostRouteProps, PostState> {
...
+      // Get all visible comment elements in DOM order
+      // Collapsed comments' children are not rendered, so they won't be in the DOM
+      // Query for article elements with IDs exactly matching 'comment-{digits}'
+      return Array.from(
+        document.querySelectorAll(SELECTOR.COMMENT_ARTICLES),
+      ).filter(el => PATTERN.COMMENT_ID.test(el.id)) as HTMLElement[];
+    },
+    () => this.state.highlightedCommentId,
+    id => this.setState({ highlightedCommentId: id }),
+  );
```
dessalines, line 278: DOM queries just to set the highlighted comment? None of this is necessary.

Just set highlighted after the comments fetch.

CyberShadow:
> Adding a `KeyboardEvents` react component (doesn't even have to be generic, since all the functionality and events are so different), loaded at the **screen level**, IE `home`, `community`, `post.tsx`, etc, **not at the sub-component level**.

Not sure how that would look like, since we have to deal with a good variety of objects (just posts on home/community page, post listing + comments on post pages, mixed items on the user page). Maybe have these expose a common interface? What do you have in mind?

> * Instantianting keyboard navigation for every individual comment is overkill and doesn't make much sense.

For what it's worth, the approach I was going for was to separate actions (upvoting, saving, etc.) from navigation. Actions are handled by the object the action applies to, while navigation is handled by its container.

src/shared/components/post/post-action-dropdown-wrapper.tsx
CyberShadow: Yes, it's a fully mechanical refactoring of the parts of `PostActionBar` that we want to also use `PostListingList`. I can check if we can just push these into `PostActionDropdown`.

src/shared/components/post/post-listing.tsx
CyberShadow: Just to clarify, are you saying we should cycle `postBodyType` to `card` when we want to show the image? I can see that working for `small_card`, but not for `list`, where the layout of everything else is different.

src/shared/components/comment/comment-node.tsx
CyberShadow: Oops! Will fix.

src/shared/components/mixins/keyboard-shortcuts-mixin.ts
CyberShadow: In principle, the navigation models are different - comments are arranged in a tree and have an additional movement axis. (Aware that the current implementation flattens everything in a list anyway - left that as-is to avoid premature optimization.)

src/shared/components/post/post-listing.tsx
CyberShadow: There is a pre-existing bug (visible on e.g. `voyager.lemmy.ml`) where entering edit mode runs `masonryUpdate` too early, causing the bottom of the edit form to overlap with the post below it.

image

Based on your comment, I'm guessing you don't have a better idea for how to fix it either :)

dessalines:
> Not sure how that would look like, since we have to deal with a good variety of objects (just posts on home/community page, post listing + comments on post pages, mixed items on the user page). Maybe have these expose a common interface? What do you have in mind?

On the `home.tsx` page for instance, near the top of the page you'd do

```tsx
<KeyboardCommentEvents
  focusedComment={this.state.focusedComment}
  onCommentSave={form => handleSaveComment(this, form)}
  onVote={...}
  onFocusNext={...}
  onNavigateToComment={...}
  ...
/>
...
<KeyboardPostEvents
  ...
```

CyberShadow:
> On the `home.tsx` page for instance, near the top of the page you'd do

Thanks. Would you mind illustrating how you envision this to work e.g. on the user page, which has both posts and comments? Would we have both `KeyboardCommentEvents` and `KeyboardPostEvents`, both registering keyboard handlers, but only acting if the currently selected item is of their preferred type? How would hand-over be done when navigating from a post to a comment, would these components share state about which item is selected?

dessalines:
> Would you mind illustrating how you envision this to work e.g. on the user page, which has both posts and comments? Would we have both KeyboardCommentEvents and KeyboardPostEvents, both registering keyboard handlers, but only acting if the currently selected item is of their preferred type?

That does seem like the best way, although the navigation is beyond me. The `personContentRes: RequestState<ListPersonContentResponse>;` is an ordered array of items that are either comments or posts with a tagged type, so you could key off the index to decide which `Events` to be active.

Nutomic:
Closing as outdated. Feel free to open a new PR for this feature.

PR #3628, opened 2025-11-17
Show list of active plugins in sidebar

Nutomic:
I put it like this under the admin list for now, with plugin description as mouseover text. It could also go into a separate box like the subscriptions list, with plugins descriptions always visible.
Screenshot_20251117_103015

Requires https://github.com/LemmyNet/lemmy-translations/pull/225

### DIFF
--- a/src/shared/components/home/home.tsx
+++ b/src/shared/components/home/home.tsx
@@ -478,7 +478,14 @@ export class Home extends Component<HomeRouteProps, HomeState> {
   }
 
   get mobileView() {
-    const { siteRes, showSubscribedMobile, showSidebarMobile } = this.state;
+    const {
+      siteRes: {
+        site_view: { local_site, site },
+        admins,
+      },
+      showSubscribedMobile,
+      showSidebarMobile,
+    } = this.state;
 
     return (
       <div className="row">
@@ -497,10 +504,14 @@ export class Home extends Component<HomeRouteProps, HomeState> {
           />
           {showSidebarMobile && (
             <SiteSidebar
-              site={siteRes.site_view.site}
-              siteRes={siteRes}
+              site={site}
+              admins={admins}
+              localSite={local_site}
               isMobile
               myUserInfo={this.isoData.myUserInfo}
+              allLanguages={this.state.siteRes.all_languages}
+              siteLanguages={this.state.siteRes.discussion_languages}
+              activePlugins={this.state.siteRes.active_plugins}
             />
           )}
           {showSubscribedMobile && (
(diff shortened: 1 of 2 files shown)


wiki-me:
Does clicking on it lead to a link like a github read me or some other provided link?

src/shared/components/home/home.tsx
```diff
@@ -529,22 +519,14 @@ export class Home extends Component<HomeRouteProps, HomeState> {
...
+    const { siteRes } = this.state;
 
     return (
       <div id="sidebarContainer">
         <SiteSidebar
-          site={site}
-          admins={admins}
-          localSite={local_site}
+          site={siteRes.site_view.site}
+          siteRes={siteRes}
```
dessalines: Since this is a case where site sidebar is also used on the community page, I'd prefer you pass `activePlugins` as a prop to it. 

In general with components (to avoid rerendering), you want to pass as little as possible (and move up a level if every caller needs the same data), and especially don't pass potentially redundant / overlapping fields like this one now does.

src/shared/components/home/site-sidebar.tsx
```diff
@@ -23,12 +23,9 @@ import {
 
 interface SiteSidebarProps {
   site: Site;
-  localSite?: LocalSite;
-  admins?: PersonView[];
+  siteRes?: GetSiteResponse;
   isMobile?: boolean;
   myUserInfo: MyUserInfo | undefined;
-  allLanguages?: Language[];
-  siteLanguages?: number[];
```
dessalines, line 31: By removing all these its now confusing where you should use siteRes vs site. Just restore this.

src/shared/components/home/site-sidebar.tsx
```diff
@@ -152,6 +157,23 @@ export class SiteSidebar extends Component<SiteSidebarProps, SiteSidebarState> {
     );
   }
 
+  plugins(plugins: PluginMetadata[]) {
```
dessalines, line 158: Lets try to be good and use functional components instead of class field components.

IE 

```tsx
type ActivePluginsProps = {plugins: PluginMetadata[]};

function ActivePlugins({plugins}: ActivePlugins) {
  return (
  ...
```

src/shared/components/home/home.tsx
Nutomic: Okay done, although its quite verbose this way.

src/shared/components/home/site-sidebar.tsx
Nutomic: Left this as is because its only used internally, besides it would be inconsistent with the rest of the code.

PR #3655, opened 2025-12-02
Community settings

### LINKED ISSUE #3565: Move community editing to its own `community/settings` page
### Requirements

Currently, we have the community editing as in-place in the community.tsx sidebar, which looks terrible on desktop:

Image

Just like with admin settings and person settings, community settings should really be its own page, rather than trying to do in-place editing.

dessalines:
This PR adds 

- A community settings page
- A moderators tab, with the ability to 
  - add mods via search
  - remove mods
  - transfer the community
- Moves modlog to the community sidebar buttons.
- Fixes #3565 

This sets up more features that will be added in other PRs:

- A multi community settings page. Will be straightforward as it'll be ripped from this. #3760 
- Adding tags. #3566 

This was a large undertaking, and there's a lot of cleanup here in the process.

## Community settings 

https://github.com/user-attachments/assets/9054c13f-fb62-42c2-822f-2e151de07a2b

Also cleaned up some sidebars:

## Sidebar

Screenshot_2026-01-15-14-33-34-090_mark via

## Sidebar of a community you mod

Screenshot_2026-01-15-14-33-19-366_mark via


Nutomic:
The only problem I notice is that the community sidebar as too many buttons now.
- Remove "Leave mod team", its already available in the settings
- Moderation related actions are rarely used, they can be moved further down and smaller, rendered as `btn-outline` or even a dropdown (create report, modlog, remove, purge)
- As admin I see some of the options even for communities which I dont moderate, these should be hidden (settings, remove)

dessalines:
> As admin I see some of the options even for communities which I dont moderate, these should be hidden (settings, remove)

Admins should have full powers over communities, it is their instance after all. Its also been an annoying problem where mods go absent, and the sidebar needs to be changed, so now admins should be able to do that.

> Moderation related actions are rarely used, they can be moved further down and smaller, rendered as btn-outline or even a dropdown (create report, modlog, remove, purge)

I'll try some other colors for the mod buttons. I do want them to be mostly uniform with the other block buttons tho. Don't want to make another dropdown for them.

> Remove "Leave mod team", its already available in the settings

K, I can do that.

dessalines:
The only one that looks decent, is `btn-outline-secondary`. I gotta say I don't like it as much as the regular button tho. And I'm trying to keep colors muted, and use as few colors as possible (except for core items like people and communities)

Screenshot_2026-01-16-15-47-32-029_mark via

Nutomic:
> Admins should have full powers over communities, it is their instance after all. Its also been an annoying problem where mods go absent, and the sidebar needs to be changed, so now admins should be able to do that.

Makes sense for local communities, but the settings button is also shown for remote communities where this wont work reliably. And just like create report or modlog, it wont be used often and should be further down, eg in a dropdown.

dessalines:
I can hide it for remote communities, but I prefer settings being on the sidebar here.

Nutomic (approved):
Gonna merge this, maybe we can do more fine-tuning later.

PR #3657, opened 2025-12-02
Adding API updates to support PagedResponse.

dessalines:
Support new `PagedResponse` types. Tested and it seems to work.


Nutomic:
You should be able to remove `directionalCursor()` and `cursorComponents()`.

dessalines:
Done. I tested it using a low limit on `home.tsx` and it seemed to work fine.

@matc-pub might want to check it at some point tho.

matc-pub:
Looks good to me.

PR #3660, opened 2025-12-03
Remove nightly build block for woodpecker

dessalines:
Going to merge as I want to test this out.


Nutomic:
The [woodpecker docs](https://woodpecker-ci.org/docs/usage/workflow-syntax#when---conditional-execution) mention this:

> If at least one of the conditions in the when block evaluate to true the step is executed,

PR #3665, opened 2025-12-04
Revert "Development instructions using remote instance"

Nutomic:
After deploying 0.19.14-beta.3 to lemmy.ml the site stopped working (even though its fine on enterprise.lemmy.ml). The backend logged lots of errors `actix_http::h1::dispatcher: stream error: request parse error: invalid Header provided`. Most likely there is something wrong with this commit, and its not really worth debugging so Im simply reverting it.


dessalines:
Good idea. I'm sure it has to do with the `https` needing to be unset or something, but ya this could potentially break a lot of sites.

dessalines:
I started another build for lemmy-ui with this removed: `0.19.14-beta.4`

PR #3667, opened 2025-12-05
Display resolved objects at top of search results (fixes #3656)

### LINKED ISSUE #3656: Support new `resolve` field in search bar from `SearchResponse`
### Requirements

In my API update rework, I haven't supported the resolve field yet in the search bar. This needs to be done.

Nutomic:
Looks like this:
Screenshot_20251205_124049

And here with a comment which contains the resolved url (so it is returned as search result):
Screenshot_20251205_124102

### DIFF
--- a/src/shared/components/search.tsx
+++ b/src/shared/components/search.tsx
@@ -246,7 +246,7 @@ const multiCommunityListing = (
 const personListing = (
   persons: PersonView[],
   myUserInfo: MyUserInfo | undefined,
-  showHeader: boolean = false,
+  showHeader: boolean = true,
 ) => {
   return (
     persons.length > 0 && (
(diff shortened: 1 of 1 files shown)


src/shared/components/search.tsx
```diff
@@ -244,11 +246,12 @@ const multiCommunityListing = (
 const personListing = (
   persons: PersonView[],
   myUserInfo: MyUserInfo | undefined,
+  showHeader: boolean = false,
```
dessalines: This one should probably be true to match the others.

src/shared/components/search.tsx
```diff
@@ -1059,7 +1094,10 @@ export class Search extends Component<SearchRouteProps, SearchState> {
   get resultsCount(): number {
     const { searchRes: r } = this.state;
 
-    const searchCount = r.state === "success" ? r.data.search.length : 0;
+    const searchCount =
+      r.state === "success"
+        ? r.data.search.length + Number(r.data.resolve !== undefined)
```
dessalines: Should probably just do a ` let resolveCount = r.data.resolve !== undefined ? 1 : 0;` to be clearer

PR #3669, opened 2025-12-05
Host address environment variable change

matc-pub:
Backend and frontend address can be specified as url (without path), protocol is optional and the default depends on LEMMY_UI_HTTPS. This allows to use different protocols for the frontend and backend, and the internal and external backend addresses.

If the frontend is unspecified, protocol and host values from the request headers are used instead.

Frontend urls are used for the meta tags set by `HtmlTags` and to detect local urls. Backend urls are used for rss feeds and pictrs images.

This also removes helper methods that return different hosts based on `isBrowser()`, SSR and CSR should always render the same urls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,16 +69,16 @@ Finally open `http://localhost:1234` in your browser. This uses the public test 
 
 The following environment variables can be used to configure lemmy-ui:
 
-| `ENV_VAR`                      | type     | default            | description                                                                                                                      |
-| ------------------------------ | -------- | ------------------ | -------------------------------------------------------------------------------------------------------------------------------- |
-| `LEMMY_UI_HOST`                | `string` | `0.0.0.0:1234`     | The IP / port that the lemmy-ui isomorphic node server is hosted at.                                                             |
+| `ENV_VAR`                      | type     | default            | description                                                                                                                        |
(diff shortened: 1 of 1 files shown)


SleeplessOne1917 (approved):

README.md
```diff
@@ -69,15 +69,16 @@ Finally open `http://localhost:1234` in your browser. This uses the public test
...
-| `LEMMY_UI_DISABLE_CSP`         | `bool`   | `false`          | Disables CSP security headers                                                       |
-| `LEMMY_UI_CUSTOM_HTML_HEADER`  | `string` | `undefined`      | Injects a custom script into `<head>`.                                              |
+| `ENV_VAR`                      | type     | default            | description                                                                                                                        |
+| ------------------------------ | -------- | ------------------ | ---------------------------------------------------------------------------------------------------------------------------------- |
+| `LEMMY_UI_HOST`                | `string` | `0.0.0.0:1234`     | The IP / port that the lemmy-ui isomorphic node server is hosted at.                                                               |
+| `LEMMY_UI_FRONTEND`            | `string` | `undefined`        | The base url where browsers can reach the lemmy-ui server, is inferred from request headers when undefined. (Protocol is optional) |
```
dessalines, line 75: Nice, it'll be good not to have to set this.

dessalines:
Thx, this is much cleaner.

PR #3671, opened 2025-12-05
Run prettier on a few more files

matc-pub:
This makes `pnpm lint` also check `README.md`,`package.json` and a few other files in the root directory.

I also added `pnpm prettier:write` to fix those same files.

### DIFF
--- a/.woodpecker.yml
+++ b/.woodpecker.yml
@@ -14,7 +14,7 @@ steps:
   prettier_check:
     image: jauderho/prettier:3.7.3-alpine
     commands:
-      - prettier -c .
+      - prettier --no-config --arrow-parens avoid -c .
     when:
       - event: pull_request
 


.prettierrc.json
```diff
@@ -1,4 +1,5 @@
 {
   "arrowParens": "avoid",
+  "plugins": ["prettier-plugin-packagejson"],
```
matc-pub, line 3: We also have `prettier-plugin-organize-imports` as an existing dependency that could be enabled here.

.woodpecker.yml
```diff
@@ -14,7 +14,7 @@ steps:
   prettier_check:
     image: jauderho/prettier:3.7.3-alpine
     commands:
-      - prettier -c .
+      - prettier --no-config --arrow-parens avoid -c .
```
dessalines, line 17: Why not just let this use the existing prettierrc.json?

package.json
```diff
@@ -137,7 +139,6 @@
     "prettier-plugin-packagejson": "^2.5.19",
     "qs": "^6.13.0",
     "rimraf": "^6.0.1",
-    "sort-package-json": "^3.4.0",
```
dessalines, line 140: Thx, this was abandoned I believe, and caused lots of annoying issues.

.woodpecker.yml
matc-pub: The docker image doesn't have access to the plugin. Our config only has two options, and `semi: true` happens to be the default.

https://woodpecker.join-lemmy.org/repos/135/pipeline/8438/4
```
+ prettier -c .
Checking formatting...
[error] Cannot find package 'prettier-plugin-packagejson' imported from /woodpecker/src/github.com/LemmyNet/lemmy-ui/noop.js
```

package.json
matc-pub: One problem probably was that only `lint-staged` ran it. `prettier-plugin-packagejson` depends `sort-package-json`.

SleeplessOne1917 (approved):

PR #3701, opened 2025-12-15
Fix ARM build by switching to Debian image (fixes #3682)

### LINKED ISSUE #3682: Fix ARM build
https://github.com/LemmyNet/lemmy-ui/pull/3680#event-21428425474

Nutomic:
As mentioned by b2c on Matrix, [this](https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/11169) seems to be the relevant issue. Its already fixed but it seems we need to wait for a new Alpine release, and then wait for Docker image `node:alpine` to be updated. Or alternatively switch to a Debian-based build image like `node:current-slim`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lemmy-ui",
-  "version": "1.0.0-try-arm-build.3",
+  "version": "1.0.0-try-arm-build.4",
   "description": "An isomorphic UI for lemmy",
   "repository": "https://github.com/LemmyNet/lemmy-ui",
   "license": "AGPL-3.0",


dessalines:
I'd be fine switching to a debian-based image.

Nutomic:
Build is working now by using Debian:
- https://woodpecker.join-lemmy.org/repos/135/pipeline/8627
- https://hub.docker.com/r/dessalines/lemmy-ui/tags

Image is currently too large because I was using the wrong deps for runtime. Doing another build now.

Dockerfile
```diff
@@ -59,15 +62,18 @@ RUN rm -rf ./node_modules/import-sort-parser-typescript
...
 
-RUN \
-  --mount=type=cache,target=/var/cache/apk,id=apk-${TARGETARCH},sharing=private \
-  apk update && apk add curl vips-cpp
+RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
+    --mount=target=/var/cache/apt,type=cache,sharing=locked \
+    rm -f /etc/apt/apt.conf.d/docker-clean \
+    && apt-get update \
+    && apt-get -y --no-install-recommends install \
+           curl
```
Nutomic, line 76: Dont know whats the right vips package for Debian here, there are a bunch of [packages with somewhat related names](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=vips).

dessalines: libvips-dev is probably the right one. A build will probably fail if its incorrect.

Nutomic: That seems to be the build-time dependency (already installed above), but this part is for the runtime dep. So if its wrong that will only be noticable when executing the image.

If you or anyone else has an ARM device, it would be good if you could test running the published [Docker image](https://hub.docker.com/layers/dessalines/lemmy-ui/1.0.0-try-arm-build.4/images/sha256-a3fec0e63d1226ee3ae9999a52940acf134eff3b275c6c1525ca7c86e61c23a9).

PR #3705, opened 2025-12-17
Increase cache duration for static assets to one year

Nutomic:
[Pagespeed Insights](https://pagespeed.web.dev/analysis/https-voyager-lemmy-ml-post-33791/8z88cwehs3?form_factor=mobile) considers one day too short, and it actually doesnt make sense to reload client.js every day. In any case there is a hash in the filename so that the new file is loaded after Lemmy upgrade.

Screenshot_20251217_141327

### DIFF
--- a/src/server/middleware.ts
+++ b/src/server/middleware.ts
@@ -39,10 +39,6 @@ export function setCacheControl(
   res: Response,
   next: NextFunction,
 ) {
-  if (process.env.NODE_ENV !== "production") {
-    return next();
-  }
-
   let caching: string;
 
   // Only allow caching for success responses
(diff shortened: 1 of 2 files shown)


src/server/middleware.ts
```diff
@@ -39,10 +39,6 @@ export function setCacheControl(
   res: Response,
   next: NextFunction,
 ) {
-  if (process.env.NODE_ENV !== "production") {
-    return next();
-  }
-
```
Nutomic, line 45: In index.tsx there is already the same check.

dessalines:
Seems fine, especially since these are all under a hash heading.

dessalines (approved):

PR #3744, opened 2026-01-09
Use pattern to validate password length (fixes #3724)

### LINKED ISSUE #3724: [Bug]: Password is truncated on signup
### Requirements

- [x] Is this a bug report? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a single bug? Do not put multiple bugs in one issue.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.

### Summary

During signup, a user can enter a password that exceeds the 60-character limit. This input is not validated by either the frontend or the backend, resulting in an account creation where the password is truncated to 60 characters.

### Steps to Reproduce

1. Go to the `/signup` page.
2. Enter a password that is 61 characters long.
3. Fill the rest of the form.
4. Submit.

**Result:**

1. An account is created. 
2. The actual password is truncated to 60 characters.
3. The user is unable to log in with the originally entered password.

**Expected:** the backend should return a clear error when the password exceeds the allowed limit, allowing the user to correct the issue.

### Technical Details

The issue was discovered from a user perspective; no server-side logs are available.

I used Chromium browser, and no errors were noted in the console.

### Version

BE 0.19.14, BE 0.19.13

### Lemmy Instance URL

lemmy.zip, beehaw.org

Nutomic:
The maxLength attribute seems to work as intended, the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/maxlength) mention "the browser will generally prevent user from entering more text than the maxlength attribute allows". And Chrome has the same behaviour, so its unlikely that we will get them to change it.

Instead we can use pattern for validation which works as expected in Chromium.

### DIFF
--- a/src/shared/components/common/password-input.tsx
+++ b/src/shared/components/common/password-input.tsx
@@ -102,10 +102,8 @@ class PasswordInput extends Component<PasswordInputProps, PasswordInputState> {
                 onInput={onInput}
                 value={value}
                 required={required !== false}
-                pattern=".+"
+                pattern=".{10,60}"
                 title={I18NextService.i18n.t("invalid_password")}
-                minLength={10}
-                maxLength={60}
               />
               <button
                 className="btn btn-outline-dark"


dessalines:
This means that whatever browser the user is using, is not mozilla or chrome based then, since its ignoring maxlength. 

We have a lot of other maxLengths in the codebase, and I don't think we should handle it with pattern matching.

matc-pub:
We could also keep maxLength and the pattern.

Nutomic:
> This means that whatever browser the user is using, is not mozilla or chrome based then, since its ignoring maxlength.

Not sure what you mean by this. I tested with Chromium, and it doesnt let you enter more than `maxLength` characters, so longer passwords are effectively truncated as described in #3724.

> We have a lot of other maxLengths in the codebase, and I don't think we should handle it with pattern matching.

Other uses of `maxLength` look fine as they simply limit the length of various text fields. Only password needs to show a warning if it is too long.

> We could also keep maxLength and the pattern.

That wouldnt fix the issue, then password would be truncated again.

dessalines (approved):
Alright I think I understand, we don't want truncation, we want an error.

PR #3756, opened 2026-01-13
Post title on listing page should always open comments (fixes #3755)

### LINKED ISSUE #3755: [1.0] Clicking title of image post opens the image url instead of comments
### Requirements

### Summary

You can test it here: https://voyager.lemmy.ml/u/thelsim@sh.itjust.works

With 0.19 clicking the title of a post always opens the post page with comments. With 1.0, only for image posts it opens the image url instead.

### Lemmy Instance Version

1.0

Nutomic:


src/shared/components/post/common.tsx
```diff
@@ -34,12 +33,9 @@ type PostNameProps = {
 export function PostName({ post, showBody }: PostNameProps) {
   const url = post.url;
 
-  // Only navigate to external, if its media, or the full post
-  const navigateExternal = (url && isMedia(url)) || showBody === "full";
-
   return (
     <h1 className="post-name h5 d-inline text-break">
-      {navigateExternal ? (
```
dessalines, line 42: Just get of this function then, since its only used here.

Nutomic: What do you mean?

dessalines: Delete the `navigateExternal` function, its only used in one place, and now that's gone.

Nutomic: Its not a function but a variable, and its already removed.

dessalines: Whoops sry, brainfart.

PR #3757, opened 2026-01-14
Show warning message to mods if community is not federated yet

Nutomic:
I came across a [comment](https://lemmy.world/post/41435198/21484593) mentioning that many new moderators dont know that they need to promote their community to different instances, in order to make it federate:

> 75% of small communities, if not higher, don’t use lemmy-federate to expand the visibility of their community. The user makes the community, broadcasts a few posts locally and then gets sad that no-one replies (because it can only be seen locally). Nor do they make use of [!newcommunities@lemmy.world](https://lemmy.world/c/newcommunities@lemmy.world) or [!communitypromo@lemmy.ca](https://lemmy.world/c/communitypromo@lemmy.ca) to advertise.

So I had the idea of showing a warning message to the moderator if the community has no remote follows:Screenshot_20260114_120646

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -14,6 +14,7 @@ import {
   updateCommunityBlock,
   updatePersonBlock,
 } from "@utils/app";
+import { T } from "inferno-i18next-dess";
 import {
   getQueryParams,
   getQueryString,
@@ -115,6 +116,7 @@ import { nowBoolean } from "@utils/date";
 import { NoOptionI18nKeys } from "i18next";
 import { TimeIntervalSelect } from "@components/common/time-interval-select";
 import { PostListingModeSelect } from "@components/common/post-listing-mode-select";
+import { communityName } from "./community-link";
 
 type CommunityData = RouteDataResponse<{
   communityRes: GetCommunityResponse;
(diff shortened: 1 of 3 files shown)


src/shared/components/community/community.tsx
```diff
@@ -443,11 +443,32 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+              <div className="alert alert-warning text-bg-warning" role="alert">
+                <h4 className="alert-heading">Community not federated yet</h4>
+                <div className="card-text">
+                  Only users on example.com can currently view this comunity. To
+                  promote the community and get more followers, talk about it in
+                  other communities and include the link `!main@example.com`.
+                  You can also use the following: <br />-
+                  https://lemmy-federate.com <br />-
+                  https://lemmy.world/c/newcommunities <br />-
+                  https://communitypromo@lemmy.ca
```
Nutomic: Any suggestions to improve the text? Also need to move it to translations and use proper variables.

dessalines: I think something like,

"Only local users can see this community. You can use lemmy-federate.com to make sure other servers can see it."

I'd rather not link specific servers and communities.

Nutomic:
Updated: https://github.com/LemmyNet/lemmy-translations/pull/234

dessalines (approved):

PR #3762, opened 2026-01-16
Show error for failed reports.

### LINKED ISSUE #3700: If create report fails, show toast with error message
### Requirements

### Summary

When creating a new report fails, there is no indication in the UI and only the browser console shows an error message. It should display a toast. 

Related to https://github.com/LemmyNet/lemmy/issues/6214. 

### Steps to Reproduce

1. Create a new post report
2. This should work fine and show a success toast
3. Create a second report for the same post

### Technical Details

Creating a duplicate report will fail as expected, however there is no error toast in the UI.

### Lemmy Instance Version

1.0.0

### Lemmy Instance URL

_No response_

dessalines:
- Fixes #3700

### DIFF
--- a/src/shared/utils/app.ts
+++ b/src/shared/utils/app.ts
@@ -30,6 +30,10 @@ import {
   PostListingMode,
   MultiCommunity,
   MultiCommunityView,
+  CommentReportResponse,
+  PostReportResponse,
+  PrivateMessageReportResponse,
+  CommunityReportResponse,
 } from "lemmy-js-client";
 import {
   CommentNodeI,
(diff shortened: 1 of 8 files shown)


src/shared/utils/app.ts
```diff
@@ -668,6 +674,21 @@ export async function pictrsDeleteToast(filename: string) {
...
 }
 
+export function reportToast(
+  res: RequestState<
+    | CommentReportResponse
+    | PostReportResponse
+    | PrivateMessageReportResponse
+    | CommunityReportResponse
+  >,
+) {
```
Nutomic, line 684: You can change this to take `CreateCommentReport | CreatePostReport | ...` and then make the API call here. This way you can get rid of all the identical `handleCommentReport()` etc methods. It needs an if else chain to call the right endpoint for each form, but thats still less code than now.

dessalines: That's not advisable, because API calls should always be made from the screens, and also because it wouldn't be too easy to set the different states (sometimes necessary), from the result of API calls.

I know its tempting because I've dealt with duped API calls here and with jerboa, but it can't really be done cleanly when you also need to account for state.

Nutomic (approved):

PR #3763, opened 2026-01-16
Fixed Development Setup Instructions in README

dankxiaobong:
## Description

`pnpm translations:init` needs to be run before running pnpm dev in order to set up the dev environment, otherwise you get an error that translations are missing

Added the information to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,9 @@ You can find more information in the [Backend Repository](https://github.com/Lem
 You need to have [pnpm](https://pnpm.io/installation) installed. Then run the following:
 
 ```bash
-git clone https://github.com/LemmyNet/lemmy-ui.git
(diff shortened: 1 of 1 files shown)


Nutomic:
Thanks, good catch! Instead of this you can change the first line to `git clone https://github.com/LemmyNet/lemmy-ui.git --recursive`. That is one line less, and ensures you have the same submodule version like in the repo. Because `translations:update` runs `git submodule update --remote --recursive` which pulls the latest version from https://github.com/LemmyNet/lemmy-translations, which will show as uncommitted change with `git status`. 

Its also worth adding a short explanation how `git submodule` works (mainly running `git submodule update` if lemmy-translations shows as modified. And mention that new translations need to be added in a separate PR to https://github.com/LemmyNet/lemmy-translations.

Issue #3592, opened 2025-11-05
Get rid of button groups and selects in favor of dropdowns with icons


dessalines:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

I'll try to follow [google's material 3 guidelines](https://m3.material.io) when I do #3592 

https://m3.material.io/components/chips/guidelines

- [ ] Create a FilterChip component, that uses `btn-light` with [bootstraps radio checks and buttons](https://getbootstrap.com/docs/5.3/forms/checks-radios/#radio-toggle-buttons). Make it look like material filter chips.
- [ ] Create a FilterChipDropdown if its more than two items.
- [ ] Get rid of all radio button groups, and ones that don't flex-reflow.
- [ ] Get rid of most select dropdowns

### Describe the feature you'd like

In lemmy-ui, we have a few cases of bootstrap button groups, which don't look good on mobile due to wrapping, and take up too much space. 

One example is the new instances type filter, but there are others on home and community:

![Image](https://github.com/user-attachments/assets/ee0dfbb8-f761-4ee7-b3b7-53798f27190c)

This should be done like jerboa, using dropdown menus, probably with an icon instea of text:

![Image](https://github.com/user-attachments/assets/633f49ab-a93b-4f15-98c6-8c683c043438)

Nutomic:
Button groups make sense for commonly used actions like switching between Subscribed/Local/All feeds because they require less clicks. For less frequently used actions it makes sense to switch to dropdown.

Its also worth noting that the wrapping and spacing problem only exists on mobile, on large screens it works fine. So what we could do is change `radio-button-group.tsx` to render as button group on medium and large screens, while rendering as dropdown on small screens.

dessalines:
To add some more context here: I'd also like to get rid of most selects (except maybe the searchable / input ones), in favor of these dropdowns with icons. 

Examples:

android menu | bootstrap dropdown | bootstrap select
--- | --- | ---
[...]

PR #3766, opened 2026-01-16
Don't allow post_url_only and post_title_only to be checked at same time in search.

### LINKED ISSUE #3619: [1.0] Dont allow setting post_title_only and post_url_only at the same time
Noticed this just after merging https://github.com/LemmyNet/lemmy-ui/pull/3616

dessalines:
- Fixes #3619


Nutomic:
In terms of UX this would be better as a button group or similar.

dessalines:
True, the filters bars need to be reworked anyway to be more responsive. I'll probably to that as part of #3592

PR #3767, opened 2026-01-16
Add /feed urls to home, profile, community, and multicommunity.

### LINKED ISSUE #3215: Add feed / rss url redirects
### Requirements

- [x] Is this a feature request? For questions or discussions use https://lemmy.ml/c/lemmy_support or the [matrix chat](https://matrix.to/#/#lemmy:matrix.org).
- [x] Did you check to see if this issue already exists?
- [x] Is this only a feature request? Do not put multiple feature requests in one issue.
- [x] Is this a backend issue? Use the [lemmy-ui](https://github.com/LemmyNet/lemmy-ui) repo for UI / frontend issues.
- [x] Do you agree to follow the rules in our [Code of Conduct](https://join-lemmy.org/docs/code_of_conduct.html)?

Many sites have a simple rss /feed that can be added to the end of existing urls.

Although we have rss buttons in the UI, it'd be nice to also have these redirects.

Add /feed redirects for the front page, communities, and users.

NA

### Additional context

_No response_

dessalines:
- You can now go to any of the above pages, add `/feed` to the url, and it will redirect to the correct back-end lemmy rss feed.
- Fixes #3215

### DIFF
--- a/src/server/index.tsx
+++ b/src/server/index.tsx
@@ -87,10 +87,10 @@ server.get("/manifest.webmanifest", ManifestHandler);
 server.get("/css/themes/:name", ThemeHandler);
 server.get("/css/code-themes/:name", CodeThemeHandler);
 server.get("/css/themelist", ThemesListHandler);
-server.get("/feed", FrontPageFeedHandler);
-server.get("/u/:name/feed", ProfileFeedHandler);
-server.get("/c/:name/feed", CommunityFeedHandler);
-server.get("/m/:name/feed", MultiCommunityFeedHandler);
+server.get(["/feed", "/.rss"], FrontPageFeedHandler);
+server.get(["/u/:name/feed", "/u/{:name}.rss"], ProfileFeedHandler);
+server.get(["/c/:name/feed", "/c/{:name}.rss"], CommunityFeedHandler);
+server.get(["/m/:name/feed", "/m/{:name}.rss"], MultiCommunityFeedHandler);
 server.get("/{*splat}", CatchAllHandler);
 
 const listener = server.listen(Number(port), hostname, () => {


Nutomic:
I tried this with Mastodon and Discourse, they both use .rss:
- https://mastodon.social/@pixelfed.rss
- https://socialhub.activitypub.rocks/c/welcome.rss

src/server/handlers/profile-feed-handler.ts
```diff
@@ -0,0 +1,7 @@
+import { profileRSSUrl } from "@utils/app";
+import type { Request, Response } from "express";
+
+export default async (req: Request, res: Response) => {
+  const name = req.params.name;
+  res.redirect(profileRSSUrl(name));
+};
```
Nutomic: No need to make a separate file for each of these, just use `src/server/handlers/feed-handlers.ts`

src/shared/utils/app.ts
```diff
@@ -153,6 +162,37 @@ export function multiCommunityRSSUrl(
...
+export function profileRSSUrl(
+  username: string,
+  sort: SearchSortType = "new",
+): string {
+  return httpBackendUrl(`/feeds/u/${username}.xml${getQueryString({ sort })}`);
+}
+
+export function notificationsRSSUrl(auth: string): string {
+  return httpBackendUrl(`/feeds/notifications/${auth}.xml`);
+}
```
Nutomic, line 200: Better to define all this directly in the handler file, then its all in one place, easy to find and its not in the way here.

src/server/handlers/profile-feed-handler.ts
dessalines: K, will do shortly.

src/shared/utils/app.ts
dessalines: Some of them are used elsewhere, and not used by the feed handler. So I wanted to keep all the RSS shortcuts together in one place.

Nutomic:
What about this?

> I tried this with Mastodon and Discourse, they both use .rss:
> - https://mastodon.social/@pixelfed.rss
> - https://socialhub.activitypub.rocks/c/welcome.rss

dessalines:
I could add those in addition I suppose. 

The `/feed` seems to be standard on most news / blogsites. IE go to https://www.androidpolice.com/ , then add https://www.androidpolice.com/feed and you get the rss feed.

I'll add a `.rss` extension now tho, IE things like `/c/{community_name}.rss`

Nutomic (approved):

PR #3780, opened 2026-01-17
Removing all `bind`s from codebase 

### LINKED ISSUE #3578: Remove all constructor `bind`s and internal class method event handlers

dessalines:
See below issue for context

- Dependent on #3655 
- Fixes #3578

This was a larger refactor done over the weekend, but I've removed all but a few `.bind`s from the codebase, and also fixed a lot of previously hidden bugs in the process.

The diffs aren't clean because all the handler functions were moved to the bottom, but very little logical changes were necessary. 

I tested and everything seems to work fine.

### DIFF
--- a/src/shared/components/common/markdown-textarea.tsx
+++ b/src/shared/components/common/markdown-textarea.tsx
@@ -554,7 +554,7 @@ function handleKeyBinds(i: MarkdownTextArea, event: KeyboardEvent) {
       }
       case "Enter": {
         if (!i.isDisabled) {
-          handleSubmit(i);
+          handleSubmit(i, event);
         }
 
         break;


dessalines:
That lint looks like its for something else (not allowing binds in jsx), but I prefer never allowing binds even in regular code.

I'll merge as this is a big refactor.

PR #3808, opened 2026-01-26
Show SSO buttons on registration page (fixes #3285)

### LINKED ISSUE #3285: Verify that new oauth functionality works.
### Requirements

We added new `createOauthProvider` and `authenticateWithOauth` endpoints, but these need to be fully tested for 1.0

Nutomic:
Screenshot_20260126_143859

https://github.com/LemmyNet/lemmy-translations/pull/239

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "lemmy-translations"]
 	path = lemmy-translations
 	url = https://github.com/lemmynet/lemmy-translations
-	branch = community-tags
+	branch = main


src/shared/components/home/signup.tsx
```diff
@@ -140,13 +130,38 @@ export class Signup extends Component<SignupRouteProps, State> {
             {this.registerForm()}
           </div>
         </div>
+        {(this.state.siteRes.oauth_providers?.length || 0) > 0 && (
```
dessalines: Isn't this equivalent to what's on the login.tsx page? Would be good to extract into a component.

Nutomic: Makes sense, done.

src/shared/components/home/oauth/oauth-login.tsx
```diff
@@ -0,0 +1,85 @@
...
+              )}
+            </div>
+          </div>
+        </>
+      )
+    );
+  }
+}
+
+export async function handleLoginWithProvider(
```
dessalines: I don't think this needs to be exported, or async.

src/shared/components/home/oauth/oauth-login.tsx
```diff
@@ -0,0 +1,85 @@
+import { I18NextService } from "@services/I18NextService";
+import { setIsoData } from "@utils/app";
+import { NoOptionI18nKeys } from "i18next";
+import { Component } from "inferno";
+import { PublicOAuthProvider } from "lemmy-js-client";
+import { v4 as uuidv4 } from "uuid";
+
+type OAuthLoginProps = { title: NoOptionI18nKeys };
+
+export default class OAuthLogin extends Component<OAuthLoginProps, object> {
```
dessalines: If there's no state, then you don't need to use a class / inferno component, you can do a functional component IE `export function OauthLogin({title}: OAuthLoginProps) {...`. There are lots of these in the codebase.

src/shared/components/home/oauth/oauth-login.tsx
```diff
@@ -0,0 +1,85 @@
...
+  public isoData = setIsoData(this.context);
```
dessalines: Don't do this, pass the oauth providers as a prop to this component.

src/shared/components/person/settings.tsx
```diff
@@ -403,7 +403,7 @@ export class Settings extends Component<SettingsRouteProps, SettingsState> {
         id="settings-tab-pane"
       >
         <div className="row">
-          {!userNotLoggedInOrBanned(this.isoData.myUserInfo) && (
+          {!userNotLoggedInOrBanned(this?.isoData?.myUserInfo) && (
```
dessalines: Changes here probably aren't needed.

src/shared/components/home/oauth/oauth-login.tsx
dessalines: Also try to avoid using default, these become really annoying to trace down.

Nutomic: Changed, although there are many other classes doing this.

Nutomic:
Made these changes, and also changed it to use only a single string for signup/login.

dessalines:
Going to fix and merge this, because the removed translations is breaking other PRs.

PR #3810, opened 2026-01-26
Responsive design rework for dropdowns and filters

### LINKED ISSUE #3592: Get rid of button groups and selects in favor of dropdowns with icons

### LINKED ISSUE #3807: [1.0] User settings page is broken
https://voyager.lemmy.ml/settings

```
TypeError: Cannot read properties of undefined (reading 'isoData')
at userSettings (webpack://lemmy-ui/src/shared/components/person/settings.tsx:406:42)
at _ref3 (webpack://lemmy-ui/src/shared/components/common/tabs.tsx:58:20)
at Array.map (<anonymous>)
at Tabs.render (webpack://lemmy-ui/src/shared/components/common/tabs.tsx:57:28)
at renderVNodeToString (/home/felix/workspace/lemmy/lemmy-ui/node_modules/.pnpm/inferno-server@9.0.10/node_modules/inferno-server/dist/index.cjs:189:35)
at renderVNodeToString (/home/felix/workspace/lemmy/lemmy-ui/node_modules/.pnpm/inferno-server@9.0.10/node_modules/inferno-server/dist/index.cjs:279:29)
at renderVNodeToString (/home/felix/workspace/lemmy/lemmy-ui/node_modules/.pnpm/inferno-server@9.0.10/node_modules/inferno-server/dist/index.cjs:200:14)
at renderVNodeToString (/home/felix/workspace/lemmy/lemmy-ui/node_modules/.pnpm/inferno-server@9.0.10/node_modules/inferno-server/dist/index.cjs:200:14)
at renderVNodeToString (/home/felix/workspace/lemmy/lemmy-ui/node_modules/.pnpm/inferno-server@9.0.10/node_modules/inferno-server/dist/index.cjs:276:27)
at renderVNodeToString (/home/felix/workspace/lemmy/lemmy-ui/node_modules/.pnpm/inferno-server@9.0.10/node_modules/inferno-server/dist/index.cjs:200:14)
```

Probably related to https://github.com/LemmyNet/lemmy-ui/pull/3780

dessalines:
## Description

- lemmy-ui needed a responsive design rework related to the home page using large radio buttons, and other pages using radio button groups, selects, and checkboxes that don't look good on mobile. We should be responsive-design first, as most ppl are using lemmy with mobile devices anyway.
  - To fix this I followed the [google material design 3 docs on filter chips and filter chip dropdowns](https://m3.material.io/components/chips/guidelines), and replaced all our existing checkboxes and selects with consistent material 3 style chips.
- Adds `FilterChipDropdown` and `FilterChipCheckbox` components, with a consistent UI, replacing all the other html selects. 
- Replaced a lot of the other selects like `sort-select` with components like `sort-dropdown` that call `FilterChipDropdown`, and type-check all the options and i18n strings.
- Adds i18n strings for the modlog actions.
- All our buttons now are `btn-light border-light-subtle` instead of `btn-secondary`. These are much softer and less jarring.
- All filter rows are now using `row row-cols align-items-center` for a smart flex reflowing row.
- Fixes #3592
- Fixes #3807 

### Default strings

A lot of chips just show "all" as the default / selected option, and its not clear what type of filter it is until you click into it. Personally I think its fine since these can be discovered once and then ignored. But some could benefit from "all_ITEM". I've done so with the community search and creator searchable filters on the search page by having the all option show `All communities`. But in a lot of cases, its not clear what the string would be.

### Icons

Currently I've chosen to not use any icons (except the X or check), either for the dropdown options, or the chip itself. IMO it looks cleaner, but icons could be added to these in the future.

# Pics

## Home

Screenshot_2026-01-27-14-08-29-120_mark via

## Home mobile

Screenshot_2026-01-27-14-08-29-121_mark via

## Communities 

Screenshot_2026-01-27-14-08-41-102_mark via

## Multi-communities 

Screenshot_2026-01-27-14-08-51-675_mark via

## Search with dropdown 

Screenshot_2026-01-27-14-09-24-454_mark via

## Notifications

Screenshot_2026-01-27-14-09-37-832_mark via

## Reports

Screenshot_2026-01-27-14-09-53-230_mark via

## Profile

Screenshot_2026-01-27-14-10-14-225_mark via

## Settings

Screenshot_2026-01-27-14-10-36-555_mark via


src/shared/components/common/modlog-kind-dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+      className={className}
+    />
+  );
+}
+
+// TODO this should be got rid of.
+export function modlogKindOrAllToModlogKind(
+  t: ModlogKindOrAll,
+): ModlogKind | undefined {
+  return t === "all" ? undefined : t;
```
dessalines: Dependent on https://github.com/LemmyNet/lemmy/pull/6300

Nutomic:
- The select elements are too small and need more padding. They should have the same height as buttons and text inputs.
- In darkly theme the buttons should be more rounded like before. Litely looks okay or could be a bit less rounded
- Comment sort still uses button group
- Would be neat to use `<select>` design as described in https://github.com/LemmyNet/joinlemmy-site/issues/568
- The new design for checkboxes might be confusing at first, but probably not hard to get used to it
- The buttons for post sort etc on the homepage definitely need more work. We should make a post "Help Design Lemmy" to get feedback on that (eg which ones are most commonly used).
- Similarly gather feedback to improve settings design.

dessalines:
> The select elements are too small and need more padding.

I'd like to keep these smaller than the normal buttons, and google's material 3 recommendations also do selects / filters as chips, and smaller than normal buttons. Normal size buttons for actions, smaller chips for filters.

> In darkly theme the buttons should be more rounded like before. Litely looks okay or could be a bit less rounded

I haven't changed anything with the themes or the button roundness, they're the same as before. I believe some of those things are theme designed, so shouldn't be a part of this PR. I'd like to just keep the buttons as `btn-light border-light-subtle`, and not override too many things, and let the themes do most of their styling.

> Would be neat to use select design as described in ...

I'm not a big fan of those, and far prefer the google material 3 filter chips. There are some places where I could make the default option / title a bit clearer tho, as mentioned in my caveats section. Should be a case-by-case basis.

> The buttons for post sort etc on the homepage definitely need more work.

I'm guessing this is the same issue as above.

> Comment sort still uses button group

Whoops, good catch, thx. I'll fix this shortly.

dessalines:
K this is ready to go.

The multi-select I'd rather do in another ticket.

Nutomic:
> I'd like to keep these smaller than the normal buttons, and google's material 3 recommendations also do selects / filters as chips, and smaller than normal buttons. Normal size buttons for actions, smaller chips for filters.

Had a look at the classes, problem is that you are using `btn-small` everywhere. That makes sense for secondary actions, but not for primary actions like switching between subscribed/local/all feed. Especially on mobile this will be very hard to use. And on `/settings` the tiny dropdown buttons look completely out of place compared to the normal size inputs for file upload, text, and submit button. So just get rid of the `btn-small`, that also fixes the roundness problem.

> I'm not a big fan of those, and far prefer the google material 3 filter chips. There are some places where I could make the default option / title a bit clearer tho, as mentioned in my caveats section. Should be a case-by-case basis.

How does this solve the problem of tag multi-select on mobile? And the problem that Ctrl+click on desktop is not intuitive?

Btw tag selection is missing from the create/edit post form now. Could be related to the api changes in https://github.com/LemmyNet/lemmy-ui/pull/3814

dessalines:
> Especially on mobile this will be very hard to use.

I've tested these on mobile, and they work fine. Again I'm following google's design docs here: smaller chips for selects / filters, and larger buttons for actions like create. Even for the settings filters they're perfectly clickable and fine. I could start adding `isSmall` props but it seems totally unecessary.

> How does this solve the problem of tag multi-select on mobile?

I'd like to push multi-select component to another ticket, this is large enough as it is. 

> Btw tag selection is missing from the create/edit post form now. Could be related to the api changes in https://github.com/LemmyNet/lemmy-ui/pull/3814

Yep I'm waiting on the API changes from that PR. I guess I'll mark this as draft until that is merged.

dessalines:
This is ready. 

I tested the post create tags, and they're working. You have to select the community first before the tags multi-select is shown.

PR #3814, opened 2026-01-28
Updating to account for changed tag and sidebars.

dessalines:
Some logical changes were necessary, such as:

- Removing markdowns forms and renderer for multi-community (it only has a summary now)
- Using summary instead of sidebar where it makes sense (like HtmlTags and document titles)

I'm not looking forward to the merge conflicts, so this should probably come after #3810

### DIFF
--- a/src/server/utils/generate-manifest-json.ts
+++ b/src/server/utils/generate-manifest-json.ts
@@ -52,7 +52,7 @@ export default async function (site: Site) {
 
   return {
     name: site.name,
-    sidebar: site.sidebar ?? "A link aggregator for the fediverse",
+    description: site.summary ?? "A link aggregator for the fediverse",
     start_url: "/",
     scope: "/",
     display: "standalone",


src/server/utils/generate-manifest-json.ts
```diff
@@ -52,7 +52,7 @@ export default async function (site: Site) {
 
   return {
     name: site.name,
-    description: site.description ?? "A link aggregator for the fediverse",
+    sidebar: site.sidebar ?? "A link aggregator for the fediverse",
```
dessalines: This field seems unecessary, its just some untyped random thing in the manifest json, I'd be fine with removing it, or renaming it to summary and using the summary instead.

Nutomic: Description is part of the webmanifest standard, but sidebar is not.
- https://www.w3.org/TR/appmanifest/#description-member
- https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/description

dessalines:
Changed my mind, this should come before #3810 since the backend changes are already pushed to voyager.lemmy.ml

PR #3817, opened 2026-01-29
Enable typecheck lint (fixes #3812)

### LINKED ISSUE #3812: Type checked linting and check for unused await
There is a lint to warn about unused `async` in function definitions: https://typescript-eslint.io/rules/require-await/

To use this requires [typed linting](https://typescript-eslint.io/getting-started/typed-linting/) which also seems worth enabling. However it reports 1730 errors.

Nutomic:


eslint.config.mjs
```diff
@@ -45,6 +52,15 @@ export default [
...
     rules: {
       "@typescript-eslint/no-explicit-any": 0,
+      "@typescript-eslint/no-unsafe-assignment": 0,
+      "@typescript-eslint/no-unsafe-call": 0,
+      "@typescript-eslint/no-unsafe-member-access": 0,
+      "@typescript-eslint/no-unsafe-argument": 0,
+      "@typescript-eslint/no-unsafe-return": 0,
+      "@typescript-eslint/unbound-method": 0,
+      // TODO: the following two are caused by calling async functions from sync functions
+      "@typescript-eslint/no-misused-promises": 0,
```
Nutomic, line 62: Try to remove this line and run lint, you will see a lot of errors about calling async functions from event handlers, or from Component functions which are sync. Is there any proper way to call async from sync methods?

dessalines: Either [async blocks](https://stackoverflow.com/a/54758455/1655478) or `.then(d => ...`

dessalines:
This PR also changes like 50+ files so #3810 should have precedence first.

The only other big refactoring PR I have planned is #3791 , but I'll probably do that after I finish some of the smaller issues here.

Nutomic:
Conflicts fixed.

PR #3818, opened 2026-01-29
Lint unbound method

Nutomic:


dessalines:
Mark stacked PRs as draft.

Nutomic:
Had to redo this because of conflicts, now ready to merge.

src/shared/components/common/all-or-banned-dropdown.tsx
```diff
@@ -9,7 +9,7 @@ const options: FilterOption<AllOrBanned>[] = [
 
 type AllOrBannedDropdownProps = {
   currentOption: AllOrBanned;
-  onSelect(val: AllOrBanned): void;
+  onSelect(this: void, val: AllOrBanned): void;
```
dessalines: https://typescript-eslint.io/rules/unbound-method/

> Class method functions don't preserve the class scope when passed as standalone variables ("unbound"). If your function does not access this, [you can annotate it with this: void](https://www.typescriptlang.org/docs/handbook/2/functions.html#declaring-this-in-a-function), ***or consider using an arrow function instead.*** Otherwise, passing class methods around as values can remove type safety by failing to capture this.

It says we should probably be using arrow functions instead, then we can avoid the pointless `this: void` . I should've probably known this because it works this way in kotlin also.

If this is easy for you to do lmk, otherwise I can do.

Nutomic: Its not clear to me how this works for the routes, plus I dont want to redo the same thing a third time. So better you do it.

dessalines: I got it.

dessalines: https://github.com/LemmyNet/lemmy-ui/pull/3870

dessalines:
Oh nm, this did get merged to main.

PR #3819, opened 2026-01-29
Adding links and titles for community/multi-community settings.

### LINKED ISSUE #3796: On (multi-) community settings page show community name
At the moment you cannot tell which community or multi-community you are editing without looking at the url. It should also include a link to go back to the community.

dessalines:
- Fixes #3796

Screenshot_2026-01-29-11-43-45-287_mark via

### DIFF
--- a/src/shared/components/modlog.tsx
+++ b/src/shared/components/modlog.tsx
@@ -580,7 +580,7 @@ const Filter = ({
     value={value ?? 0}
     options={[
       {
-        label: I18NextService.i18n.t(title),
+        label: I18NextService.i18n.t(title) as string,
         value: "0",
       },
     ].concat(options)}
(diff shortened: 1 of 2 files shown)


src/shared/components/community/community-settings.tsx
```diff
@@ -222,34 +224,44 @@ export class CommunitySettings extends Component<RouteProps, State> {
...
+                    <>
+                      <div className="row justify-content-md-center">
+                        <h1 className="col-12 col-md-6 h4 mb-4">
+                          <CommunityLink
+                            community={getCommunityRes.community_view.community}
+                            myUserInfo={myUserInfo}
+                          />{" "}
+                          {lowerCaseFirstLetter(
+                            I18NextService.i18n.t("settings"),
+                          )}
```
Nutomic: This is not the right way to do localization. In German the noun should remain uppercase, and some languages would use a different word order. So it should use a new string with parameter for name.

dessalines: Good call.

Nutomic (approved):

PR #3820, opened 2026-01-29
Add accept terms and conditions / legal checkbox on signup.

### LINKED ISSUE #3797: Accept legal information during signup
If the instance has any non-empty `site.legal_information`, this should be shown during signup, with buttons to accept or go back. See [Mastodon for reference](https://flipboard.social/auth/sign_up).

dessalines:
- Fixes #3797

Screenshot_2026-01-29-12-16-01-355_org mozilla fennec_fdroid


src/shared/components/home/signup.tsx
```diff
@@ -276,6 +277,26 @@ export class Signup extends Component<SignupRouteProps, State> {
...
+                className="form-check-input"
+                id="register-accept-legal"
+                type="checkbox"
+                required
+              />
+              <label
+                className="form-check-label"
+                htmlFor="register-accept-legal"
+              >
+                <Link to="/legal" target="_blank">
```
dessalines: Not sure if I should be opening it in a new tab or not.

Nutomic:
Its not obvious that it is a link, and using a checkbox label as link is unusual. Instead it should show the full legal text in a box directly on the signup page, with scrollbars if its too long.

Also the one on [Mastodon](https://flipboard.social/auth/sign_up) is not only for legal terms, but also for the instance rules. It would make sense to do this as well and move the rules out of the sidebar. We could call the db field `local_site.rules_and_policies` or `rules_and_legal_info`.

dessalines:
I suppose I can slam in the markdown, but it still needs the checkbox anyway, because otherwise there's no *" agree to this*.

Nutomic:
Yes of course, I just mean its strange to have a checkbox label which is also a link.

dessalines:
Looks like this. 

Screenshot_2026-02-02-15-19-23-533_org mozilla fennec_fdroid-edit

I still kept the link because I think its useful, if they want to see the full page.

src/assets/css/main.css
```diff
@@ -516,6 +516,10 @@ em-emoji-picker {
   max-height: 100px;
 }
 
+.legal-info-box {
+  max-height: 200px;
+}
```
Nutomic, line 521: Looks like Bootstrap really doesnt have any predefined classes for fixed height, odd.

src/shared/components/home/signup.tsx
Nutomic: Again remove the link, because clicking a checkbox label should toggle the checkbox and nothing else. The entire text is already there.

src/assets/css/main.css
dessalines: They do but not for specific sizes, and only relative to the parent: https://getbootstrap.com/docs/5.3/utilities/sizing/#relative-to-the-parent

I tried out `mh-25` and it didn't do anything so I needed to add this.

Nutomic:
Still the same problem as before, clicking the checkbox label doesnt toggle the checkbox as expected. Instead it opens a new tab. If you want to include the link it clearly needs to be a separate element. For example make only the words "Terms and conditions" a link, and make it underlined with different text color.

dessalines:
K I removed the link now, the full label checks the form.

Issue #3781, opened 2026-01-17
Add quick toggle on home and community pages for hiding read posts


dessalines:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Seems to be a common complaint, that it isn't easy to hide read posts from the main pages.

This should be added.

Nutomic:
We should be careful not to add too many options though. Maybe hide some of them by default, and have an "Expand" button (arrow down) to show more options. Or a setting for advanced UI.

dessalines:
Yep, we have a LOT of filters, and they need to be made more responsive, and possibly have an expand like you're saying.

PR #3821, opened 2026-01-29
Add show read post filter to home and community pages.

dessalines:
- This lets users quickly hide read posts.
- Must come after https://github.com/LemmyNet/lemmy-ui/pull/3810
- Fixes https://github.com/LemmyNet/lemmy-ui/issues/3781

# Pics

Screenshot_2026-01-29-15-18-40-602_mark via

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -58,6 +58,7 @@ import { CommentNodes } from "./comment-nodes";
 import { BanUpdateForm } from "../common/modal/mod-action-form-modal";
 import CommentActionDropdown from "../common/content-actions/comment-action-dropdown";
 import { canAdmin } from "@utils/roles";
+import ActionButton from "@components/common/content-actions/action-button";
 
 type CommentNodeState = {
   showReply: boolean;
@@ -230,7 +231,9 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
               mark: this.isCommentNew || distinguished,
             })}
           >
-            <div className="ms-2">
+            <div
+              className={classNames({ "ms-2": this.props.viewType === "tree" })}
+            >
               {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
               <div
                 className="row text-muted small"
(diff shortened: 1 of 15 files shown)


Nutomic:
I prefer to call it "Hide read". Thats shorter, and also makes more sense with default value false.

dessalines:
Tough call.... in both the settings and the back end, its called `show`, and keeps with the positive language... but I do agree `hide` (turned off) makes more sense as the default. I'll try it out.

dessalines:
Its a little confusing to have a `hide` right next to a `show` button, but I guess I'm okay with it.

Screenshot_2026-02-02-15-47-02-667_mark via

Nutomic:
I didnt even notice that before, its very confusing. I dont use the "hide post" feature at all, so it took me some time to figure out how it works. It seems you can manually hide posts in the context menu, and with this button you can show manually hidden posts again? This is niche functionality and doesnt belong on the main page. It should look similar to user blocks.

dessalines:
> I didnt even notice that before, its very confusing. I dont use the "hide post" feature at all, so it took me some time to figure out how it works. It seems you can manually hide posts in the context menu, and with this button you can show manually hidden posts again? 

Yes I agree `read` and `hidden` are somewhat overlapping, but it is an explicit option that we should have on the front page, because sometimes people don't want to view specific posts, whether they have read them or not.

It can't be similar to user blocks, because the only and best place to view and "unhide" these posts, are from the post listings pages like home and community anyway.

I also don't use it but it is functionality we already have in the back end, and I'm sure some people use it.

Nutomic:
That doesnt make sense, if a post is explicitly hidden by the user then it should always be hidden. Or should we also add buttons to show blocked communities and blocked users? 

To unhide posts there can be a simple list in the user settings, which shows only the post title and `X` button to unhide it.

dessalines:
> if a post is explicitly hidden by the user then it should always be hidden.

You still need a place to unhide them.

> To unhide posts there can be a simple list in the user settings, which shows only the post title and X button to unhide it.

This isn't easy, it means we would have to add an entire post listing view tab to the settings page, only for hidden posts, which is way too much work. It'd essentially be re-creating the home page. Much easier to just use the filter button on the existing page.

Another solution I'd consider, is adding a `show_hidden_posts` to the user settings and `local_user` table, alongside the already existing `show_read_posts`. But it'd still be preferable to have the `show hidden posts` on the homepage as an override filter IMO, just like `show_read`.

dessalines:
I spose I'd be willing to remove the `show hidden posts` button from the main page, but only if we add a `local_user` setting in the back-end that allows you to see them. I really don't want to add this as its own view on your settings page.

Nutomic:
> This isn't easy, it means we would have to add an entire post listing view tab to the settings page, only for hidden posts, which is way too much work. It'd essentially be re-creating the home page. Much easier to just use the filter button on the existing page.

Thats not difficult at all, it only needs a function similar to `blockedUsersList()` in settings.tsx, which loops through the posts and shows the post title + maybe community name. No need to show everything like in a normal post listing.

dessalines:
Open that up as a separate issue then and copy paste your comments, because its separate from this read posts filter.

Nutomic:
On second thought my suggestion might be too complicated and would potentially require a backend change to allow fetching only hidden posts. Still the problem is that we dont have much space up there, especially on mobile. 

So I would move the button "show hidden posts" down to the bottom, near the pagination buttons. That makes sense because you can see and click it after looking through the post list and not finding the post you want. Similarly it makes sense to add "show hidden posts" button in the search results.

dessalines:
> So I would move the button "show hidden posts" down to the bottom, near the pagination buttons.

I don't think that's a good idea. All the filters should be centralized in one place, in the filter bar.

Look at the [UI examples on this page for filters chips](https://m3.material.io/components/chips/guidelines), they all look clean, and are mobile friendly, which is what I'm trying to create.

Nutomic:
Then some of these buttons should be hidden by default, using a button to expand (and state can be saved in local storage). Because there are simply too many options now especially for small mobile screens. But also on large screens it can be confusing to see so many buttons.

Screenshot_20260204_144740

Besides it doesnt use horizontal space properly as you can see.

dessalines:
Open up a different issue with this, and I can maybe add an expand button for mobile.

Nutomic:
https://github.com/LemmyNet/lemmy-ui/issues/3848

dessalines:
Bump.

PR #3823, opened 2026-01-30
UX: Updated Color and Added Icons to Site-Sidebar and Community-Sidebar Buttons

dankxiaobong:
## Description
UX: Updated Color and Added Icons to Site-Sidebar and Community-Sidebar Buttons

### Before
Site-Sidebar
image
Community-Sidebar
image

### After
Site-Sidebar
image
Community-Sidebar
image

### DIFF
--- a/src/shared/components/common/content-actions/create-item-buttons.tsx
+++ b/src/shared/components/common/content-actions/create-item-buttons.tsx
@@ -36,7 +36,7 @@ export function CreatePostButton({
   communityView,
   myUserInfo,
 }: CreatePostButtonProps) {
-  const classes = classNames("btn btn-secondary d-block mb-2 w-100", {
+  const classes = classNames("btn btn-primary d-block mb-2 w-100", {
     "no-click":
       communityView?.community.deleted ||
       communityView?.community.removed ||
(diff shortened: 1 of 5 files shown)


dessalines:
This will have to come after the other current PRs (like #3810), and will likely have conflicts anyway.

I'm not sure how I feel about the icons (I suppose they're okay), but I'm trying to minimize our color usage and use a more muted look, and reserve using colors only for our core entities: communities and users.

dankxiaobong:
Thank you for looking at the PR! I'll look into adding the icons after #3810. 

My 2 cents would be:
- Adding Icons to the sidebar buttons makes adds consistency as the icons are used with similar actions on other parts of the site (e.g. comment/post actions)
- Giving the  `Create Post`, `Create Community` and `Create Multi-Community` some visual weight compared to the other sidebar buttons, makes it visually easier for user/persons to perform the more commonly used action
- The "Danger" actions as `remove`and `purge community` are currently to heavy and using the outline for the buttons instead is consistent with the visual language of other site (e.g. github when deleting a repo)

dessalines:
I do agree with a lot of these. You can always open a new PR based on that branch.

Issue #3581, opened 2025-10-30
Consistent design for notification items


Nutomic:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

Notification items reuse the existing views which are not consistent. For example button sizes and margins are different, private messages have a border on top.

Here is an example, first post then comment and finally private message:
<img width="1146" height="514" alt="Image" src="https://github.com/user-attachments/assets/e1760e65-4386-4ec7-aef0-58ff18fb6d02" />

dessalines:
It could be very difficult to try to create a generic notification component, because you might lose a lot of the functionality included in the typed components. The dropdowns, actions, votes, special displays, context...

I have either another issue or a comment about this, but the `markable` / mark as read capability is the main thing that needs to be uniform. Whether that exists inside the `CommentNode` / `PostListing`, or whether it lives outside as a checkbox somewhere outside the item, is the main thing.

Also tho, this is a good use-case for continuing to make the components have a uniform style. The comment node looks out of place because its action buttons are on the left, unlike the post listing (which should also allow voting there btw).

The private message also doesn't have action bars or the time component aligned to the right, and has a pointless collapse button.

Nutomic:
It would make sense to handle mark as read via a button on the notification item, instead of using the buttons for marking post/comment as read. Because all of them need to go through `/api/v4/account/notifications/mark_as_read` and can reuse the same logic, avoiding conversions from post_id to notification_id etc. But then we would have to merge the notification and post action buttons together somehow.

dessalines:
We might just be forced to use / add the `markable` prop on all these, and put it in their own action bars.

dessalines:
Gonna start working on this next.

Issue #3793, opened 2026-01-21
Add `createOrSaveLoading` and `deleteLoading` props to all necessary forms


dessalines:
### Requirements

- [x] This is a feature request and not a bug report. Otherwise, please create a new [bug report](https://github.com/LemmyNet/lemmy-ui/issues/new?assignees=&labels=bug%2Ctriage&projects=&template=BUG_REPORT.yml) instead.
- [x] Please [check](https://github.com/LemmyNet/lemmy-ui/issues) to see if this request (or a similar one) already exists.
- [x] It's a single feature. Please don't request multiple features in one issue.

### Describe the feature you'd like

The loading indicators on a few forms are broken, because they aren't using loading states from the API calls.

I added a lot of these in a recent refactor, but there are still some missing.

PR #3842, opened 2026-02-03
Adding ability to mark comments as read from the notifications screen.

dessalines:
- Fixes https://github.com/LemmyNet/lemmy-ui/issues/3783
- Adds a `ShowMarkReadType` with hide, dropdown, or main_bar options as
a helper type.

Going to leave the visual improvements on the notifications scree for #3581 , which I'll do next

Screenshot_2026-02-03-14-06-04-192_mark via

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -3,7 +3,7 @@ import { numToSI } from "@utils/helpers";
 import { futureDaysToUnixTime } from "@utils/date";
 import classNames from "classnames";
 import { isBefore, parseISO, subMinutes } from "date-fns";
-import { Component, InfernoMouseEvent, InfernoNode, linkEvent } from "inferno";
+import { Component, InfernoMouseEvent, InfernoNode } from "inferno";
 import { Link } from "inferno-router";
 import {
   AddAdmin,
@@ -234,7 +234,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
               {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions */}
               <div
                 className="row text-muted small"
-                onClick={linkEvent(this, handleCommentCollapse)}
+                onClick={event => handleCommentCollapse(this, event)}
                 aria-label={this.expandText}
                 role="group"
               >
--- a/src/shared/components/app/app.tsx
+++ b/src/shared/components/app/app.tsx
@@ -1,5 +1,5 @@
 import { isAnonymousPath, isAuthPath, setIsoData } from "@utils/app";
-import { Component, createRef, linkEvent } from "inferno";
+import { Component, createRef } from "inferno";
 import { Provider } from "inferno-i18next-dess";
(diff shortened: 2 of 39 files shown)


Nutomic:
- Bug: if there are multiple unread notifications and you click the checkmark on one, it shows a loading indicator on all checkmark icons (also those of other notifications).
- After clicking mark as read/unread, it changes the state of a different notification item. Not sure if it sends the wrong id to the server, or updates the wrong element in the response handler.
- Dropdown for Unread/All doesnt make sense with only two options. Should be a checkbox or button. 
- "Mark as read" button doesnt trigger reload.

src/shared/components/comment/comment-node.tsx
```diff
@@ -64,6 +65,7 @@ type CommentNodeState = {
   collapsed: boolean;
   viewSource: boolean;
   showAdvanced: boolean;
+  // TODO get rid
   fetchChildrenLoading: boolean;
```
Nutomic, line 69: Dont forget this.

dessalines: These are for #3793 , I just add TODOs where I see them.

dessalines:
> Bug: if there are multiple unread notifications and you click the checkmark on one, it shows a loading indicator on all checkmark icons (also those of other notifications).

I checked this out, and I think it has to do with a notification bug where some notifs are duped. Did you check to make sure the two notifs you checked have different ids?

> After clicking mark as read/unread, it changes the state of a different notification item. Not sure if it sends the wrong id to the server, or updates the wrong element in the response handler.

This seems like the same issue as above, but I can't replicate this:


https://github.com/user-attachments/assets/1214dc69-5aea-45a5-b3b3-b79eb88675eb

> Dropdown for Unread/All doesnt make sense with only two options. Should be a checkbox or button.

This one I want to keep a dropdown to clarify the two options. If it were a boolean button, it would only be "[x] Show all", and the unclicked state of that wouldn't be clear. 

I'll fix the mark all as read one now.

Nutomic:
> I checked this out, and I think it has to do with a notification bug where some notifs are duped. Did you check to make sure the two notifs you checked have different ids?

Checked it, the notifications have different ids but they all refer to the same comment. Will check the backend code and try to fix that.

> This one I want to keep a dropdown to clarify the two options. If it were a boolean button, it would only be "[x] Show all", and the unclicked state of that wouldn't be clear.

Its too complicated because now you need two clicks, instead of one for a checkbox. Its also confusing to have two buttons next to each other both saying "All" (see https://github.com/LemmyNet/joinlemmy-site/issues/568).

Nutomic:
Strange, I cant reproduce the duplicate notifications at all. At least mark as read buttons are working correctly now.

dessalines:
> Its also confusing to have two buttons next to each other both saying "All"

`Unread` is always the default for that component anyway. Anyways a boolean button would either have to be `x Show all` (which is ambiguous when you don't see the other option), or `x Show read`, which isn't clear either.

Nutomic:
Opened another issue, we can merge this one.

https://github.com/LemmyNet/lemmy-ui/issues/3849

PR #3846, opened 2026-02-04
Show warning in community if instance is dead

Nutomic:
Screenshot_20260204_133031

Suggested in https://lemmy.ml/post/42651625/23746897. It would also be good to disable post, comment, vote and other actions in this case, but thats more complicated to implement.

https://github.com/LemmyNet/lemmy-translations/pull/247

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "lemmy-translations"]
 	path = lemmy-translations
 	url = https://github.com/lemmynet/lemmy-translations
-	branch = hide_read_posts
+	branch = main


src/shared/components/community/community.tsx
```diff
@@ -539,6 +527,40 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+        ),
+      };
+    }
+
+    const oneWeekAgo = new Date(new Date().getTime() - 7 * 24 * 60 * 60 * 1000);
+    const deadInstance =
+      res.site && new Date(res.site?.last_refreshed_at) < oneWeekAgo;
+    if (!local && deadInstance) {
+      return {
+        title: "comunity_dead_instance_title",
```
dessalines: `community`

src/shared/components/community/community.tsx
```diff
@@ -539,6 +527,40 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+        body: I18NextService.i18n.t("comunity_dead_instance_body"),
```
dessalines: Same

src/shared/components/community/community.tsx
```diff
@@ -201,6 +201,11 @@ function getShowHiddenFromQuery(hidden: string | undefined): boolean {
   return hidden === "true";
 }
 
+interface Warning {
+  title: NoOptionI18nKeys;
+  body; // TODO: type view
```
dessalines: This type isn't necessary.

src/shared/components/community/community.tsx
```diff
@@ -539,6 +527,40 @@ export class Community extends Component<CommunityRouteProps, State> {
     );
   }
 
+  checkShowWarning(res: GetCommunityResponse): Warning | undefined {
```
dessalines: Make this a functional component (or an internal function) that return tsx. There's no point in having it return a weird `Warning` type.

Then it'll look like `<ShowWarnings res={communityRes} />`

MrKaplan-lw:
related: https://github.com/LemmyNet/lemmy/issues/5789

src/shared/components/community/community.tsx
```diff
@@ -1356,3 +1331,46 @@ function updateModerators(
...
+      <T className="d-inline" i18nKey="community_not_federated_message">
+        #{communityName(community)}
+        <a href="https://lemmy-federate.com">#</a>
+      </T>
+    );
+  }
+
+  const oneWeekAgo = new Date(new Date().getTime() - 7 * 24 * 60 * 60 * 1000);
+  const deadInstance =
+    res.site && new Date(res.site?.last_refreshed_at) < oneWeekAgo;
```
Nutomic, line 1398: Should use `community.last_refreshed_at`, because site may be null for other Fediverse platforms.

Edit: Actually that could also be wrong if its a very inactive community. Should really check both of them.

src/shared/components/community/community.tsx
```diff
@@ -1356,3 +1331,47 @@ function updateModerators(
     return s;
   });
 }
+
+const ShowWarning = ({ res }: { res: GetCommunityResponse }) => {
```
Nutomic: Not sure whats the right syntax here.

dessalines: For all functional components in typescript, you type the props like this. 

Screenshot_2026-02-06-09-16-07-271_com termux

The destructuring is optional, you could also do `function ....(props: ComponentProps)`

Also don't use `const`, these are functions.

src/shared/components/community/community.tsx
```diff
@@ -1356,3 +1331,47 @@ function updateModerators(
...
+  if (!res || !res.community_view) {
+    return;
+  }
```
dessalines: Doesn't seem necessary. Anyways you should make sure the res state is `success` before displaying this block anyway.

src/shared/components/community/community.tsx
```diff
@@ -1356,3 +1331,47 @@ function updateModerators(
...
+  const community = res.community_view.community;
+  // Show a message to the moderator if this community is not federated yet (ie it has no
+  // remote followers).
+  const notFederated =
+    res.community_view.can_mod &&
+    community.subscribers === community.subscribers_local &&
+    community.visibility !== "local_only_public" &&
+    community.visibility !== "local_only_private";
+  let title: NoOptionI18nKeys | undefined;
+  let body: InfernoNode;
```
dessalines: Don't do this, just use the body below in the tsx block.

src/shared/components/community/community.tsx
```diff
@@ -1356,3 +1331,47 @@ function updateModerators(
...
+  if (community.local && notFederated) {
+    title = "community_not_federated_title";
+    body = (
+      <T className="d-inline" i18nKey="community_not_federated_message">
+        #{communityName(community)}
+        <a href="https://lemmy-federate.com">#</a>
+      </T>
+    );
+  }
```
dessalines: All this should be inside the tsx block

src/shared/components/community/community.tsx
Nutomic: You mean like `I18NextService.i18n.t(notFederated ? "community_not_federated_title" : "dead_community_title")`? Tried it but its much harder to read, and wont work if we add another warning type. So left it like this.

dessalines: I'll do a PR to this one. It isn't good to do tsx outside of the render / return block.

Nutomic: Sadly a lot of the alert HTML is duplicated now.

src/shared/components/home/home.tsx
```diff
@@ -900,7 +900,7 @@ export class Home extends Component<HomeRouteProps, HomeState> {
             </div>
             <div className="col">
               <FilterChipCheckbox
-                option={"hide_read_posts"}
+                option={"show_read_posts"}
```
dessalines: This got messed up in the merge, I'll fix.

dessalines:
```suggestion
                option={"hide_read_posts"}
```

src/shared/components/community/community.tsx
```diff
@@ -790,7 +766,7 @@ export class Community extends Component<CommunityRouteProps, State> {
             </div>
             <div className="col">
               <FilterChipCheckbox
-                option={"hide_read_posts"}
+                option={"show_read_posts"}
```
dessalines:
```suggestion
                option={"hide_read_posts"}
```

PR #3855, opened 2026-02-06
Add dropdown labels

dessalines:
## Description

Adds an optional dropdown label for the `FilterChipDropdown`.

I've only added it for sort and type, but others could be added later on a case-by-case basis

Screenshot_2026-02-06-10-47-19-126_org cromite cromite

Also fixes the showing suggested if none exists.


Nutomic:
It looks a bit like a button group instead of a single button with label. Try to get rid of the vertical line in the middle and add `:` (ie `Type: Local`). Also the current value can be bold like on https://joinmastodon.org/servers

dessalines:
Makes sense: here that is:

Screenshot_2026-02-06-21-47-23-803_org cromite cromite

The selected item is already bolded in the dropdown menu. I'd rather not bold it in the button title.

Nutomic:
Looks good, but most of the dropdowns are still missing labels. Eg in notifications, user profile, multi-communities, list vs grid view.

In the settings there are labels on the dropdown now, which looks inconsistent as there is already a separate label next to the button.

dessalines:
> In the settings there are labels on the dropdown now, which looks inconsistent as there is already a separate label next to the button.

I'll fix. 

> Eg in notifications, user profile, multi-communities, list vs grid view.

I'll try to get some of these, although its not always clear what the title should be.

dessalines:
K, fixed.

dessalines:
Yep I don't have a good idea for that one either, and I'm trying to avoid long label names.

PR #3866, opened 2026-02-09
Dead instance warning dess

dessalines:
## Description

This extracts the two types of warnings into components so as not to inline tsx bodies.

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -53,6 +53,7 @@ import {
   PagedResponse,
   CommentView,
   GetCommunity,
+  Community as CommunityI,
   GetCommunityResponse,
   GetPosts,
   PostView,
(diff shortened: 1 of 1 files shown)


src/shared/components/community/community.tsx
```diff
@@ -1343,33 +1344,48 @@ function ShowWarning({ res }: ShowWarningProps) {
...
     community.visibility !== "local_only_private";
-  let title: NoOptionI18nKeys | undefined;
-  let body: InfernoNode;
-  if (community.local && notFederated) {
-    title = "community_not_federated_title";
-    body = (
-      <T className="d-inline" i18nKey="community_not_federated_message">
-        #{communityName(community)}
-        <a href="https://lemmy-federate.com">#</a>
-      </T>
```
dessalines, line 1354: Doing this is okay for simple titles or strings, but its bad practice when you have to inject entire tsx blocks, and indicates that components should be split up.

PR #3870, opened 2026-02-12
lint-unbound-method: Use arrow functions for events

dessalines:


Nutomic:
Seems to be missing the change in eslint config.

dessalines:
Its a PR to yours, and in that one you just removed the `unbound-method: 0` line, so it is checking.

PR #3871, opened 2026-02-13
Put show subscribed and show_sidebar on the same line as other selects.

### LINKED ISSUE #3848: Rework post listing options
Screenshot_20260204_144740

There are too many buttons which is confusing especially for new users. Some of them like "show read posts" should be hidden by default with an expand button (and expand state saved in local storage). Others like sort and feed type should be more emphasized.

Also it doesnt use all the available horizontal space.

dessalines:
- Fixes #3848

## Description

This puts the show subscribed and show sidebar on the same line as other selects, and also uses the `FilterChipCheckbox` like the others.

https://github.com/user-attachments/assets/8cc851de-5056-411f-839b-714fb67345a5

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -53,7 +53,6 @@ export default [
     rules: {
       "@typescript-eslint/no-explicit-any": 0,
       "@typescript-eslint/no-unsafe-assignment": 0,
-      "@typescript-eslint/no-unsafe-call": 0,
       "@typescript-eslint/no-unsafe-member-access": 0,
       "@typescript-eslint/no-unsafe-argument": 0,
       "@typescript-eslint/no-unsafe-return": 0,
--- a/src/embedded/index.ts
+++ b/src/embedded/index.ts
@@ -6,11 +6,13 @@ import { lazyLoad } from "unlazy";
 
 // document.body doesn't exist yet
 window.requestAnimationFrame(() => {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   const cleanup = lazyLoad('img[loading="lazy"]');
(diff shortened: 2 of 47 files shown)


Nutomic:
Looks better than before, also the buttons are a bit bigger. Only thing is that "Show subscribed" and "show sidebar" should keep using the same expand icon as before. And "Type: Posts" should be after the subscribed/sidebar buttons.

dessalines:
> Only thing is that "Show subscribed" and "show sidebar" should keep using the same expand icon as before.

I'm not a fan of that previous expand icon, I'd rather use the new boolean chip component I made, and it makes them look much more consistent anyway.

> also the buttons are a bit bigger.

I didn't change anything there.

> And "Type: Posts" should be after the subscribed/sidebar buttons.

Will do now.

Nutomic:
> I'm not a fan of that previous expand icon, I'd rather use the new boolean chip component I made, and it makes them look much more consistent anyway.

Thats the problem, they look so consistent that "Show sidebar" appears like a filter for the post listing now. Its not clear at all that it will open a different view.

dessalines:
True but I'd rather it just look consistent like the other buttons. They all have to be on the same row anyway.

Nutomic:
The [Material Design Guidelines](https://m3.material.io/components/chips/guidelines) also use different icons for each button.

dessalines:
Its optional, but also is this just because you prefer an `+` / `-` to an `X` and `Check`? 

I could add that as an option to the `FilterChipBoolean` but it seems really minor.

dessalines:
Okay it looks like this now:

Screenshot_2026-02-19-15-38-02-64_5dca195d7d09adf043e42a8f3ad62b19

Issue #3881, opened 2026-02-16
Replace `inferno-helmet` library


Nutomic:
https://github.com/Schiavi/inferno-helmet

It hasnt received any update in 8 years and is missing type definitions which is a problem for https://github.com/LemmyNet/lemmy-ui/issues/3840.

dessalines:
We'll probably have to fork that repo and add our own typings, or just add ignore lints for that import.

Nutomic:
Or put the necessary logic directly into our repo.

Nutomic:
Gave it a try to pull this code in and convert it to Typescript, but its very complicated. Its about 1000 lines of code which is completely outdated, and there is no indication what the correct types may be. Looking at this code it is very likely to contain many bugs. Maybe the best solution is to rewrite it from scratch.

Nutomic:
Looks like upgrading the library is a better option. The latest active fork is [birdofpreyru/react-helmet](https://github.com/birdofpreyru/react-helmet).

The latest react version also has [native support for metadata](https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags) which is available in Inferno, but this probably doesnt work for setting theme and language on the `<html>` tag.

dessalines:
That seems like the best option, but its fairly low priority, shouldn't be for 1.0 IMO. Its also something that ideally someone else should do, as it could take some time.

PR #3882, opened 2026-02-16
Fix lints for unsafe call

Nutomic:
There are a few failures remaining which I'm not sure to fix. Part of these are missing type definitions in `unlazy` and `inferno-helmet` (https://github.com/LemmyNet/lemmy-ui/issues/3881).

### DIFF
--- a/src/shared/components/common/time-interval-filter.tsx
+++ b/src/shared/components/common/time-interval-filter.tsx
@@ -1,4 +1,4 @@
-import { Component } from "inferno";
+import { Component, FormEvent } from "inferno";
 import { I18NextService } from "../../services";
 import { NoOptionI18nKeys } from "i18next";
 
@@ -101,7 +101,10 @@ export class TimeIntervalFilter extends Component<Props, State> {
   }
 }
 
-function handleTimeIntervalNumChange(i: TimeIntervalFilter, event: any) {
+function handleTimeIntervalNumChange(
+  i: TimeIntervalFilter,
+  event: FormEvent<HTMLInputElement>,
+) {
   const interval = {
     num: Number(event.target.value),
     unit: i.state.interval.unit,
--- a/src/shared/utils/markdown.ts
+++ b/src/shared/utils/markdown.ts
@@ -21,8 +21,6 @@ import markdown_it_highlightjs from "markdown-it-highlightjs/core";
 import { getStaticDir } from "./env";
 import mila from "markdown-it-link-attributes";
 
-let Tribute: any;
-
 export let md: MarkdownIt = new MarkdownIt();
 
 export let mdNoImages: MarkdownIt = new MarkdownIt();
@@ -300,12 +298,10 @@ export function getEmojiMart(
   };
   return new Picker(pickerOptions);
 }
+
 export async function setupTribute() {
-  // eslint-disable-next-line eqeqeq
-  if (Tribute == null) {
-    console.debug("Tribute is null, importing...");
-    Tribute = (await import("tributejs")).default;
-  }
+  const Tribute = (await import("tributejs")).default;
+
   return new Tribute({
     noMatchTemplate: function () {
       return "";
@@ -349,7 +345,7 @@ export async function setupTribute() {
           const it: PersonTribute = item.original;
           return it.key;
         },
-        values: debounce(async (text: string, cb: any) => {
+        values: debounce(async (text: string, cb) => {
           cb(await personSearch(text));
         }),
         allowSpaces: false,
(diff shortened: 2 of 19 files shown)


src/shared/components/common/time-interval-filter.tsx
```diff
@@ -52,7 +52,7 @@ export class TimeIntervalFilter extends Component<Props, State> {
           className="form-control interval-filter-input border-light-subtle"
           aria-label={I18NextService.i18n.t("time_interval")}
           value={num}
-          onInput={event => handleTimeIntervalNumChange(this, event)}
+          onInput={linkEvent(this, handleTimeIntervalNumChange)}
```
dessalines: Don't use linkEvent anymore

src/shared/components/community/communities.tsx
```diff
@@ -392,15 +393,16 @@ function handleListingTypeChange(i: Communities, val: ListingType) {
   });
 }
 
-function handleSearchChange(i: Communities, event: any) {
-  i.setState({ searchText: event.target.value });
+function handleSearchChange(i: Communities, event: Event) {
+  i.setState({ searchText: (event.target as HTMLInputElement).value });
```
dessalines, line 400: No need to cast here, just change the type of `event`. IDE should be able to guess it correctly.

src/shared/components/home/login-reset.tsx
```diff
@@ -106,11 +107,16 @@ export class LoginReset extends Component<
     );
   }
 
-  handleEmailInputChange(i: LoginReset, event: any) {
-    i.setState(s => ((s.form.email = event.target.value.trim()), s));
+  handleEmailInputChange(i: LoginReset, event: Event) {
+    i.setState(
+      s => (
+        (s.form.email = (event.target as HTMLInputElement).value.trim()),
```
dessalines: Same, no need to cast.

src/shared/components/person/password-change.tsx
```diff
@@ -101,17 +101,17 @@ export class PasswordChange extends Component<
     );
   }
 
-  handlePasswordChange(i: PasswordChange, event: any) {
-    i.state.form.password = event.target.value;
+  handlePasswordChange(i: PasswordChange, event: Event) {
+    i.state.form.password = (event.target as HTMLInputElement).value;
```
dessalines: Same

src/shared/components/person/settings.tsx
```diff
@@ -1925,8 +1926,11 @@ async function handleChangePasswordSubmit(
   }
 }
 
-function handleImportFileChange(i: Settings, event: any) {
-  i.setState({ settingsFile: event.target.files?.item(0) });
+function handleImportFileChange(i: Settings, event: Event) {
+  i.setState({
+    settingsFile:
+      (event.target as HTMLInputElement).files?.item(0) ?? undefined,
```
dessalines: Same

src/shared/utils/markdown.ts
```diff
@@ -301,13 +300,7 @@ export function getEmojiMart(
   return new Picker(pickerOptions);
 }
 
-export async function setupTribute() {
-  // eslint-disable-next-line eqeqeq
-  if (Tribute == null) {
-    console.debug("Tribute is null, importing...");
-    Tribute = (await import("tributejs")).default;
-  }
-
```
dessalines, line 310: Test tribute to make sure its working correctly. 

This is a late import that should probably not be removed.

tsconfig.json
```diff
@@ -24,7 +24,14 @@
...
       "@components/*": ["shared/components/*"]
-    }
+    },
+    "plugins": [
+      {
+        "name": "@manuth/typescript-eslint-plugin",
+        "ignoreJavaScript": true,
+        "configFile": "custom.eslintrc.js"
+      }
+    ]
```
dessalines: Why is this necessary now.

dessalines (commented):
Instead of adding `event: Event`, what you can do, is 

- Change it to `event`, without any type.
- Your IDE "fix" command should automatically be able to guess the inferno type based on its usage.

tsconfig.json
Nutomic: Strange I have no idea where this came from. Removed.

src/shared/utils/markdown.ts
Nutomic: Right I had to revert this because it broke the entire site.

Nutomic:
Addressed your comments. The remaining lint errors are difficult to fix so I will definitely need help with them. Or should I simply ignore all of them?

dessalines:
I got them, and some other corrections: https://github.com/LemmyNet/lemmy-ui/pull/3887

PR #3883, opened 2026-02-16
Add admin setting for suggested communities (fixes #3434)

### LINKED ISSUE #3434: Add `suggested_communities` to site settings
### Requirements

Add `suggested_communities` to site settings

Nutomic:
Screenshot_20260216_154340


src/shared/components/home/site-form.tsx
```diff
@@ -94,6 +96,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
       default_comment_sort_type: ls?.default_comment_sort_type,
       default_post_sort_type: ls?.default_post_sort_type,
       default_post_time_range_seconds: ls?.default_post_time_range_seconds,
+      suggested_communities: ls?.suggested_communities,
```
Nutomic: The name is not so clear now, seeing `suggested_communities: 7` in the api response doesnt tell you that its an id for multi-community. `suggestions_multi_community_id` would be clearer but sounds a bit strange. Or `suggested_multi_community_id`, though its not really the multi-comm which is suggested but the communities inside.

src/shared/components/multi-community/multi-community-select.tsx
```diff
@@ -0,0 +1,127 @@
...
+    return (
+      <SearchableSelect
+        id="multi-community-select"
+        value={this.props.value}
+        options={[
+          {
+            label: I18NextService.i18n.t("suggested"),
+            value: "",
+            disabled: true,
+          } as Choice,
```
Nutomic, line 54: This kind of label as default value doesnt make much sense here as we already show the label next to it. But its needed to avoid a crash in `searchable-select:68` which calls `options[selectedIndex].label` when the options are not initialized yet.

dessalines:
I should be able to test this tmrw.

src/shared/components/home/site-form.tsx
dessalines: I def agree that it should be `suggested_multi_community_id`

Its clearer and that's what it actually is.

EDIT: The backend is done here now: https://github.com/LemmyNet/lemmy/pull/6351

And API updates done: #3900

src/shared/components/multi-community/multi-community-select.tsx
dessalines: I think its fine, but the label should probably be `Suggested multi community`

Also we need a way to clear this out, IE set it to null in the DB. That means:

- Add an X button here, probably to the right of this dropdown, if `suggested_multi_community` exists
- In the back end, let a `suggested_multi_community_id: 0` allow a DB clear.

I'll do the back-end portion now.

EDIT: this is done: https://github.com/LemmyNet/lemmy/pull/6351

src/shared/components/home/site-form.tsx
```diff
@@ -554,6 +557,18 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
             </div>
           </div>
         )}
+        <div className="mb-3 row">
```
dessalines: Add `align-items-center` like the rest.

dessalines:
K that's deployed: `1.0.0-suggested-community-fixes.0`

- https://github.com/LemmyNet/lemmy/pull/6351
- https://github.com/LemmyNet/lemmy-js-client/pull/857

dessalines:
I've got the API updates done here: #3900

Nutomic:
Instead of an X button to clear, I set the default dropdown value to "None" which can be selected to clear it.

Screenshot_20260223_121707

https://github.com/LemmyNet/lemmy-translations/pull/252

dessalines (approved):
Looks good, feel free to merge when you get the translations conflicts.

PR #3887, opened 2026-02-17
Lint unsafe call2 dess

dessalines:


src/shared/components/common/language-select.tsx
```diff
@@ -152,15 +151,6 @@ export class LanguageSelect extends Component<
...
 
-  handleSelectLanguageChange(i: LanguageSelect, event: any) {
-    const options: HTMLOptionElement[] = Array.from(event.target.options);
-    const selected: number[] = options
-      .filter(o => o.selected)
-      .map(o => Number(o.value));
-
-    i.props.onChange(selected);
-  }
-
```
dessalines, line 163: I don't know what this was about, because this was called on opening the button dropdown menu (not selecting items), and event.target.options didn't exist. Removing it didn't seem to break the language selections.

Nutomic: Confirmed still working.

PR #3888, opened 2026-02-18
Fix lint no-floating-promises

Nutomic:


src/shared/components/community/pending-follows.tsx
```diff
@@ -296,12 +296,9 @@ async function handleApproveFollower(
...
   });
-  i.setState(s => {
-    if (s.appsRes.state === "success" && res.state === "success") {
-      i.refetch(i.props);
-    }
-    return s;
-  });
+  if (i.state.appsRes.state === "success" && res.state === "success") {
+    await i.refetch(i.props);
+  }
```
Nutomic, line 301: Confirmed manually that this and the similar code in notifications.tsx are still working.

PR #3899, opened 2026-02-21
Add ability for mods to edit post nsfw and tags.

### LINKED ISSUE #3274: Add ability for mods to edit posts (nsfw and tags only).
### Requirements

This is `modEditPost` in the js-client.

dessalines:
- Fixes #3274

https://github.com/user-attachments/assets/41ae95ae-384a-49a7-a56d-9a13c7dd74f4


Nutomic:
- The menu item is a bit unclear, better call it "Edit as mod"
- While editing, the post title and body should not be hidden if possible. With the normal post edit for you can see them in the inputs, but here its very difficult to tell which post you are editing (eg when switching between multiple open tabs)
- Looks like NSFW checkbox is only available if the instance has content warning enabled, not sure if that makes sense. But can be changed later.
- Also there seems to be a backend issue with NSFW, even with content warning enabled, enabling NSFW in a community or post always returns `error: "nsfw_not_allowed"` (tested as admin)
- If NSFW is not allowed and the community has no tags, it shows an empty edit form with no inputs. Would be better to hide the menu entry in that case.

Edit: Just saw that https://github.com/LemmyNet/lemmy-ui/pull/3903 addresses part of this.

dessalines:
K done:

Screenshot_2026-02-23-15-01-50-906_mark via

Nutomic:
Looks good, only need to fix conflicts.

PR #3910, opened 2026-02-23
Adding choices for multi-select. Removing SearchableSelect

### LINKED ISSUE #3813: Better input for tag selection
Currently uses a default browser multi-select which is not intuitive (eg Ctrl + click to select multiple or unselect), and doesnt work well on mobile. Same problems as the 0.19 language select which people complained a lot about.

It should be replaced by something like `LanguageSelect` using checkboxes. `SearchableSelect` could also work, though dropdown + multi-select can get finicky, and we dont really need search for a handful of values. 

https://github.com/LemmyNet/lemmy-ui/pull/3795#issuecomment-3807389090

dessalines:
- Removes all our custom select code for `SearchableSelect` and `LanguageSelect`, in favor of a `FilterChipSelect`, which uses `choices.js` behind the scenes. Its the best library for searchable selects (either via a static list or a fetched one), as well as multi-selects.
- It took a long time but I got everything to look good, the only limitation is that loading indicators don't work (but someone else could probably figure them out, as choices does let you use promises when setting the choices).
- Fixes #3813

# Appearance

## Setting post community and tags:

https://github.com/user-attachments/assets/3020ded6-0c10-44ac-9749-17253c21caac

## Language dropdown

https://github.com/user-attachments/assets/3a166426-42bd-41d9-a4c8-12e1052755d3

## Multi-language select

https://github.com/user-attachments/assets/f39fb418-e1d3-4945-9677-d6e1a146af43

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -543,6 +543,12 @@ em-emoji-picker {
   background-color: unset;
 }
 
+.choices__list--dropdown {
+  border-width: var(--bs-border-width);
+  border-color: var(--bs-border-color);
+  border-radius: var(--bs-border-radius) !important;
+}
+
 .choices__list--dropdown .choices__item--selectable.is-highlighted,
 .choices__list[aria-expanded] .choices__item--selectable.is-highlighted {
   background-color: var(--bs-highlight-bg);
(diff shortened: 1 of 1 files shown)


Nutomic:
Functionality is fine, but the styling is really inconsistent with the rest of the page. The selection dropdown should have more rounded corners. And selected items in multi-select look very strange, should look more like `btn-sm`.

dessalines:
Because its an external library with its own custom css, there's only so much I can do. I'll try to devote some more time to it later this week, but its not easy.

dessalines:
Okay I did what I could with some styling. This image shows both the rounded borders of the dropdown, and smaller multiselected options.

Screenshot_2026-02-27-14-54-47-830_org cromite cromite

PR #3913, opened 2026-02-24
Adding viewer.js as an image expander.

### LINKED ISSUE #3841: Expand images
At the moment it is not possible to expand images within the site like you can do in 0.19. Instead clicking a thumbnail always opens the image url directly. You can try here:
- https://voyager.lemmy.ml/c/gaming@lemmy.ml
- https://voyager.lemmy.ml/post/33852
- https://voyager.lemmy.ml/c/photography@lemmy.ml

Additionally, when opening a post with image it would make sense show the expanded image initially, so theres no need to click the thumbnail.

dessalines:
- Fixes #3841

# Thumbnail viewer

https://github.com/user-attachments/assets/61365bdd-86b0-4696-8bc4-5371f5969ca0

# Full card viewer

https://github.com/user-attachments/assets/2acd1045-0846-4a68-8a5c-054983c5871d

# Images in post bodies and comment bodies

https://github.com/user-attachments/assets/1801983d-df52-4a64-bdb9-01f045595bba

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -194,7 +194,7 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     // Load the viewer for every image in the comment
     const id = this.commentView.comment.id;
     const images = document.querySelectorAll(
-      `#comment-${id} > div > div.comment-content > div > p:nth-child(2) > img`,
+      `#comment-${id} > div > div.comment-content > div > p > img`,
     );
     const viewerjss: Viewer[] = [];
     images.forEach((i: HTMLElement) => {
--- a/src/shared/components/post/post-listing.tsx
+++ b/src/shared/components/post/post-listing.tsx
@@ -130,7 +130,7 @@ export class PostListing extends Component<PostListingProps, PostListingState> {
     // Load the image viewer for every image in the post body
     const id = this.props.postView.post.id;
     const images = document.querySelectorAll(
-      `#post-listing-${id} > div > article > div:nth-child(3) > article > div > div > p:nth-child(2) > img`,
+      `#post-listing-${id} > div > article > div > article > div > div > p > img`,
     );
     const viewerjss: Viewer[] = [];
     images.forEach((i: HTMLElement) => {


src/shared/components/comment/comment-node.tsx
```diff
@@ -187,6 +191,37 @@ export class CommentNode extends Component<CommentNodeProps, CommentNodeState> {
     }
   }
 
+  loadViewerJsForImages(setState: boolean = true) {
+    // Load the viewer for every image in the comment
+    const images = document.querySelectorAll(
+      "div.comment-content > div > p:nth-child(2) > img",
```
dessalines: These selectors for post-listings and comment-nodes are working, but they're probably unfortunately selecting a lot more than just this comment. I'll probably have to tweak some more.

dessalines: edit: fixed this now.

Nutomic:
Personally I like the 0.19 way of showing expanded images directly inline in the feed much more. That way it doesnt take you out of the context, so you can view the image while voting or writing a reply at the same time.

For markdown images an overlay makes more sense, as inline expand is probably not possible. Still the current overlay doesnt work well. It has too much delay before being shown. And the controls are useless (rotate, mirror, zoom image). There are also thumbnails at the bottom to show previous/next images, but it doesnt work even if there are multiple images in the same comment. So better to use a simpler expander without any of that extra functionality.

dessalines:
I'm not a fan of the inline expanding, both because it can't work in certain contexts and parts of pages, or on mobile, and because its not actually a proper image viewer. I originally took the idea as a hack from the old Reddit Expansion Suite, but all apps (and this one as well) should use proper full-screen image viewers now. Inline images aren't good for mobile either, which most ppl are using now.

> Still the current overlay doesnt work well. It has too much delay before being shown. And the controls are useless (rotate, mirror, zoom image).

I can tweak those now. 

> but it doesnt work even if there are multiple images in the same comment. 

I'll double check that, but I'm pretty sure I had that working. 

EDIT:

I was able to fix the multiple images issue, and remove the toolbar, but there's no way to speed up the transition.

Nutomic (approved):

PR #3920, opened 2026-02-27
Fix lint no-misused-promises

Nutomic:


src/shared/utils/app.ts
```diff
@@ -943,3 +944,10 @@ export function mark_as_read_i18n(read: boolean): string {
     ? I18NextService.i18n.t("mark_as_unread")
     : I18NextService.i18n.t("mark_as_read");
 }
+
+export function sync<T>(
+  promise: Promise<T>,
+  callback: (arg0: T) => void = () => {},
+) {
+  promise.then(callback).catch(console.error);
+}
```
Nutomic, line 953: Wrapping code which fails the lint check with this function makes it pass. At the moment there are 430 errors for this lint, is it good to rewrite all of them in this way?

dessalines: IMO nah, if its really 430 errors then just disable the lint. This wrapper is just hiding them.

Nutomic: From what I found our existing code is really incorrect and results in floating promises. Using then/catch is the correct solution.

https://stackoverflow.com/questions/76391254/safety-of-using-and-approach-to-floating-promises

dessalines: I'd really not use a wrapper function for these, plus this one is converting the newer `.await()` to  the older `.then`

We have two options for error handling all these async functions: we can either wrap them all in try catch blocks, or better, just add `.catch` after them, like 

`(await someAction(...)).catch(e => console.log(e))`

https://javascript.info/async-await#error-handling

This is all because javascript is a terrible language and doesn't have the concept of `Result`, which we would've been forced to handle in rust.

dessalines: The other bad thing about a `sync` wrapper function, is what if we want to handle errors in specific ways depending on where its found.

matc-pub: The lint tries to prevent assigning `() => Promise<void>` to `() => void`. But our current config also allows this for other types. E.g assigning `() => number` to `() => void` is fine, because nothing changes for the caller.

```ts
export const fff : () => void = () => 1234;
```

At least we should disable parts of the lint. This config only shows 4 problems.

```diff
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -56,7 +56,16 @@ export default [
       "@typescript-eslint/no-unsafe-member-access": 0,
       "@typescript-eslint/no-unsafe-return": 0,
       // TODO: the following is caused by calling async functions from sync functions
-      "@typescript-eslint/no-misused-promises": 0,
+      "@typescript-eslint/no-misused-promises": [
+        "error",
+        {
+          checksVoidReturn: {
+            arguments: false, // allow async callbacks
+            inheritedMethods: false, // allow async lifecycles in class components
+            attributes: false, // allow async event handlers
+          },
+        },
+      ],
       "no-console": ["error", { allow: ["warn", "error", "debug", "assert"] }],
       "@typescript-eslint/no-useless-constructor": "error",
       "inferno/jsx-boolean-value": "error",
```

matc-pub: In #3938  `arguments`  is no longer disabled.

Nutomic:
Made all the changes to fix this lint everywhere. With AI help it wasnt so bad. Also did a manual review of all changes and made some minor adjustments.

src/shared/utils/app.ts
Nutomic: > `(await someAction(...)).catch(e => console.log(e))`

This does not work, it still throws a lint error.

> The other bad thing about a sync wrapper function, is what if we want to handle errors in specific ways depending on where its found.

At the moment these errors are silently ignored. If they should be handled differently, its very easy to add another parameter `error_callback` to the sync function.

Nutomic: > The lint tries to prevent assigning () => Promise<void> to () => void. But our current config also allows this for other types. E.g assigning () => number to () => void is fine, because nothing changes for the caller.

Returning number instead of void is fine, because you are simply ignoring the result. But returning Promise instead of void is not fine, because it leads to [floating promises](https://dev.to/alex_aslam/the-ghost-in-the-machine-a-journey-into-the-floating-promise-pitfall-3ck3). This post actually mentions that unhandled rejections will make the whole process crash.

matc-pub: > > `(await someAction(...)).catch(e => console.log(e))`
> 
> This does not work, it still throws a lint error.

That syntax is wrong. It tries to call `catch` on the resolved value, not the promise.

> ... unhandled rejections will make the whole process crash.

Looks like this can be solved with a `unhandledRejection` handler.

From https://typescript-eslint.io/rules/no-misused-promises/#when-not-to-use-it :

> ... Alternately, if you're not worried about crashes from floating or misused Promises -such as if you have global unhandled Promise handlers registered- then in some cases it may be safe to not use this rule. ...

Adding the following to src/server/index.ts prevents the crash:

```typescript
process.on("unhandledRejection", (error: any) => {
  console.error("Unhandled Rejection:", error);
});
```

dessalines: I'd still really rather not wrap every async function in `sync` just to appease this lint. Maybe @matc-pub could try a PR to handle this.

Nutomic: > Looks like this can be solved with a unhandledRejection handler.

That wont be necessary as this PR fixes all occurences of the problem.

> I'd still really rather not wrap every async function in sync just to appease this lint. 

The lint exists for a reason, and it is part of the recommended checks. As mentioned in the links I posted above, floating promises are real problems which can lead to bugs. So lets merge this and solve the problems.

dessalines: I'd rather @matc-pub take a look. Degrading every async / await to a `.then` function via a wrapper function is not a good way to handle this.

dessalines: Ah I see matc did already. #3938 is a far better solution than this one.

Nutomic: There is no degrading. The existing code is broken and this PR fixes it. #3938 is only a partial fix.

PR #3931, opened 2026-02-28
Fixing images being fetched at full size.

### LINKED ISSUE #3922: Full-size image displayed as thumbnail
- https://voyager.lemmy.ml/post/34021
- https://voyager.lemmy.ml/post/34013

Thumbnails are rendered at 80x80 pixels. Yet lemmy-ui downloads the full image with 1000x1000 pixels or more. Problem occurs both for local uploads and proxied images.

dessalines:
- This ensures that all post images (proxied, thumbnails, or full size), uses the correct `max_size` param in the lemmy backend (it was called `thumbnail` before), so that smaller images are fetched.
  - Icons are 96px, thumbnails are 256px, and full cards are 512px.
- Also removes any of the format conversions, and just lets them be viewed in the original format.
- For viewerjs, it removes the `max_size` param, so that when clicked, it views full size images.
- This also works for images in markdown bodies now, which use the `thumbnail` size by default (and can be clicked into to show the full size).
- Requires https://github.com/LemmyNet/lemmy/pull/6369
- Fixes #3922

### DIFF
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -107,7 +107,7 @@ export class PictrsImage extends Component<Props, State> {
           <img
             ref={this.imageRef}
             src={base64Placeholder(width, height)}
-            data-src={this.src()}
+            data-src={buildPictrsSrc(this.state.src, type)}
             data-blurhash={imageDetails?.blurhash}
             alt={this.alt()}
             title={this.alt()}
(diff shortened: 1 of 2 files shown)


src/shared/components/common/pictrs-image.tsx
```diff
@@ -147,7 +143,7 @@ export class PictrsImage extends Component<Props, State> {
     );
   }
 
-  src(format: string): string {
+  src(): string {
```
dessalines: ~~Might need to externalize this function so that it can be used with markdown images.~~

Nutomic:
Tried to view the posts in the linked issue with your changes, in both cases the thumbnails fail to load.
- In case of `/post/34021` it attempts to load https://voyager.lemmy.ml/api/v4/image/02d0cca5-bf6e-4264-bba2-9f58c9998aff.jpeg?max_size=256 which fails. This should be fixed by https://github.com/LemmyNet/lemmy/pull/6369.
- And in case of `/post/34013` it loads https://voyager.lemmy.ml/api/v4/image/proxy?url=https://i0.wp.com/dbzer0.com/wp-content/uploads/2025/12/small3-1.png?resize=980%2C980&ssl=1&max_size=256 which is invalid with two times `?` (needs urlencode for url param)
    ```json
    {"error":"not_found","cause":"NotFound"}
    ```

Fullsize images are both working.

dessalines:
Both those are dependent on https://github.com/LemmyNet/lemmy/pull/6369 which isn't deployed to voyager.lemmy.ml 's backend yet. Should be testable tomorrow.

Nutomic:
Both of the posts look good now, only small thumbnails are loaded. Urlencode is also fine, just the browser decodes it before rendering which was confusing.

PR #3937, opened 2026-03-03
Add bg-secondary-subtle to make placeholders more subtle.

### LINKED ISSUE #3932: Decrease skeleton loading opacity
### Requirements

The skeleton loader currently has a high contrast with the website background. The idea of this feature is to reduce it's opacity (maybe from 1 to 0.6) to make it visually smoother. Here's a current picture of the skeleton loader on lemmy-ui 1.0.0:

Image

dessalines:
- Fixes #3932

You shouldn't mess with the opacity, since `placeholder-glow` needs to do that: https://getbootstrap.com/docs/5.3/components/placeholders/#animation

But you can change the base text color, so I made it more subtle:

Screenshot_2026-03-03-09-19-48-479_org cromite cromite

### DIFF
--- a/src/shared/components/common/loading-skeleton.tsx
+++ b/src/shared/components/common/loading-skeleton.tsx
@@ -44,13 +44,13 @@ class PostsLoadingSkeletonItem extends Component<any, any> {
       <div className="my-3">
         <div className="col flex-grow-1">
           <div className="row">
-            <div className="col flex-grow-0 order-last order-sm-first">
-              <PostThumbnailLoadingSkeleton />
-            </div>
             <div className="col flex-grow-1">
               <LoadingSkeletonLine size={12} />
               <LoadingSkeletonLine size={8} />
               <LoadingSkeletonLine size={4} />
+            </div>
+            <div className="col flex-grow-0 order-last">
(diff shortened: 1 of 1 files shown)


rodrigo-fm:
Thank you very much for that :)

Nutomic mentioned a [problem](https://github.com/LemmyNet/lemmy-ui/issues/3932#issuecomment-3990597320) with the loading skeleton on different listing views on lemmy 1.0.0.

I might be able to create new loading skeleton formats in the future, but I'm focusing on the infinite scroll toggle on jerboa.

dessalines:
No probs! Its no rush on the other placeholder views, there are a lot more important issues.

Nutomic:
Looks good. Although the skeleton on desktop for normal list view is also wrong now, because thumbnails were moved to the right. On mobile the skeleton already has thumbnails on the right as it should.

src/shared/components/common/loading-skeleton.tsx
Nutomic: Need to change the order here for correct thumbnail position in skeleton.

dessalines: Done 

Screenshot_2026-03-04-20-40-27-738_org cromite cromite

Issue #3840, opened 2026-02-02
Fix remaining lint errors


Nutomic:
After https://github.com/LemmyNet/lemmy-ui/pull/3817 there are still various lint errors which are being ignored, such as various unsafe checks:

https://github.com/LemmyNet/lemmy-ui/blob/f2770ebde4a46543c1607c519fbb195974fc4d63/eslint.config.mjs#L54-L63

We should remove them one by one to improve the codebase. For async functions called from sync functions, use [async blocks](https://stackoverflow.com/a/54758455/1655478) or `.then(d => ...`

Nutomic:
For the lint [no-unsafe-return](https://typescript-eslint.io/rules/no-unsafe-return/) it is necessary to specify the return type for all the render methods. It seems the correct return type is `InfernoNode`, but even with that lint throws "Unsafe return of a value of type error". Do we need to use a different return type, or even define it ourselves?

dessalines:
Might be good to ask this in the [inferno repo](https://github.com/infernojs/inferno/issues). 

InfernoNode might also not be the correct type... could be something like `JSX.Element`

Nutomic:
`InfernoNode` as return type works, but it requires a manual conversion every time with `as InfernoNode`. Started a discussion in the inferno repo to ask about this (https://github.com/infernojs/inferno/discussions/1695).

matc-pub:
I found this bug report regarding the no-unsafe-return lint: https://github.com/typescript-eslint/typescript-eslint/issues/7151

Declaring `JSX.Element` type as `InfernoNode` seems to work.
```diff
--- a/src/shared/utils/types.ts
+++ b/src/shared/utils/types.ts
@@ -12,6 +12,7 @@ import {
 } from "lemmy-js-client";
 import { RequestState } from "@services/HttpService";
 import { Match } from "inferno-router/dist/Route";
+import { InfernoNode } from "inferno";
 
 /**
  * This contains serialized data, it needs to be deserialized before use.
@@ -40,6 +41,10 @@ declare global {
   interface String {
     toLowerCase<T extends string>(this: T): Lowercase<T>;
   }
+  /* eslint-disable-next-line @typescript-eslint/no-namespace */
+  namespace JSX {
+    type Element = InfernoNode;
+  }
 }
 
 export interface InitialFetchRequest<
```

The compiler then finds 35 errors (on main branch), mostly `Not all code paths return a value`. They can probably all be fixed with `InfernoNode | void`.

[...]

PR #3938, opened 2026-03-03
Limited no-misused-promises lint

matc-pub:
Limited version of #3920

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -60,6 +60,7 @@ export default [
         "error",
         {
           checksVoidReturn: {
+            // An unhandledRejection handler prevents NodeJS from exiting on unhandled rejections.
             inheritedMethods: false, // allow async lifecycles in class components
             attributes: false, // allow async event handlers in tsx
           },
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -10,6 +10,11 @@ import "bootstrap/js/dist/collapse";
 import "bootstrap/js/dist/dropdown";
 import "bootstrap/js/dist/modal";
 
+window.addEventListener("unhandledrejection", (ev: PromiseRejectionEvent) => {
+  ev.preventDefault();
+  console.error("Unhandled promise rejection:", ev.reason);
+});
+
 async function startClient() {
   // Allows to test imports from the browser console.
   window.checkLazyScripts = async () => {


Nutomic:
All of these are already handled in #3920. It would be better if you work on some of the other disabled lints instead (https://github.com/LemmyNet/lemmy-ui/issues/3840).

dessalines:
I far prefer this one to the other PR.

src/server/index.tsx
```diff
@@ -82,6 +82,11 @@ if (
   server.use(setDefaultCsp);
 }
 
+// Stop NodeJS from exiting on unhandled promise rejections. Browsers just log an error.
+process.on("unhandledRejection", (error: any) => {
+  console.error("Unhandled promise rejection:", error);
+});
```
Nutomic, line 88: On the client side this also needs to be handled.

https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event

eslint.config.mjs
```diff
@@ -56,7 +56,15 @@ export default [
...
       "@typescript-eslint/no-unsafe-return": 0,
       // TODO: the following is caused by calling async functions from sync functions
-      "@typescript-eslint/no-misused-promises": 0,
+      "@typescript-eslint/no-misused-promises": [
+        "error",
+        {
+          checksVoidReturn: {
+            inheritedMethods: false, // allow async lifecycles in class components
+            attributes: false, // allow async event handlers in tsx
+          },
```
Nutomic, line 66: Also put a comment to mention that this can be safely ignored thanks to unhandledRejection handler.

src/server/index.tsx
matc-pub: Firefox now logs the error twice.

From the compatibility table:

> event.preventDefault() does not prevent Firefox from logging the error message in the console, see bug 1642147.

Nutomic: Not ideal but better than crashing. I would still put preventDefault() here in case it gets fixed later.

matc-pub: The handler for node isn't an event handler. The one in the browser does call preventDefault.

dessalines:
Seems good to me, not sure if you still wanted to work more on it, as its marked as draft.

PR #3939, opened 2026-03-03
Fixup some types

matc-pub:
I noticed some contexts being typed as State in component constructor. 

I also looked over most type casts and fixed the easy ones.

Then there were also a few`getElementId` calls that did not account for `undefined`.  I switched those over to `RefObject` for stronger typing.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
       "@babel/preset-env",
       {
         "loose": true,
-        "browserslistEnv": "package.json"
+        "browserslistEnv": "package.json" // matches webpack target config
       }
     ],
     ["@babel/typescript", { "isTSX": true, "allExtensions": true }]
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
       "prettier --write"
     ]
   },
-  "browserslist": "fully supports es6",
+  "browserslist": "fully supports es6 and supports async-functions",
   "dependencies": {
     "@babel/plugin-proposal-decorators": "^7.24.7",
     "@babel/plugin-transform-class-properties": "^7.25.4",
@@ -57,7 +57,7 @@
     "compression": "^1.8.1",
     "cookie": "^1.0.0",
     "cookie-parser": "^1.4.6",
-    "copy-webpack-plugin": "^13.0.0",
+    "copy-webpack-plugin": "^14.0.0",
     "css-loader": "^7.1.2",
     "date-fns": "^4.0.0",
     "emoji-mart": "^5.6.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -34,7 +34,7 @@ importers:
         version: 6.0.1
       babel-loader:
         specifier: ^10.0.0
-        version: 10.0.0(@babel/core@7.29.0)(webpack@5.105.3)
+        version: 10.1.0(@babel/core@7.29.0)(webpack@5.105.4)
       babel-plugin-inferno:
         specifier: ^6.7.1
         version: 6.8.5(@babel/core@7.29.0)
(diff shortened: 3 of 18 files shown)


src/server/utils/get-error-page-data.ts
```diff
@@ -10,7 +10,7 @@ export function getErrorPageData(error: Error, site?: GetSiteResponse) {
 
   const adminMatrixIds = site?.admins
     .map(({ person: { matrix_user_id } }) => matrix_user_id)
-    .filter(id => id) as string[] | undefined;
+    .filter<string>(id => id !== undefined);
```
Nutomic, line 13: Good, would be nice to have a lint for this too. There is `@typescript-eslint/consistent-type-assertions` but it throws 120 errors.

https://stackoverflow.com/questions/77395532/eslint-warning-for-dangerous-use-of-as

src/shared/components/app/theme.tsx
```diff
@@ -16,7 +16,7 @@ interface State {
 
 export class Theme extends Component<Props, State> {
   private lightQuery?: MediaQueryList;
-  constructor(props: Props, context: State) {
+  constructor(props: Props, context: any) {
```
Nutomic: Good catch, really annoying that there is no lint error thrown. However using any will make `@typescript-eslint/no-explicit-any` once we get to that. It seems you can put `object` instead to avoid that.

src/shared/components/common/filter-chip-select.tsx
```diff
@@ -30,7 +32,7 @@ export class FilterChipSelect extends Component<Props, State> {
   async componentDidMount() {
     // Only load choices on first mount
     const Choices = (await import("choices.js")).default;
-    const element = document.querySelector("#" + this.state.id);
+    const element = this.selectRef.current;
```
Nutomic, line 35: It seems we can add a lint rule to disallow `.querySelector` and `.getElementById` using `no-restricted-syntax`.

https://stackoverflow.com/questions/40977713/how-to-forbid-a-specific-named-function-with-eslint

src/shared/components/home/oauth/oauth-callback.tsx
```diff
@@ -101,14 +100,11 @@ export class OAuthCallback extends Component<OAuthCallbackRouteProps, State> {
         }
         case "failed": {
           let err_redirect = "/login";
-          switch (loginRes.err.message) {
+          switch (loginRes.err.name) {
```
Nutomic, line 103: This change is wrong, Lemmy errors look like this: `{"error":"domain_blocked","message":"test","cause":"DomainBlocked"}`

matc-pub: `lemmy-js-client` transforms the error. We are using `I18NextService.i18n.t(res.err.name as NoOptionI18nKeys)` in a lot of places.

```ts
      let err = new LemmyError(
        json.error ?? response.statusText,
        response.status,
        json.message,
      );
```

```ts
/**
 * A Lemmy error type.
 *
 * The name is the i18n translatable error code.
 * The message is either an empty string, or extra non-translatable info.
 */
export class LemmyError extends Error {
  name: string;
  status: number;
  message: string;
  cause: any;

  constructor(name: string, status: number, message: string = "", cause?: any) {
```

src/shared/components/person/settings.tsx
```diff
@@ -212,7 +212,7 @@ export class Settings extends Component<SettingsRouteProps, SettingsState> {
     exportSettingsRes: EMPTY_REQUEST,
   };
 
-  constructor(props: SettingsRouteProps, context: SettingsState) {
+  constructor(props: SettingsRouteProps, context: any) {
     super(props, context);
```
dessalines: I suppose these being `object` would fix the no any lint.

PR #3940, opened 2026-03-03
Target same js version for babel and webpack

matc-pub:
Relevant: #3908

This allows babel to use async/wait too, and reduces client.js size by about 30kb uncompressed.

We did have mismatched targets in the past, where babel would convert all arrow functions, just for webpack to wrap everything inside one.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
       "@babel/preset-env",
       {
         "loose": true,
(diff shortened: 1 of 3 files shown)


Nutomic (approved):
Looks good, pnpm dev doesnt show any warnings.

PR #3942, opened 2026-03-04
add priv msg delete by recipient button

smorks:
## Description

This adds the ability for a recipient to delete private messages. 

backend: https://github.com/LemmyNet/lemmy/pull/6378
js-client: https://github.com/LemmyNet/lemmy-js-client/pull/875

Not 100% sure how to handle `lemmy-js-client`, where it's linked to a tag, but I'm guessing the tag can only done once the PR is merged. So I don't have that change yet here in the `package.json`.

### DIFF
--- a/src/shared/components/person/notifications.tsx
+++ b/src/shared/components/person/notifications.tsx
@@ -938,6 +938,7 @@ async function handleCreateMessage(
         // FIXME: maybe just let it disappear, comments do too (own comments don't show in notifs)
         notification: {
           id: 0,
+          creator_id: 0,
           recipient_id: 0,
           read: true,
           published_at: nowBoolean(true) ?? "",


src/shared/components/private_message/private-message.tsx
```diff
@@ -225,6 +266,13 @@ export class PrivateMessage extends Component<
           show={this.state.showReportDialog}
           loading={false}
         />
+        <ConfirmationModal
+          message={`Are you sure you want to ${this.props.private_message_view.private_message.deleted_by_recipient ? "restore" : "delete"} this message?`}
+          loadingMessage={`${this.props.private_message_view.private_message.deleted_by_recipient ? "Restoring" : "Deleting"}...`}
+          onNo={() => handleToggleShowConfirmDelete(this)}
+          onYes={() => handleDeleteByRecipientClick(this)}
+          show={this.state.showConfirmDelete}
+        />
```
Nutomic: We dont use any confirmation modals for deletion, its unnecessary seeing as it can easily be restored.

src/shared/components/private_message/private-message.tsx
```diff
@@ -60,6 +60,13 @@ export class PrivateMessage extends Component<
     );
   }
 
+  get is_recipient(): boolean {
```
dessalines: `isRecipient` , typescript uses `camelCase`. We should try to do a cleanup of these at some point.

src/shared/components/private_message/private-message.tsx
```diff
@@ -152,6 +159,40 @@ export class PrivateMessage extends Component<
...
+                        onClick={() =>
+                          handleDeleteClick(
+                            this,
+                            message_view.private_message.deleted_by_recipient,
+                          )
+                        }
+                        data-tippy-content={
+                          !message_view.private_message.deleted_by_recipient
+                            ? I18NextService.i18n.t("delete")
+                            : I18NextService.i18n.t("restore")
```
dessalines: Might as well extract this into a variable above, since its used twice.

src/shared/components/private_message/private-message.tsx
```diff
@@ -152,6 +159,40 @@ export class PrivateMessage extends Component<
...
+                        }
+                      >
+                        {this.props.deleteLoading ? (
+                          <Spinner />
+                        ) : (
+                          <Icon
+                            icon="trash"
+                            classes={`icon-inline ${
+                              message_view.private_message
+                                .deleted_by_recipient && "text-danger"
```
dessalines: Use the `classNames` helper:

```ts
classNames("icon-inline",
  {"text-danger": message_view.private_message.deleted_by_recipient}
)
```

src/shared/components/private_message/private-message.tsx
```diff
@@ -257,7 +303,8 @@ export class PrivateMessage extends Component<
 
   get messageUnlessRemoved(): string {
     const message = this.props.private_message_view.private_message;
-    return message.deleted
+    return message.deleted ||
+      (message.deleted_by_recipient && this.is_recipient)
```
dessalines: I'm guessing this won't be necessary once your back-end PR hides these from the recipients anyway, but its fine to leave it in.

Nutomic: It should be removed as it adds unnecessary complexity.

src/shared/components/private_message/private-message.tsx
smorks: I extracted this to a function. so it could be used by both delete buttons. I can change to a variable if that makes more sense.

src/shared/components/private_message/private-message.tsx
smorks: It appears those lines are necessary in order for it to change to "deleted by creator" when the recipient deletes the message. Otherwise it just stays there. Unless I'm missing something?

Nutomic: The onDelete handler should update `props.private_message_view` with a new one returned from the delete api call. That one already has the deleted flag set. So with that it should automatically update the view to render it as deleted.

smorks: Yes, that does happen, but it has the `deleted_by_recipient` flag set, which is why the above code is necessary, I believe. Because (I think) we only want it to show as deleted for the recipient, but the sender will still see it?

Nutomic: When listing private messages for the sender, we should probably override `deleted_by_recipient` with false in the backend, so the sender cannot tell that it was deleted.

dessalines: Would be a back-end issue.

Nutomic: https://github.com/LemmyNet/lemmy/pull/6398

Nutomic:
There is an error in notifications.tsx that you need to fix.

smorks:
> There is an error in notifications.tsx that you need to fix.

I don't believe that error was created by my changes, but I added a fix that I think should be ok? Also simplified the logic for showing the message text, as was suggested.

Nutomic:
The error probably came from updating the js client, which pulled in some other changes. Thank you!

PR #3944, opened 2026-03-06
Hide some feed select buttons by default

Nutomic:
These are rarely used and shouldnt take up space unnecessarily.

collapsed:

Screenshot_20260306_123814

expanded:

Screenshot_20260306_123750

mobile expanded:
Screenshot_20260306_123900

Same change also needs to be made in community.tsx.


src/shared/components/home/home.tsx
```diff
@@ -872,32 +874,105 @@ export class Home extends Component<HomeRouteProps, HomeState> {
     const { showSubscribedMobile, showSidebarMobile } = this.state;
 
     return (
-      <div className="row row-cols-auto align-items-center g-3 mb-3">
-        {/* Only show these two selects on mobile */}
-        {this.hasFollows && (
+      <div className="g-3 mb-3">
+        <div className="row row-cols-auto align-items-center">
```
dessalines: There's something weird going on with the gutters here, they aren't being used on the mobile row.

src/assets/symbols.svg
dessalines, line 1: Create a variant or prop to the expand chip checkbox that uses these icons. Because your expand button should look similar to the other already existing ones.

src/shared/components/home/home.tsx
```diff
@@ -872,32 +874,105 @@ export class Home extends Component<HomeRouteProps, HomeState> {
...
+              listingType ??
+                this.state.siteRes.site_view.local_site
+                  .default_post_listing_type,
+              sort,
+            )}
+          </div>
+          <button
+            className="col btn btn-ghost"
+            onclick={_ => handleHideSelectButtons(this)}
+          >
```
dessalines: Use the `ExpandChipCheckbox`

dessalines (commented):
Something weird going on with the diff, this really only needs to add that button for mobile, as well as `{this.showMobileExpanded && filter....` around the ones you want to hide / show.

src/shared/components/home/home.tsx
Nutomic: Fixed

Nutomic:
The expand should also be used on desktop. Reason is that these options ("show hidden posts" and "hide read posts") are rarely used, and shouldnt distract the user from more important actions like changing the sort type.

I dont want to use ExpandChipCheckbox because its already intuitive with the arrow icon, and doesnt need any text. And btn-ghost doesnt catch so much attention, it shouldnt have any background or border.

src/shared/components/community/community.tsx
```diff
@@ -763,8 +763,8 @@ export class Community extends Component<CommunityRouteProps, State> {
     const myUserInfo = this.isoData.myUserInfo;
 
     return (
-      <div className="g-3 mb-3">
-        <div className="row row-cols-auto align-items-center">
+      <div>
```
dessalines: This extra wrapper div is probably not necessary.

dessalines:
I tested and this isn't working at all for some reason, neither on mobile nor desktop views.

The diff is still strange, it should only need to add `{this.showExpanded && filter....` around the existing filters you want to hide.

src/shared/components/community/community.tsx
```diff
@@ -761,22 +763,82 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+            <Icon icon="chevrons-down" />
+          </button>
         </div>
         {postOrCommentType === "post" && myUserInfo && (
-          <>
+          <div
+            className={
+              this.state.selectButtonsHidden
+                ? "d-none"
+                : "row row-cols-auto mt-2"
```
dessalines: Here's the culprit. Don't do this, just surround the specific filters you want to hide with `{!this.state.selectButtonsHidden && <div `

Nutomic: Okay fixed

dessalines: Still had some issues, I got it: https://github.com/LemmyNet/lemmy-ui/pull/3962

Nutomic:
Screenshot_20260313_123845

---

Updated with changes from #3962:
- Arrow up icon to collapse view
- Hide if logged out
- Adjust margins
- Keep separate row diff to avoid button repositioning

I tried to keep the expanded buttons in the same row but it looks strange:

---
Screenshot_20260313_123801

Nutomic:
Merged in https://github.com/LemmyNet/lemmy-ui/pull/3986

PR #3957, opened 2026-03-10
Lint forbid get element

Nutomic:
Requires https://github.com/LemmyNet/lemmy-ui/pull/3939/changes

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -56,7 +56,16 @@ export default [
       "@typescript-eslint/no-unsafe-member-access": 0,
       "@typescript-eslint/no-unsafe-return": 0,
       // TODO: the following is caused by calling async functions from sync functions
-      "@typescript-eslint/no-misused-promises": 0,
+      "@typescript-eslint/no-misused-promises": [
+        "error",
+        {
+          checksVoidReturn: {
+            // An unhandledRejection handler prevents NodeJS from exiting on unhandled rejections.
+            inheritedMethods: false, // allow async lifecycles in class components
+            attributes: false, // allow async event handlers in tsx
+          },
+        },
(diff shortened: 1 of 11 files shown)


eslint.config.mjs
```diff
@@ -77,6 +77,19 @@ export default [
           ],
         },
       ],
+      "no-restricted-properties": [
+        "error",
+        {
+          object: "document",
+          property: "querySelector",
+          message: "Use createRef instead",
```
matc-pub: `querySelector` is also available on elements. https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector

Nutomic: Done. I hope thats all, otherwise it gets very verbose.

matc-pub: This rule is not type aware. It doesn't catch e.g. `this.commentSectionRef.current?.querySelector("#foo");`. 

For querySelector it should be ok to block access on all objects by removing the "object" property from the rule.

Nutomic: Done, now it works.

Nutomic:
It seems that this broke something with the language and community dropdowns, in the previous commit on main branch they look normal. However the diff doesnt show any changes in that area.

Screenshot_20260312_163632

dessalines:
It was probably [this one here](https://github.com/LemmyNet/lemmy-ui/pull/3939/changes#diff-0e808d319616bdf290f1fa1be8054c521824c223fc890c4e36ec60132277129b) 

If that select doesn't work with ref, then there's nothing we can do but undo this lint.

Nutomic:
Fix here: https://github.com/LemmyNet/lemmy-ui/pull/3966/

PR #3958, opened 2026-03-10
New image changes (fixes #3921, fixes #3956)

### LINKED ISSUE #3921: Image uploads disabled
1.0 has a new config option [image_upload_disabled](https://github.com/LemmyNet/lemmy/blob/main/config/defaults.hjson#L72). This is enabled on voyager.lemmy.ml. Still lemmy-ui lets you upload images in the create post screen, and doesnt show any error message when that fails.

Ideally image upload should be hidden in this case, but its tricky because the server config is not available to the client.

### LINKED ISSUE #3956: Add more image settings (about sizes and such) to the UI
### Requirements

The PR below moves a lot of pictrs-related configuration from the config, to the UI. 

We need to add a tab and form for setting them in the admin panel.

Context: https://github.com/LemmyNet/lemmy/issues/6371

Nutomic:
Everything works in my testing.

Screenshot_20260310_133830

Translations: https://github.com/LemmyNet/lemmy-translations/pull/258

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -86,6 +86,18 @@ export default [
           ],
         },
       ],
+      "no-restricted-properties": [
+        "error",
+        {
+          object: "document",
+          property: "getElementById",
+          message: "Use createRef instead",
+        },
+        {
+          property: "querySelector",
+          message: "Use createRef instead",
+        },
+      ],
     },
   },
 ];
--- a/src/client/index.tsx
+++ b/src/client/index.tsx
@@ -9,6 +9,7 @@ import { setupMarkdown } from "@utils/markdown";
 import "bootstrap/js/dist/collapse";
 import "bootstrap/js/dist/dropdown";
 import "bootstrap/js/dist/modal";
+import { createRef } from "inferno";
 
 window.addEventListener("unhandledrejection", (ev: PromiseRejectionEvent) => {
   ev.preventDefault();
@@ -35,18 +36,19 @@ async function startClient() {
     loadLanguageInstances(fallbackLanguages, interfaceLanguage),
   ]);
 
+  const rootRef = createRef<HTMLDivElement>();
   const wrapper = (
     <BrowserRouter>
-      <App dateFnsLocale={dateFnsLocale} i18n={i18n} />
+      <App dateFnsLocale={dateFnsLocale} i18n={i18n} rootRef={rootRef} />
     </BrowserRouter>
   );
 
-  const root = document.getElementById("root");
+  if (rootRef.current) {
+    hydrate(wrapper, rootRef.current);
 
-  if (root) {
-    hydrate(wrapper, root);
-
-    root.dispatchEvent(new CustomEvent("lemmy-hydrated", { bubbles: true }));
+    rootRef.current.dispatchEvent(
+      new CustomEvent("lemmy-hydrated", { bubbles: true }),
+    );
   }
 }
 
--- a/src/server/handlers/catch-all-handler.tsx
+++ b/src/server/handlers/catch-all-handler.tsx
@@ -30,6 +30,7 @@ import { parsePath } from "history";
 import { getQueryString } from "@utils/helpers";
 import { adultConsentCookieKey } from "@utils/config";
 import { loadLanguageInstances } from "@services/I18NextService";
+import { createRef } from "inferno";
 
 const specifiedFrontend = process.env.LEMMY_UI_FRONTEND
   ? getBaseUrl(process.env.LEMMY_UI_FRONTEND)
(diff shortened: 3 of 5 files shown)


Nutomic:
Needs a new js client to resolve conflicts, but publishing is currently failing: https://woodpecker.join-lemmy.org/repos/133/pipeline/4139/3#L664

dessalines:
The npm token expired, I generated a new one so it deployed now.

src/shared/components/home/site-form.tsx
```diff
@@ -110,6 +123,17 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
 
   render() {
     const siteSetup = this.props.siteRes?.site_view.local_site.site_setup;
+    const image_mode_options: Choice[] = [
```
dessalines: Make this a `FilterOption<ImageMode>[]`

And its probably best to have this `const` outside of the class.

src/shared/components/home/site-form.tsx
```diff
@@ -540,6 +564,161 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
             />
           </div>
         </div>
+        <div className="mb-3 row">
+          <FilterChipSelect
```
dessalines: Use the `FilterChipDropdown` , not the select

src/shared/components/home/site-form.tsx
```diff
@@ -540,6 +564,161 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+            className="col-12 col-form-label"
+            htmlFor="image_proxy_bypass_domains"
+          >
+            {I18NextService.i18n.t("image_proxy_bypass_domains")}
+          </label>
+          <div className="col-12">
+            <input
+              type="text"
+              id="image_proxy_bypass_domains"
+              placeholder="example.com,sample.org"
```
dessalines, line 600: This is pry fine.

src/shared/components/home/site-form.tsx
```diff
@@ -540,6 +564,161 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+            className="col-12 col-form-label"
+            htmlFor="image_upload_timeout_seconds"
+          >
+            {I18NextService.i18n.t("image_upload_timeout_seconds")}
+          </label>
+          <div className="col-12">
+            <input
+              type="number"
+              id="image_upload_timeout_seconds"
+              placeholder="example.com,sample.org"
```
dessalines: There are a bunch of these placeholders on the number inputs. They should only be there for the text one.

src/shared/components/home/site-form.tsx
```diff
@@ -540,6 +564,161 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+          </label>
+          <div className="col-12">
+            <input
+              type="number"
+              id="image_max_avatar_size"
+              placeholder="example.com,sample.org"
+              className="form-control"
+              value={this.state.siteForm.image_max_avatar_size}
+              onInput={e => handleImageMaxAvatarSize(this, e)}
+              minLength={3}
```
dessalines: I don't think minLength is right for number inputs. These should have `min` and `max`

src/shared/components/home/site-form.tsx
```diff
@@ -771,3 +960,84 @@ function handleSelectSuggestedMultiComm(
 ) {
   i.setState(s => ((s.siteForm.suggested_multi_community_id = suggested), s));
 }
+
+function handleChangeImageMode(i: SiteForm, image_mode: Choice[]) {
```
dessalines: `imageMode: ImageMode`

src/shared/components/post/post-form.tsx
```diff
@@ -371,37 +371,41 @@ export class PostForm extends Component<PostFormProps, PostFormState> {
...
-              </div>
+                    {capitalizeFirstLetter(I18NextService.i18n.t("image"))}
+                    <Icon icon="image" classes="icon-inline ms-1" />
+                  </label>
+                  <div className="col-sm-10">
+                    <input
+                      id="file-upload"
+                      type="file"
+                      accept="image/*,video/*"
+                      name="file"
```
dessalines, line 389: grep the entire codebase for `type="file"` to see if there are any others that need to be surrounded by this check now too.

src/shared/components/home/site-form.tsx
Nutomic: Copy-paste fail, fixed.

src/shared/components/post/post-form.tsx
Nutomic: Right there is also the image upload in markdown-textarea. This one doesnt take any SiteView as parameter but instead various separate props. I ended up changing this to take `IsoData` directly, to avoid having so many separate parameters. This had to be adjusted in many different places too.

dessalines: No undo that one. Don't use global state, especially isodata, except for the top-level pages. Pass around the specific things you need into components as params. Yes its more verbose but its far more explicit, better type-checked, and makes components more portable and easier to check. Isodata should only ever be used by the top level pages, never ever as props.

Nutomic: Okay, I suppose this will lead to unnecessary rerendering if unrelated data changes.

PR #3960, opened 2026-03-10
Post time range input

Nutomic:
This one is much more intuitive to use:
Screenshot_20260311_111819

TODO:
- [x] Add more time values (those which are currently available in 0.19)
- [x] Hide time input if current sort is Old or New
- [x] Avoid using vertical space for label (tooltip could work, but doesnt support live update)
- [x] Reorder buttons
- [ ] Same input for community feed
- [ ] Put the existing `TimeIntervalFilter` in the expanded menu, or otherwise delete it

Requires https://github.com/LemmyNet/lemmy-ui/pull/3944


dessalines:
I really don't like this, it takes up way too much space, and loses all the custom flexibility. Please keep it the way it was.

Nutomic:
It only takes 24px of vertical space, while the filter buttons which are now hidden by default take 31px. Also this has much better usability especially on mobile. I added back TimeIntervalFilter now (without presets), so you can also set custom intervals. At the moment both time inputs are not synchronized yet.

Screenshot_20260312_110044

dessalines:
I'm still not a big fan of it, and I'd like to get other input from @SleeplessOne1917 or @matc-pub . A bar like this (especially where the intervals are random), tells you nothing by glancing at it (compare that with the existing dropdown telling you 5 weeks). Every UI element should be immediately comprehensible as to what its showing you, and a bar tells you nothing (unless its like a percentage or something)

I spose I'd be fine with it as long as its hidden by default with the expand dropdown, and the time interval filter dropdown stays.

Nutomic:
There is a label with the current value at the top right, that could be moved somewhere closer to the slider. It would also be possible to add step markers and labels, something like this:
Screenshot_20260312_160617

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#examples

dessalines:
That's kinda why you don't see hardly any UIs using these range inputs nowadays, because they don't look good. Dropdowns are far better.

dessalines:
This can probably be closed due to matc's PR.

PR #3961, opened 2026-03-10
Lint to disallow use of any (ref #3840)

Nutomic:
https://typescript-eslint.io/rules/no-explicit-any/

In general we should use `unknown` instead which is safer. In some cases it is also possible to use generics or specific types instead of any.

The remaining errors are from tributejs and emoji-mart which dont provide type definitions (https://github.com/missive/emoji-mart/issues/576). Same problem with tributejs. Most others are from `Record`. Unless you have any suggestions how to fix them, I will mark all these as ignored.

Requires https://github.com/LemmyNet/lemmy-ui/pull/3957

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "inferno-router": "^9.0.11",
     "inferno-server": "^9.0.11",
     "jwt-decode": "^4.0.0",
-    "lemmy-js-client": "1.0.0-delete-pm-recipient.0",
+    "lemmy-js-client": "1.0.0-delete-pm-recipient.3",
     "markdown-it": "^14.1.0",
     "markdown-it-bidi": "^0.2.0",
     "markdown-it-container": "^4.0.0",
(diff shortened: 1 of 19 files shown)


src/shared/components/common/media-uploads.tsx
matc-pub: Typescript is ok with this `setState` for an `unknown` state. `Record<string, never>` or `never` should prevent this.

MultiCommunitySidebar also has such a `setState` call that should fail.

dessalines:
Just ignore the tribute and emoji-mart errors.

src/shared/components/common/media-uploads.tsx
Nutomic: Okay changed all these to `never`, or `object` in some cases and removed the invalid `setState` calls.

Nutomic:
strictTypeChecked could be the next step once all of recommendedTypeChecked is passing without ignored errors. Or at least the most useful lints from that group.

PR #3962, opened 2026-03-12
Fixing hide-select-buttons

dessalines:
- Adds a chevron up, styles those buttons correctly
- hides it if not logged in (because otherwise those buttons do nothing)
- Fixes the diff


Nutomic:
Right now this doesnt work well because the buttons move around when you click expand, quite disorienting. With my PR you can expand and collapse the extra buttons repeatedly without moving the mouse at all.

dessalines:
You could put them after the expand button, but the main thing is to not create a brand new flowrow like with the other PR. I prefer this because it keeps the checkboxes near each other tho, and the dropdowns all after.

Nutomic:
Merged in https://github.com/LemmyNet/lemmy-ui/pull/3986

PR #3999, opened 2026-03-25
Search adjustments (fixes #3995)

Nutomic:
- Rename fields eg require_email_verification -> email_verification_required
- For post_form, modlog etc search using endpoints listCommunities etc (this could be reverted as different search types are available again, probably makes no difference)
- Search pagination and sorting removed
- Requires https://github.com/LemmyNet/lemmy-js-client/pull/906

There is currently an error thrown from setHeaders:
```
Uncaught (in promise) TypeError: ___WEBPACK_IMPORTED_MODULE_7__.HttpService.client.setHeaders is not a function
    _setAuthInfo2 UserService.ts:88
    init UserService.ts:29
    ts UserService.ts:101
    a async module:70
    ts client.js:37427
    Webpack 19
UserService.ts:88:34
```

### DIFF
--- a/src/shared/utils/app.ts
+++ b/src/shared/utils/app.ts
@@ -6,9 +6,7 @@ import {
   MyUserInfo,
   PostView,
   RegistrationApplicationView,
-  Search,
   Comment,
-  SearchType,
   PersonView,
   Language,
   Instance,
@@ -35,6 +33,9 @@ import {
   PostReportResponse,
   PrivateMessageReportResponse,
   CommunityReportResponse,
+  ListCommunities,
+  ListPersons,
+  ListMultiCommunities,
 } from "lemmy-js-client";
 import {
   CommentNodeI,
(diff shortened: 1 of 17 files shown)


src/shared/utils/app.ts
```diff
@@ -332,26 +329,29 @@ export function enableDownvotes(siteRes: GetSiteResponse): boolean {
...
-    type_,
-    sort: "top",
-    listing_type: "all",
+export async function searchMultiCommunities(search_term: string) {
+  const form: ListMultiCommunities = {
+    search_term,
+    sort: "subscribers",
+    type_: "all",
   };
+  const res = await HttpService.client.listMultiCommunities(form);
```
Nutomic, line 352: These could be changed back to use search endpoint, probably doesnt make much difference.

dessalines (approved):
Seems good, but I still do want to rework the search page and filter bars to account for the different search types and filters.

src/shared/utils/app.ts
dessalines: The individual ones like you've done here are much preferred.

PR #4025, opened 2026-04-10
Walk prototype chain in wrapClient

matc-pub:
With the changes in lemmy-js-client the setHeaders function became inaccessible in the wrapped client.

The type for the wrapped client did miss a `ReturnType<T>` for sync methods with async wrapper. I also noticed that the wrapper function can't return a "success" state for void results.

### DIFF
--- a/src/shared/services/HttpService.ts
+++ b/src/shared/services/HttpService.ts
@@ -32,17 +32,29 @@ export type RequestState<T> =
   | EmptyRequestState
   | LoadingRequestState
   | FailedRequestState
-  | SuccessRequestState<T>;
+  // undefined and null produce an EmptyRequestState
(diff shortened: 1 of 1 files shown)


Nutomic (approved):
Confirmed working, although I dont understand why this is necessary. Would it be easier to change the code in lemmy-js-client, also so that other clients dont need this extra logic?

matc-pub:
It's necessary because we map all api responses to a `RequestState` and assumed that `LemmyHttp` did not inherit anything. lemmy-js-client shouldn't have to match that assumption.

Nutomic:
Then why not move RequestState into the js client?

dessalines (approved):

dessalines:
We should really move the request state in there, and get rid of all this overly complicated wrappedclient logic. I'll create an issue.

PR #4032, opened 2026-04-11
Define variables from _variables-dark.scss for dark themes

matc-pub:
Dark themes did define some variables that bootstrap ignored in dark mode. I just added the `-dark` suffix to defined variables available in `_variables-dark.scss`.

Most noticeable are the body background colors for i386-dark (now blue again) and darkly-pureblack (now `#000` instead of `#111`)


dessalines (approved):
Sweet, thx.

PR #4040, opened 2026-04-16
Trying to fix arm builds 2.

dessalines:
Turns out that although my regex check is correct, I think the `when` block wasn't, as the evaluate was considered an `OR` condition.

https://woodpecker.join-lemmy.org/repos/135/pipeline/10425

Going to try this out, merge it right away, and manually start the nightly build to see if it fixes it.


dessalines:
Verified, that fixed it.

PR #4056, opened 2026-04-23
Search pagination

Nutomic:


src/shared/utils/config.ts
```diff
@@ -24,7 +24,7 @@ export const markdownFieldCharacterLimit = 10000;
 export const maxUploadImages = 20;
 export const concurrentImageUpload = 4;
 export const updateUnreadCountsInterval = 30000;
-export const fetchLimit = 20;
+export const fetchLimit = 1;
```
Nutomic: For testing, need to remove later.

dessalines (approved):

PR #4085, opened 2026-05-11
Persist markdown content in local storage (fixes #4084)

### LINKED ISSUE #4084: Store content of markdown editor in local storage
This way the already written text doesnt disappear if the browser gets closed by accident (eg browser killed due to OS out of memory).

https://stackoverflow.com/a/23142165

https://lemmy.ml/post/47157639

Nutomic:


src/shared/components/common/markdown-textarea.tsx
```diff
@@ -115,6 +118,13 @@ export class MarkdownTextArea extends Component<
     }
   }
 
+  componentWillUnmount() {
```
Nutomic: This hook is not called at all, and I couldnt find another one that looks relevant. I am seeing the confirmation dialog for leaving the page, but no idea where that is defined because there isnt any event listener for `beforeunload`.

src/shared/components/common/markdown-textarea.tsx
```diff
@@ -99,6 +99,9 @@ export class MarkdownTextArea extends Component<
       const tribute = await setupTribute();
       const textarea = this.textAreaRef.current;
       if (textarea) {
+        const content: string | undefined =
+          localStorage.getItem("lemmy-markdown") ?? undefined;
```
Nutomic: In this way it will store a single text item which is shared for all markdown inputs. So the text from an unfinished comment would also show up in the profile bio editor for example. We could use a different identifier depending where this textarea is rendered, but that would require a new prop.

src/shared/components/common/markdown-textarea.tsx
matc-pub: `<Prompt />` from inferno-router shows the confirmation dialog.

You can't rely on beforeunload and componentWillUnmount to be called on browser crashes. `componentWillUnmount` should probably remove the content from local storage, as the user decided to abandon it. The content should be stored from `onInput` using debounce.

Nutomic: Okay changed it to store the text in `handleContentChange`. Debounce is probably not necessary as its only storing data locally. And made it to remove the local storage item on submit.

dessalines (commented):
Also add the clearing localstorage to `handleReplyCancel`, L638

That's the only issue I can think of, but I'm sure there are other cases that might cause issues, and I need to test it a bit. But this is a great idea.

Fixes #4084

Nutomic:
Done

dessalines (approved):
Tested and it works well. 👍

PR #4086, opened 2026-05-11
Fix hide post from profile page (fixes #4071)

### LINKED ISSUE #4071: Hide post option from profile does nothing
### Requirements

### Summary

When viewing your own posts from the profile and selecting 'Hide Post' from the options, no request is sent out and nothing happens at all.

For added context, the reason I wanted to do this was to hide posts I've deleted from my profile to not see them in my history, but the same happens if I do it on a post that's not deleted.

### Steps to Reproduce

1. Go to user profile
2. Open options for a post
3. Click 'Hide Post'
4. Check inspect network, nothing happens

### Technical Details

Browser: Firefox 150.0.1 (64 bit)
Arch Linux with kernel version: 7.0.3-arch1-1

### Lemmy Instance Version

0.19.18, nightly-2026-05-06

### Lemmy Instance URL

https://lemmy.ml, https://voyager.lemmy.ml

Nutomic:
Functionality is working now, but the profile seems to render hidden posts as well. So it seems like nothing happens after clicking the button.

To fix that I added an icon for hidden posts, although `eye-slash` looks quite ugly like this.
Screenshot_20260511_153815

### DIFF
--- a/src/shared/components/post/common.tsx
+++ b/src/shared/components/post/common.tsx
@@ -25,31 +25,28 @@ import { CommunityTag } from "@components/community/community-tag";
 import { InfernoNode } from "inferno";
 
 type PostNameProps = {
-  post_view: PostView;
+  post: Post;
   showBody: ShowBodyType;
 };
-export function PostName({ post_view, showBody }: PostNameProps) {
-  const url = post_view.post.url;
+export function PostName({ post, showBody }: PostNameProps) {
+  const url = post.url;
 
   return (
     <h1 className="post-name h5 d-inline text-break">
       {showBody === "full" ? (
         <a
           className={
-            !post_view.post.featured_community && !post_view.post.featured_local
+            !post.featured_community && !post.featured_local
               ? "text-body"
               : "link-primary"
           }
           href={url}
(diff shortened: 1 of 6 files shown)


dessalines:
> Functionality is working now, but the profile seems to render hidden posts as well. So it seems like nothing happens after clicking the button.

This is because listing a person's content in `crates/db_views/person_content_combined/src/impls.rs` doesn's seem to be filtering for hidden content. I'll add that now, then you can remove the weird `eye-slash` thing.

src/shared/components/post/common.tsx
dessalines, line 1: You can remove the changes in this file, since I did the back-end PR to hide them. 

Also in the future I don't think `PostName` is the right component for additions like this. There's a `PostBadges` for these.

dessalines:
Github is messing up and not showing the linked issue for some reason.

Fixes #4071

Nutomic:
Okay reverted that commit

dessalines (changes requested):
Perfect, thx.

dessalines (approved):

PR #4102, opened 2026-05-18
Hide mod/admin badges on post and comment listings

Nutomic:
With this change only the user profile indicates that someone is an admin. This should make admins and mods less hesitant to participate in discussions as normal user. And people wont think that comments are official communication when they are not. It will also make "mark comment distinguished" more useful (https://github.com/LemmyNet/lemmy/issues/6308).

### DIFF
--- a/src/shared/components/comment/comment-node.tsx
+++ b/src/shared/components/comment/comment-node.tsx
@@ -810,10 +810,8 @@ function CommentHeader({
   myUserInfo,
 }: CommentHeaderProps) {
   const {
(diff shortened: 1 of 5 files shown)


dessalines:
I'm still somewhat torn on this, because I think its more transparent to have the admin / mod status shown. But I'm willing to try it out anyway. Not many people have chimed in with arguments for or against. 

We can revert if people don't like it.

dessalines (approved):

PR #4110, opened 2026-05-19
Change large card view to use full width

Nutomic:
This seems better to view full-size images on desktop, without separately expanding each image.

Screenshot_20260519_143024

### DIFF
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -195,9 +195,12 @@ export function buildPictrsSrc(src: string, type: PictrsImageType): string {
       url.searchParams.set("max_size", thumbnailSize.toString());
       break;
     case "icon":
+    case "icon_without_banner":
       url.searchParams.set("max_size", iconThumbnailSize.toString());
       break;
     case "banner":
+    case "icon_and_banner":
+    case "card_top":
       url.searchParams.set("max_size", bannerSize.toString());
       break;
     case "full_size":
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -200,6 +200,8 @@ export function buildPictrsSrc(src: string, type: PictrsImageType): string {
     case "banner":
       url.searchParams.set("max_size", bannerSize.toString());
       break;
+    case "full_size":
+      break;
     default:
       url.searchParams.set("max_size", defaultImgSize.toString());
       break;
--- a/src/shared/components/common/pictrs-image.tsx
+++ b/src/shared/components/common/pictrs-image.tsx
@@ -18,7 +18,7 @@ const defaultImgSize = 512;
 const bannerSize = 2048;
 
 type PictrsImageType =
-  | "full_size"
+  | "large_thumbnail"
   | "thumbnail"
   | "icon"
   | "banner"
@@ -203,7 +203,9 @@ export function buildPictrsSrc(src: string, type: PictrsImageType): string {
     case "card_top":
       url.searchParams.set("max_size", bannerSize.toString());
       break;
-    case "full_size":
+    case "large_thumbnail":
+      // Use bannerSize here for slight downscaling, but larger than thumbnail
+      url.searchParams.set("max_size", bannerSize.toString());
       break;
     default:
       url.searchParams.set("max_size", defaultImgSize.toString());
--- a/src/shared/components/post/post-listings.tsx
+++ b/src/shared/components/post/post-listings.tsx
@@ -222,8 +222,8 @@ export class PostListings extends Component<PostListingsProps, never> {
 function postListingModeCols(mode: PostListingMode): string {
   switch (mode) {
     case "list":
-      return "col-12";
     case "card":
+      return "col-12";
     case "small_card":
       return "col-12 col-md-6";
   }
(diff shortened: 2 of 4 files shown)


dessalines:
Not a fan of this, especially for wide devices. Then you get stuff like this on really wide devices:


Screenshot_2026-05-19-11-12-48-656_com jerboa


 If you want to increase the double column breakpoint from `md` to `lg` (or possibly `xl` tho, that might work.

Nutomic:
You mean that images get cut off? Doesnt happen in my testing, it always shows the full image. Although theres a problem that tall images can take up the entire screen, possibly needs a max height.

---

Screenshot_20260525_125813

Screenshot_20260525_125826

---

Using `col-xl-6` doesnt seem to work, even with a rather narrow browser window (1040px wide) it shows two columns.

dessalines:
> Although theres a problem that tall images can take up the entire screen, possibly needs a max height.

Yes this is the case, and its why wider screens should be using the 2-column layout.

> Using col-xl-6 doesnt seem to work, even with a rather narrow browser window (1040px wide) it shows two columns.


xl should be >1200 px wide: https://getbootstrap.com/docs/5.3/layout/breakpoints/#available-breakpoints

There's several places where this should be done I think, but I'll test it out.

dessalines:
I've verified that `col-xl-6` works fine.

Nutomic:
Ah making the browser window narrower does switch it to single column. But `col-xl-6` still isnt so helpful, because the main content has a maximum width even on very large screens. So on wider browser window/bigger screen you actually get smaller images:

Screenshot_20260526_115842
Screenshot_20260526_115855

---

With the change in this PR you can see larger images also on big screens, and dont need to expand every single image:
Screenshot_20260526_120000

dessalines:
> because the main content has a maximum width even on very large screens

That's an issue with the top level [containers](https://getbootstrap.com/docs/5.3/layout/containers/#how-they-work) not something to be fixed with this PR. 

The card view should not be ignoring the double columns, that's something that I want to keep.

Nutomic:
The problem is that "Card" and "Small Card" modes are almost identical on large screens. While "Card" does show images a bit larger, its not large enough to view them properly, and so you still need to expand manually for every single image.

src/shared/components/common/pictrs-image.tsx
```diff
@@ -195,11 +195,16 @@ export function buildPictrsSrc(src: string, type: PictrsImageType): string {
...
+    case "icon_without_banner":
       url.searchParams.set("max_size", iconThumbnailSize.toString());
       break;
     case "banner":
+    case "icon_and_banner":
+    case "card_top":
       url.searchParams.set("max_size", bannerSize.toString());
       break;
+    case "full_size":
+      break;
```
Nutomic, line 209: Noticed a bug here, in large card mode it was always loading images downscaled to 512px because it was entering the default case here. Also added other variants where it seems to make sense.

src/shared/components/post/post-listings.tsx
```diff
@@ -222,8 +222,8 @@ export class PostListings extends Component<PostListingsProps, never> {
 function postListingModeCols(mode: PostListingMode): string {
   switch (mode) {
     case "list":
-      return "col-12";
     case "card":
+      return "col-12";
     case "small_card":
       return "col-12 col-md-6";
```
dessalines: Just change this:

```suggestion
      return "col-12 col-xl-6";
```

The cards should still be doing a 2-column masonry layout, its what differentiates it from a list view.

src/shared/components/common/pictrs-image.tsx
dessalines: Hrm... not sure about this one. `512px` seems plenty large enough, and we def want to downscale no matter, even for the card views, to save on initial page load bandwidth. These images could be 10MB or more, we don't know.

The only time we shouldn't be downscaling IMO, is for the full image viewer.

dessalines:
> The problem is that "Card" and "Small Card" modes are almost identical on large screens. While "Card" does show images a bit larger, its not large enough to view them properly, and so you still need to expand manually for every single image.

I'm not really sure that's the case:
 
Screenshot_2026-05-28-14-07-42-273_org cromite cromite

This to me is plenty large enough for scrolling, and my screen isn't even that large.

src/shared/components/common/pictrs-image.tsx
Nutomic: Okay changed the name to `large_thumbnail` and set it to 2048px.

Nutomic:
Sure the text is still barely readable. But if you use Card view on mobile it shows images practically in fullscreen size. Then switch to desktop with the same mode and suddenly you only get medium-sized thumbnails.

In the Lemmy settings you can only specify a single Listing Mode which applies to all devices. So it should actually have similar functionality everywhere, otherwise you need to change Listing Mode all the time.

dessalines:
If ppl want an ultra-wide view, that's what `list` view mode is for. The card views should use a masonry layout (2 columns for xl screens and higher). This PR as it is undoes that.

In the future we could possibly make the number of columns configurable, but IMO that's a bad idea, because it should be done automatically by the current device screen size. 

We should also probably even add 3 columns or more for `xxl` screens.

Nutomic:
More columns makes no sense if the container has such a limited width. On large screen with browser in fullscreen, Lemmy only uses about half the available width.

dessalines (approved):

PR #4112, opened 2026-05-19
Clear markdown from local storage when submitting post and others

Nutomic:
Some of the markdown forms dont use the Textarea's onSubmit function, so they are not clearing the markdown in localstorage when submitted. You can reproduce this by creating a new post with body text, the body text will be prefilled in the comment form.

Means we need to clear it explicitly in every handler.


dessalines (approved):

dessalines:
We might discover others, but that'll have to be after testing.

Nutomic:
I searched the code MarkdownTextarea and made adjustments wherever it was specified without onSubmit parameter, so these should be all.

PR #4114, opened 2026-05-19
Update theme and language after login/register

Nutomic:


src/shared/components/home/signup.tsx
```diff
@@ -472,6 +472,11 @@ async function handleRegisterSubmit(
 
           if (myUserRes.state === "success") {
             updateMyUserInfo(myUserRes.data);
+            refreshTheme();
+            await I18NextService.reconfigure(
+              window.navigator.languages,
+              myUserRes.data.local_user_view.local_user.interface_language,
+            );
```
Nutomic, line 479: Important for https://github.com/LemmyNet/lemmy/issues/6154

dessalines (approved):

Issue #4132, opened 2026-05-22
Create post page shows too many related posts


Nutomic:
<img width="795" height="1274" alt="Image" src="https://github.com/user-attachments/assets/82c8f9bc-bfe2-48c6-938a-8f8c1f3bf291" />

It seems to use the default post listing limit of 20 items. This means you need to scroll down a lot to reach the other inputs. Should be changed to a lower limit (around 3-5), or use max-height for related posts with scrollbar.

And actually none of the posts here are actually related. It should use title only search instead.

PR #4133, opened 2026-05-22
Better related posts in post form (fixes #4123)

### LINKED ISSUE #4123: Notification picker overflows community sidebar
The text "Replies and mentions (default)" is too long so the text overflows. Changing the text is not a good solution because other languages may also have overflows. Instead we should truncate the button text, and only show the full text inside the dropdown.

Image

Nutomic:
Screenshot_20260522_145224


dessalines (approved):

dessalines:
Fixes #4132

PR #4148, opened 2026-05-25
1.0 prefix for nightly version

Nutomic:


Dockerfile
```diff
@@ -48,7 +48,7 @@ COPY .git .git
 # If the CI is the cron, then use nightly, otherwise use the tag.
 ARG CI_PIPELINE_EVENT
 RUN \
-  VERSION_OUT=$([ "$CI_PIPELINE_EVENT" = "cron" ] && echo "nightly-$(date -u +"%Y-%m-%d")" || echo $(git describe --tag)); \
+  VERSION_OUT=$([ "$CI_PIPELINE_EVENT" = "cron" ] && echo "1.0.0-nightly-$(date -u +"%Y-%m-%d")" || echo $(git describe --tag)); \
```
dessalines: Should probably be `$(git describe --tag --abbrev=0)-nightly-...`

Nutomic: That would generate `1.0.0-beta.0-nightly-2026-05-26` which looks wrong, cant be nightly and beta at the same time.

Nutomic: Found a solution with sed

dessalines: Either one works for me. The sed isn't that necessary because it is the correct tag. But this is fine, the main thing is that nightly is appended to the end.

dessalines (approved):

Dockerfile
dessalines: I'm also thinking that even the full git version (ie not abbrev=0), might also not be bad for the cron builds. Gives a specific commit to reference.

Issue #4131, opened 2026-05-22
Suggestions for Lemmy-ui Mobile Layout


qiancsff:
1、Post filtering controls on mobile look cluttered; they should follow the layout shown in the right image.

2、If a post has no image, do not display the blank placeholder in the list view.

3、The navigation bar should be moved to the bottom to better align with mobile usage habits.

<img width="870" height="830" alt="Image" src="https://github.com/user-attachments/assets/a12b959d-f2e9-4a59-963e-053a22c0b5dd" />

dessalines:
> Post filtering controls on mobile look cluttered; they should follow the layout shown in the right image.

As a mobile first app, we're following the material 3 filter chip guidelines: https://m3.material.io/components/chips/overview . The reason they're on multiple rows, is because in the example you linked, there are only 4 filters (and all can fit on one row), whereas we have 7. I also originally wanted to hide the label for the filters (to save on horizontal space), but it made things a bit less comprehensible. There's discussion in another issue about that.

> If a post has no image, do not display the blank placeholder in the list view.

I agree with this. It really doesn't make sense to show a blank placeholder that contains no info.

> The navigation bar should be moved to the bottom to better align with mobile usage habits.

When there were plans to rework the UI with leptos, I had planned to do this, and use a proper navigation bar: https://m3.material.io/components/navigation-bar/overview . We have this in jerboa (the lemmy android app), but not in lemmy-ui , as it'd require a lot of work.

Nutomic:
> As a mobile first app, we're following the material 3 filter chip guidelines: https://m3.material.io/components/chips/overview . The reason they're on multiple rows, is because in the example you linked, there are only 4 filters (and all can fit on one row), whereas we have 7. I also originally wanted to hide the label for the filters (to save on horizontal space), but it made things a bit less comprehensible. There's discussion in another issue about that.

Yet its true that these take a lot of space on small screens like this. We should reduce the margins between items, and potentially move more of them into the dropdown to save space.

Also the buttons for "Show subscribed" and "Show sidebar" should have different colors to be highlighted.

PR #4149, opened 2026-05-25
Reduce margins for select inputs (ref #4131)

Nutomic:
Screen Shot 2026-05-25 at 12 17 47

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -759,20 +759,20 @@ export class Community extends Component<CommunityRouteProps, State> {
     return (
       <>
         <div className="row row-cols-auto align-items-center g-3 mb-3">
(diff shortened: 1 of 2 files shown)


src/shared/components/community/community.tsx
```diff
@@ -759,20 +759,20 @@ export class Community extends Component<CommunityRouteProps, State> {
     return (
       <>
         <div className="row row-cols-auto align-items-center g-3 mb-3">
```
dessalines: Just decrease the gaps to `g-3` to `g-2` or something.

https://getbootstrap.com/docs/5.3/layout/gutters/#how-they-work

Nutomic: Right thats much easier.

dessalines (approved):

PR #4150, opened 2026-05-25
Change "Show sidebar", "Show subscribed" buttons (ref #4131)

Nutomic:
Use shorter strings to save space, and highlight these buttons with different colors. 

Screen Shot 2026-05-25 at 12 26 23

Screenshot includes https://github.com/LemmyNet/lemmy-ui/pull/4149

Requires https://github.com/LemmyNet/lemmy-translations/pull/263

### DIFF
--- a/src/shared/components/common/filter-chip-checkbox.tsx
+++ b/src/shared/components/common/filter-chip-checkbox.tsx
@@ -40,7 +40,10 @@ function ChipCheckbox({
         onChange={e => onCheck(e.target.checked)}
       />
       <label
-        className={classNames("btn btn-sm btn-outline-primary", className)}
+        className={classNames(
+          "btn btn-sm btn-light border-light-subtle",
+          className,
+        )}
         for={`filter-check-${id}`}
       >
         <Icon icon={icon} classes="icon-inline me-1" />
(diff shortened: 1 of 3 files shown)


src/shared/components/common/filter-chip-checkbox.tsx
```diff
@@ -40,10 +40,7 @@ function ChipCheckbox({
         onChange={e => onCheck(e.target.checked)}
       />
       <label
-        className={classNames(
-          "btn btn-sm btn-light border-light-subtle",
-          className,
-        )}
+        className={classNames("btn btn-sm btn-outline-primary", className)}
```
Nutomic: `btn-outline-secondary` doesnt look ideal, but `btn-outline-primary` or `btn-outline-info` also dont look better.

dessalines: Remove these color changes. These buttons should all look uniform.

I've gone to a muted look, and the only colors that should be shown (outside of a few rare cases) are for communities or persons.

dessalines (commented):
The translation changes are fine, but def don't change the button styles.

src/shared/components/common/filter-chip-checkbox.tsx
Nutomic: I do think that these two buttons should have a slightly different style, because they have a different functionality. Maybe a darker grey tone. Reverted for now because I cant find anything that looks good in the theme.

dessalines (approved):
Feel free to merge once you get the translation conflicts.

PR #4161, opened 2026-05-25
Don't show thumbnails for text posts.

dessalines:
- This removes the pointless message-square thumbnail for text-only posts.
- #4131

### Before

Screenshot_2026-05-25-13-25-58-425_mark via

### After

Screenshot_2026-05-25-13-26-18-912_mark via

### DIFF
--- a/src/shared/components/person/person-listing.tsx
+++ b/src/shared/components/person/person-listing.tsx
@@ -104,13 +104,13 @@ function AvatarAndName({
     !showAvatars(myUserInfo);
 
   return (
-    <>
+    <span className="overflow-wrap-anywhere">
       {!hideAvatar_ && !banned && avatar && (
         <PictrsImage src={avatar} type="icon" />
       )}
       <span className={nameClasses}>{name}</span>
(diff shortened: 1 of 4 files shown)


Nutomic:
On mobile this makes sense to have more space for the title. But on desktop there is plenty of space available, and it looks more consistent with these thumbnails.

dessalines:
K I fixed it, mainly was due to some incorrect non-overflowed wrappings:

Screenshot_2026-05-26-15-04-26-76_e4424258c8b8649f6e67d283a50a2cbc

Nutomic:
Good! You could also hide the thumbnail for external posts on mobile (grey icon with arrow). Actual image thumbnails should be shown though.

On the other hand, link thumbnails can be useful to see that it opens another site. I will let you decide.

Nutomic (approved):

dessalines:
Ya I considered it, but realized its not the best idea, because the post title always goes to the comments. That big link icon is the main way you can go externally to the link.

PR #4164, opened 2026-05-26
Force show private comm sidebar on mobile (fixes #4163)

### LINKED ISSUE #4163: Cant subscribe to private community from mobile
Problem is that the sidebar on mobile is only available through the nav button, but that is hidden in case of private community.

Image

Nutomic:
Screen Shot 2026-05-26 at 13 17 16

### DIFF
--- a/src/shared/components/community/community.tsx
+++ b/src/shared/components/community/community.tsx
@@ -525,6 +525,7 @@ export class Community extends Component<CommunityRouteProps, State> {
                       )}
                     </div>
                   </div>
+                  {/** Force show sidebar on mobile if it's a private unfollowed community, so that it can be followed.**/}
                   {this.mobileSidebar(true)}
                 </>
               )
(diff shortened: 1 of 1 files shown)


src/shared/components/community/community.tsx
```diff
@@ -511,16 +511,22 @@ export class Community extends Component<CommunityRouteProps, State> {
...
+                  >
+                    <h4 className="alert-heading">
+                      {I18NextService.i18n.t("community_visibility_private")}
+                    </h4>
+                    <div className="card-text">
+                      {I18NextService.i18n.t(
+                        "cant_view_private_community_message",
+                      )}
+                    </div>
                   </div>
```
dessalines, line 527:
```suggestion
                  </div>
                  {/** Force show sidebar on mobile if it's a private unfollowed community, so that it can be followed.**/}
```

dessalines (approved):
Feel free to merge after the suggestion for the comment

PR #4173, opened 2026-05-27
Community tag colors

### LINKED ISSUE #4171: Add map from tag colors to theme colors
### Requirements

We added `color01`, `color02` to the tags, but haven't done any mapping or theming of them yet in lemmy-ui.

dessalines:
## Description

Adds 6 tag colors, as well as a mapping to the bootstrap ones.

I tried to make the badges more subtle (like the buttons), but it seemed a bit dangerous to mess with bootstraps contrasts. As you can see the badge text color changes based on the badge background color, and this is built in to the bootstrap badge class theme. Its better not to mess with it, and just do it the official way: https://getbootstrap.com/docs/5.3/components/badge/

https://github.com/user-attachments/assets/23048268-ea4e-49cc-8994-9fa64a3943e4

Fixes #4171

### DIFF
--- a/src/shared/components/common/tag-color-dropdown.tsx
+++ b/src/shared/components/common/tag-color-dropdown.tsx
@@ -1,0 +1,64 @@
+import { TagColor } from "lemmy-js-client";
+import { FilterChipDropdown, FilterOption } from "./filter-chip-dropdown";
+
+// There are actually 10 color options in the back end, but squash these down
+const tagColorOptions: FilterOption<TagColor>[] = [
+  { noI18n: "1", value: "color01" },
+  { noI18n: "2", value: "color02" },
+  { noI18n: "3", value: "color03" },
+  { noI18n: "4", value: "color04" },
+  { noI18n: "5", value: "color05" },
+  { noI18n: "6", value: "color06" },
+];
+
+type TagColorDropdownProps = {
+  currentOption: TagColor;
+  onSelect: (val: TagColor) => void;
+  className?: string;
+};
+
+export function tagColorToColorClass(tagColor: TagColor): string {
+  switch (tagColor) {
+    case "color01":
+      return "light";
+    case "color02":
+      return "primary";
+    case "color03":
+      return "info";
+    case "color04":
+      return "success";
+    case "color05":
+      return "warning";
+    case "color06":
+      return "danger";
+    case "color07":
+    case "color08":
+    case "color09":
+    case "color10":
+      return "light";
+  }
+}
+
+function tagColorToBtnClasses(tagColor: TagColor): string {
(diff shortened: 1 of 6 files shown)


src/shared/components/common/tag-color-dropdown.tsx
```diff
@@ -0,0 +1,64 @@
...
+    case "color05":
+      return "warning";
+    case "color06":
+      return "danger";
+    case "color07":
+    case "color08":
+    case "color09":
+    case "color10":
+      return "light";
+  }
```
Nutomic, line 39: This doesnt make sense, if we cant handle all colors in the frontend, we need to reduce the number of colors in the backend.

Nutomic:
Choosing between color 1, color 2, color 3 like this looks really unintuitive. Plus it doesnt make sense to use theme colors like success, warning, danger for tags. After all these dont represent any error messages. It would really be better to allow free color choice like Github labels and Piefed tags.

src/shared/components/common/tag-color-dropdown.tsx
dessalines: The number of colors is entirely dependent on different UI frameworks. Some may have only 2-3, some may have 5+. We used 10 in the backend arbitrarily, understanding that most UI frameworks will squash this number. 

In jerboa I'll probably only use 3 out of the 10. A good case could be made here also for restricting it down more.

dessalines:
> Choosing between color 1, color 2, color 3 like this looks really unintuitive.

I prefer these over trying to map the names to bootstrap ones. Especially because on other UI frameworks, a `Warning` color might not exist. So its better to use the enum name, and have that name be consistent across different apps.

> It would really be better to allow free color choice like Github labels and Piefed tags.

Hardcoding colors is a terrible idea, and UI frameworks + themes should be using their own theme colors. This is what pretty much every UI framework does nowadays, and its why I demanded that we switched away from hardcoding hex colors to using `Color0X` enums, and let themes pick the colors that go with their theme design. A good rule of thumb in designing UI's is to never hardcode colors, always let theme designers do that.

Nutomic:
For colors you should use [these here](https://getbootstrap.com/docs/5.3/customize/color/#all-colors) (blue-400, orange-400 etc). These will fit with all themes considering that tags are rendered very small. And it wont clash with actual error or info messages.

dessalines:
Those are not theme colors, and most themes won't (and shouldn't) define those. They aren't guaranteed to look good across light / dark themes. Best to use the theme colors as badges suggest:

https://getbootstrap.com/docs/5.3/components/badge/#background-colors

Nutomic (approved):

Nutomic:
Okay lets merge like this, but I dont think its the best solution.

PR #4179, opened 2026-05-28
Notification display fixes

### LINKED ISSUE #4169: Notification display
Image

On mobile, notifications for new private community follows are not shown in the top bar. They are only visible when opening the menu.

A possible solution would be to hide notification/report/application buttons from the navbar if they have zero items. So only types with 1+ unread items would be shown in navbar.

dessalines:
## Description

- Removes pointless duped counts in mobile navbar
- Uses feather icons for lock, unlock (most of our icons are from this)
- Add pending subs to mobile navbar.
- Fixes #4169 

I thought about hiding the mobile navbar icons also (if they have no notifs), but it seems like a bad idea, as these are shortcuts to get to those pages quickly.

Screenshot_2026-05-28-15-02-52-17_5dca195d7d09adf043e42a8f3ad62b19


Nutomic:
For notifications and maybe reports it makes sense to show the shortcuts always. But for unread applications and pending follows they are really irrelevant if there are no unread items. So at least these two can be hidden if empty.

dessalines:
I realized only the notif button is best to always show. The others I'm hiding now if the counts are zero.

Nutomic (approved):

PR #4181, opened 2026-05-28
Post / Comment listing fixes

### LINKED ISSUE #4167: Inconsistent design for posts and comments
Image

This one is particularly noticable in mixed listings like the user profile. 
Posts: Avatar + Highlighted Username + "to" + Community with Icon
Comments: Community with Icon + "by" + Username without Avatar and not highlighted

Both of these should use the same word order (ie always community first, or always username first). The avatar should always be shown. Post usernames shouldnt have orange background.

dessalines:
## Description

- Only show the OP badged username for the post page. 
- Made muted for community-link and person-listing mandatory, and changed those values where it made sense. Main rule of thumb is to not highlight both at the same time, since its too many colors. In short: for post listings, mute only the poster name and don't mute the community, and for comments, don't mute the name. Either way they're tweakable now.
- If showing the community in a title line, always put it first, so its **(avatar) community** by **user**. Now post listings on the home page match the notifs. I tried it the other way around (IE user TO community), and it looked terrible, and I checked a few other apps (photon, jerboa), and they all do it the first way. Highlighting only the community looks best, mute everything else.
- I tried re-adding back user avatars for these post title lines, but it looked ugly. There's a reason why most apps don't do them. Best to only highlight one item per line.
- Externalized `hideAvatar`, rather than inferring it from a `showCommunity` prop. This lets us hide it from the profile view.
- Fixes #4167 

## Screenshots

Home:

Screenshot_2026-05-28-16-58-00-484_org cromite cromite

Notifs:

Screenshot_2026-05-28-16-57-44-382_org cromite cromite


src/shared/components/comment/comment-node.tsx
```diff
@@ -88,6 +88,10 @@ type CommentNodeProps = {
   allLanguages: Language[];
   siteLanguages: number[];
   hideImages: boolean;
+  showBadgeForPostCreator: boolean;
+  mutedPersonName: boolean;
+  mutedCommunityName: boolean;
+  hideAvatar: boolean;
```
Nutomic, line 94: These are false most of the time, so make them nullable with `?` to avoid writing them out every time.

src/shared/components/post/post-listing.tsx
```diff
@@ -73,6 +73,9 @@ type PostListingProps = {
   markReadLoading: boolean;
   voteLoading: boolean;
   topBorder: boolean;
+  mutedPersonName: boolean;
+  mutedCommunityName: boolean;
+  hideAvatar: boolean;
```
Nutomic, line 78: Also make these nullable. Same for other boolean fields you added.

src/shared/components/comment/comment-nodes.tsx
```diff
@@ -59,6 +59,10 @@ interface CommentNodesProps {
   showCommunity: boolean;
   viewType: CommentViewType;
   showMarkRead: ShowMarkReadType;
+  showBadgeForPostCreator: boolean;
+  mutedPersonName: boolean;
+  mutedCommunityName: boolean;
```
Nutomic:
```suggestion
  mutePersonName: boolean;
  muteCommunityName: boolean;
```

Right now it sounds like a username string which should be muted. As a boolean the name should start with `mute`. 

Also make these optional.

Nutomic:
The design is good, much better than before!

src/shared/components/comment/comment-nodes.tsx
dessalines: Makes sense, I'll change the names shortly.

As for the optionals though, I'm trying to move away from those, and convert all to required. Both in jerboa and here, I found its best (even tho its verbose), to try to make every component param required, even if its optional. It forces you to look at every usage of that component, and handle every case. When they're optional, you end up overlooking a lot of cases. And when you add that component somewhere else, it makes you think about which params make sense, rather than just forgetting about optionals.

Nutomic (approved):

Nutomic:
Need to resolve some conflicts.

PR #4206, opened 2026-06-02
Make community title optional, reduce min actor name length

Nutomic:
Requires https://github.com/LemmyNet/lemmy-js-client/pull/976

### DIFF
--- a/src/shared/components/common/tag-color-dropdown.tsx
+++ b/src/shared/components/common/tag-color-dropdown.tsx
@@ -1,5 +1,6 @@
 import { TagColor } from "lemmy-js-client";
 import { FilterChipDropdown, FilterOption } from "./filter-chip-dropdown";
+import { NoOptionI18nKeys } from "i18next";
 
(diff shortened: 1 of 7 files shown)


src/shared/components/common/tag-color-dropdown.tsx
```diff
@@ -55,7 +56,7 @@ export function TagColorDropdown({
   return (
     <FilterChipDropdown
       allOptions={tagColorOptions}
-      label={"color"}
+      label={"color" as NoOptionI18nKeys}
```
dessalines: This isn't right, color should be translated now.  Update the git submodules then do this:

```suggestion
      label={"color"}
```

dessalines (commented):
After addressing my comment below, feel free to merge.

src/shared/components/common/tag-color-dropdown.tsx
Nutomic: Removed

PR #4207, opened 2026-06-02
Hide notification icons without unread items on desktop (fixes #4205)

### LINKED ISSUE #4205: On desktop hide icons for registration application, pending follows etc
Image

All these icons add unnecessary clutter. If the respective numbers are zero, the buttons should be hidden and moved to the dropdown menu instead. 

Similar to https://github.com/LemmyNet/lemmy-ui/pull/4179 but that is only for mobile.

Nutomic:
Basically the same change as https://github.com/LemmyNet/lemmy-ui/pull/4179 but for desktop.

Screenshot_20260602_130818

Fixes #4205

### DIFF
--- a/src/shared/components/app/navbar.tsx
+++ b/src/shared/components/app/navbar.tsx
@@ -303,50 +303,51 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
                           formattedCount: numToSI(this.state.unreadNotifsCount),
                         })}
                       </span>
+                      {this.state.unreadNotifsCount > 0 && (
+                        <span className="mx-1 badge rounded-pill text-bg-danger">
+                          {numToSI(this.state.unreadNotifsCount)}
+                        </span>
+                      )}
                     </NavLink>
                   </li>
-                  {moderatesSomething(this.props.myUserInfo) && (
-                    <li id="navModeration" className="nav-item">
-                      <NavLink
-                        className="nav-link d-inline-flex align-items-center d-md-inline-block"
-                        to="/reports"
-                        title={I18NextService.i18n.t("unread_reports", {
-                          count: Number(this.state.unreadReportCount),
-                          formattedCount: numToSI(this.state.unreadReportCount),
-                        })}
-                        onMouseUp={() => handleCollapseClick(this)}
-                      >
-                        <Icon icon="shield" />
-                        <span className="badge rounded-pill text-bg-danger d-inline ms-1 d-md-none ms-md-0">
-                          {I18NextService.i18n.t("unread_reports", {
+                  {moderatesSomething(this.props.myUserInfo) &&
+                    this.state.unreadReportCount > 0 && (
+                      <li id="navModeration" className="nav-item">
+                        <NavLink
+                          className="nav-link d-inline-flex align-items-center d-md-inline-block"
+                          to="/reports"
+                          title={I18NextService.i18n.t("unread_reports", {
                             count: Number(this.state.unreadReportCount),
                             formattedCount: numToSI(
                               this.state.unreadReportCount,
                             ),
                           })}
-                        </span>
-                      </NavLink>
-                    </li>
-                  )}
-                  {amAdmin(this.props.myUserInfo) && (
-                    <li id="navApplications" className="nav-item">
-                      <NavLink
-                        to="/registration_applications"
-                        className="nav-link d-inline-flex align-items-center d-md-inline-block"
-                        title={I18NextService.i18n.t(
-                          "unread_registration_applications",
-                          {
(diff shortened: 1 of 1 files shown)


src/shared/components/app/navbar.tsx
```diff
@@ -303,50 +303,51 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
...
+                              formattedCount: numToSI(
+                                this.state.unreadReportCount,
+                              ),
+                            })}
+                          </span>
+                          {this.state.unreadReportCount > 0 && (
+                            <span className="mx-1 badge rounded-pill text-bg-danger">
+                              {numToSI(this.state.unreadReportCount)}
+                            </span>
+                          )}
```
Nutomic, line 313: These badges were removed in https://github.com/LemmyNet/lemmy-ui/pull/4179/changes but they should remain unchanged.

dessalines:
The PR didn't get linked again. I'm convinced its because github can't handle when you put `fixes ...` in the commit title. 

Move it to the commit body / comment if you can from now on.

src/shared/components/app/navbar.tsx
```diff
@@ -432,6 +461,54 @@ export class Navbar extends Component<NavbarProps, NavbarState> {
...
+                            >
+                              <Icon icon="clipboard" classes="me-1" />
+                              {I18NextService.i18n.t(
+                                "registration_applications",
+                              )}
+                            </NavLink>
+                          </li>
+                        )}
+                        {moderatesPrivateCommunity(this.props.myUserInfo) &&
+                          this.state.unreadPendingFollowsCount && (
```
dessalines: This one has a count check (but without actually `> 0`), and the other ones in this block don't.

dessalines:
I'm not a huge fan of the complexity this adds, and its not really too bad a thing to always see mod type items on a wide screen anyway.

src/shared/components/app/navbar.tsx
Nutomic: Removed, dropdown items should always be shown.

Nutomic:
Its not a problem, but the main design changes in 1.0 were to declutter the UI by moving post actions etc to dropdown menus. Here its very similar, always showing these items adds clutter and doesnt look clean. It makes sense to use the dropdown menu more, which was practically empty before.

To reduce complexity its better to move some of this stuff into components and reuse the logic. I did that for some of the elements, but others seem to be slightly different so its hard to get everything correct.

src/shared/components/app/navbar.tsx
```diff
@@ -501,3 +487,55 @@ async function handleLogOut(i: Navbar) {
...
+  label: NoOptionI18nKeys;
+}
+function Mynavlink({ i, to, icon, label }: NavlinkProps) {
+  return (
+    <li>
+      <NavLink
+        to={to}
+        className="dropdown-item px-2"
+        title={I18NextService.i18n.t(label)}
+        onClick={() => handleCollapseClick(i)}
```
Nutomic: This is actually only needed for mobile, on desktop the menu closes automatically. So it could be left out here.

src/shared/components/app/navbar.tsx
```diff
@@ -501,3 +487,55 @@ async function handleLogOut(i: Navbar) {
   await UserService.Instance.logout();
   handleCollapseClick(i);
 }
+
+interface NavlinkProps {
+  i: Navbar;
```
dessalines: Never inject a component into another component.

Add an `onClick: () => void;` if you need to handle that action.

src/shared/components/app/navbar.tsx
```diff
@@ -501,3 +487,55 @@ async function handleLogOut(i: Navbar) {
...
+  to: string;
+  icon: string;
+  label: NoOptionI18nKeys;
+}
+function Mynavlink({ i, to, icon, label }: NavlinkProps) {
```
dessalines: `MyNavLink`

src/shared/components/app/navbar.tsx
```diff
@@ -501,3 +487,55 @@ async function handleLogOut(i: Navbar) {
   await UserService.Instance.logout();
   handleCollapseClick(i);
 }
+
+interface NavlinkProps {
```
dessalines: Would be better to call this `MyNavLinkProps`

src/shared/components/app/navbar.tsx
Nutomic: Fixed

src/shared/components/app/navbar.tsx
Nutomic: Changed, although the name is not great.

dessalines (approved):
👍

PR #4208, opened 2026-06-02
Fixing a few typos

dessalines:


src/shared/components/common/modal/create-or-edit-oauth-provider-modal.tsx
```diff
@@ -294,7 +294,7 @@ export default class CreateOrEditOAuthProviderModal extends Component<
                 <div className="row">
                   <div className="col">
                     <ProviderCheckboxField
-                      id="auto-verfiy-email"
+                      id="auto-verify-email"
```
Nutomic, line 297: This id is not used anywhere else, might as well remove it.

dessalines: The component requires it for some reason. Might be useful for front ends also.

Nutomic (approved):

PR #4228, opened 2026-06-10
Adding stacked mobile tables

### LINKED ISSUE #4226: Make tables easier to read on mobile
Image

Its not clear which labels and numbers belong together. Maybe it would be better to hide the table headers on mobile, and instead show labels right next to each number.

dessalines:
- Replaces the weird tables in favor of a "stacked" method, where column name and data each take half the screen.
- I tried some sticky header alternatives but they all looked ugly.
- Wide / tables above `md` stay the same.
- Fixes #4226

# Examples

## Communities

Screenshot_2026-06-10-19-49-31-90_e4424258c8b8649f6e67d283a50a2cbc

## Instances

Screenshot_2026-06-10-19-49-44-01_e4424258c8b8649f6e67d283a50a2cbc

## Modlog

Screenshot_2026-06-10-19-50-01-48_e4424258c8b8649f6e67d283a50a2cbc

### DIFF
--- a/src/shared/components/common/tables.tsx
+++ b/src/shared/components/common/tables.tsx
@@ -1,1 +1,18 @@
+import { I18NextService } from "@services/I18NextService";
+import { NoOptionI18nKeys } from "i18next";
+
 export const TableHr = () => <hr className="border border-secondary" />;
+
+type ResponsiveTableRowHeaderProps = {
+  title: NoOptionI18nKeys;
+};
+
+export function ResponsiveTableRowHeader({
+  title,
+}: ResponsiveTableRowHeaderProps) {
+  return (
+    <div className="d-md-none col-6">
+      <div className="fw-bold">{I18NextService.i18n.t(title)}</div>
+    </div>
+  );
+}
--- a/src/shared/components/common/media-uploads.tsx
+++ b/src/shared/components/common/media-uploads.tsx
@@ -13,6 +13,7 @@ import { PictrsImage } from "./pictrs-image";
 import { httpBackendUrl } from "@utils/env";
 import { toast } from "@utils/app";
 import { ResponsiveTableRowHeader, TableHr } from "./tables";
+import classNames from "classnames";
 
 interface Props {
   uploads: PagedResponse<LocalImageView>;
@@ -26,6 +27,7 @@ export class MediaUploads extends Component<Props, never> {
     const images = this.props.uploads.items;
 
     const cols = "col-6 col-md-3";
+    const imageCols = "col-12 col-md-3";
 
     return (
       <div className="media-uploads">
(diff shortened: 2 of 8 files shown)


src/shared/components/common/tables.tsx
```diff
@@ -1 +1,18 @@
...
+
+type ResponsiveTableRowHeaderProps = {
+  title: NoOptionI18nKeys;
+};
+
+export function ResponsiveTableRowHeader({
+  title,
+}: ResponsiveTableRowHeaderProps) {
+  return (
+    <div className="d-md-none col-6">
```
dessalines, line 14: Currently this is always taking `col-md-6`, although I spose it could be configurable. 

Especially for the modlog screen, where the column names like time, mod, and action should probably take less than half the screen.

Nutomic:
Definitely much easier to read like this. Its basically a type of card layout. Some thoughts:
- For community list I would hide the "Name" label, and just put the icon + name larger as title
- The other labels (users, posts etc) dont need to be bold
- For modlog put the action on top or as title. Same for instances etc, put the name as title without any label.
- Names in modlog are always hidden if current user is not a mod/admin, so hide the line "Moderator - moderator" in that case
- For uploads half the screen is empty now and only shows the delete button. It would look better with a floating delete button on top of the image.

dessalines:
> For community list I would hide the "Name" label, and just put the icon + name larger as title

I tried this but it looks ugly, and not aligned. The stacked table looks better when the column name and the item are aligned.

> The other labels (users, posts etc) dont need to be bold

The column name should always be bold, to mark it out from the items. This is how it works in regular tables also, so its better to keep to that standard.

> For modlog put the action on top or as title. Same for instances etc, put the name as title without any label.

I'll do this shortly. Probably action, mod, time makes the most sense. But I think I will keep the title label, because it looks better when its aligned.

> Names in modlog are always hidden if current user is not a mod/admin, so hide the line "Moderator - moderator" in that case

I'll fix shortly.

> For uploads half the screen is empty now and only shows the delete button. It would look better with a floating delete button on top of the image.

I'd like to keep the delete button because its explicit, but I'll try moving the image to its own row for mobile.

dessalines:
> so hide the line "Moderator - moderator" in that case

Actually for this one, we do indicate whether its a mod or admin, without having the persons name. So its better to keep for both mobile and desktop.

dessalines:
modlog:

Screenshot_2026-06-11-09-53-40-15_e4424258c8b8649f6e67d283a50a2cbc

uploads:

Screenshot_2026-06-11-09-53-12-28_e4424258c8b8649f6e67d283a50a2cbc

Nutomic:
>> For community list I would hide the "Name" label, and just put the icon + name larger as title
>
> I tried this but it looks ugly, and not aligned. The stacked table looks better when the column name and the item are aligned.

Would you show a screenshot how it looks? 

>> so hide the line "Moderator - moderator" in that case
> 
> Actually for this one, we do indicate whether its a mod or admin, without having the persons name. So its better to keep for both mobile and desktop.

Then we should show "Moderator" or "Administrator" only once, not twice.

dessalines:
Here's what it looks when removing the title only for the first row. IMO its more confusing than having a 2x2 table.

Screenshot_2026-06-16-10-54-04-15_a87fd7db6caa850b517aa6fa9d2fcd0e

> Then we should show "Moderator" or "Administrator" only once, not twice.

IMO its the same issue. Its better if every line has 2 items, column name and value.

Nutomic:
That would look fine if you center the name horizontally, and add some padding below. Basically like a card title.

dessalines:
I'd really rather just keep the layout as a 2x2 table, rather than centering.

Nutomic (approved):

PR #4274, opened 2026-07-15
Add report resolve reason

### LINKED ISSUE #4273: Resolve Report with Comment (UI part)
### Requirements

UI part of https://github.com/LemmyNet/lemmy/issues/6438

The way I'm thinking it is when the user clicks 'Resolve' then a follow up text box opens up for the resolve reason.
I'm not sure in the case of unresolving a report if the resolve reason should also be set to null or not.

iByteABit256:
## Description

Fixes #4273

### Before

**No input for resolve reason, button sends the PUT request straight away**

image

### After

**You can set the reason for resolving and unresolving**

image

image

image

### DIFF
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -160,10 +160,7 @@ export class CommentReport extends Component<
             )}
             {conclusion && (
               <div>
-                {I18NextService.i18n.t(
-                  resolved ? "resolve_reason" : "unresolve_reason",
-                )}
-                : {conclusion}
+                {I18NextService.i18n.t("conclusion")}: {conclusion}
               </div>
             )}
           </div>
@@ -239,7 +236,7 @@ export class CommentReport extends Component<
         <ConcludeReportModal
           isResolved={resolved}
           conclusion={conclusion}
-          onSubmit={reason => handleResolveReport(this, reason)}
+          onSubmit={conclusion => handleResolveReport(this, conclusion)}
           onCancel={() => this.setState({ showResolveReportDialog: false })}
           show={this.state.showResolveReportDialog}
           loading={this.props.loading}
@@ -273,12 +270,12 @@ function handleAdminBan(i: CommentReport) {
   });
 }
 
-function handleResolveReport(i: CommentReport, reason?: string) {
+function handleResolveReport(i: CommentReport, conclusion?: string) {
   const cr = i.props.report.comment_report;
   i.props.onResolveReport({
     report_id: cr.id,
     resolved: !cr.resolved,
-    conclusion: reason,
+    conclusion: conclusion,
   });
   i.setState({ showResolveReportDialog: false });
 }
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -49,19 +49,6 @@ export class CommentReport extends Component<
     showResolveReportDialog: false,
   };
 
-  handleResolveReport = (form: {
-    reason?: string;
-    action: "resolve" | "unresolve";
-  }) => {
-    const cr = this.props.report.comment_report;
-    this.props.onResolveReport({
-      report_id: cr.id,
-      resolved: !cr.resolved,
-      resolve_reason: form.reason,
-    });
-    this.setState({ showResolveReportDialog: false });
-  };
-
   render() {
     const r = this.props.report;
     const comment = r.comment;
@@ -176,11 +163,6 @@ export class CommentReport extends Component<
                 {I18NextService.i18n.t("resolve_reason")}: {resolveReason}
               </div>
             )}
-            {!resolved && resolveReason && (
-              <div>
-                {I18NextService.i18n.t("unresolve_reason")}: {resolveReason}
-              </div>
-            )}
           </div>
         )}
         <div className="row row-cols-auto align-items-center gx-3 my-2">
@@ -254,7 +236,7 @@ export class CommentReport extends Component<
         {this.state.showResolveReportDialog && (
           <ResolveReportModal
             isResolved={resolved}
-            onSubmit={this.handleResolveReport}
+            onSubmit={reason => handleResolveReport(this, reason)}
             onCancel={() => this.setState({ showResolveReportDialog: false })}
             show
             loading={this.props.loading}
@@ -287,4 +269,14 @@ function handleAdminBan(i: CommentReport) {
     person: i.props.report.comment_creator,
     ban: !i.props.report.creator_banned,
   });
+}
+
+function handleResolveReport(i: CommentReport, reason?: string) {
+  const cr = i.props.report.comment_report;
+  i.props.onResolveReport({
+    report_id: cr.id,
+    resolved: !cr.resolved,
+    resolve_reason: reason,
+  });
+  i.setState({ showResolveReportDialog: false });
 }
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -53,7 +53,7 @@ export class CommentReport extends Component<
     const r = this.props.report;
     const comment = r.comment;
     const resolved = r.comment_report.resolved;
-    const resolveReason = r.comment_report.resolve_reason;
+    const resolveReason = r.comment_report.conclusion;
 
     // Set the original post data ( a troll could change it )
     comment.content = r.comment_report.original_comment_text;
@@ -278,7 +278,7 @@ function handleResolveReport(i: CommentReport, reason?: string) {
   i.props.onResolveReport({
     report_id: cr.id,
     resolved: !cr.resolved,
-    resolve_reason: reason,
+    conclusion: reason,
   });
   i.setState({ showResolveReportDialog: false });
 }
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -39,8 +39,6 @@ interface CommentReportState {
   showResolveReportDialog: boolean;
 }
 
-const REASON_DELIMITER = "|";
-
 @tippyMixin
 export class CommentReport extends Component<
   CommentReportProps,
@@ -56,7 +54,6 @@ export class CommentReport extends Component<
     const comment = r.comment;
     const resolved = r.comment_report.resolved;
     const resolveReason = r.comment_report.resolve_reason;
-    const displayReason = getDisplayReason(resolveReason, resolved);
 
     // Set the original post data ( a troll could change it )
     comment.content = r.comment_report.original_comment_text;
@@ -161,12 +158,9 @@ export class CommentReport extends Component<
                 />
               </T>
             )}
-            {displayReason && (
+            {resolved && resolveReason && (
               <div>
-                {I18NextService.i18n.t(
-                  resolved ? "resolve_reason" : "unresolve_reason",
-                )}
-                : {displayReason}
+                {I18NextService.i18n.t("resolve_reason")}: {resolveReason}
               </div>
             )}
           </div>
@@ -242,7 +236,6 @@ export class CommentReport extends Component<
         {this.state.showResolveReportDialog && (
           <ResolveReportModal
             isResolved={resolved}
-            resolveReason={resolveReason}
             onSubmit={reason => handleResolveReport(this, reason)}
             onCancel={() => this.setState({ showResolveReportDialog: false })}
             show
@@ -276,20 +269,6 @@ function handleAdminBan(i: CommentReport) {
     person: i.props.report.comment_creator,
     ban: !i.props.report.creator_banned,
   });
-}
-
-function getDisplayReason(
-  resolveReason?: string,
-  isResolved: boolean = true,
-): string | undefined {
-  if (!resolveReason) return undefined;
-
-  const parts = resolveReason.split(REASON_DELIMITER);
-  if (parts.length === 2) {
-    // Return the appropriate part based on resolution state
-    return isResolved ? parts[0] : parts[1];
-  }
-  return resolveReason;
 }
 
 function handleResolveReport(i: CommentReport, reason?: string) {
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -39,6 +39,8 @@ interface CommentReportState {
   showResolveReportDialog: boolean;
 }
 
+const REASON_DELIMITER = "|";
+
 @tippyMixin
 export class CommentReport extends Component<
   CommentReportProps,
@@ -54,6 +56,7 @@ export class CommentReport extends Component<
     const comment = r.comment;
     const resolved = r.comment_report.resolved;
     const resolveReason = r.comment_report.resolve_reason;
+    const displayReason = getDisplayReason(resolveReason, resolved);
 
     // Set the original post data ( a troll could change it )
     comment.content = r.comment_report.original_comment_text;
@@ -158,9 +161,12 @@ export class CommentReport extends Component<
                 />
               </T>
             )}
-            {resolved && resolveReason && (
+            {displayReason && (
               <div>
-                {I18NextService.i18n.t("resolve_reason")}: {resolveReason}
+                {I18NextService.i18n.t(
+                  resolved ? "resolve_reason" : "unresolve_reason",
+                )}
+                : {displayReason}
               </div>
             )}
           </div>
@@ -236,6 +242,7 @@ export class CommentReport extends Component<
         {this.state.showResolveReportDialog && (
           <ResolveReportModal
             isResolved={resolved}
+            resolveReason={resolveReason}
             onSubmit={reason => handleResolveReport(this, reason)}
             onCancel={() => this.setState({ showResolveReportDialog: false })}
             show
@@ -269,6 +276,20 @@ function handleAdminBan(i: CommentReport) {
     person: i.props.report.comment_creator,
     ban: !i.props.report.creator_banned,
   });
+}
+
+function getDisplayReason(
+  resolveReason?: string,
+  isResolved: boolean = true,
+): string | undefined {
+  if (!resolveReason) return undefined;
+
+  const parts = resolveReason.split(REASON_DELIMITER);
+  if (parts.length === 2) {
+    // Return the appropriate part based on resolution state
+    return isResolved ? parts[0] : parts[1];
+  }
+  return resolveReason;
 }
 
 function handleResolveReport(i: CommentReport, reason?: string) {
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -50,7 +50,7 @@ export class CommentReport extends Component<
   };
 
   handleResolveReport = (form: {
-    reason: string;
+    reason?: string;
     action: "resolve" | "unresolve";
   }) => {
     const cr = this.props.report.comment_report;
@@ -66,6 +66,7 @@ export class CommentReport extends Component<
     const r = this.props.report;
     const comment = r.comment;
     const resolved = r.comment_report.resolved;
+    const resolveReason = r.comment_report.resolve_reason;
 
     // Set the original post data ( a troll could change it )
     comment.content = r.comment_report.original_comment_text;
@@ -169,6 +170,16 @@ export class CommentReport extends Component<
                   muted={false}
                 />
               </T>
+            )}
+            {resolved && resolveReason && (
+              <div>
+                {I18NextService.i18n.t("resolve_reason")}: {resolveReason}
+              </div>
+            )}
+            {!resolved && resolveReason && (
+              <div>
+                {I18NextService.i18n.t("unresolve_reason")}: {resolveReason}
+              </div>
             )}
           </div>
         )}
--- a/src/shared/components/comment/comment-report.tsx
+++ b/src/shared/components/comment/comment-report.tsx
@@ -19,6 +19,7 @@ import {
   BanFromSiteData,
 } from "@components/person/reports";
 import ModActionFormModal from "@components/common/modal/mod-action-form-modal";
+import ResolveReportModal from "@components/common/modal/resolve-report-modal";
 import { commentToFlatNode } from "@utils/app";
 
 interface CommentReportProps {
@@ -35,6 +36,7 @@ interface CommentReportProps {
 
 interface CommentReportState {
   showRemoveCommentDialog: boolean;
+  showResolveReportDialog: boolean;
 }
 
 @tippyMixin
@@ -44,14 +46,26 @@ export class CommentReport extends Component<
 > {
   state: CommentReportState = {
     showRemoveCommentDialog: false,
+    showResolveReportDialog: false,
+  };
+
+  handleResolveReport = (form: {
+    reason: string;
+    action: "resolve" | "unresolve";
+  }) => {
+    const cr = this.props.report.comment_report;
+    this.props.onResolveReport({
+      report_id: cr.id,
+      resolved: !cr.resolved,
+      resolve_reason: form.reason,
+    });
+    this.setState({ showResolveReportDialog: false });
   };
 
   render() {
     const r = this.props.report;
     const comment = r.comment;
-    const tippyContent = I18NextService.i18n.t(
-      r.comment_report.resolved ? "unresolve_report" : "resolve_report",
-    );
+    const resolved = r.comment_report.resolved;
 
     // Set the original post data ( a troll could change it )
     comment.content = r.comment_report.original_comment_text;
@@ -135,7 +149,7 @@ export class CommentReport extends Component<
         </div>
         {r.resolver && (
           <div>
-            {r.comment_report.resolved ? (
+            {resolved ? (
               <T i18nKey="resolved_by">
                 #
                 <PersonListing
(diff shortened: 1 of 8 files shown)


iByteABit256:
Woodpecker is failing because of the lemmy-js-client version, I need a new tag there to be able to fix it

Edit: Also I forgot about displaying the reason as well, I'll mark it draft for now and add that too

Nutomic:
I tagged js-client `1.0.0-resolve-reason.0` for you.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+    const action = this.props.isResolved ? "unresolve" : "resolve";
+    let reason = this.state.reason.trim();
+
+    // Only set default if no reason provided
+    if (!reason) {
+      reason =
+        action === "resolve"
+          ? "Resolved by moderator"
+          : "Unresolved by moderator";
```
Nutomic: If no reason is provided it should simply be undefined.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    this.props.onSubmit({ reason, action });
```
Nutomic: If you unresolve a report, this will quietly overwrite the existing reason correct? Would be better to append if there is a reason already.

src/shared/components/comment/comment-report.tsx
```diff
@@ -44,14 +46,26 @@ export class CommentReport extends Component<
 > {
   state: CommentReportState = {
     showRemoveCommentDialog: false,
+    showResolveReportDialog: false,
+  };
+
+  handleResolveReport = (form: {
+    reason: string;
+    action: "resolve" | "unresolve";
+  }) => {
```
dessalines: Make this a top level function, not a closure. Look at any other handler in this codebase.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { Spinner } from "../icon";
+import { randomStr } from "@utils/helpers";
+import type { Modal } from "bootstrap";
+import classNames from "classnames";
+import { modalMixin } from "../../mixins/modal-mixin";
+import { LoadingEllipses } from "../loading-ellipses";
+
+interface ResolveReportModalProps {
+  isResolved: boolean;
+  onSubmit: (form: { reason: string; action: "resolve" | "unresolve" }) => void;
```
dessalines: Just pass up the reason. You can't resolve something that's already been resolved, so there's no need for that action.

If ppl mistyped the reason or something, they can just unresolve / resolve again.

src/shared/components/common/modal/conclude-report-modal.tsx
dessalines, line 1: If you want to see a good example of a simple modal, look at `src/shared/components/common/modal/person-note-modal.tsx` .

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+    this.modalDivRef = createRef();
+    this.reasonRef = createRef();
+  }
+
+  handleReasonChange = (event: FormEvent<HTMLInputElement>) => {
+    this.setState({ reason: event.currentTarget.value });
+  };
+
+  handleSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
```
dessalines: Top level functions.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+                          isResolved
+                            ? "unresolve_reason_optional"
+                            : "resolve_reason_optional",
+                        )}
+                        required={false}
+                        value={reason}
+                        onInput={this.handleReasonChange}
+                        ref={this.reasonRef}
+                      />
+                      {isResolved && (
```
dessalines: All this is unecessary. You only need a text box with an optional reason, and a single button that says either resolve or unresolve.

src/shared/components/common/modal/conclude-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+                {I18NextService.i18n.t("cancel")}
+              </button>
+            </footer>
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  handleShow() {
```
dessalines, line 129: I looked at a lot  of the other `handleShow`'s in the codebase, and none of them do this. It's always just `this.xRef.current?.focus();`, or just lets the existing mixin handle it.

src/shared/components/community/community-report.tsx
dessalines, line 1: Same issues as above.

src/shared/components/comment/comment-report.tsx
```diff
@@ -224,18 +240,20 @@ export class CommentReport extends Component<
             loading={false}
           />
         )}
+        {this.state.showResolveReportDialog && (
```
dessalines: I'm not sure this are correct, looking at the current examples of modals. You always render it, and pass `show=state.showDialog`.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,206 @@
...
+@modalMixin
+export default class ResolveReportModal extends Component<
+  ResolveReportModalProps,
+  ResolveReportModalState
+> {
+  modalDivRef: RefObject<HTMLDivElement>;
+  private reasonRef: RefObject<HTMLInputElement>;
+  modal?: Modal;
+  state: ResolveReportModalState = {
+    reason: "",
```
dessalines: Instead of a blank reason, you should probably pass the previous reason to it, as @Nutomic mentioned.

src/shared/components/common/modal/conclude-report-modal.tsx
iByteABit256: The reason for this was that I always get errors like this on Firefox:

```
Uncaught TypeError: can't access property "current", this.reasonRef is undefined
    handleShow resolve-report-modal.tsx:133
```

I left it the simple way for now though, functionally it works just the same

src/shared/components/comment/comment-report.tsx
iByteABit256: It didn't work for my modal, I also see other instances where a boolean check is needed like in the person note modal as well:

```typescript
{renderPersonNoteDialog && (
          <PersonNoteModal
            note={person_actions?.note}
            personId={creator.id}
            show={showPersonNoteDialog}
            onSubmit={form => {
              handleHideAllDialogs(this);
              onPersonNote(form);
            }}
            loading={false}
            onCancel={() => handleHideAllDialogs(this)}
          />
        )}
```

iByteABit256:
I fixed the code style issues and removed a lot of the unnecessary stuff.

I concluded that it was much simpler to just not include a separate reason for unresolving a report, it complicates everything a bit too much and I think appending unresolve reasons to the resolve_reason column with some delimiter would cause a lot of trouble.

I changed it to just not send any resolve_reason in the unresolve case.
However there is a bug in the backend implementation, if resolve_reason is non existent, it still overrides the existing value and sets the value to NULL.
After that is fixed though, the previous resolve reason will be persisted after unresolving, and the moderator can re-resolve and fix a typo in the previous reason for example very easily.

src/shared/components/common/modal/conclude-report-modal.tsx
dessalines: That's because of the issue above, where you aren't rendering the report dialog, so the button isn't there.

You should just always render the dialog, but only show it based on the state, `show={this.state.showReportDialog}`

src/shared/components/comment/comment-report.tsx
dessalines: This is going to cause problems, see above.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -97,70 +70,29 @@ export default class ResolveReportModal extends Component<
...
+                          {I18NextService.i18n.t("reason")}
+                        </label>
+                        <input
+                          type="text"
+                          id={reasonId}
+                          className="form-control my-2 my-lg-0"
+                          placeholder={I18NextService.i18n.t("reason")}
+                          required={false}
+                          value={reason}
+                          onInput={event => handleReasonChange(this, event)}
```
dessalines: We try to use `e` in this codebase instead of event, because event is sometimes a special keyword. Replace all usages of `event` with `e` for this PR.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -196,11 +130,33 @@ export default class ResolveReportModal extends Component<
...
+  i: ResolveReportModal,
+  event: FormEvent<HTMLFormElement>,
+) {
+  event.preventDefault();
+
+  const trimmedReason = i.state.reason?.trim();
+  const resolved = i.props.isResolved;
+
+  // Only pass reason if resolving and a reason was provided
+  const reason = !resolved && trimmedReason ? trimmedReason : undefined;
```
dessalines: Hrm... not sure this is correct. You can give a reason when unresolving a report if you want.

src/shared/components/community/community-report.tsx
```diff
@@ -115,6 +103,11 @@ export class CommunityReport extends Component<Props, State> {
                 />
               </T>
             )}
+            {resolved && resolveReason && (
+              <div>
+                {I18NextService.i18n.t("resolve_reason")}: {resolveReason}
```
dessalines: You can give a reason when resolving / unresolving if you want. So this should really just be `resolveReason &&` 

Same for the other cases.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -97,70 +70,29 @@ export default class ResolveReportModal extends Component<
...
-                              "resolve_report_confirmation",
-                            )}
-                          </small>
-                        </div>
-                      </div>
-                    </div>
-                  )}
-                  {isResolved && (
-                    <div className="row">
+                    <div className="row mb-3">
```
dessalines: You should always show the reason box.

dessalines (commented):
Nice job, this is very close, but there's something going on with unresolve report. Its not showing the reason field. You should be able to give a reason when unresolving a report also:

Image

src/shared/components/common/modal/resolve-report-modal.tsx
iByteABit256: All of this is because of the approach I mentioned above https://github.com/LemmyNet/lemmy-ui/pull/4274#issuecomment-4996902341

If we're going to have a reason for unresolving reports, I suppose we're going to use some delimiter to store them both in the same column? For example `Resolve reason text|Unresolve reason text` which might also be `Resolve reason text|` or `|Unresolve reason text` if one of the two was kept empty.

It feels pretty clunky to do it this way, I believe if we're going to have an unresolve reason too then we should have a separate column for it as well.

The other approaches are the one I went with initially, which was the resolve/unresolve reason completely overwriting whatever reason exists currently, which is bad if the user wants to correct something in the resolve reason. Or just to remove unresolve reason altogether like I went with here, which simplifies things at the cost of some extra functionality.

Nutomic: Storing them in the same field is okay, otherwise we would need a separate database table which would be too complicated. Anyway you definitely shouldnt quietly throw away user input like this.

iByteABit256: This was in combination with the modal conditionally showing the text box so the user would not have the ability to add an unresolve reason anyway, so there wouldn't be user input being thrown away. Though this specific part is probably not needed in that case as well on second thought.

Nevermind that though, I'll go with the same column with delimiter approach as you say

iByteABit256:
I added back the unresolve reason field and removed all the conditional logic from before.

Now you can add both a resolve and unresolve reason, they both get persisted, and they have a '|' delimiter in the stored value since they share the same column.

I updated the PR description with a demo video, and I opened a [new PR for the unresolve reason translation](https://github.com/LemmyNet/lemmy-translations/pull/276).

I also fixed the 'event' param name to 'e'.

src/shared/components/comment/comment-report.tsx
```diff
@@ -259,3 +277,27 @@ function handleAdminBan(i: CommentReport) {
...
+): string | undefined {
+  if (!resolveReason) return undefined;
+
+  const parts = resolveReason.split(REASON_DELIMITER);
+  if (parts.length === 2) {
+    // Return the appropriate part based on resolution state
+    return isResolved ? parts[0] : parts[1];
+  }
+  return resolveReason;
+}
```
dessalines: What is going on here, none of this is necessary. Just show the `resolve_reason` regardless of whether its a resolve or unresolve.

src/shared/components/comment/comment-report.tsx
```diff
@@ -35,23 +36,27 @@ interface CommentReportProps {
 
 interface CommentReportState {
   showRemoveCommentDialog: boolean;
+  showResolveReportDialog: boolean;
 }
 
+const REASON_DELIMITER = "|";
```
dessalines: Not necessary.

src/shared/components/comment/comment-report.tsx
dessalines: This is still not addressed. Use the state for the `show={}`

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+function getCurrentReasonFromProps(
+  props: ResolveReportModalProps,
+): string | undefined {
+  const { resolveReason, isResolved } = props;
+  if (!resolveReason) return undefined;
+
+  const parts = resolveReason.split(REASON_DELIMITER);
```
dessalines: Get rid of all this combining reason logic. Just load the initial reason from the props.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,212 @@
...
+  // Format: "resolve_reason|unresolve_reason"
+  if (parts.length === 2) {
+    return !isResolved ? parts[0] : parts[1];
+  }
+  return resolveReason;
+}
+
+function combineReasons(
```
dessalines: Not necessary.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,212 @@
...
+  render() {
+    const { reason } = this.state;
+    const { isResolved, loading } = this.props;
+    const reasonId = `resolve-reason-${randomStr()}`;
+    const formId = `resolve-form-${randomStr()}`;
+
+    // Determine placeholder text based on current action
+    const placeholderText = I18NextService.i18n.t(
+      isResolved ? "unresolve_reason" : "resolve_reason",
+    );
```
dessalines: You can probably get rid of this also. Just use the `reason` i18n string for the placeholder.

dessalines (commented):
None of the *combining reason* or delimiter logic is necessary.

When you load the modal, regardless of whether it was resolved or unresolved before, just put existing `resolve_reason` into the reason input field.

iByteABit256:
I guess there was a miscommunication before. I went back with the initial logic and fixed the remaining unresolved issues.
Hopefully it should be ok now.

src/shared/components/common/modal/resolve-report-modal.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  <div className="row mb-3">
+                    <div className="col-12">
+                      <label className="visually-hidden" htmlFor={reasonId}>
+                        {I18NextService.i18n.t("reason")}
+                      </label>
+                      <input
+                        type="text"
+                        id={reasonId}
+                        className="form-control my-2 my-lg-0"
+                        placeholder={I18NextService.i18n.t("reason")}
```
Nutomic: The placeholder is not visible if reason is already prefilled. Better to remove this, and make the label always visible by removing `visually-hidden`.

Nutomic:
Looks pretty good now. But what we have now is not necessarily a `resolve_reason`, it can also be unresolve reason. This is confusing so I propose to rename the API field and labels simply to `reason`. Thoughts?

iByteABit256:
I agree but `reason` is already taken from the report reason field, so it has to be something more specific implying both resolve and unresolve, not sure what would be a fitting name for it. Maybe something like `resolution_change_reason`?

Nutomic:
Ah youre right these are both on the same table. Then I would go with `decision`, `conclusion` or another synonym:

https://www.merriam-webster.com/thesaurus/decision

iByteABit256:
Check also the backend changes, I renamed it to `conclusion` and updated it here too, and I also removed the placeholder as you said and made the label always visible

Note related to the update form change in the backend, the UI will still send an empty string instead of null so that the value is overridden each time.

Edit: translations seem merged, not sure why the lint step is still not finding them

dessalines:
You might need to run `git submodule update --remote`

dessalines (approved):
Looks good btw. Once the back-end is merged, we can merge this.

src/shared/components/comment/comment-report.tsx
```diff
@@ -156,17 +158,27 @@ export class CommentReport extends Component<
                 />
               </T>
             )}
+            {conclusion && (
+              <div>
+                {I18NextService.i18n.t(
+                  resolved ? "resolve_reason" : "unresolve_reason",
```
Nutomic: This should only use a single string `conclusion` (same text whether its resolved or not).

src/shared/components/common/modal/conclude-report-modal.tsx
```diff
@@ -0,0 +1,152 @@
...
+  isResolved: boolean;
+  conclusion?: string;
+  onSubmit: (reason?: string) => void;
+  onCancel: () => void;
+  show: boolean;
+  loading: boolean;
+}
+
+interface ConcludeReportModalState {
+  reason?: string;
```
Nutomic:
```suggestion
  conclusion?: string;
```

src/shared/components/common/modal/conclude-report-modal.tsx
```diff
@@ -0,0 +1,152 @@
...
+  e: FormEvent<HTMLInputElement>,
+) {
+  i.setState({ reason: e.currentTarget.value });
+}
+
+function handleSubmit(i: ConcludeReportModal, e: FormEvent<HTMLFormElement>) {
+  e.preventDefault();
+  const trimmedReason = i.state.reason?.trim();
+  // This will still send an empty string instead of None to the backend
+  // so that it overrides the value, this is intended
```
Nutomic, line 145: Why is it intended to clear the conclusion here?

src/shared/components/common/modal/conclude-report-modal.tsx
```diff
@@ -0,0 +1,152 @@
...
+  reason?: string;
+}
+
+@modalMixin
+export default class ConcludeReportModal extends Component<
+  ConcludeReportModalProps,
+  ConcludeReportModalState
+> {
+  modalDivRef: RefObject<HTMLDivElement>;
+  private reasonRef: RefObject<HTMLInputElement>;
```
Nutomic:
```suggestion
  private conclusionRef: RefObject<HTMLInputElement>;
```

src/shared/components/common/modal/conclude-report-modal.tsx
```diff
@@ -0,0 +1,152 @@
...
+      </div>
+    );
+  }
+
+  handleShow() {
+    this.reasonRef.current?.focus();
+  }
+}
+
+function handleReasonChange(
```
Nutomic:
```suggestion
function handleConclusionChange(
```

src/shared/components/community/community-report.tsx
```diff
@@ -95,27 +103,47 @@ export class CommunityReport extends Component<Props, object> {
                 />
               </T>
             )}
+            {conclusion && (
+              <div>
+                {I18NextService.i18n.t(
+                  resolved ? "resolve_reason" : "unresolve_reason",
```
Nutomic: Also use single string `conclusion`

src/shared/components/community/community-report.tsx
```diff
@@ -95,27 +103,47 @@ export class CommunityReport extends Component<Props, object> {
...
+          show={this.state.showResolveReportDialog}
+          loading={this.props.loading}
+        />
       </div>
     );
   }
 }
 
-function handleResolveReport(i: CommunityReport) {
+function handleResolveReport(i: CommunityReport, reason?: string) {
```
Nutomic:
```suggestion
function handleResolveReport(i: CommunityReport, conclusion?: string) {
```

src/shared/components/post/post-report.tsx
```diff
@@ -163,15 +165,25 @@ export class PostReport extends Component<PostReportProps, PostReportState> {
             )}
           </div>
         )}
+        {conclusion && (
+          <div>
+            {I18NextService.i18n.t(
+              resolved ? "resolve_reason" : "unresolve_reason",
+            )}
```
Nutomic:
```suggestion
            {I18NextService.i18n.t("conclusion")}
```

src/shared/components/post/post-report.tsx
```diff
@@ -163,15 +165,25 @@ export class PostReport extends Component<PostReportProps, PostReportState> {
...
+          </div>
+        )}
         <div className="row row-cols-auto align-items-center gx-3 my-2">
           <div className="col">
             <ActionButton
-              label={tippyContent}
-              icon={r.post_report.resolved ? "check" : "x"}
+              label={I18NextService.i18n.t(
+                resolved ? "unresolve_report" : "resolve_report",
+              )}
```
Nutomic, line 178:
```suggestion
            {I18NextService.i18n.t("conclusion")}
```

src/shared/components/post/post-report.tsx
```diff
@@ -226,18 +238,19 @@ export class PostReport extends Component<PostReportProps, PostReportState> {
             loading={false}
           />
         )}
+        <ConcludeReportModal
+          isResolved={resolved}
+          conclusion={conclusion}
+          onSubmit={reason => handleResolveReport(this, reason)}
```
Nutomic:
```suggestion
          onSubmit={conclusion => handleResolveReport(this, conclusion)}
```

src/shared/components/post/post-report.tsx
```diff
@@ -261,3 +274,14 @@ function handleAdminBan(i: PostReport) {
     ban: !i.props.report.creator_banned,
   });
 }
+
+function handleResolveReport(i: PostReport, reason?: string) {
```
Nutomic:
```suggestion
function handleResolveReport(i: PostReport, conclusion?: string) {
```

src/shared/components/private_message/private-message-report.tsx
```diff
@@ -83,28 +91,47 @@ export class PrivateMessageReport extends Component<Props, object> {
                 />
               </T>
             )}
+            {conclusion && (
+              <div>
+                {I18NextService.i18n.t(
+                  resolved ? "resolve_reason" : "unresolve_reason",
+                )}
```
Nutomic: Here also `conclusion`

src/shared/components/private_message/private-message-report.tsx
```diff
@@ -83,28 +91,47 @@ export class PrivateMessageReport extends Component<Props, object> {
...
+            )}
           </div>
         )}
         <div className="mt-2">
           <ActionButton
-            label={tippyContent}
-            icon={pmr.resolved ? "check" : "x"}
+            label={I18NextService.i18n.t(
+              resolved ? "unresolve_report" : "resolve_report",
+            )}
```
Nutomic, line 105: Here too

src/shared/components/private_message/private-message-report.tsx
```diff
@@ -83,28 +91,47 @@ export class PrivateMessageReport extends Component<Props, object> {
...
-            onClick={() => handleResolveReport(this)}
-            iconClass={`text-${pmr.resolved ? "success" : "danger"}`}
+            onClick={() => this.setState({ showResolveReportDialog: true })}
+            iconClass={`text-${resolved ? "success" : "danger"}`}
           />
         </div>
+        <ConcludeReportModal
+          isResolved={resolved}
+          conclusion={conclusion}
+          onSubmit={reason => handleResolveReport(this, reason)}
```
Nutomic: And here

src/shared/components/private_message/private-message-report.tsx
```diff
@@ -83,28 +91,47 @@ export class PrivateMessageReport extends Component<Props, object> {
...
+          loading={this.props.loading}
+        />
       </div>
     );
   }
 }
 
-function handleResolveReport(i: PrivateMessageReport) {
-  i.setState({ loading: true });
+function handleResolveReport(i: PrivateMessageReport, reason?: string) {
```
Nutomic: And here

src/shared/components/common/modal/conclude-report-modal.tsx
iByteABit256: Because after fixing the backend update form bug, sending None (null/undefined) will leave the conclusion as is in the database despite it being a conclusion for the wrong action. So in that case, unresolving a report after resolving it with conclusion 'foo' and leaving the input empty would send null and result in the UI displaying "Unresolved because foo", is seems very confusing that way while using it.

Clearing it each time is the most straight forward way it can work for the user imo given the fact that we only use a single column for both reasons.

src/shared/components/comment/comment-report.tsx
iByteABit256: Doesn't that make it a bit confusing though when you unresolve a report and it just displays 'conclusion'? It makes sense when you've resolved a report, but in the unresolve case it's not that obvious to what this is a conclusion to if you don't show a different string.

src/shared/components/common/modal/conclude-report-modal.tsx
Nutomic: Its not a good idea to quietly lose data which cannot be restored. You can send None for now even if its confusing, and we can find a better solution later.

Nutomic:
There is also a conflict that you need to resolve, and it looks like CI is failing.

dessalines:
I'll see if I can finish it up.

src/shared/components/common/modal/conclude-report-modal.tsx
Nutomic: @dessalines You merged without fixing this one.

PR #4288, opened 2026-08-03
[0.19] Upgrade dependencies

Nutomic:
Deps for 0.19 are very outdated so we should update them and make a new release sooner or later.

Didnt make any major version updates as that could cause problems.


dessalines (approved):
Some of the babel upgrades scare me, because I remember having to mess with these a lot in `main`. 

As long as you test it locally and it works, then feel free to merge.

Nutomic:
Did a quick test and everything seems to work normally. Do you remember what broke during the previous upgrade? Anyway we should discover any problems when its deployed on enterprise and lemmy.ml

dessalines:
I don't remember, but its nbd.

PR #4289, opened 2026-08-03
Fix for restoring scroll position (fixes #4225)

### LINKED ISSUE #4225: Scroll position not restored when navigating back
1. Scroll down on the homepage
2. Open a post
3. Navigate back

With Lemmy 0.19 it will restore the previous scroll position, but in 1.0 it is broken. Instead it scrolls to the top of the page again.

Nutomic:
Used LLM to resolve the bug. This involved checking changes from Lemmy 0.19, and changes in InfernoJS, and adding extra logging. Apparently the problem is caused by a race condition due to restructuring of code. Now scroll position is again restored correctly.

Heres the LLM summary:
> Problem: Navigating back restores the scroll position momentarily, but the browser then resets the page to the top. Save/restore logic itself works — the value is stored and read correctly.
> 
> Root cause: On same-document back/forward navigation, browsers asynchronously reset scroll to the top after the page renders, even with history.scrollRestoration = "manual". This race fires ~90ms after the app's restore (which runs in componentDidUpdate once data loads), overriding it.
> 
> Fix (scroll-mixin.ts): restore() now re-applies the saved position if the browser resets it to top within 500ms, detected via a scroll event with scrollY === 0. The listener cleans up after itself or when re-applied, so it never fights the user.

### DIFF
--- a/src/shared/components/mixins/scroll-mixin.ts
+++ b/src/shared/components/mixins/scroll-mixin.ts
@@ -2,17 +2,32 @@ import { isBrowser, nextUserAction, snapToTop } from "../../utils/browser";
 import { Component, InfernoNode } from "inferno";
 import { Location, History, Action } from "history";
 
-function restoreScrollPosition(props: {
-  location: Location;
-}): number | undefined {
+function restoreScrollPosition(props: { location: Location }) {
   const key: string = props.location.key;
-  const y = sessionStorage.getItem(`scrollPosition_${key}`);
+  const scrollPosition = sessionStorage.getItem(`scrollPosition_${key}`);
 
-  if (y !== null) {
-    window.scrollTo({ left: 0, top: Number(y), behavior: "instant" });
-    return Number(y);
+  if (scrollPosition !== null) {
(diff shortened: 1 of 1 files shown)


src/shared/components/mixins/scroll-mixin.ts
```diff
@@ -115,8 +119,27 @@ export function scrollMixin<
     }
 
     restore() {
-      restoreScrollPosition(this.props);
+      const target = restoreScrollPosition(this.props);
       this.preventRestore();
+
+      if (target !== undefined && target > 0) {
+        // Browsers restore the scroll position they stored for the history
+        // entry shortly after a traversal. This happens asynchronously and can
```
dessalines: This is just duping the already existing restoreScrollPosition function. All the new cleanup logic should go inside there.

Nutomic: Its not duplicating anything. Still I moved the new logic into that function.

dessalines (approved):

PR #4290, opened 2026-08-04
Serve all CSS files from `/static` (fixes #3706)

### LINKED ISSUE #3706: Move all css files under `/static`
Some css files are loaded like `/static/5b6438cf/styles/styles.css` but others as `/css/themes/browser.css`. These should all be moved under `/static` with the hash, so they can be served as immutable. 

This will let us simplify the cache-control logic: https://github.com/LemmyNet/lemmy-ui/blob/0b3edaa709afa090325e52d63e9069bc7156222b/src/server/middleware.ts#L47

Nutomic:
Tested and works. Also made adjustments for caching.


src/server/handlers/code-theme-handler.ts
dessalines, line 1: Why are these deleted?

Nutomic: My bad I thought they were unused, but it seems they are needed for custom themes. Restored them now.

dessalines (approved):

PR #4291, opened 2026-08-04
Add description for private instance setting (fixes #4271)

### LINKED ISSUE #4271: Description for "Private instance" admin setting
We keep receiving some security advisories which report that content on private instances is leaked via federation. But that is actually working as intended, because the private instance setting only affects access via the local api (to lock out LLM scrapers for example). To make content completely private requires disabling federation or changing community visibility settings.

So in `/admin` there should be a description text for this setting, something like this:

> Prevent access to this instance for users without login. Note that data can still be from other instances via federation. For even stricter access control, you can disable federation or change the visibility of communities to private or local only.

Later all settings should have descriptions, see https://github.com/LemmyNet/lemmy-ui/issues/3916

Nutomic:
Adding this one specifically because it is often confusing for admins.

Screenshot 2026-08-04 at 14-55-56 Admin settings - Voyager

### DIFF
--- a/src/shared/components/home/site-form.tsx
+++ b/src/shared/components/home/site-form.tsx
@@ -436,7 +436,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
               option={"private_instance"}
               isChecked={this.state.siteForm.private_instance ?? false}
               onCheck={val => handleSitePrivateInstance(this, val)}
-              description="Only allow logged-in users to browse the site. Helpful to stop web crawlers. Note that content is still accessible over federation (use community visibility settings to prevent this)."
+              description="private_instance_description"
             />
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


src/shared/components/home/site-form.tsx
```diff
@@ -436,6 +436,7 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
               option={"private_instance"}
               isChecked={this.state.siteForm.private_instance ?? false}
               onCheck={val => handleSitePrivateInstance(this, val)}
+              description="Only allow logged-in users to browse the site. Helpful to stop web crawlers. Note that content is still accessible over federation (use community visibility settings to prevent this)."
```
Nutomic: Need to move this to translations. Is the text good like this or any suggestions?

dessalines: Yes seems fine.

Nutomic:
Translation done in https://github.com/LemmyNet/lemmy-translations/pull/278

dessalines (approved):

PR #4293, opened 2026-08-04
Moving css files under /static. (#3706)

### LINKED ISSUE #3706: Move all css files under `/static`

JoeJoeflyn:
Moves CSS routes under `/static/{hash}/` for immutable caching. Closes #3706.


Nutomic:
Duplicate of https://github.com/LemmyNet/lemmy-ui/pull/4290

JoeJoeflyn:
tks

PR #4297, opened 2026-08-05
Fix sidebar spacing and registration applications unread counter

JoeJoeflyn:
## Summary
- Replace `mb-2`/`my-1` margins with `d-flex flex-column gap-2` on sidebar containers for consistent spacing
- Remove redundant `.btn.d-block + .btn.d-block` margin-top CSS rule (was marked as TODO in main.css)
- Switch badges and mods list from `ul`/`li` to `div`/`span` with `d-flex flex-wrap gap-1` to avoid default list margins and padding
- Add `mb-0` to alert and `<p>` elements inside gap containers to prevent double spacing
- Fix registration applications `unread_only` filter to use `registrationStateFromQuery` instead of raw view string
- Decrement `UnreadCounterService` unread application count on approve instead of manually editing list state

#### Test plan
- [x] Sidebar buttons have consistent spacing with no double margins
- [x] Badges wrap correctly with `gap-1` spacing, no extra padding
- [x] Mods list renders without default list margins
- [x] Registration applications unread filter works correctly
- [x] Approving an application decrements the unread counter

Before:
screenshot-2026-08-05_03-41-50
After:
screenshot-2026-08-05_17-31-37

Video:
Before:
screenrecording-2026-08-05_04-44-46-ezgif com-video-to-gif-converter
After:
screenrecording-2026-08-05_04-53-33-ezgif com-video-to-gif-converter

### DIFF
--- a/src/assets/css/main.css
+++ b/src/assets/css/main.css
@@ -314,14 +314,6 @@ hr {
   -ms-transform: scale(1.2);
 }
 
-/**
- * TODO: Fix this in markup rather than this overly specific selector:
- * https://getbootstrap.com/docs/5.3/components/buttons/#block-buttons
- */
-.btn.d-block + .btn.d-block {
-  margin-top: 0.5rem;
-}
-
 .mini-overlay {
   display: block;
   height: 1.5em;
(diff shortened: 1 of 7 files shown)


dessalines (approved):
Nice, thx for these fixes.

PR #4298, opened 2026-08-05
feat: add invitation system for require_invitation registration mode (fixes #4128)

### LINKED ISSUE #1777: Add Invitation for registration
For front end issues, use [lemmy-ui](https://github.com/LemmyNet/lemmy-ui)

I don't know, if it's a frontend- or a backendissue... Registration in lemmy is very "basic". Open registration invites spammers to spam. Ok. Now there are captures... 

But i can not have closed registration and create manually a user.

Add "invitations" to login-process

For my usecase i will prefer "Invites". Create a link which is valid for a due-date and it can be used only $COUNT times. Pleroma has this feature, and it is great. I can create for example one invitation-link with count 4 and or without duedate.
so i can send this link in my family-group (4 members except me), and they can use ist. if they give it to someone i don't like to have on my server, i wil recognize ist.

Make it possible, to create a user manually as admin, even with closed registration.

### LINKED ISSUE #4128: Invite-only registration mode
See https://github.com/LemmyNet/lemmy/pull/6450 for details, and https://github.com/LemmyNet/lemmy-js-client/pull/957 for js-client.

cc @kryoseu

JoeJoeflyn:
Closes LemmyNet/lemmy#1777

## Summary
- Add `require_invitation` option to `RegistrationModeDropdown`
- Add invite management UI in user settings (create, list, revoke, copy link)
- Add invite token field to signup page, reads `?token=` from URL
- Add `max_invites_per_user_allowed` field to admin site form
- Add tab persistence via URL hash in `Tabs` component
- Add translation keys for invite UI

#### Test plan
- [x] Set registration mode to "Require invitation" in admin settings
- [x] Set max invites per user allowed
- [x] Go to user settings > Invites tab, create an invite
- [x] Copy invite link, open in incognito — signup page shows token pre-filled
- [x] Register with the invite token — succeeds
- [x] Revoke an invite — that token no longer works
- [x] Reload settings page — stays on Invites tab

screenshot-2026-08-05_19-39-06
screenshot-2026-08-05_19-38-43
screenshot-2026-08-06_16-12-18
screenshot-2026-08-06_16-11-59


Nutomic:
- Translations are missing (same problem as your other PR)
- Changing the dropdown values for max_uses and expires_in_hours doesnt work, the always go back to the default
- Expire invites after 30min by default doesnt make sense, use unlimited as default. Maybe also make number unlimited by default
- The request to list invites (GET /api/v4/account/invite?limit=20) fails, it should actually be /api/v4/account/invite/list. Probably needs fix in lemmy-js-client. Did you really test this?

JoeJoeflyn:
The list works locally (see screenshots), but you're right about the lemmy-js-client path. I'll fix it

poVoq:
Would it be possible to have invites in addition to admin approval?

Like if an existing member invites someone they can use their account immediatly, but if a new user doesn't have an invite token they can still register and wait for admin approval?

Also, the admin approval text question is kinda nice to have regardless, even if not needed with an invite. Looking up what people said about themselves when registering is often useful. But I guess the wording would need to change a bit if it is optional instead of mandatory.

JoeJoeflyn:
> Would it be possible to have invites in addition to admin approval?
> 
> Like if an existing member invites someone they can use their account immediatly, but if a new user doesn't have an invite token they can still register and wait for admin approval?
> 
> Also, the admin approval text question is kinda nice to have regardless, even if not needed with an invite. Looking up what people said about themselves when registering is often useful. But I guess the wording would need to change a bit if it is optional instead of mandatory.

Let's wait for the maintainers' opinions first.

Nutomic:
Like I said you need to make a PR in lemmy-js-client repo to fix the api path. Then we will make a new tag there that you can use to make this PR work. And again, please dont use `git push --fork`.

@poVoq So basically bypass registration application with invite. The problem is that a spammer could get one legitimate account approved, and then use it to invite 10 more accounts (each of which can invite another 10). So it would first require a new feature for admins to easily see who invited a certain user, and ban these related accounts together (see https://github.com/LemmyNet/lemmy/issues/1777#issuecomment-4269729217). So please open an issue about this in the backend repo, and we focus on the basic implementation for now.

src/shared/components/home/authenticate/signup.tsx
```diff
@@ -79,8 +81,19 @@ export class Signup extends Component<SignupRouteProps, State> {
...
+      }
+    }
+  }
+
+  extractInviteToken() {
+    const params = new URLSearchParams(window.location.search);
+    const token = params.get("token") ?? undefined;
+    if (token && this.state.form.token !== token) {
+      this.setState(s => ({ ...s, form: { ...s.form, token } }));
     }
```
dessalines: Make this a top-level function.

src/shared/components/home/authenticate/signup.tsx
```diff
@@ -233,7 +247,11 @@ export class Signup extends Component<SignupRouteProps, State> {
   renderCaptcha(): InfernoNode | void {
     switch (this.state.captchaRes.state) {
       case "loading":
-        return <Spinner />;
+        return (
+          <div className="d-flex align-items-center justify-content-center py-4 my-2 w-100">
+            <Spinner large />
+          </div>
```
dessalines: This is a lot of custom classes. If you're going to do this a lot, make it its own component, something like `<CenteredSpinner`, or a param on the existing `Spinner` component.

src/shared/components/home/authenticate/signup.tsx
```diff
@@ -268,6 +286,37 @@ export class Signup extends Component<SignupRouteProps, State> {
...
   }
 
+  renderInviteTokenInput(siteView?: SiteView) {
+    const isRequireInvitation =
+      siteView?.local_site.registration_mode === "require_invitation";
+    const hasTokenInState = !!this.state.form.token;
+
+    if (!isRequireInvitation && !hasTokenInState) {
+      return null;
+    }
```
dessalines: Don't return null early. Change this logical block  to its opposite and move it into the return below.

src/shared/components/person/invites.tsx
```diff
@@ -0,0 +1,386 @@
...
+              onChange={e => {
+                const val = e.target.value;
+                this.setState(s => ({
+                  ...s,
+                  form: {
+                    ...s.form,
+                    max_uses: val === "unlimited" ? undefined : Number(val),
+                  },
+                }));
+              }}
```
dessalines: Extract into a handler function.

src/shared/components/person/invites.tsx
```diff
@@ -0,0 +1,386 @@
...
+            >
+              <option value="1" selected={this.state.form.max_uses === 1}>
```
dessalines: I spose hardcoding a specific set of options is a good idea.

src/shared/components/person/invites.tsx
```diff
@@ -0,0 +1,386 @@
...
+              onChange={e => {
+                const val = e.target.value;
+                this.setState(s => ({
+                  ...s,
+                  form: {
+                    ...s.form,
+                    expires_at: val === "never" ? undefined : Number(val),
+                  },
+                }));
+              }}
```
dessalines: top level handler function.

src/shared/components/person/invites.tsx
```diff
@@ -0,0 +1,386 @@
...
+      </form>
+    );
+  }
+
+  invitesList() {
+    switch (this.state.invitesRes.state) {
+      case "loading":
+        return (
+          <div className="d-flex flex-column align-items-center justify-content-center py-5 text-body-secondary w-100 text-center">
+            <Spinner large />
```
dessalines: See comment about spinner above.

src/shared/components/person/invites.tsx
```diff
@@ -0,0 +1,386 @@
...
+      case "success": {
+        const invites = this.state.invitesRes.data.items;
+        if (!invites.length) {
+          return (
+            <p className="text-body-secondary mb-0">
+              {I18NextService.i18n.t("no_invites")}
+            </p>
+          );
+        }
+        return (
```
dessalines: else return { ...

or even better, do this inside the jsx:

`invites.length ? .... : ....`

src/shared/components/person/invites.tsx
```diff
@@ -0,0 +1,386 @@
...
+          <div className="table-responsive">
+            <table className="table table-hover align-middle mb-0">
+              <thead className="table-light">
+                <tr className="small text-body-secondary">
+                  <th
+                    scope="col"
+                    className="py-3"
+                    style={{ "min-width": "260px" }}
```
dessalines: Don't hardcode widths or styles like this. Also don't use tables, we did a lot of work to get rid of them because they don't work well on mobile.

There's an example you can paste from `src/shared/components/multi-community/multi-communities.tsx` , and the `<ResponsiveTableRowHeader` component also.

dessalines (commented):
This is looking quite good, nice job. My main issue is that we should keep the pattern of staying away from tables, which I did a lot of work to convert all of them to responsive rows a few months ago. There's examples in `src/shared/components/community/communities.tsx` and `src/shared/components/multi-community/multi-communities.tsx` you can copy-paste from.

src/shared/components/common/icon.tsx
```diff
@@ -35,10 +35,19 @@ export function Icon({ icon, classes, inline, small }: IconProps) {
...
 
-export function Spinner({ large, className }: SpinnerProps) {
+export function Spinner({ large, centered, className }: SpinnerProps) {
+  if (centered) {
+    return (
+      <div className="d-flex align-items-center justify-content-center py-4 my-2">
+        <Spinner large={large} className={className} />
+      </div>
+    );
+  }
```
dessalines, line 49: Don't do an early return. Add the classes necessary below to the `classNames` function if centered is true.

src/shared/components/home/authenticate/signup.tsx
```diff
@@ -475,3 +520,11 @@ export function signupTitleName(siteView?: SiteView): string {
     siteView?.local_site.private_instance ? "apply_to_join" : "sign_up",
   );
 }
+
+function extractInviteToken(i: Signup) {
+  const params = new URLSearchParams(window.location.search);
+  const token = params.get("token") ?? undefined;
+  if (token && i.state.form.token !== token) {
+    i.setState(s => ({ ...s, form: { ...s.form, token } }));
```
dessalines: Its much easier to read to do : 

```typescript
i.state.form.token = token;
i.setState(i.state);
```

src/shared/components/home/site-form.tsx
```diff
@@ -370,6 +371,34 @@ export class SiteForm extends Component<SiteFormProps, SiteFormState> {
...
+                type="number"
+                id="max-invites-per-user"
+                className="form-control"
+                min={0}
+                value={this.state.siteForm.max_invites_per_user_allowed ?? 0}
+                onInput={e =>
+                  handleMaxInvitesPerUserChange(
+                    this,
+                    Number(e.target.value) || 0,
+                  )
```
dessalines: Would probably be better to handle this conversion inside the handler function, but its up to you.

dessalines (commented):
Sry for my slowness in reviewing. Everything looks pretty good, just a few more things below.

PR #4302, opened 2026-08-06
Some simple logic fixes

Nutomic:
Had LLM analyze code changes since 0.19 which found a number of simple logic bugs. Should be mostly self-explanatory.

### DIFF
--- a/src/shared/components/post/create-post.tsx
+++ b/src/shared/components/post/create-post.tsx
@@ -178,9 +178,7 @@ export class CreatePost extends Component<
         initialCommunitiesRes,
       });
 
-      if (
-        communityId?.toString() !== this.state.selectedCommunity?.community.id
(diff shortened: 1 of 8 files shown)


src/shared/components/post/create-post.tsx
```diff
@@ -178,9 +178,7 @@ export class CreatePost extends Component<
         initialCommunitiesRes,
       });
 
-      if (
-        communityId?.toString() !== this.state.selectedCommunity?.community.id
-      ) {
+      if (communityId !== this.state.selectedCommunity?.community.id) {
```
Nutomic, line 181: This comparison would always fail because its comparing string to int. Unfortunately lints dont catch it because both sides share type `undefined`.

dessalines (approved):

PR #4305, opened 2026-08-07
Fix for ViewerJS leak

Nutomic:
The existing code recreates viewers on every render (which is correct), but on rerender never destroys the previous viewers which leads to memory leak. Also viewers should not be stored in state because changing them should not cause extra render.


dessalines (commented):
Should probably also be added to the `componentWillUnmount` hooks of both those screens.

Nutomic:
In both files `  componentWillUnmount()` calls `this.unloadViewerJs();` so that looks fine.

dessalines (approved):

PR #4307, opened 2026-08-14
[0.19] Add `referrerpolicy` to video iframe

flamingos-cant:
## Description

This change makes it so that YouTube video embeds are actually playable. According to MDN, `strict-origin-when-cross-origin` is suppose to be the default, so I'm not sure why this change fixes this but it does.

Before: https://slrpnk.net/post/41462932  
After: https://feddit.uk/post/53830092


Nutomic (approved):

Nutomic:
I guess that InfernoJS internally uses a different default value. Thanks for the fix!

PR #4322, opened 2026-09-14
feat: handle email verification failures

Pdzly:
## Description

This PR clarifies when the verification link the user tries to access is invalid, already used or expired.

Potential linked issue: #6653

Linked translation PR: https://github.com/LemmyNet/lemmy-translations/pull/286

### Before
grafik

### After
grafik


dessalines:
Looks good, thx!

src/shared/components/person/verify-email.tsx
```diff
@@ -31,26 +30,35 @@ export class VerifyEmail extends Component<
...
       this.props.history.push("/login");
     }
   }
 
-  async componentWillMount() {
-    if (isBrowser()) {
-      await this.verify();
-    }
+  componentDidMount() {
+    void this.verify();
```
dessalines: What is this void about? You can remove them.

And why did the async get removed?

src/shared/components/person/verify-email.tsx
```diff
@@ -31,26 +30,35 @@ export class VerifyEmail extends Component<
   };
 
   async verify() {
+    if (this.state.verifyRes.state === "loading") {
+      return;
+    }
```
dessalines: What is this about? Don't early return on a loading... this async function needs to set loading to true, then API call the verify.

Pdzly: Thanks this is my first time working with the framework will fix and update it.

Pdzly: Done

src/shared/components/person/verify-email.tsx
```diff
@@ -35,24 +34,30 @@ export class VerifyEmail extends Component<
...
 
-  async componentWillMount() {
-    if (isBrowser()) {
-      await this.verify();
-    }
+  async componentDidMount() {
+    await this.verify();
   }
 
+  handleRetry = async () => {
```
dessalines: Use an async function. We try to avoid fake closure functions in this codebase as much as possible.

src/shared/components/person/verify-email.tsx
```diff
@@ -74,9 +79,62 @@ export class VerifyEmail extends Component<
...
           </div>
         </div>
       </div>
     );
   }
+
+  verificationError() {
+    const { verifyRes } = this.state;
+    if (verifyRes.state !== "failed") {
+      return;
```
dessalines: Early return is still here.

src/shared/components/person/verify-email.tsx
```diff
@@ -35,24 +34,30 @@ export class VerifyEmail extends Component<
...
+    const verifyRes = await HttpService.client.verifyEmail({
+      token: this.props.match.params.token,
+    });
     this.setState({
-      verifyRes: await HttpService.client.verifyEmail({
-        token: this.props.match.params.token,
-      }),
+      verifyRes:
+        verifyRes.state === "empty"
+          ? { state: "failed", err: new Error("empty_response") }
```
dessalines: Why is this error getting overridden with `new Error`? Why not just handle the failed state in your component? Let the res be set normally.

